PR #1, opened 2017-11-30
Add test and review infrastructure files

jayconrod:


.test-bazelrc
```diff
@@ -0,0 +1,9 @@
+# This file contains options passed to Bazel when running tests.
+# They are used by Travis CI and by non-Bazel test scripts.
+
+build --verbose_failures
```
achew22: Bazel has a cool flag `--config` which is available in (I think) every command. If you pass it in you can change the bazelrc options that are active.

For example if you had a `tools/bazel.rc` you could fill it with

```
build:ci --verbose_failures
build:ci --sandbox_debug
build:ci --span_strategy=standalone
build:ci --genrule_strategy=standalone
test:ci --test_strategy=stanalone
test:ci --test_output=errors
```

Then you could invoke bazel build with: 

```
    bazel \
       --output_base=$HOME/.cache/bazel \
       --batch \
       --host_jvm_args=-Xmx500m \
       --host_jvm_args=-Xms500m \
       --config=ci \
       test \
       --local_resources=400,1,1.0 \
       //...
```

I ended up doing this last night which is why it's on the mind. If you don't like it don't do it, but it seemed nice to me last night when I was reading the source. If you wanna see an example, here is the CL.
https://github.com/achew22/bazel-watcher/commit/d703d18900d21834ecb87ba8140b82f3bca47079

jayconrod: Done. This sounds like a more canonical solution for CI configurations.

PR #38, opened 2017-12-18
Rewrite packages.PlatformStrings (again)

### LINKED ISSUE #28: Duplicated dependency in Gazelle generated BUILD.bazel
My project has [go-iatty](https://github.com/mattn/go-isatty) as a dependency. When building using gazelle generated BUILD files I get the following error:

```
ERROR: /Users/kkiningh/go/src/ci/vendor/github.com/mattn/go-isatty/BUILD.bazel:3:1: Label '//vendor/golang.org/x/sys/unix:go_default_library' is duplicated in the 'deps' attribute of rule 'go_default_library'
ERROR: /Users/kkiningh/go/src/ci/vendor/github.com/gin-gonic/gin/BUILD.bazel:3:1: Target '//vendor/github.com/mattn/go-isatty:go_default_library' contains an error and its package is in error and referenced by '//vendor/github.com/gin-gonic/gin:go_default_library'
```

Looking at the generated build rules, I see the following declaration

```Python
go_library(                                                                                                                                                                                                     
    name = "go_default_library",                                                                                                                                                                                
    srcs = [
        # omitted
    ],
    importpath = "github.com/mattn/go-isatty",
    deps = select({
        "@io_bazel_rules_go//go/platform:solaris": [
            "//vendor/golang.org/x/sys/unix:go_default_library",
        ],
        "//conditions:default": [],
    }) + select({
            "@io_bazel_rules_go//go/platform:linux_ppc64": [
            "//vendor/golang.org/x/sys/unix:go_default_library",
        ],
        "@io_bazel_rules_go//go/platform:linux_ppc64le": [
            "//vendor/golang.org/x/sys/unix:go_default_library",
        ],
        "//conditions:default": [],
    }),
)
```

I'm not sure why this would cause a duplicated dependency, but when I remove the entire ```dep``` declaration the build proceeds file. I'm building on OSX High Sierra with Intel i7.

jayconrod:
PackageStrings needed a new invariant: strings must not be duplicated
across sets. For example, the same string cannot be OS-specific and
platform-specific. Bazel rejects build files where this is the case
for deps.

This change introduces platformStringsBuilder. Strings are added to
sets, and information is stored about which set each string belongs
to. The Map and Clean methods are removed; when PlatformStrings is
built, it is already clean.

This affects option strings somewhat. Unprocessed groups of options
are stored as strings with OptSeparator between each option. rules is
now responsible for splitting and quoting these.

This means order is no longer preserved for option strings (they will
be sorted when PlatformStrings is built). We couldn't actually
guarantee order before: for example, if a file contained an
arch-specific #cgo directive, followed by an OS-specific #cgo
directive, they would already appear in the wrong order.

Fixes #28

### DIFF
--- a/packages/package.go
+++ b/packages/package.go
@@ -16,8 +16,8 @@ limitations under the License.
 package packages
 
 import (
-	"errors"
 	"fmt"
+	"log"
 	"path"
 	"sort"
 	"strings"
@@ -67,6 +67,13 @@ type ProtoTarget struct {
 // PlatformStrings contains a set of strings associated with a buildable
 // Go target in a package. This is used to store source file names,
 // import paths, and flags.
+//
+// Strings are stored in four sets: generic strings, OS-specific strings,
+// arch-specific strings, and OS-and-arch-specific strings. A string may not
+// be duplicated within a list or across sets; however, a string may appear
+// in more than one list within a set (e.g., in "linux" and "windows" within
+// the OS set). Strings within each list should be sorted, though this may
+// not be relied upon.
 type PlatformStrings struct {
 	// Generic is a list of strings not specific to any platform.
 	Generic []string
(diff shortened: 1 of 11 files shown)


packages/package.go
```diff
@@ -192,72 +209,127 @@ func (p *Package) addFile(c *config.Config, info fileInfo, cgo bool) error {
...
+	goSrcs := []platformStringsBuilder{
+		pb.library.sources,
+		pb.binary.sources,
+		pb.test.sources,
+		pb.xtest.sources,
 	}
+	for _, sb := range goSrcs {
+		if sb.strs != nil {
+			for s, _ := range sb.strs {
+				if strings.HasSuffix(s, ".go") {
```
ianthehat: Isn't a package buildable if it only has .s files?

jayconrod: Nope. See [build.go:890](https://golang.org/src/go/build/build.go#L890).

I guess at least one .go file is needed to determine the package name. Not sure if there are other reasons.

PR #40, opened 2017-12-18
Improve performance of gazelle --mode=diff

yasushi-saito:
Don't invoke the "diff" command if the contents haven't changed.
Don't call Sync() on every write. Just closing the file suffices.

Before this change, diffing a repo with 1600 BUILD files would take 60 seconds.
After the change, it takes 2 seconds.

### DIFF
--- a/cmd/gazelle/diff.go
+++ b/cmd/gazelle/diff.go
@@ -16,28 +16,33 @@ limitations under the License.
 package main
 
 import (
+	"bytes"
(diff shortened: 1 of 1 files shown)


yasushi-saito:
trying the CLA thing again.

jayconrod:
Thanks! LGTM.

I confirmed internally that you have signed the CLA. Not sure why the bot doesn't know that.

Issue #29, opened 2017-12-15
Import repository rules from vendoring tools


jayconrod:
(Migrated from bazelbuild/rules_go#1082)

Gazelle should be able to transcribe lock files from popular vendoring tools like dep into repository rules in WORKSPACE. These tools produce lock files, which contain import paths and commit ids for each remote repository. It should be straightforward to generate repository rules from these.

* We should understand lock file formats of a few popular tools. dep will be first. Support for more tools can be added on demand.
* The import functionality should be available via a new command. For example:

```
gazelle migrate-deps Gopkg.lock
```

* Gazelle may initially add `go_repository` rules for imported dependencies. However, it is likely we will deprecate `go_repository` soon, so Gazelle should add `git_repository` and `http_archive` rules with pre-generated build files.
  * There isn't a viable replacement for `go_repository` yet. `new_git_repository` and `new_http_archive` only accept a single flat build file, which can be difficult to maintain and causes migration problems for repositories that eventually migrate to Bazel. We need replacements for these rules that can overlay a tree of build files on a downloaded repository.
* Gazelle should not add new rules if an existing rule checks out the same repository. The import command must be idempotent. If dependencies change or are updated in the lock file, Gazelle should update rules in place when possible (and report an error otherwise).
* Gazelle may not be able to remove dependencies that were removed from the lock file.

robfig:
Gazelle supports Go Modules

PR #47, opened 2017-12-21
New command: migrate-deps

jayconrod:
With this change, Gazelle supports a new command: migrate-deps. This
can be used to migrate dependencies from a vendoring lock file (only
dep's Gopkg.lock for now, but more to come) to go_repository rules in
WORKSPACE.

Related #29


merger/merger.go
```diff
@@ -484,52 +509,34 @@ func shouldKeep(e bf.Expr) bool {
...
-	if kind := kind(c); kind == "load" {
-		if len(c.List) == 0 {
-			return -1, nil
-		}
-		m = &loadMatcher{stringValue(c.List[0])}
-	} else {
-		m = &nameMatcher{kind, name(c)}
-	}
+// match looks for the matching CallExpr in stmts. If a rule with a matching
+// "name" attribute is found, and the kind matches, the rule and its index
```
ianthehat: the other way around, index then rule

merger/merger.go
```diff
@@ -484,52 +509,34 @@ func shouldKeep(e bf.Expr) bool {
...
+// are returned. If a rule with a matching name is found but the kind does not
+// match, the index of the rule and nil are returned. If no match is found,
+// -1 and nil are returned.
+func match(stmts []bf.Expr, x *bf.CallExpr) (int, *bf.CallExpr) {
```
ianthehat: The api is not great, you can't do _, rule = match(...) when you don't care about type.
Not needed to fix for this PR (it's a private rarely called function) but consider changing to something that maybe returns index, rule, bool where index and rule are both always set or not together, and the final bool indicates a full or partial match (or something)

repos/repo.go
```diff
@@ -0,0 +1,76 @@
...
+}
+
+type byName []repo
+
+func (s byName) Len() int           { return len(s) }
+func (s byName) Less(i, j int) bool { return s[i].name < s[j].name }
+func (s byName) Swap(i, j int)      { s[i], s[j] = s[j], s[i] }
+
+var lockFileParsers = map[string]func(string) ([]repo, error){
+	"Gopkg.lock": importRepoRulesDep,
```
ianthehat: I think you should add an extra step of indirection in here.
Have a function that peeks at a file to guess the "type" and returns a handler for that type.
The import them becomes a method on the returned object.
At the moment the only peeking is checking the base name, but you could also imagine looking inside the file for some magic to disambiguate. Also gives you a useful value to hang other things off in the future.

merger/merger.go
jayconrod: Done, that seems sensible.

This function will need to change in the future anyway, since we'll need to match on more attributes (e.g., importpath).

repos/repo.go
jayconrod: Done

PR #52, opened 2017-12-26
resolve: detect when a rule imports itself

### LINKED ISSUE #45: Circular dependency in generated proto_library rules
In case a package contains multiple proto files:

- these files have the same `go_package` option, and
- some of these files are importing some others (no circular dependency)

gazelle will put them in the same proto_library rule with a dependency to itself.  The generated go_proto_library will have the same issue.

Take `google/api/{http,annotations}.proto` for instance.  They both have `option go_package = "google.golang.org/genproto/googleapis/api/annotations;annotations"`, and `annotations.proto` imports `http.proto`.  gazelle will generate the follow code:

```
load("@io_bazel_rules_go//proto:def.bzl", "go_proto_library")

proto_library(
    name = "annotations_proto",
    srcs = [
        "annotations.proto",
        "http.proto",
    ],
    deps = [
        ":annotations_proto",      <---------- CIRCULAR DEPENDENCY
        "@com_google_protobuf//:descriptor_proto",
    ],
)

go_proto_library(
    name = "annotations_go_proto",
    importpath = "google/api",
    proto = ":annotations_proto",
    deps = [
        ":go_default_library",      <---------- CIRCULAR DEPENDENCY
        "@com_github_golang_protobuf//protoc-gen-go/descriptor:go_default_library",
    ],
)
```

jayconrod:
Add selfImportError, which is now returned by
RuleIndex.findRuleByImport and related methods when a rule imports
itself (which is common with protos).

Resolver methods treat this error the same as standardImportError: no
dependency is emitted.

Fixes #45

### DIFF
--- a/resolve/index.go
+++ b/resolve/index.go
@@ -190,7 +190,7 @@ func (ix *RuleIndex) skipGoEmbds() {
 		}
 
 		for _, l := range embedLabels {
-			embed, ok := ix.findRuleByLabel(l, r.label.Pkg)
+			embed, ok := ix.findRuleByLabel(l, r.label)
 			if !ok {
 				continue
 			}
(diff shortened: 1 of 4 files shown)


resolve/index.go
```diff
@@ -260,21 +270,19 @@ func (ix *RuleIndex) findRuleByImport(imp importSpec, lang config.Language, from
...
-			if m.label.Repo == "" {
-				parts := strings.Split(m.label.Pkg, "/")
-				for i, part := range parts {
-					if part == "vendor" {
-						isVendored = true
-						vendorRoot = strings.Join(parts[:i], "/")
-						break
-					}
+			parts := strings.Split(m.label.Pkg, "/")
+			for i, part := range parts {
```
ashi009: How about
```
if idx := strings.Index(m.label.Pkg, "/vendor/"); idx > 0 {
  isVendored = true
  vendorRoot = m.label.Pkg[:idx]
}
```

ashi009: Please ignore.  This is not covering some corner cases.

PR #62, opened 2017-12-28
def.bzl: fix extra_args.

diegs:
The extra_args (and args, via the fallback logic) are no longer being
added to the generated bash script.


diegs:
Sorry, the other deltas were performed automatically by `buildifier`. I can back them out.

PR #69, opened 2018-01-03
resolve: add rules to index after pre-resolve merge

### LINKED ISSUE #67: Add rules to index after pre-resolve merge
`resolve.RuleIndex` provides two public methods that basically do the same thing: `AddRulesFromFile` and `AddGeneratedRules`. These methods need to account for naming conflicts between generated rules and existing rules.

Instead, we should just have one method that indexes the rules in a file *after* the pre-resolve merge has been performed. The pre-resolve step will combine generated and existing rules and will delete empty rules.

This should also simplify `merger.MergeFile`, which currently returns a list of merged rules to be passed to `AddGeneratedRules`.

jayconrod:
runFixUpdate now merges generated rules into existing build files
before adding the rules in the merged file to the library index with
AddRulesFromFile. This avoids the need to add existing and generated
rules to the index separately. It also simplifies the index, since we
don't need to track generated and replaced rules.

Also: the '# gazelle:ignore' directive is now handled earlier in
packages.Walk instead of merger.MergeFile. This avoids extra work and
spurious error messages.

Fixes #67

### DIFF
--- a/cmd/gazelle/fix-update.go
+++ b/cmd/gazelle/fix-update.go
@@ -91,13 +91,11 @@ func runFixUpdate(cmd command, args []string) error {
 		// Fix any problems in the file.
 		if file != nil {
 			file = merger.FixFileMinor(c, file)
+			fixedFile := merger.FixFile(c, file)
 			if cmd == fixCmd {
-				file = merger.FixFile(c, file)
-			} else {
-				fixedFile := merger.FixFile(c, file)
-				if fixedFile != file {
-					log.Printf("%s: warning: file contains rules whose structure is out of date. Consider running 'gazelle fix'.", file.Path)
-				}
+				file = fixedFile
+			} else if fixedFile != file {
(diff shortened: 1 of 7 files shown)


cmd/gazelle/fix-update.go
```diff
@@ -79,53 +79,58 @@ func runFixUpdate(cmd command, args []string) error {
...
-				file = merger.FixFileMinor(c, file)
-				if cmd == fixCmd {
-					file = merger.FixFile(c, file)
-				} else {
-					fixedFile := merger.FixFile(c, file)
-					if fixedFile != file {
-						log.Printf("%s: warning: file contains rules whose structure is out of date. Consider running 'gazelle fix'.", file.Path)
-					}
+			file = merger.FixFileMinor(c, file)
+			if cmd == fixCmd {
```
ianthehat: fixedFile := merger.FixFile(c, file)
if cmd == fixCmd {
    file = fixedFile
} else if fixedFile != file {

jayconrod: Done

PR #70, opened 2018-01-04
common: extract PathHasPrefix and PathTrimPrefix functions

jayconrod:
Fixed a bug in dependency resolution for vendored libraries. Vendor
directories at the repository root were not visible.

Logic for path prefix checking and trimming was replicated in several
places. These are now extracted into functions in a new common package.


jayconrod:
Yeah, common is probably a bad idea. I already thought of another thing I want to stick in here, so I'll make small separate packages instead.

PR #71, opened 2018-01-04
Vendor github.com/pelletier/go-toml at v1.0.1

jayconrod:
When rules_go builds Gazelle at master for use with `go_repository`, it fails because the dependency on `github.com/pelletier/go-toml` is missing. Adding `go-toml` to `go_repository_tools` (which downloads an extracts an additional tarball) seems way too gross, so I'm vendoring it instead.

This is temporary. As soon as we remove `go_repository`, we will be able to switch back to an external repository.

### DIFF
--- a/deps.bzl
+++ b/deps.bzl
@@ -22,11 +22,10 @@ def gazelle_dependencies():
   )
 
   # TODO(jayconrod): restore when gazelle is no longer built for go_repository.
-  # _maybe(go_repository,
-  #     name = "com_github_pelletier_go_toml",
-  #     importpath = "github.com/pelletier/go-toml",
-  #     commit = "16398bac157da96aa88f98a2df640c7f32af1da2", # v1.0.1 as of 2017-12-19
-  # )
+  _maybe(native.local_repository,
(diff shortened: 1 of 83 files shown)


deps.bzl
```diff
@@ -21,11 +21,12 @@ def gazelle_dependencies():
       commit = "f3dd8fd95a7d078cb10fd7fb475b22c3cdbcb307", # 0.2.0 as of 2017-12-04
   )
 
-  _maybe(go_repository,
-      name = "com_github_pelletier_go_toml",
-      importpath = "github.com/pelletier/go-toml",
-      commit = "16398bac157da96aa88f98a2df640c7f32af1da2", # v1.0.1 as of 2017-12-19
-  )
+  # TODO(jayconrod): restore when gazelle is no longer built for go_repository.
```
ianthehat: Seems like the perfect opportunity to try
```
local_repository(
  name = "com_github_pelletier_go_toml",
  path = "endor/github.com/pelletier/go-toml",
)
```

:)

jayconrod: Done. I'm pretty surprised this works.

PR #73, opened 2018-01-04
Check minimum rules_go version when running the "fix" command

### LINKED ISSUE #51: Check minimum rules_go version
With #48, Gazelle generates build files that depend on the `compilers` attribute of `go_proto_library`, which was introduced after 0.8.1. We should check that the minimum version is at least 0.9.0.

This should be checked in the `gazelle` rule (error) and in Gazelle itself (warning).

jayconrod:
* Added a small internal/version package for parsing and comparing
  versions.
* Added repo.FindExternalRepo, which locates the directory of an
  external repository. This will be used for indexing later.
* Added checkRulesGoVersion which parses RULES_GO_VERSION out of
  def.bzl and compares it to a minimum.
* Updated io_bazel_rules_go to master.

Fixes #51


internal/repos/repo.go
```diff
@@ -91,3 +93,28 @@ func generateRepoRule(repo repo) bf.Expr {
...
+// the external repository with the given name. An error is returned if the
+// repository directory cannot be located.
+func FindExternalRepo(repoRoot, name string) (string, error) {
+	// See https://docs.bazel.build/versions/master/output_directories.html
+	// for documentation on Bazel directory layout.
+	// We expect the bazel-out symlink in the workspace root directory to point to
+	// <output-base>/execroot/<workspace-name>/bazel-out
+	// We expect the external repository to be checked out at
+	// <output-base>/external/<name>
+	externalPath := strings.Join([]string{repoRoot, "bazel-out", "..", "..", "..", "external", name}, string(os.PathSeparator))
```
ianthehat: I know some people that modify this name by setting --symlink_prefix in their bazelrc, and also if the build is failing the convenience symlinks might not be made
`bazel info output_base` would be more correct, but has the cost of invoking bazel...

cmd/gazelle/fix-update.go
```diff
@@ -70,7 +70,14 @@ func runFixUpdate(cmd command, args []string) error {
 	if err != nil {
 		return err
 	}
-	c.ShouldFix = cmd == fixCmd
+	if cmd == fixCmd {
+		c.ShouldFix = cmd == fixCmd
+
+		// Only check the version when "fix" is run. Generated build files
```
ianthehat: If the version is wrong, what is the user path to making it right, and can we print an error message that tells them exactly what to do?
Ideally this would be a feature update-repos where it updates the version of rules_go in use as well?

jayconrod: I've added a sentence to the message to update io_bazel_rules_go in WORKSPACE.

`update-repos` should be able to this in the future. It might need a special case, since rules_go isn't a normal Go repository, but the command could be the same as updating any other repository.

internal/repos/repo.go
jayconrod: --symlink_prefix doesn't change the name of bazel-out. This is mentioned in the documentation, so it's intentional, hopefully stable behavior. Annoyingly inconsistent, but useful in this case.

I expect this function may be called frequently (at least once per external repo on each fix/update run), so I'd like to avoid invoking bazel here if at all possible.

PR #74, opened 2018-01-05
Switch to normal vendoring for go-toml

jayconrod:
local_repository paths are relative to the directory containing the
main WORKSPACE file. Gazelle declared a local_repository in the
gazelle_dependencies macro for go-toml with a relative path pointing
to the vendor directory. This works inside the bazel_gazelle
workspace, but it breaks other repositories that import and use the
gazelle_dependencies macro.

### DIFF
--- a/deps.bzl
+++ b/deps.bzl
@@ -22,10 +22,11 @@ def gazelle_dependencies():
   )
 
   # TODO(jayconrod): restore when gazelle is no longer built for go_repository.
-  _maybe(native.local_repository,
-      name = "com_github_pelletier_go_toml", # v1.0.1 as of 2017-12-19
-      path = "vendor/github.com/pelletier/go-toml",
-  )
+  # _maybe(go_repository,
+  #     name = "com_github_pelletier_go_toml",
+  #     importpath = "github.com/pelletier/go-toml",
+  #     commit = "16398bac157da96aa88f98a2df640c7f32af1da2", # v1.0.1 as of 2017-12-19
+  # )
 
 def _maybe(repo_rule, name, **kwargs):
   if name not in native.existing_rules():
--- a/internal/repos/BUILD.bazel
+++ b/internal/repos/BUILD.bazel
@@ -11,8 +11,8 @@ go_library(
     deps = [
         "//internal/resolve:go_default_library",
         "//internal/rules:go_default_library",
+        "//vendor/github.com/pelletier/go-toml:go_default_library",
         "@com_github_bazelbuild_buildtools//build:go_default_library",
-        "@com_github_pelletier_go_toml//:go_default_library",
     ],
 )
 
--- a/vendor/BUILD.bazel
+++ b/vendor/BUILD.bazel
@@ -1,1 +1,0 @@
-# gazelle:exclude github.com/pelletier/go-toml
--- a/vendor/github.com/pelletier/go-toml/BUILD.bazel
+++ b/vendor/github.com/pelletier/go-toml/BUILD.bazel
@@ -1,5 +1,3 @@
-# gazelle:prefix github.com/pelletier/go-toml
-
 # Exclude test files, since they pull in additional dependencies
 # gazelle:exclude benchmark_test.go
 # gazelle:exclude cmd/tomljson/main_test.go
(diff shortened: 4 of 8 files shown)


jayconrod:
cc @katre 

I think it would need to be fixed on the Bazel side. `local_repository` accepts paths, not labels. Actually, the documentation [says it must be an absolute path](https://docs.bazel.build/versions/master/be/workspace.html#local_repository.path), though maybe that's a documentation bug.

Maybe there should be a separate attribute that allows you to point to the repository's WORKSPACE file with a label?

katre:
The documentation is wrong, a relative path to the location of the WORKSPACE is allowed. Please file a separate bug in the bazel repo and I'll address that.

katre:
There are two problems with using labels for local repositories:
* Absolute paths can't be labels, and how do we distinguish?
* The contents of a local repository should only be addressed via the local repository, not as a package in the main repository. Using "//vendor/github.com/pelletier/go-toml" for the local repository is just wrong, that's an invalid label and will cause an error.

What's the actual problem here? go-toml is declaring a local repository, and it has the wrong path?

jayconrod:
@katre I've filed bazelbuild/build#4391 for the documentation bug. Thanks for looking at that.

The problem is this. Let's say I have a repository that depends on Gazelle, so I put this in my WORKSPACE:

```
git_repository(
    name = "bazel_gazelle",
    ...
)

gazelle_dependencies()
```

Gazelle vendors go-toml, so `gazelle_dependencies` is defined like this:

```
def gazelle_dependencies():
  _maybe(native.local_repository,
      name = "com_github_pelletier_go_toml",
      path = "vendor/github.com/pelletier/go-toml",
  )

...
def _maybe(repo_rule, name, **kwargs):
  if name not in native.existing_rules():
    repo_rule(name=name, **kwargs)
```

In my own repository, I try to build Gazelle:

```
$ bazel build @bazel_gazelle//cmd/gazelle
____Loading package: @bazel_gazelle//cmd/gazelle
____Loading package: @io_bazel_rules_go//
____Loading package: @io_bazel_rules_go//go/platform
____Loading package: @com_github_bazelbuild_buildtools//tables
ERROR: /usr/local/google/home/jayconrod/.cache/bazel/_bazel_jayconrod/1df8d8ad4cb35ee83522d81953b4d69f/external/bazel_gazelle/internal/repos/BUILD.bazel:3:1: no such package '@com_github_pelletier_go_toml//': /usr/local/google/home/jayconrod/.cache/bazel/_bazel_jayconrod/1df8d8ad4cb35ee83522d81953b4d69f/external/com_github_pelletier_go_toml must be an existing directory and referenced by '@bazel_gazelle//internal/repos:go_default_library'.
ERROR: Analysis of target '@bazel_gazelle//cmd/gazelle:gazelle' failed; build aborted: no such package '@com_github_pelletier_go_toml//': /usr/local/google/home/jayconrod/.cache/bazel/_bazel_jayconrod/1df8d8ad4cb35ee83522d81953b4d69f/external/com_github_pelletier_go_toml must be an existing directory.
____Elapsed time: 1.366s

$ ls -l /usr/local/google/home/jayconrod/.cache/bazel/_bazel_jayconrod/1df8d8ad4cb35ee83522d81953b4d69f/external/com_github_pelletier_go_toml
lrwxrwxrwx 1 jayconrod eng 81 Jan  5 14:32 /usr/local/google/home/jayconrod/.cache/bazel/_bazel_jayconrod/1df8d8ad4cb35ee83522d81953b4d69f/external/com_github_pelletier_go_toml -> /usr/local/google/home/jayconrod/Code/scratch/vendor/github.com/pelletier/go-toml
```

The directory that the `com_github_pelletier_go_toml` symlink points to, `/usr/local/google/home/jayconrod/Code/scratch/vendor/github.com/pelletier/go-toml`, does not exist. That's a directory in my own repository, not a directory in `@bazel_gazelle`, which is what I want.

katre:
I believe there's a way to get the current path to your deps.bzl, and use that in the _maybe call:
```
def gazelle_dependencies():
  _maybe(native.local_repository,
      name = "com_github_pelletier_go_toml",
      path = __local_path__ + "vendor/github.com/pelletier/go-toml",
  )
```

That way, the path is correct both directly, and when included as a remote repository.

@dslomov Do you know how to do that?

PR #78, opened 2018-01-05
Disable proto rule generation by default in vendor

### LINKED ISSUE #77: Disable proto rule generation by default in vendor
Original issue: bazelbuild/rules_go#1219, bazelbuild/rules_go#1022

Vendored .proto files rarely compile without modification in Bazel. If a vendored proto A imports another vendored proto B, the import path in A must be adjusted manually to include a "vendor" prefix. Bazel `proto_library` assumes all proto import paths are relative to the root of a repository. There's currently no way to adjust import paths via `proto_library`.

Since working `proto_library` rules can't be generated for vendored protos in most cases, Gazelle should not attempt to do so by default. We should automatically disable proto rule generation when entering a vendor directory.

Proto rule generation can be explicitly turned on by adding a directive like the one below to a build file. The directive applies to the build file's package and subdirectories.

```
# gazelle:proto default
```

jayconrod:
Fixes #77

### DIFF
--- a/internal/config/directives.go
+++ b/internal/config/directives.go
@@ -17,6 +17,7 @@ package config
 
 import (
 	"log"
+	"path"
 	"regexp"
 	"strings"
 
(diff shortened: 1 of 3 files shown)


jayconrod:
cc @mattmoor

PR #85, opened 2018-01-09
internal/repo: use declared repository names in external resolver

### LINKED ISSUE #13: Associate default prefixes with external repositories
(Migrated from bazelbuild/rules_go#606)

We should associate prefixes with external repository names. These prefixes would be used transform import paths to Bazel labels for repositories that haven't been fetched and can't be indexed yet (see #12).

In the examples below, we would transform dependencies on `github.com/example/project` to `@repo_with_custom_name//:go_default_library`.

For `go_repository`, we can infer this from `importpath` or `remote`.

```
go_repository(
    name = "repo_with_custom_name",
    importpath = "github.com/example/project",
)
```

We may be able to do this for `git_repository`, too, in some cases.

```
git_repository(
    name = "repo_with_custom_name",
    remote = "https://github.com/example/project",
)
```

This wouldn't work with import path redirection. In particular, it wouldn't work with golang.org, gopkg.in, any fork, or private repositories. Gazelle should recognize directives in WORKSPACE to override the import path.

```
# gazelle:importpath github.com/example/project
git_repository(
    name = "repo_with_custom_name",
    remote = "https://github.com/fork/project",
)
```

If Gazelle can't discover the repository by parsing WORKSPACE (i.e., if a repository is declared in a macro), we should be able to read standalone declarations that map the prefix to the repo name.

```
# gazelle:importpath repo_with_custom_name github.com/example/project
```

jayconrod:
Gazelle now reads go_repository rules from WORKSPACE when starting
up. When externalResolver encounters an import path that is covered by
one of the declared external repositories, it will emit a label using
that repository's name. This allows custom repository names.

Fixes #13

### DIFF
--- a/internal/label/label.go
+++ b/internal/label/label.go
@@ -31,6 +31,10 @@ type Label struct {
 	Relative        bool
 }
 
+func New(repo, pkg, name string) Label {
+	return Label{Repo: repo, Pkg: pkg, Name: name}
+}
+
 // NoLabel is the nil value of Label. It is not a valid label and may be
 // returned when an error occurs.
 var NoLabel = Label{}
@@ -41,9 +45,9 @@ var (
 	labelNameRegexp = regexp.MustCompile(`^[A-Za-z0-9_/.+=,@~-]*$`)
 )
 
-// ParseLabel reads a label from a string.
+// Parse reads a label from a string.
 // See https://docs.bazel.build/versions/master/build-ref.html#lexi.
-func ParseLabel(s string) (Label, error) {
+func Parse(s string) (Label, error) {
 	origStr := s
 
 	relative := true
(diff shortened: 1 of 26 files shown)


internal/label/label.go
```diff
@@ -13,7 +13,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package resolve
+package label
```
ianthehat: This make the typename label.Lablel
The traditional approach here is to rename the type to a single letter (label.L)

Also label.ParseLabel should just be label.Parse

internal/label/label.go
```diff
@@ -130,9 +130,26 @@ func (l Label) Equal(other Label) bool {
 		l.Relative == other.Relative
 }
 
-func packageContains(repo, pkg string, label Label) bool {
+// PackageContains tests whether a label is contained within a given repository
+// and package. The label must not be relative.
+func PackageContains(repo, pkg string, label Label) bool {
```
ianthehat: I don't like this interface, seems ugly
I think I would have
```
func (l L) Package() {
    return L{
        Repo:     repo,
        Pkg:      pkg,
        Relative: relative,
    }
}
```

and then use
```
mylabel.Package().Equal(label.L{Repo:repo, Pkg: pkg})
```
and ideally the arg to equal would already be a label anyway

internal/label/label.go
jayconrod: @alandonovan Can you settle a readability question? Does a single letter type name (L) make sense when the type would otherwise have the same name as the package (label.Label)? This seems overly brief to me, but Ian points out this policy works well for testing.T and should have been applied to context.Context. WDYT?

alandonovan: There are plenty of precedents for "single-type packages" whose name is the same as their sole type (html/template, math/rand, context, etc).  I think label.Label is fine, as is template.Template, though perhaps context.Context should have followed testing.T if only for the frequency with which users have to write this name.

jayconrod: Decided to keep `label.Label`. That feels more readable to me.

Renamed `label.ParseLabel` to `Parse` and added `New` constructor.

internal/label/label.go
jayconrod: Changed this to `Contains` method.

Issue #99, opened 2018-01-24
Add ability to keep custom binary names


jmmv:
Because of https://github.com/bazelbuild/rules_go/issues/1239, I'm trying to find a way to get my previously-working build to work again. The problem is that I have a bunch of tools (not tests!) that depend on `go_binary` targets as data and need to find the binaries to use them. I also have user-facing instructions that tell people to run `./bazel-bin/foo/bar`, which don't work any more and cannot be written in a generic manner due to the unpredictable platform-specific component.

For the most part, I got a workaround by using a `genrule` that just copies the binary to the expected location, thus eliminating the platform-specific part of the path. These look like:

```python
def stage_go_binary(name, src, visibility=None):
  native.genrule(
      name = "stage-%s" % name,
      srcs = [src],
      outs = [name],
      cmd = "cp $(location %s) $@" % src,
      output_to_bindir = 1,
      visibility = visibility,
  )

def staged_go_binary(name, visibility=None, **kwargs):
  go_binary(
      name = "%s-bin" % name,
      **kwargs
  )
  stage_go_binary(
      name = name,
      src = ":%s-bin" % name,
      visibility = visibility,
  )
```

so that, instead of calling `go_binary` from my build files I call `staged_go_binary` that does the magic thing in the background. However, as you can imagine this confuses Gazelle as it does not know that `staged_go_binary` is equivalent to `go_binary` and thus Gazelle reverts all my changes. Fine.

So I went with the simpler alternative of not replacing the binary targets, but instead doing this:

```python
go_binary(
    name = "foo-bin",
    ...
)

stage_go_binary(
    name = "foo",
    src = ":foo-bin",
)
```

in the hopes that this extra amount of manual work would keep Gazelle happy... but it doesn't either because Gazelle cannot recognize that my `foo-bin` is really the `foo` it wants to generate.

All these hacks are awful, but I don't see other solutions based on the original issue quoted above. Therefore, I was wondering if Gazelle could be extended to have an extra tag to tell it that a specific target was renamed. E.g.:

```python
go_binary(
    name = "foo-bin",  # gazelle:name foo
    ...
)
```

This feels like a feature that would be useful in other situations as well. Thoughts?

jayconrod:
This seems like a good temporary measure, but I'm worried it will interfere with some planned changes.

[...]

PR #108, opened 2018-01-30
internal/merger: more flexible rule matching

### LINKED ISSUE #17: Detect and update rules in non-standard layouts
(Migrated from bazelbuild/rules_go#746)

We have examples in this repository that are in non standard directory layouts (see examples/cgo/sub).
It would be good if gazelle automatically spotted these and knew what to do, at the moment you have to add ignore directives, and gazelle will not update the rule for you.

### LINKED ISSUE #106: It needs to be possible to rename a go_binary rule
At the moment gazelle does not work correctly with go_binary rules that are not the same name as the folder.
While that should be the default when generating new rules, if there is an existing go_binary rule then gazelle should probably presume it is maintaining that one rather than adding a new one. If there is more than one, it should probably just give up and print a warning?

jayconrod:
Previously, Gazelle would merge a generated rule with an existing rule
if the existing rule had the same name and kind. This change makes the
matching algorithm more flexible.

* The kind must always match. This may be relaxed in the future.
* If the name doesn't match any existing rule, Gazelle will consider
  other attributes based on the kind. For example, when matching
  go_library, Gazelle will look for matching importpath attributes.
* go_binary rules will match any go_binary rules in the same
  build file.

Fixes #106
Related #99
Fixes #17

### DIFF
--- a/internal/merger/fix.go
+++ b/internal/merger/fix.go
@@ -293,15 +293,7 @@ func removeLegacyProto(c *config.Config, oldFile *bf.File) *bf.File {
 		sort.Ints(deletedIndices)
 	}
 	fixedFile := *oldFile
-	fixedFile.Stmt = make([]bf.Expr, 0, len(oldFile.Stmt)-len(deletedIndices))
-	di := 0
-	for i, stmt := range oldFile.Stmt {
-		if di < len(deletedIndices) && i == deletedIndices[di] {
-			di++
-			continue
-		}
-		fixedFile.Stmt = append(fixedFile.Stmt, stmt)
-	}
+	fixedFile.Stmt = deleteIndices(oldFile.Stmt, deletedIndices)
 	return &fixedFile
 }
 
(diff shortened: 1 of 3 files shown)


jayconrod:
Just realized there is a problem here. When a library is renamed, binaries and tests that reference it via `embed` won't be updated. Same deal with `proto`. The merger will need to update labels in generated rules in these attributes.

jayconrod:
@ianthehat Some large additional changes, PTAL.

I've added functionality to replace labels in generated rules that refer to other generated rules. For example, if a `go_library` has a custom name, the `embed` attribute of the internal `go_test` will be updated to use the new label before merging.

This continues the general pattern of copying and updating rules, rather than mutating them in place. I'm concerned that mutating in place may be a source of bugs since rules may be aliased in a few places, and we'll likely want to do this concurrently in the future. However, this is getting very clunky; the buildifier library makes mutation easy and copying hard. Might want to rewrite this in the future. Let me know if you have any thoughts.

I've changes the matching logic to ignore `proto` and `embed` attributes, since they can change between matching and merging. This means `go_test` rules cannot have custom names, since there's nothing to match. I may need to add a directive for this, as @jmmv originally suggested in #99. Although once Gazelle squashes internal / external tests, maybe I could just match whatever `go_test` is present, same as `go_binary`.

ianthehat:
Yes, one go_test rule per package is the goal I think.

PR #110, opened 2018-01-31
run gazelle in pure mode

shariat:
I don't see any reason not to. I do it manually by passing `--features` to `bazel run`, since my `libstdc++` is broken.

### DIFF
--- a/cmd/gazelle/BUILD.bazel
+++ b/cmd/gazelle/BUILD.bazel
@@ -31,6 +31,12 @@ go_library(
 go_binary(
     name = "gazelle",
     embed = [":go_default_library"],
+    visibility = ["//visibility:public"],
+)
+
+go_binary(
+    name = "gazelle_pure",
+    embed = [":go_default_library"],
     pure = "on",
     visibility = ["//visibility:public"],
 )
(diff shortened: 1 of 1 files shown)


jayconrod:
I think this would require an extra copy of the standard library to be built in workspaces that don't have `pure` binaries (@ianthehat, could you confirm that?). It might be better to have two `go_binary` rules, one pure, and one not.

@shariat In your situation, do you need to build all `go_binary` rules in pure mode? If so, a better approach may be to add `--features=pure` to a bazelrc file so you don't have to type it out. There are [several places](https://docs.bazel.build/versions/master/user-manual.html#bazelrc) you can put this, but a `tools/bazel.rc` file checked into your repo would probably work best (bazel-gazelle and rules_go both have a file like this for CI configuration).

shariat:
@jayconrod I do have `--features=pure` in `bazelrc` and it works. just wanted to say maybe we should run in pure mode.

jayconrod:
I added a new `go_binary` rule, `//cmd/gazelle:gazelle_pure`, which is the same thing but in pure mode. You can set the `gazelle` rule to use this:

```
gazelle(
    name = "gazelle",
    gazelle = "@bazel_gazelle//cmd/gazelle:gazelle_pure",
)
```

shariat:
great, I just wanted to ask that

PR #111, opened 2018-02-01
Address some minor warning thrown by linters.

Helcaraxan:
- 1 missing format argument in test error message
- 1 removal of variable shadowing
- 2 loop range simplifications


Helcaraxan:
Just came across these things will having the project open in my IDE and wanted to suggest the necessary (small) modifications to address them.

PR #113, opened 2018-02-02
cmd/move_labels: add tool for updating labels after move

### LINKED ISSUE #63: Gazelle should fix labels in rules in Bazel-native projects imported through Go-native vendoring
As Bazel is starting to get more popular, we're running into problems with Bazel-native Go projects that we import through Go-native vendoring (i.e. `dep` + the `vendor/` directory).

An explicit example: https://github.com/kubernetes/repo-infra/pull/54 imports `github.com/bazelbuild/buildtools`. The root `BUILD.bazel` file in `bazelbuild/buildtools` has a `test_suite` which makes sense in its repo:
```python
test_suite(
    name = "tests",
    tests = [
        "//api_proto:api.gen.pb.go_checkshtest",
        "//build:go_default_test",
        "//build:parse.y.go_checkshtest",
        "//build_proto:build.gen.pb.go_checkshtest",
        "//deps_proto:deps.gen.pb.go_checkshtest",
        "//edit:go_default_test",
        "//extra_actions_base_proto:extra_actions_base.gen.pb.go_checkshtest",
        "//lang:tables.gen.go_checkshtest",
        "//tables:go_default_test",
        "//wspace:go_default_test",
    ],
)
```

but when imported into our repo, the labels are all wrong; they should instead be something like
```python
test_suite(
    name = "tests",
    tests = [
        "//vendor/github.com/bazelbuild/buildtools/api_proto:api.gen.pb.go_checkshtest",
        "//vendor/github.com/bazelbuild/buildtools/build:go_default_test",
        "//vendor/github.com/bazelbuild/buildtools/build:parse.y.go_checkshtest",
        "//vendor/github.com/bazelbuild/buildtools/build_proto:build.gen.pb.go_checkshtest",
        "//vendor/github.com/bazelbuild/buildtools/deps_proto:deps.gen.pb.go_checkshtest",
        "//vendor/github.com/bazelbuild/buildtools/edit:go_default_test",
        "//vendor/github.com/bazelbuild/buildtools/extra_actions_base_proto:extra_actions_base.gen.pb.go_checkshtest",
        "//vendor/github.com/bazelbuild/buildtools/lang:tables.gen.go_checkshtest",
        "//vendor/github.com/bazelbuild/buildtools/tables:go_default_test",
        "//vendor/github.com/bazelbuild/buildtools/wspace:go_default_test",
    ],
)
```

Gazelle will already fix the go rules (I believe), but other things (like this `test_suite`) remain broken.

An (untested) naive idea is to have Gazelle or some other tool examine all labels in build rules under `vendor/`, adding the necessary path prefix if it's missing. I'm not sure how safe this is, and whether it'd work 100%, though.

jayconrod:
This is a tool to update labels in a tree of build files after it's
been moved from another place. This is especially useful for vendoring
repositories that already have build files.

Fixes #63

### DIFF
--- a/cmd/move_labels/BUILD.bazel
+++ b/cmd/move_labels/BUILD.bazel
@@ -1,0 +1,25 @@
+load("@io_bazel_rules_go//go:def.bzl", "go_binary", "go_library", "go_test")
+
+go_library(
+    name = "go_default_library",
+    srcs = ["move_labels.go"],
+    importpath = "github.com/bazelbuild/bazel-gazelle/cmd/move_labels",
+    visibility = ["//visibility:private"],
+    deps = [
+        "//internal/label:go_default_library",
+        "//internal/pathtools:go_default_library",
+        "@com_github_bazelbuild_buildtools//build:go_default_library",
+    ],
+)
+
+go_binary(
+    name = "move_labels",
+    embed = [":go_default_library"],
+    visibility = ["//visibility:public"],
+)
+
+go_test(
+    name = "go_default_test",
+    srcs = ["move_labels_test.go"],
+    embed = [":go_default_library"],
+)
--- a/cmd/move_labels/move_labels.go
+++ b/cmd/move_labels/move_labels.go
@@ -25,6 +25,7 @@ import (
 	"os"
 	"path"
 	"path/filepath"
+	"regexp"
 	"strings"
 
 	"github.com/bazelbuild/bazel-gazelle/internal/label"
@@ -118,7 +119,12 @@ func moveLabelsInFile(file *build.File, from, to string) {
 			return nil
 		}
 		label := str.Value
-		moved := moveLabel(from, to, label)
+		var moved string
+		if strings.Contains(label, "$(location") {
+			moved = moveLocations(from, to, label)
+		} else {
+			moved = moveLabel(from, to, label)
+		}
 		if moved == label {
 			return nil
 		}
@@ -138,6 +144,25 @@ func moveLabel(from, to, str string) string {
 	}
 	l.Pkg = path.Join(to, pathtools.TrimPrefix(l.Pkg, from))
 	return l.String()
+}
+
+var locationsRegexp = regexp.MustCompile(`\$\(locations?\s*([^)]*)\)`)
+
+// moveLocations fixes labels within $(location) and $(locations) expansions.
+func moveLocations(from, to, str string) string {
+	matches := locationsRegexp.FindAllStringSubmatchIndex(str, -1)
+	buf := new(bytes.Buffer)
+	pos := 0
+	for _, match := range matches {
+		buf.WriteString(str[pos:match[2]])
+		label := str[match[2]:match[3]]
+		moved := moveLabel(from, to, label)
+		buf.WriteString(moved)
+		buf.WriteString(str[match[3]:match[1]])
+		pos = match[1]
+	}
+	buf.WriteString(str[pos:])
+	return buf.String()
 }
 
 func isBuiltinLabel(label string) bool {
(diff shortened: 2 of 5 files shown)


jayconrod:
@ixdy I wrote a new tool that should help with vendoring. I stepped through the workflow you described in #63 (vendoring github.com/bazelbuild/buildtools in k8s.io/repo-infra). Seems to work okay. Let me know what you think.

ixdy:
Luck @thockin ran into a similar issue on https://github.com/kubernetes/kubernetes/pull/57600, so I attempted to see if this would fix things over there (by running `move_labels -to vendor/github.com/bazelbuild/bazel-gazelle/`).

A few problems I've run into:

1. Running `move_labels` after `gazelle` can yield the wrong results, e.g. it's causing diffs like
```diff
-        "//vendor/github.com/bazelbuild/buildtools/build:go_default_library",
+        "//vendor/github.com/bazelbuild/bazel-gazelle/vendor/github.com/bazelbuild/buildtools/build:go_default_library
```
Running `gazelle` again fixes that, at least.

2. The `$(location)` functions in genrules aren't fixed, so this rule is still invalid:
```python
genrule(
    name = "std_package_list",
    srcs = ["@go_sdk//:packages.txt"],
    outs = ["std_package_list.go"],
    cmd = "$(location //resolve/internal/gen_std_package_list) $< $@",
    tools = ["//vendor/github.com/bazelbuild/bazel-gazelle/resolve/internal/gen_std_package_list"],
)
```

3. After fixing that rule manually, things still don't work; now the issue is that the `vendor/github.com/bazelbuild/bazel-gazelle/resolve/internal` package is missing, probably because Godep detected we didn't need it. (Not sure if `dep` is smarter here.)

4. Also, the `WORKSPACE` file at `vendor/github.com/bazelbuild/bazel-gazelle` is missing, which makes figuring out which packages need to be fixed trickier. (This is probably also a Godep-related issue.)

jayconrod:
> 1. Running move_labels after gazelle can yield the wrong results

I think you will need to run this tool first. It assumes the labels are based in the old location. Gazelle will scramble that assumption for *some* labels but not others.

> 2. The $(location) functions in genrules aren't fixed

I just pushed a commit to fix this. It will detect `$(location <label>)` and `$(locations <label>)` within strings using a regexp, then just replace the labels.

> 3. vendor/github.com/bazelbuild/bazel-gazelle/resolve/internal package is missing

This might be from an old version of Gazelle. I wonder if Godep is getting confused since the packages have moved? I put most of the libraries in internal/. resolve/internal became internal/resolve. Regardless, if something is missing, it sounds like a vendoring tool issue.

> 4. Also, the WORKSPACE file at vendor/github.com/bazelbuild/bazel-gazelle is missing

Also probably vendoring tool issue. Not sure why that file would be dropped.

ixdy:
I discovered a similar issue of a missing `WORKSPACE` file in https://github.com/kubernetes/test-infra/pull/4859, which is using dep. It seems like despite having a constraint on `github.com/bazelbuild/buildtools`, after pruning only the subpackages `build` and `tables` are needed, so everything else is removed (though the `LICENSE` file is left intact).

ixdy:
Regarding the missing `WORKSPACE` file and trying to determine the workspace root - it might be possible to walk `dep`'s `Gopkg.lock`, looking at each of the projects there and re-rooting based on the name of each project. (I'm not sure where this logic should live.)

`Godeps.json` appears to just list importpaths, rather than projects + packages, so this strategy won't work there, but we plan to eventually move to `dep` anyway for kubernetes.

ixdy:
OTOH, this might just be kicking the can down the road without fixing the real issue. Eventually there will probably be some bazelified project with real non-Go dependencies that `dep` will mishandle.

Do we require using a native bazel dependency then? Require that that project is imported with `dep` with no pruning? Something else?

jayconrod:
Checking this in as-is. It's somewhat useful now, and we can iterate on it more over time.

PR #119, opened 2018-02-06
internal/overlay_repository.bzl: add experimental overlay rules

jayconrod:
These are drop-in replacements for http_archive and git_repository
with the added feature that you can copy in an "overlay" set of files
using custom names. The overlay is specified as a dictionary mapping
labels to relative paths in the destination repository.

In the future, these rules will be an alternative to
go_repository. Build files may be pre-generated using Gazelle, and
then copied into repositories using an overlay.

For now, these rules are not exposed in def.bzl and are not public yet.

### DIFF
--- a/internal/overlay_repository.bzl
+++ b/internal/overlay_repository.bzl
@@ -1,0 +1,79 @@
+# Copyright 2018 The Bazel Authors. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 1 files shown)


internal/overlay_repository.bzl
```diff
@@ -0,0 +1,79 @@
...
+      type = ctx.attr.type,
+      stripPrefix = ctx.attr.strip_prefix,
+  )
+
+  _apply_overlay(ctx, ctx.attr.overlay)
+
+http_archive = repository_rule(
+    implementation = _http_archive_impl,
+    attrs = {
+        "url": attr.string(),
```
ianthehat: I would drop url, it's been deprectated an awfully long time, no need to be backwards compatible.

PR #128, opened 2018-02-13
Use new overlay repository rules for all dependencies

jayconrod:
* gazelle_dependencies now uses git_repository and http_archive with
  overlay build files checked into third_party.
* These rules are now exposed in def.bzl
* go-toml is once again an external dependency. The vendor directory
  has been deleted.


jayconrod:
Submitting this without review because it's conceptually the same as bazelbuild/rules_go#1324 and a couple other PRs are blocked on this. Let me know if I should change anything though.

PR #129, opened 2018-02-13
Copy go_repository from rules_go to Gazelle

jayconrod:
Gazelle will eventually be the permanent home for this rule. It needs
to be in both places for a while though.

* Copied rules_go fetch_repo to cmd/gazelle
* Copied rules_go repository.bzl and go_repository.bzl into
  internal/go_repository.bzl. Note that go_repository_tools is
  significantly simplified. go_repository is pretty much the same.
* Declared go_repository_tools in gazelle_dependencies.
* Exposed go_repository in def.bzl.


jayconrod:
@ianthehat PTAL. I've added a small `bazel_test` for `go_repository`.

internal/go_repository.bzl
```diff
@@ -0,0 +1,236 @@
...
+            values = [
+                "",
+                "default",
+                "disable",
+                "legacy",
+            ],
+        ),
+    },
+)
+"""See go/workspace.rst#go-repository for full documentation."""
```
ianthehat: Probably a lie?

internal/go_repository.bzl
```diff
@@ -0,0 +1,236 @@
...
+        "PATH": ctx.os.environ["PATH"],  # to find git
+    }
+    if "SSH_AUTH_SOCK" in ctx.os.environ:
+      fetch_repo_env["SSH_AUTH_SOCK"] = ctx.os.environ["SSH_AUTH_SOCK"]
+    if "HTTP_PROXY" in ctx.os.environ:
+      fetch_repo_env["HTTP_PROXY"] = ctx.os.environ["HTTP_PROXY"]
+    if "HTTPS_PROXY" in ctx.os.environ:
+      fetch_repo_env["HTTPS_PROXY"] = ctx.os.environ["HTTPS_PROXY"]
+
+    # TODO(yugui): support submodule?
```
ianthehat: We are never going to do this...

jayconrod: Removed

internal/go_repository.bzl
jayconrod: I just created a new placeholder file to link to. I'll fill it in with documentation about all three repository rules in a follow-up PR.

PR #151, opened 2018-03-05
Refactor FixFile, FixLoads, MergeFile in imperative style

### LINKED ISSUE #58: Fix command should explain what's wrong
Migrated from bazelbuild/rules_go#792

When a BUILD file needs to be fixed, `gazelle fix` just emits a warning without explaining what's wrong with it. This could be something as trivial as loaded symbols being out of order.

Our warning should explain what needs to be fixed. We should not emit a warning for trivial problems that won't interfere with later operations (loaded symbol ordering).

### LINKED ISSUE #150: Modify AST in place instead of making copies
The merger package returns copies of AST structures instead of mutating inputs. This is nice from a purity / functional programming perspective, but it results in lots of grungy, error-prone copying code in fix.go and merger.go. We should switch to a more imperative style, modifying the AST in place.

jayconrod:
* FixFile, FixLoads, and MergeFile now modify a file in place instead of returning
  a new file. This simplifies the implementation of current and future
  transformations, since it avoids the need to copy objects which
  aren't easily copied.
* FixFile is consolidated with FixFileMinor. Risky transformations are
  only performed if c.ShouldFix is true. Warnings are printed
  otherwise.
* Internal functions like squashExpr, mergeExpr are still in functional style. This
  may change in the future if they need significant modifications.

Fixes #150
Fixes #58


jayconrod:
Note for review: I moved some of the fix functions around after refactoring them. It's in a separate commit though, hopefully that helps.

PR #154, opened 2018-03-12
Replace loads of go_repository from rules_go

jayconrod:
* gazelle update-repos and gazelle fix will now change a load of
  go_repository from @io_bazel_rules_go to @bazel_gazelle.
* If a new load needs to be inserted, it will be after any calls to
  go_rules_dependencies, go_register_toolchains, and
  gazelle_dependencies.


jayconrod:
I've been thinking the same thing. The abstractions in the build package are usually not what I want, and they're really inefficient. I'm thinking about building my own layer on top of the AST, but it may take some time.

PR #160, opened 2018-03-19
Resolve Go WKT imports to rules_go

### LINKED ISSUE #159: Map Go imports of WKTs to rules_go
Originally reported in bazelbuild/rules_go#1386.

When a WKT is imported in a .proto file, we don't generate a dependency in `go_proto_library`, since the compiler is expected to provide those. The official WKTs are defined in `@io_bazel_rules_go//proto/wkt`.

When a WKT is imported in a .go file, we treat it as a normal Go import. For example, see https://github.com/GoogleCloudPlatform/google-cloud-go/blob/master/internal/protostruct/protostruct.go#L19. `github.com/golang/protobuf/ptypes/struct` will be resolved as `@com_github_golang_protobuf//ptypes/struct:go_default_library`. We should resolve it as `@io_bazel_rules_go//proto/wkt:struct_go_proto` instead.

jayconrod:
Gazelle will now resolve imports of
github.com/golang/protobuf/ptypes/X to
@io_bazel_rules_go//proto/wkt:x_go_proto.

Fixes #159, bazelbuild/rules_go#1386

### DIFF
--- a/internal/config/constants.go
+++ b/internal/config/constants.go
@@ -42,12 +42,14 @@ const (
 	// WellKnownTypesProtoRepo is the repository containing proto_library rules
 	// for the Well Known Types.
 	WellKnownTypesProtoRepo = "com_google_protobuf"
-	// WellKnownTypesGoProtoRepo is the repository containing go_library rules
-	// for the Well Known Types.
-	WellKnownTypesGoProtoRepo = "com_github_golang_protobuf"
+	// WellKnownTypeProtoPrefix is the proto import path prefix for the
+	// Well Known Types.
+	WellKnownTypesProtoPrefix = "google/protobuf"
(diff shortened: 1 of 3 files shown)


swsnider:
This breaks compatibility with rules_go 0.10.1, it seems -- proto/wkt doesn't exist in the 0.10.1 tag.

jayconrod:
Gazelle 0.11.0 will require rules_go 0.11.0 since it will add `importmap` attributes in vendored `go_proto_library` rules. That feature was also added after the 0.10 branch was cut.

I plan to cut both releases early next week.

swsnider:
Yeah, I get that, it just means that `go get github.com/bazelbuild/bazel-gazelle` breaks now :(

PR #166, opened 2018-03-23
Evaluate symlinks in repo root

jayconrod:
Gazelle usually ignores symlinks, but it may follow them if they point
to directories outside of the repo root (after all symlinks are
resolved).

With this change, Gazelle evaluates symlinks in the repo root. Without
this, Gazelle incorrectly decides to follow symlinks.


jayconrod:
cc @ixdy this should fix the bug you mentioned in kubernetes/kubernetes#61575

cc @yasushi-saito I don't think this will break anything for you, but please take a look just in case.

ixdy:
thanks!

Issue #136, opened 2018-02-21
Documentation on how to add a new language to Gazelle


achew22:
Would you be willing to put together a little document on all the places that one would need to modify in order to get Gazelle to output build files for another language?

jayconrod:
[Design.rst](https://github.com/bazelbuild/bazel-gazelle/blob/master/Design.rst) is the closest thing I have to this right now.

What languages are you thinking about?

Ideally, Gazelle can be restructured a bit so additional languages can be plugged in without too much hassle. I'm happy to describe what would need to be changed to add a new language today, but it would be better to figure out a framework that will work for lots of languages in the future.

achew22:
I was thinking about adding support for `rules_typescript`.

> Ideally, Gazelle can be restructured a bit so additional languages can be plugged in without too much hassle.

Yes, that would be ideal. It seems that the majority of the code in Gazelle would be reusable, I would just need to add a new indexer that can understand the rules_typescript naming convention and write a scanner that is able to read import statements from `.ts` files. I'm also willing to accept a subset of typescript as the target language just to get things off the ground in the first iteration.

Do you have any plans to add support for Gazelle as a framework?

achew22:
Friendliest of pings

jayconrod:
Sorry for the delay. I haven't forgotten, but perf took a long time. I'll write this up this week.

achew22:
Perf is also killing me. Could you please not do it till next week? No need to stress

PR #220, opened 2018-05-31
Refactor configuration to support multiple languages

jayconrod:
This change makes Gazelle's configuration extensible, in preparation
for supporting multiple languages.

* Config now has an Exts field, which maps from extension names
  (usually language names) to extension values (interface{}).
* Added a Configurer interface, which allows implementors to register
  and check command-line flags and handle directives. Each language
  will implement this (as well as other modules that handle
  directives, like Walk).
* config.CommonConfigurer is an implementation of Configurer for
  language-agnostic functionality in config. Currently, it just
  handles -build_file_name.
* fix-update and Walk now use Configurer.
* Several types and constants are moved from config to rule to break
  circular dependencies, including Directive and Platform. rule no
  longer depends on config.
* label.Labeler is moved to a new package, Labeler in order to break
  circular dependencies. Labeler will be deleted later.

Related #136

### DIFF
--- a/cmd/gazelle/BUILD.bazel
+++ b/cmd/gazelle/BUILD.bazel
@@ -18,6 +18,7 @@ go_library(
         "//internal/config:go_default_library",
         "//internal/generator:go_default_library",
         "//internal/label:go_default_library",
+        "//internal/labeler:go_default_library",
         "//internal/merger:go_default_library",
         "//internal/packages:go_default_library",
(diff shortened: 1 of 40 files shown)


achew22:
My local copy of this rests on my work desktop. I'm really excited for this. 

cc @mrmeku since he is a pretty cool guy. eh does typescript + Bazel and doesn't afraid of anything

PR #235, opened 2018-06-21
add vendor_go_repository workspace rule

bennyscetbun:
### **Problem:**
we are using vendors and also external go_repositories. This make issue, due to duplication, when a vendor is also an external (_//vendor/foo_ and _@foo//_).

### **Solution:**
Load the vendors like _local_repository_ and make them external.
To do so we've added a new rule to load a _go_repository_ but from the local vendors.
_vendor_go_repository_

Any ideas and comments are welcomed

### DIFF
--- a/def.bzl
+++ b/def.bzl
@@ -14,6 +14,7 @@
 
 load("@bazel_skylib//:lib.bzl", "shell")
 load("//internal:go_repository.bzl", "go_repository")
+load("//internal:vendor_go_repository.bzl", "vendor_go_repository")
 load("//internal:overlay_repository.bzl", "git_repository", "http_archive")
 
 def _gazelle_runner_impl(ctx):
--- a/deps.bzl
+++ b/deps.bzl
@@ -18,6 +18,10 @@ load(
     _go_repository_tools = "go_repository_tools",
 )
 load(
+    "@bazel_gazelle//internal:vendor_go_repository.bzl",
+    "vendor_go_repository",
+)
+load(
     "@bazel_gazelle//internal:overlay_repository.bzl",
     "git_repository",
     "http_archive",
--- a/internal/vendor_go_repository.bzl
+++ b/internal/vendor_go_repository.bzl
@@ -32,7 +32,7 @@ def _vendor_go_repository_impl(ctx):
         "PATH": ctx.os.environ["PATH"],
     }
     for file in __join_path(ctx.path(Label("@//:WORKSPACE")).dirname, "vendor/"+ctx.attr.importpath).readdir():
-        result = env_execute(ctx, ["cp", "-r", file, ctx.path(file.basename)], environment = fetch_repo_env, timeout = _GO_REPOSITORY_TIMEOUT)
+        result = env_execute(ctx, ["cp", "-RP", file, ctx.path(file.basename)], environment = fetch_repo_env, timeout = _GO_REPOSITORY_TIMEOUT)
         if result.return_code:
             fail("failed to fetch %s: %s" % (ctx.name, result.stderr))
     generate = ctx.attr.build_file_generation == "on"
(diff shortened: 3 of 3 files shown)


jayconrod:
I think this rule is too specialized to include with Gazelle. I think the file copying and label resolution may also be more expensive than necessary.

Could you elaborate on the problem you're trying to solve? If you have a package checked into vendor/, by default, Gazelle will resolve dependencies to the vendored package instead of any external repository. Imports that are not provided by any known `go_library` will be resolved to external repositories. It's still possible to get multiple copies of a package, but the `importmap` is set automatically for vendored packages, and that should resolve conflicts. This is roughly what go build does.

The main complaint I've heard about the way Gazelle and rules_go handle vendoring is the fact that build files need to be added to the vendor directory, and tools like dep remove them automatically, blowing away any manual changes. I've looked at some solutions for that, but I think most solutions are worse than the problem.

bennyscetbun:
hi jayconrod,
I can see your concern.
But we had/got issue with grpc for example. We ve got it in our vendor and rules_go also load it as external, this make bazel fail due to a miss match versions.

We are thinking of a good way to solve our vendor problem with gazelle.
Quick questions:
What would you think of an option for gazelle that will make it generate WORKSPACE files in vendor directories based on Gopkg.lock and load the vendors with local_repository as externals?

jayconrod:
I've thought about this a bit. I think it can work, but I think it's too complicated to be the common mode of operation for Gazelle. Supporting it through an opt-in feature seems reasonable though.

Bazel supports this pretty well. If you have a `local_repository` that points to a directory within your repository, Bazel will ignore that directory when walking the tree in your own repository. So you can do things like `bazel test //...` without worrying about it misinterpreting build files in that directory.

Gazelle doesn't know about that trick, and it probably should. We should read `local_repository` rules and exclude those directories automatically. Ideally, you would be able to define a `gazelle` rule in those directories and then do `bazel run @org_golang_google_grpc//:gazelle`, and that would update the build files in that repository. Or maybe `bazel run //:gazelle` would just do that for all of the vendored repositories.

You may be able to make this work right now by declaring a `local_repository` named `org_golang_google_grpc` *before* calling `go_rules_dependencies` (which will not override the repo if it already exists). You'll also need to add a `# gazelle:exclude vendor/google.golang.org/grpc` directive in your root build file (or some shortened variant of that in some parent directory of the vendored repo).

nugget:
Not sure if this will be useful to you or not, @bennyscetbun.  I ran into this same situation when working with go-swagger generated code.  I solved it by creating a repository rule which builds all my vendor dependencies in a separate Bazel package labeled `@vendor` and then I am safe to use that label for all my vendor `deps` in my project build files.  That way I can reference my vendored packages in the form:

```
@vendor//vendor/...
```

This neatly avoids any Bazel "is provided by more than one rule" warnings during build, allows passed structure namespaces to always agree, and prevents me from ever accidentally mismatching dependent versions across packages.

I pushed out the rules to https://github.com/ActiveState/rules_vendor in case they're generally useful for your situation.

This approach still requires that I post-process all the Gazelle-generated build files to rewrite the external deps, and would be assisted by a Gazelle feature that allowed the user to specify a custom external dep prefix (in lieu of the default `/vendor` prefix).  It still gets us almost all of the way there without any changes to Gazelle's behavior, though.

jayconrod:
Closing old PRs.

There are a couple new solutions to the vendored gRPC problem now. Gazelle will automatically resolve `google.golang.org/grpc` and several other proto-related imports to specific labels (not in vendor) when not in the `disable` or `disable_global` proto mode. Bazel also supports `.bazelignore` files to exclude certain directories, and Gazelle supports `# gazelle:exclude path` directives and `-exclude=path` arguments.

PR #236, opened 2018-06-21
Refactor Gazelle to enable support for new languages

### LINKED ISSUE #136: Documentation on how to add a new language to Gazelle

jayconrod:
This PR rewrites most of Gazelle to enable support for new languages.

* internal/language declares an interface for language extensions to implement. This interface covers, configuration, fixing, rule generation, and dependency resolution.
* internal/language/proto and internal/language/go are implementations of this interface.
* All language-specific logic is extracted from packages outside of internal/language. internal/merger and internal/resolve are now language-agnostic. Several packages, including internal/packages and internal/generator are deleted, since the entirety of their functionality has moved to the language extensions.

Behavioral compatibility is preserved as much as possible. There are a few small differences in directories with both proto and Go code with different package names, but for the most part, nothing has changed.

Fixes #136

Still to do: update Design.rst with new architecture.


jayconrod:
cc @achew22 Multi-language support has landed!

PR #240, opened 2018-06-24
Add code generation validator

achew22:
This change adds a new "testing" directory that contains fake source
files that allow Gazelle authors to quickly iterate on designs. The test
library works by having real shaped directory structures then
automatically running Gazelle on it and comparing the output of that
against a checked in golden file.

Additionally there is a "update.sh" script that will run all the tests
in that directory and copy the generated BUILD.bazel file into the
appropriate source directory. This should keep the maintenance burden
for developers relatively low as well as provide a convenient way for
repo maintainers to see what would happen if they were to approve a PR
by inspecting what happens to the goldens.


jayconrod:
Thanks for submitting this! I think this framework will be very useful for integration testing.

It may take me some time to fully review this. I want to explore whether it makes more sense for this to be a Go library instead of bash. I also want to look at whether this can be used for unit testing in addition to integration testing. The Go and proto extensions have something a little like this for testing rule generation (but not dependency resolution or merging), and it would be good to eliminate the duplication.

Do you have a set of tests anywhere that use this framework? It would be good to see this in action.

jayconrod:
Sorry to be slow on reviewing this. I've been sick the last couple days, and I wasn't able to get back to this as soon as I would have liked.

I think this framework is going to be really valuable both for integration and unit testing. However, I'd like to push back on a few elements of the design.

What I'd really like is something driven by tests instead of something that is itself a test. For example, in a Go test, I'd like to be able to say "run this function in this directory and check that the generated BUILD.bazel files match the checked-in BUILD.want files". This would be useful for both unit tests (e.g., the function could just perform a fix or emit generated rules) and for integration tests (e.g., the function could run all of `gazelle fix` or `gazelle update`).

I think it's valuable to keep the test framework written in Go like the rest of Gazelle. It's hard to keep bash scripts portable on Linux, macOS, and Windows, and I'd like to be able to run Gazelle tests on all of those. There tend to be a lot of problems with quoting and path separators. Keeping the framework in Go means we can continue to use flags like `--test_filter` a.k.a. `-test.run` to run individual tests or test cases. We can also continue running `go test` (which I find useful when Gazelle itself is not in a fully working state).

The ability to write back golden files is particularly useful. With `go test`, a somewhat common practice is defining an `-update` flag, then writing back the golden file when that flag is set. There's an example of this in [gofmt_test.go](https://golang.org/src/cmd/gofmt/gofmt_test.go). Bazel's sandboxing and remote execution make this tricky. I think test binaries would have to be run with `bazel run`, and we'd need to find tests that support the `-update` flag with `bazel query` (maybe using `tags = ["update"]` on the `go_test` rules). We'd probably need a bash script to orchestrate that, but that should be okay if it's not needed to execute the tests normally.

jayconrod:
Closing old PRs. Let me know if you want to move forward on this though.

There's an `internal/testtools` package now which supports creating and checking a temporary directory tree from literals inside .go files. This has been very useful for small tests across Gazelle, but I recognize it can get unwieldy for larger rule generation tests. That's probably the right place for this sort of thing though.

PR #246, opened 2018-06-27
Use symlink for overlay files.

pmuetschard:
`cp` is not a recognized command on Windows.

### DIFF
--- a/internal/overlay_repository.bzl
+++ b/internal/overlay_repository.bzl
@@ -49,7 +49,7 @@ def _git_repository_impl(ctx):
   revision = ctx.attr.commit if ctx.attr.commit else ctx.attr.tag
   _check_execute(ctx, ["git", "clone", "-n", ctx.attr.remote, "."], "failed to clone %s" % ctx.attr.remote)
   _check_execute(ctx, ["git", "checkout", revision], "failed to checkout revision %s in remote %s" % (revision, ctx.attr.remote))
-  
+
   _apply_overlay(ctx, overlay)
 
 git_repository = repository_rule(
@@ -78,7 +78,7 @@ def _apply_overlay(ctx, overlay):
   """
   # TODO(jayconrod): sanitize destination paths.
   for src_path, dst_rel in overlay:
-    _check_execute(ctx, ["cp", src_path, dst_rel], "failed to copy file from %s" % src_path)
+    ctx.template(dst_rel, src_path)
 
 def _check_execute(ctx, arguments, message):
   res = ctx.execute(arguments)


jayconrod:
I still have two concerns about using symlinks here: 1) We'll get an error when replacing existing files, and 2) it requires administrative privilege (even if Bazel does this in other places, I'd rather not contribute to the problem if they want to clean that up).

I think a better approach is to generate a .bat script with `copy` commands, then execute it with `ctx.execute`. Bonus points for getting the script to delete itself when it finishes.

pmuetschard:
> 1) We'll get an error when replacing existing files,

I think that is a good thing. As a user, I would not expect `rules_go` to silently override a file in a repository. This would be a nightmare to debug. How would a user even know this is happening?

> 2) it requires administrative privilege

Can you point me to documentation on this? I don't think it is true. Note that it does *not* create "symbolic links", but junctions on Windows. Either way, as long as you ask bazel to create the link for you, it will work, even if bazel changes how it's implemented in the future. In fact, I would think that using `repository_ctx.symnlink` is safer than copying in the file (with a `.bat` file, `bash`, or other means) as bazel will actually be aware of the file source and notice any changes.

jayconrod:
> I think that is a good thing. As a user, I would not expect rules_go to silently override a file in a repository. This would be a nightmare to debug. How would a user even know this is happening?

It's important to be able to replace incompatible or incomplete build files in external repos. rules_go has undergone some breaking API changes in the last year (hopefully no more in the foreseeable future). If you want to use an external repo with build files that use an old, incompatible version of rules_go, you need to be able to update or replace those build files.

> Can you point me to documentation on this? I don't think it is true.

I found [something on the Microsoft blog](https://blogs.windows.com/buildingapps/2016/12/02/symlinks-windows-10/#Ed9Olhkz6hJp4KWV.97). You're right, symlinks no longer require administrative privilege. They did recently though.

> Note that it does not create "symbolic links", but junctions on Windows. Either way, as long as you ask bazel to create the link for you, it will work, even if bazel changes how it's implemented in the future.

I didn't know that. Fair enough.

> In fact, I would think that using repository_ctx.symnlink is safer than copying in the file (with a .bat file, bash, or other means) as bazel will actually be aware of the file source and notice any changes.

Bazel will track changes in files referenced by labels resolved during repository rule execution. When a file changes, it will re-execute the repository rule in the next build that uses something in that repository, regardless of how the file is used. So it should be safe either way.

pmuetschard:
PTAL. I've changed it to use `repository_ctx.template`, which happily overrides existing files.

PR #274, opened 2018-07-30
More protable bash invocation

amckague:
This helps with nixos where bash is installed elsewhere by default.


amckague:
The nixos startup process links `/usr/bin/env` in where you'd expect. `/bin/sh` is also generated to support system(3). Everything else is pulled in similar way to bazel workspace with external dependencies.

PR #278, opened 2018-08-02
Explicitly specify build mode for Gazelle binary

ash2k:
This avoids the dependency on bazel --feature flags and system libc. The idea is the same as in https://github.com/bazelbuild/rules_go/pull/1615.

### DIFF
--- a/cmd/gazelle/BUILD.bazel
+++ b/cmd/gazelle/BUILD.bazel
@@ -34,13 +34,20 @@ go_library(
 go_binary(
     name = "gazelle",
     embed = [":go_default_library"],
+    msan = "off",
+    pure = "off",
+    race = "off",
+    static = "off",
     visibility = ["//visibility:public"],
 )
 
 go_binary(
     name = "gazelle_pure",
     embed = [":go_default_library"],
+    msan = "off",
     pure = "on",
+    race = "off",
+    static = "off",
     tags = ["manual"],
     visibility = ["//visibility:public"],
 )


ash2k:
I've also ran buildifier on the repo. Let me know if I should undo it.

cmd/gazelle/BUILD.bazel
```diff
@@ -34,14 +34,10 @@ go_library(
 go_binary(
     name = "gazelle",
     embed = [":go_default_library"],
-    visibility = ["//visibility:public"],
-)
-
-go_binary(
-    name = "gazelle_pure",
-    embed = [":go_default_library"],
+    msan = "off",
```
ash2k: This attribute is not documented btw.

jayconrod: Thanks for pointing that out. Fixed in #1624.

cmd/gazelle/BUILD.bazel
```diff
@@ -34,14 +34,10 @@ go_library(
...
     pure = "on",
-    tags = ["manual"],
+    race = "off",
+    static = "off",
```
jayconrod: I'm okay with disabling `msan`, `race`, and `static` in `//cmd/gazelle:gazelle` (though I'm not sure I understand the motivation), but I'm not okay with `pure = "on"` in `//cmd/gazelle:gazelle`. Keep `gazelle_pure` and use that if you need it.

Pure mode forces us to compile the standard library in a different mode so that it doesn't include any cgo code. That can slow down builds considerably if both pure and non-pure targets are being built.

def.bzl
```diff
@@ -17,42 +17,43 @@ load("//internal:go_repository.bzl", "go_repository")
...
-      ctx.attr.command,
-      "-mode", ctx.attr.mode,
-      "-external", ctx.attr.external,
-  ]
-  if ctx.attr.prefix:
-    args.extend(["-go_prefix", ctx.attr.prefix])
-  if ctx.attr.build_tags:
-    args.extend(["-build_tags", ",".join(ctx.attr.build_tags)])
-  args.extend(ctx.attr.extra_args)
+    args = [
```
jayconrod: In general, I'd prefer formatting changes to be in separate PRs. It makes other changes easier to rebase and cherry-pick. I don't think this particular change will cause any harm though, so need to split it this time.

cmd/gazelle/BUILD.bazel
ash2k: The motivation is to avoid the dependency on `--feature` flags for gazelle to avoid re-building it if a flag is used. Also (for pure) to avoid depending on libc (fixes rule usage on nixos).
I agree with the point re. building stdlib in pure mode - I've reverted the change. I guess we have to use non-rule gazelle on nix.
I've explicitly set pure=off to avoid depending to the `--feature` flag.

ash2k: Actually, I've just found out that we could do the following to run a pure version of gazelle via the rule:
```bzl
gazelle(
    name = "gazelle",
    gazelle = "@bazel_gazelle//cmd/gazelle:gazelle_pure",
)
```

jayconrod: Ah, ok, avoiding a dependency on feature flags makes more sense now.

+1 to using the `gazelle` attribute in the `gazelle` rule. Seems like the best way to do this.

PR #286, opened 2018-08-06
Experimental: read and write build files in alternate directories

jayconrod:
* The flags -experimental_{read,write}_build_files_dir may now be used
  to read and write build files to alternate directories, which may be
  outside of the repository root.
* When a build file is read from an alternate directory, the build
  file in the source directory is ignored.
* When a build file is written to an alternate directory, any existing
  build file in that directory is replaced. The build file in the
  source directory is not updated.

### DIFF
--- a/cmd/gazelle/diff.go
+++ b/cmd/gazelle/diff.go
@@ -20,21 +20,18 @@ import (
 	"io/ioutil"
 	"os"
 	"os/exec"
-
-	"github.com/bazelbuild/bazel-gazelle/internal/config"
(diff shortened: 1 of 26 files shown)


achew22:
Do you intend to use this for e2e testing? I think you have built a fantastic e2e engine here.

jayconrod:
@achew22 Mainly I built this for autogo and updating dependencies. It's not a complete solution, but it's a step in the right direction. It could be useful for e2e though.

PR #289, opened 2018-08-10
internal/repos: resolve the VCS when importing from a lock file

zllak:
This will make the `update-repos -from_file=` actually work because it was generating `go_repository` rules potentially with a `remote` but without a `vcs`.
There is one problem tho, where the repository in the Gopkg.lock is a private repository.
`dep` will create an entry with the URL like `https://github.com/.....git`, so it will end up in the WORKSPACE with this URL, but bazel is unable to resolve this.
I don't have a solution for that particular problem, but to sed the `Gopkg.lock` to transform private repositories to `git@github.com:...`.

### DIFF
--- a/internal/repos/repo.go
+++ b/internal/repos/repo.go
@@ -71,7 +71,7 @@ var lockFileParsers = map[lockFileFormat]func(string) ([]Repo, error){
 // a list of equivalent repository rules that can be merged into a WORKSPACE
 // file. The format of the file is inferred from its basename. Currently,
 // only Gopkg.lock is supported.
-func ImportRepoRules(filename string) ([]*rule.Rule, error) {
+func ImportRepoRules(filename string, remoteCache *RemoteCache) ([]*rule.Rule, error) {
 	format := getLockFileFormat(filename)
 	if format == unknownFormat {
 		return nil, fmt.Errorf(`%s: unrecognized lock file format. Expected "Gopkg.lock"`, filename)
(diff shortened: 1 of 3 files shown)


internal/repos/repo.go
```diff
@@ -85,6 +86,13 @@ func ImportRepoRules(filename string) ([]*rule.Rule, error) {
 
 	rules := make([]*rule.Rule, 0, len(repos))
 	for _, repo := range repos {
+		if repo.Remote != "" && repo.VCS == "" {
+			// here we need to resolve the GoPrefix to determine the VCS
```
jayconrod: This involves a network fetch that can take a long time, and it may not be needed in the common case where the repository rule is already in WORKSPACE. Instead of calling `vcs.RepoRootForImportPath` directly here, call `RemoteCache.RepoRootForImportPath`. You'll need a `RemoteCache` populated with the contents of WORKSPACE. That should be created in `cmd/gazelle/update-repos.go`, pre-populated with `ListRepositories`, and passed in here.

See `updateImportPaths` in `cmd/gazelle/update-repos.go` as an example of something very similar. That also parallelizes the network fetches, which is a good idea here, too.

zllak: Thanks ! I'll have a look at it.

jayconrod: I was looking at this a bit more closely today, and I should have seen this earlier: this is actually querying the original import path, not the source URL. So if you have a stanza like this (taken from import_test.go, but with the `.git` extension deleted):

```
[[projects]]
  branch = "parse-constraints-with-dash-in-pre"
  name = "github.com/Masterminds/semver"
  packages = ["."]
  revision = "a93e51b5a57ef416dac8bb02d11407b6f55d8929"
  source = "https://github.com/carolynvs/semver"
```

This will query `github.com/Masterminds/semver` instead of `https://github.com/carolynvs/semver`. So if you cloned this project into a private hg repo, this code would give us the wrong answer.

Do you know how dep figures this out? I don't think we can use `RepoRootForImportPath` to determine the import path, since the source is not an import path and might not use https. I skimmed [gps/deduce.go](https://github.com/golang/dep/blob/master/gps/deduce.go), and it seems like they use a combination of scheme, known URLs, and HTTP metadata querying. There's far more complexity there than I'm comfortable integrating into Gazelle though.

I'm okay with adding some known URL patterns at least. We know github.com is only going to have git repositories for example, so we could have more known patterns like in remote.go.

jayconrod:
Closing old PRs. Please reopen if you still want to move forward with this.

PR #310, opened 2018-09-06
Support go modules with update-repos -from_file=go.mod

### LINKED ISSUE #200: gazelle update-repos doesn't support vgo
[vgo](https://github.com/golang/go/wiki/vgo) is the next generation of dependency management for go.

I'd like to work on a PR to add vgo lock file support to the update-repos command. Plan:
 * Add vgo support to internal/repos
 * Update documentation with clear indicators that vgo is still not production-ready.

Does this PR have a chance to be accepted?

jayconrod:
With this change, Gazelle can import modules listed in go.mod files
(including transitive dependencies that are not listed) as
go_repository rules. Gazelle runs "go list -m -json all" to obtain a
list of modules, transforms the list into go_repository rules, then
merges those rules into WORKSPACE.

If Gazelle is invoked by the wrapper script generated by the gazelle
rule, it uses the go tool from the downloaded SDK. Otherwise, it uses
the SDK pointed to by GOROOT or whatever's in PATH.

Fixes #200


mghignet:
Hi,
Thank you for your work on this project!
When do you plan to release a new version including this feature? 
Thanks a lot!

terinjokes:
It doesn't look like this patch supports the [`+incompatbile`](https://go-review.googlesource.com/c/go/+/124384) version syntax, these get imported as-is, which aren't valid tags upstream.

jayconrod:
@terinjokes Could you file a new issue for that? I may lose track otherwise.

@mghignet (sorry for the slow response). There should be a new version in the next few days. Hopefully by the end of the week.

terinjokes:
Coming right up.

jayconrod:
@terinjokes I just saw an issue was filed in #343 and there's a PR to fix in #344, which I'll review soon.

terinjokes:
@jayconrod Yes, I noticed that myself when I went to go make it.

Issue #343, opened 2018-10-10
Ignore "+incompatible" and mixed versioning coming from go.mod


christian-roggia:
Currently the following generated **go.mod** rule will be translated into a wrong bazel rule:

`github.com/googleapis/gax-go v2.0.0+incompatible // indirect`

This will be converted in:

```
go_repository(
    name = "com_github_googleapis_gax_go",
    importpath = "github.com/googleapis/gax-go",
    tag = "v2.0.0+incompatible",
)
```

and this Bazel rule will fail as there is no `v2.0.0+incompatible` tag in the repository.
 
The same erroneous behavior is observed when there is a mixed versioning (tag + timestamp + commit) of the package:

`gopkg.in/asn1-ber.v1 v1.0.0-20170511165959-379148ca0225`

This will be converted in:

```
go_repository(
    name = "in_gopkg_asn1_ber_v1",
    importpath = "gopkg.in/asn1-ber.v1",
    tag = "v1.0.0-20170511165959-379148ca0225",
)
```

and this Bazel rule will fail, again, because there is no `v1.0.0-20170511165959-379148ca0225` tag in the repository.

The solution is to strip "+incompatible" from the **go.mod** rules and to keep only the **commit** (not the "v1.0.0" tag, as it will fail) whenever the version satisfies the following regexp:
`^(.*?)-([0-9]{14})-([a-fA-F0-9]{12})$`

The final result should be:

```
go_repository(
    name = "com_github_googleapis_gax_go",
    importpath = "github.com/googleapis/gax-go",
    tag = "v2.0.0",
)
go_repository(
    name = "in_gopkg_asn1_ber_v1",
    importpath = "gopkg.in/asn1-ber.v1",
    commit = "379148ca0225",
)
```

PR #344, opened 2018-10-10
Better parsing of go.mod rules

christian-roggia:
This PR fixes https://github.com/bazelbuild/bazel-gazelle/issues/343

### DIFF
--- a/internal/repos/modules.go
+++ b/internal/repos/modules.go
@@ -39,21 +39,11 @@ var regexMixedVersioning = regexp.MustCompile(`^(.*?)-([0-9]{14})-([a-fA-F0-9]{1
 
 func toRepoRule(mod module) Repo {
 	var tag, commit string
-	if strings.HasPrefix(mod.Version, "v0.0.0-") {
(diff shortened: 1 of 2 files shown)


internal/repos/modules.go
```diff
@@ -35,6 +35,35 @@ type module struct {
 	Main          bool
 }
 
+var regexMixedVersioning = regexp.MustCompile(`^(.*?)-([0-9]{14})-([a-fA-F0-9]{12})$`)
+
+func toRepoRule(mod module) Repo {
+	var tag, commit string
+	if strings.HasPrefix(mod.Version, "v0.0.0-") {
+		i := strings.LastIndex(mod.Version, "-")
```
christian-roggia: Please note that "strings.LastIndex()" **cannot** fail as we already know from line 42 that the prefix contains at least one dash.

christian-roggia:
@jayconrod don't worry, I am used to waiting much longer than this. I cleaned up the tests and removed all redundant code.

PR #350, opened 2018-10-21
Handle the "replace" directive in go.mod when running "update-repos"

joe-mann:
This PR adds handling of "replace" directives found in go.mod when running "update-repos --from_file=go.mod". This allows Gazelle to handle situations where a repository has been forked, or where a repository has been "mirrored" to an internally-hosted location, for example.

Edit: I've just realised that I forgot a small piece of refactoring to remove the tests' reliance being able to talk to the remote "replaced" repositories (when vcs.RepoRootForImportPath is called). I'll fix this issue ASAP.

### DIFF
--- a/internal/repos/import_test.go
+++ b/internal/repos/import_test.go
@@ -28,6 +28,7 @@ import (
 
 func init() {
 	goListModulesFn = goListModulesStub
+	findRemoteFn = findRemoteStub
 }
 
 func TestImports(t *testing.T) {
@@ -257,4 +258,11 @@ func goListModulesStub(dir string) ([]byte, error) {
 	"GoMod": "/usr/local/google/home/jayconrod/go/pkg/mod/cache/download/gopkg.in/yaml.v2/@v/v2.2.1.mod"
 }
 `), nil
+}
+
+func findRemoteStub(importpath string) (remote, vcsName string, err error) {
+	if importpath == "github.com/joe-mann/go-toml" {
+		return "https://github.com/joe-mann/go-toml", "git", nil
+	}
+	return "", "", nil
 }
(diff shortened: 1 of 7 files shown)


joe-mann:
I signed it!

joe-mann:
Fixed network reliance issue in tests.

joe-mann:
Fixed network reliance issue in tests.

joe-mann:
Fix tests to not use network.

internal/repos/modules.go
```diff
@@ -76,7 +112,11 @@ func importRepoRulesModules(filename string) (repos []Repo, err error) {
 			continue
 		}
 
-		repos = append(repos, toRepoRule(mod))
+		repo, err := toRepoRule(mod)
```
jayconrod: Since this is going to be hitting the network now, these conversions should happen in parallel.

internal/repos/modules_test.go
```diff
@@ -36,3 +43,22 @@ func TestGoModuleSpecialCases(t *testing.T) {
 		})
 	}
 }
+
+func TestGoModuleReplacement(t *testing.T) {
```
jayconrod: I'd prefer not to test against `toRepoRule` directly, since it's an unexported function and we should be able to restructure its functionality without breaking tests. In any case, this seems redundant with import_test.go.

internal/repos/import_test.go
```diff
@@ -248,3 +259,10 @@ func goListModulesStub(dir string) ([]byte, error) {
 }
 `), nil
 }
+
+func findRemoteStub(importpath string) (remote, vcsName string, err error) {
```
jayconrod: See the comment below about `RemoteCache`. I think `newStubRemoteCache` should be used instead.

joe-mann:
Quick update - the fixes are almost done, but I haven't had as much time as I'd have liked over the last week.

jayconrod:
Note that most libraries have moved from `internal` into the repository root directory (sorry for the merge conflicts). This work is in preparation for a rule that will create Gazelle binaries with custom languages.

PR #352, opened 2018-10-28
Add "github.com/pkg/errors" to gazelle deps

Yannic:
This change is in preparation of bazelbuild/rules_go#1614, which starts
analyzing targets of bazel_test in the host workspace.

### DIFF
--- a/deps.bzl
+++ b/deps.bzl
@@ -40,6 +40,15 @@ def gazelle_dependencies(go_sdk = "@go_sdk//:ROOT"):
         commit = "f3dd8fd95a7d078cb10fd7fb475b22c3cdbcb307",  # 0.2.0 as of 2017-12-04
     )
 
+    _maybe(
+        go_repository,
+        name = "errors_go",
+        importpath = "github.com/pkg/errors",
+        commit = "30136e27e2ac8d167177e8a583aa4c3fea5be833",
+        patches = ["@bazel_gazelle//internal:repository_rules_test_errors.patch"],
+        patch_args = ["-p1"],
+    )
+
 def _maybe(repo_rule, name, **kwargs):
     if name not in native.existing_rules():
         repo_rule(name = name, **kwargs)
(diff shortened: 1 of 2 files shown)


jayconrod:
To be sure that I understand, this is needed because when Gazelle upgrades past bazelbuild/rules_go#1614, this test will break right?

Thanks for sending this, but I can't merge this one as it is. Gazelle can't have `go_repository` dependencies, because `go_repository` depends on `bazel_gazelle_go_repository_tools`, which needs to build a Gazelle binary and cannot depend on `go_repository` to fetch dependencies.

I'm not sure what the solution is yet, but it probably involves some variation of `bazel_test` that works the old way. I've been thinking about redesigning and simplifying `bazel_test` anyway though, so maybe it's better to just cross that bridge when we come to it.

Yannic:
Yes, this is needed once Gazelle updates past bazelbuild/rules_go#1614 which will complain about missing `@errors_go//`.

I'm not completely sure if I understand your reasoning to reject this. I do understand that gazelle itself cannot depend on any `go_repository` because this would introduce a dependency cycle. However, simply having a `go_repository` in `WORKSPACE` (or in `deps.bzl`) does not mean the gazelle binary depends on it, and Bazel won't even download `github.com/pkg/errors` unless anything depends on it (i.e. `bazel_test`). I think simply adding a comment that this can only be used for tests should be sufficient. Or am I missing something here?

jayconrod:
So i`//internal:repository_rules_test` should be the only thing that's affected by this. The purpose of that test is to verify the implementation of `go_repository`. I'd like that test to be as self-contained as possible (unfortunately testing repository rules is pretty difficult). Moving the repository rule being tested out of the test WORKSPACE into Gazelle's general list of dependencies doesn't seem right to me. Especially moving this repository into `gazelle_dependencies` isn't right because it will be visible to anyone using Gazelle, and it effectively becomes a public interface.

PR #353, opened 2018-10-29
internal/language/go: replace ${SRCDIR} with "." in project root

### LINKED ISSUE #348: gazelle/go_repository should detect dependencies on header files specified via #cgo CFLAGS: -I${SRCDIR}/<path>
I have a project that uses something like so:

```
// #cgo CFLAGS: -I${SRCDIR}/internal/include -I${SRCDIR}/internal/libbacktrace
```

That project is included in the WORKSPACE of another project like so:

```python
go_repository(
    name = "com_github_cstrahan_cgosymbolizer",
    commit = "67837f878601260c69d94c2c5039a47f40bb233b",
    importpath = "github.com/cstrahan/cgosymbolizer",
)
```

Trying to compile the latter results in:

```
ERROR: /home/cstrahan/.cache/bazel/_bazel_cstrahan/0e3896d6d23d3761a1e0a1ef1f9e1eac/external/com_github_cstrahan_cgosymbolizer/BUILD.bazel:3:1: in cc_library rule @com_github_cstrahan_cgosymbolizer//:go_default_library.cgo_c_lib: The include path '/internal/libbacktrace' references a path outside of the execution root.
```

jayconrod:
Previously, if a cgo file in the project root contained a directive
like `#cgo CPPFLAGS: -I${SRCDIR}/foo`, it would be expanded to
-I/foo. It will now be expanded to -I./foo, which gets simplified to
-Ifoo.

Fixes #348

### DIFF
--- a/internal/language/go/fileinfo.go
+++ b/internal/language/go/fileinfo.go
@@ -462,6 +462,9 @@ func expandSrcDir(str string, srcdir string) (string, bool) {
 	// so convert native paths with a different delimiter
 	// to "/" before starting (eg: on windows).
 	srcdir = filepath.ToSlash(srcdir)
+	if srcdir == "" {
+		srcdir = "."
+	}
 
 	// Spaces are tolerated in ${SRCDIR}, but not anywhere else.
 	chunks := strings.Split(str, "${SRCDIR}")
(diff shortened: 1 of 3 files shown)


darmie:
This doesn't seem to work for symlinked paths

PR #359, opened 2018-11-02
Create a gazelle_plugin rule

achew22:
This rule allows one to define an additional language into the Gazelle
runner. By choosing this design, an author can write a generator that
plugs into gazelle but does not live in the Gazelle repo allowing for
independent iteration. Due to this independent iteration it is now also
possible to create a plugin that only works for a custom set of rules
that are, for example, developed in house. These in house and external
rules can then be interwoven with the rules_go provided plugins in
whatever order is necessary for the project.

Concerns I have for this PR:

- [ ] Maintain `go get`ability
    - I don't know enough about the go default packaging system to know if what I did make it impossible to `go get` the Gazelle binary by having a `package main` and a `package gazelle` in the same directory. Some input on how to do that better would be appreciated.
- [ ] Internal visibility concerns
    - I made some stuff in an `internal` directory publicly visible. I would like to either reexport those from a public path or move them to a public path directly. 
- [ ] Documentation
    - Before submitting I would like to add documentation but I don't want to write docs until I have a firm understanding of the end state.
- [ ] Private providers from `rules_go`
    -  I use a private provider (`GoLibrary`) from rules_go to get the `importpath`. Is it possible to do that without defeating the visibility of `rules_go`? (if only someone who was reviewing knew rules_go really well)


def.bzl
```diff
@@ -36,11 +49,47 @@ def _gazelle_runner_impl(ctx):
         args.extend(["-build_tags", ",".join(ctx.attr.build_tags)])
     args.extend(ctx.attr.extra_args)
 
-    out_file = ctx.actions.declare_file(ctx.label.name + ".bash")
+    # TODO: Do something intelligent here to detect hwne there is a Gazelle passed in
```
mrmeku: when

jayconrod:
So... bad news: I can't accept this PR as is. But good news: it's because I'm already working on something that accomplishes the same goal with nearly the same approach. I'm hoping to ship it next week.

Here's what I'm planning:

* Most (all?) of Gazelle internal packages will be moved into the public space. Cleaning up the documentation and removing dead stuff has taken some time, but this part is nearly ready.
* Users will be able to write extensions as regular `go_library` rules. They must define a function named `New` that returns something assignable to the interface `language.Language`. I think I'd rather just set a convention here than define a new `gazelle_plugin` rule.
* There will be an internal `gazelle_binary` rule that accepts a `languages` attribute, which is a list of `go_library`. That will generate a .go file that will be compiled together with `srcs` (and the `languages` will be made available to the compiler and linker, together with `deps`).
* The `gazelle` macro will also accept a `languages` parameter. When set, it will define an alternative `gazelle_binary` with those languages.

jayconrod:
Closing this PR. Gazelle 0.16.0 is out now with support for `gazelle_binary`.

Issue #364, opened 2018-11-06
Go extension should be able to run without proto extension


jayconrod:
If someone creates a `gazelle_binary` with `//languages/go:go_default_library` but not `//languages/proto:go_default_library`, Gazelle should not panic.

```
panic: interface conversion: interface {} is nil, not *proto.ProtoConfig

goroutine 1 [running]:
github.com/bazelbuild/bazel-gazelle/language/proto.GetProtoConfig(...)
	external/bazel_gazelle/language/proto/config.go:52
github.com/bazelbuild/bazel-gazelle/language/go.(*goLang).CheckFlags(0xc00000e0d8, 0xc000036c00, 0xc000193ab0, 0x0, 0x0)
	external/bazel_gazelle/language/go/config.go:197 +0x115
main.newFixUpdateConfiguration(0x0, 0xc00001e0e0, 0x2, 0x2, 0xc0000c4a80, 0x5, 0x8, 0xc0000d4600, 0x4, 0x4, ...)
	external/bazel_gazelle/cmd/gazelle/fix-update.go:300 +0x90a
main.runFixUpdate(0x0, 0xc00001e0e0, 0x2, 0x2, 0xc0000a1418, 0x535b01)
	external/bazel_gazelle/cmd/gazelle/fix-update.go:179 +0x804
main.run(0xc00001e0d0, 0x3, 0x3, 0xc00000e0d8, 0x89758a)
	external/bazel_gazelle/cmd/gazelle/gazelle.go:78 +0x158
main.main()
	external/bazel_gazelle/cmd/gazelle/gazelle.go:59 +0x91
```

It's okay for the Go extension to depend on the proto extension (though looser coupling is better), but it should tolerate the proto extension not being run.

Issue #363, opened 2018-11-06
Special cases for gazelle and rules_go imports


jayconrod:
Gazelle should automatically resolve Go imports starting with `github.com/bazelbuild/bazel_gazelle` to `@bazel_gazelle` and imports starting with `github.com/bazelbuild/rules_go` to `@io_bazel_rules_go`. These repositories don't follow normal naming conventions and usually won't appear directly in WORKSPACE.

PR #365, opened 2018-11-06
gazelle_binary: create gazelle binaries with custom languages

### LINKED ISSUE #311: Support for generating custom rules
Virtually all the logic in Gazelle is a good fit for me: directive support, parsing source code, etc. are all perfect. However, I need to generate slightly different rules than Gazelle's default `go_library`, `go_test`, etc. Today, the only way I can see to do this is to fork Gazelle and change `internal/language/go/kinds.go` (and any related code). Is this correct?

If so, do you have plans to make Gazelle more extensible in the future? I suspect that this use case is relatively common, since many large-ish organizations have their own ideas about how Go should be built, tested, linted, and formatted by default. It'd be great if Gazelle exposed a way to let users supply their own set of rules.

jayconrod:
gazelle_binary is a new rule that creates binaries to be used with the
gazelle rule. It accepts a list of languages (the default languages
are proto and go). It generates a .go source file that imports the
language packages and initializes each extension. That source file is
compiled into the main package in a new gazelle binary. The rest of
the main package does not reference any specific languages.

Fixes #311


jayconrod:
## TODO

- [ ] #363 - special cases for gazelle and rules_go imports
- [ ] #364 - go extension should be able to run without proto extension
- [ ] Documentation
- [ ] Test

Issue #376, opened 2018-11-15
Allow vendor mode to have customized vendor directory


linuxerwang:
When specify -external=vendor, it would be better to have a way to customize the vendor directory.

At present, gazelle always assumes //vendor as the place. But for many large projects, code files with different languages might be put in the same source repo, and it's not always working to assume //vendor for Go.

To support customized vendor directory, the -external flag can be augmented with a colon, like "-external=vendor:path/to/vendor". Or introduce a new flag like "-vendor_dir" should also work great.

linuxerwang:
My bad. Just realized that what made it work was my change to add -vendor_dir, not importmap_prefix. A pull request is coming after.

jayconrod:
Sorry for the slow response; just got back from vacation.

Could you explain more about the problem you're trying to solve? `-external=vendored` should almost never be needed since Gazelle indexes library rules and resolves dependencies based on that index. `-external=vendored` would only be needed when your vendor directory is not actually present when you run Gazelle.

linuxerwang:
Hi, Jay

It comes from our non-standard folder layout:

TOP/
    |-- WORKSPACE
    |-- proj1/
    |-- proj2/
    |-- proj3/
    |-- ...
    |-- third-party-go/
             |-- vendor/
                    |-- github.com/

When using gazelle to generate BUILD files for third-party libraries, its deps session contains lines like "//github.com/..." instead of "//third-party-go/vendor/github.com", thus it fails to compile. My change attempted to make it work. It works for third-party BUILD files, but not for the top level folders like "proj1". Still digging in gazelle's code.

Sounds like I mis-used gazelle?

jayconrod:
Something doesn't seem right here. Let's say you have some .go files in `third-party-go/vendor/github.com/foo-org/foo`. Gazelle should generate a `go_library` in that directory like this:

```
go_library(
    name = "go_default_library",
    importpath = "github.com/foo-org/foo",
    importmap = "third-party-go/vendor/github.com/foo-org/foo",
    srcs = ["foo.go"],
    visibility = ["//visibility:public"],
)
```

If you have a .go file anywhere else in `third-party-go/` or its subdirectories that imports `github.com/foo-org/foo`, Gazelle should resolve that to `//third-party-go/vendor/github.com/foo-org/foo:go_default_library` in the `deps` for that library.

[...]

PR #377, opened 2018-11-15
Add flag -vendor_dir to control vendor directory when -external is set to vendored.

linuxerwang:
See issue #376.


linuxerwang:
As discussed in #376 this pull request is not needed.

PR #380, opened 2018-11-22
Explicitly export symbols that need it, reformat files with Buildifier

laurentlb:
Progress towards https://github.com/bazelbuild/rules_go/issues/1829

### DIFF
--- a/def.bzl
+++ b/def.bzl
@@ -19,9 +19,15 @@ load(
 )
 load("@bazel_skylib//lib:shell.bzl", "shell")
 load("//internal:go_repository.bzl", _go_repository = "go_repository")
-load("//internal:overlay_repository.bzl", "git_repository", "http_archive")
+load(
+    "//internal:overlay_repository.bzl",
+    _git_repository = "git_repository",
+    _http_archive = "http_archive",
+)
 
 go_repository = _go_repository
+git_repository = _git_repository
+http_archive = _http_archive
 
 def _gazelle_runner_impl(ctx):
     go = _go_context(ctx)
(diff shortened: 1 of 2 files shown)


def.bzl
```diff
@@ -18,9 +18,11 @@ load(
     _go_rule = "go_rule",
 )
 load("@bazel_skylib//lib:shell.bzl", "shell")
-load("//internal:go_repository.bzl", "go_repository")
+load("//internal:go_repository.bzl", _go_repository = "go_repository")
 load("//internal:overlay_repository.bzl", "git_repository", "http_archive")
```
jayconrod: We'll also need to re-export `git_repository` and `http_archive` from here and deps.bzl below.

vendor/github.com/bazelbuild/buildtools/build/build_defs.bzl
```diff
@@ -17,36 +17,35 @@ distributed under the License is distributed on an "AS IS" BASIS,
...
-      srcs = [src],
-      outs = [out],
-      tools = [_GO_YACC_TOOL],
-      cmd = ("export GOROOT=$$(dirname $(location " + _GO_YACC_TOOL + "))/..;" +
-             " $(location " + _GO_YACC_TOOL + ") " +
-             " -o $(location " + out + ") $(SRCS)"),
-      visibility = visibility,
-      local = 1,
-  )
+def go_yacc(src, out, visibility = None):
```
jayconrod: Please revert this file. This is part of a local copy of `github.com/bazelbuild/buildtools`, so any changes here may be erased when we update to a newer version.

At the moment, Gazelle uses a specific older version of this library since the API broke a couple times. I would guess it's safe to upgrade now though.

laurentlb:
Comments should be resolved

PR #382, opened 2018-11-27
Make helper tool build deterministic

andrew-d:
Currently, the builds for helper utilities `fetch_repo` and `gazelle` aren't deterministic, and have hashes that change depending on where they're built.  Since these are built as repository rules, it makes it tricky to prefetch dependencies using [`bazel sync`](https://blog.bazel.build/2018/07/09/bazel-sync-and-resolved-file.html), since the hash of dependencies will change per-build.

These changes build the helper utilities with the appropriate set of flags to ensure that they're deterministic; this can be tested by running `strings path-to-gazelle | egrep '(\/bazel\/|\/tmp\/)'`.

Let me know if this works for you folks!  All tests pass locally on Linux with these changes.


andrew-d:
Thanks for the quick response!

PR #406, opened 2018-12-20
Update boilerplate for 0.16.0 [skip ci]

jayconrod:


prattmic:
The sha256 file on https://github.com/bazelbuild/bazel-gazelle/releases/tag/0.16.0 is incorrect. It contains the rules_go 0.16.5 sha256.

prattmic:
Yup, that is what I see now.

PR #409, opened 2018-12-31
prioritizing external libraries

linzhp:
When Gazelle is running in external mode, and there is a `go_repository` matching the import path not otherwise found, Gazelle should point the dependency to the `go_repository`, regardless of whether the import path has the `go_prefix` or not.

The scenario is an organization may have project A and B in repo X, and project c in repo Y, but A, B and C all share example.com as their prefix. As a result, the `go_prefix` in repo X is example.com. However, project c is not in repo X, and it is defined as a go_repository in repo X. When Gazelle sees example.com/c, it should first attempt to look up the known repositories in WORKSAPCE file, before producing //c:go_default_library, which doesn't exist.

@jayconrod What do you think?


jayconrod:
I don't think I agree with this behavior. I think users would be confused if they've told Gazelle to do one thing (with `# gazelle:prefix example.com`), and it's doing something else (resolving to an external repository). This would also break users who run Gazelle with `-index=false` (in particular, autogazelle users).

But maybe it's possible to solve this problem with different configuration directives?

* In repo X, A/BUILD.bazel, you could set `# gazelle:prefix example.com/a`
* In repo X, B/BUILD.bazel, you could set `# gazelle:prefix example.com/b`
* In the `go_repository` rule for Y, you could set `importpath = "example.com/c"`, which is equivalent to setting `# gazelle:prefix example.com/c` in the root build file of that repository.

Gazelle should resolve imports starting with those prefixes to the correct locations. Ideally, there shouldn't be any locations with overlapping prefixes.

linzhp:
Following your suggestion, I found package level `gazelle:prefix` directives don't work because of two reasons:

* I don't see any map from directory/package to prefix. So it looks like the same prefix is shared across the whole repo
* Because of [this line](https://github.com/bazelbuild/bazel-gazelle/blob/master/walk/walk.go#L245), package level config are not propagated outside the [visit function](https://github.com/bazelbuild/bazel-gazelle/blob/master/walk/walk.go#L118) except the root level config. Package level walk directives (`exclude`, `follow` and `ignore`) still works because they are used right after being discovered in [visit function](https://github.com/bazelbuild/bazel-gazelle/blob/master/walk/walk.go#L137)

I am pretty new to Gazelle, so I don't know whether these two behaviors are expected, or should I fix them?

linzhp:
My change in this PR only affects Go packages not found in the main source tree. In such case, could you explain when users would want Gazelle to generate a label pointing to non-existent target in the main source tree, instead of some external library defined as `go_repository` containing the missing package?

jayconrod:
In general, Gazelle directives apply to the directory where they appear and subdirectories. The `gazelle:prefix` directive tells Gazelle how to generate `importpath` attributes for `go_library` and other Go rules. So it makes sense that they only apply within a directories and subdirectories. See [Directives](https://github.com/bazelbuild/bazel-gazelle#directives) for more details.

When Gazelle runs, it builds an index of library targets in the repository. For Go, that's a map from `importpath` strings to labels. Gazelle generates `deps` attributes using this index. Local, external, or vendor labels are constructed for imports that can't be resolved using the index. See [Dependency resolution](https://github.com/bazelbuild/bazel-gazelle#dependency-resolution) for more details.

> My change in this PR only affects Go packages not found in the main source tree. In such case, could you explain when users would want Gazelle to generate a label pointing to non-existent target in the main source tree, instead of some external library defined as go_repository containing the missing package?

I don't think the current behavior is ideal, but I disagree that the proposed behavior is correct.

Before Gazelle used an index for dependency resolution, it just relied on prefixes and external modes to resolve dependencies. There wasn't any way to check whether a dependency actually existed, so it wasn't an error. This is still the behavior when `-index=false`, which autogazelle sets. We should probably print an error if indexing is enabled when a dependency can't be resolved but we expect it should be, based on the prefix.

jayconrod:
#414 was a duplicate of this PR. @siddharthab's example shows that this problem can't be fixed with more uses of `# gazelle:prefix`.

> There are legitimate cases when an import path can have the current
repository's go_prefix as a prefix and yet can belong to a different
repository.
>
> For example, v.io is github.com/vanadium/core, and v.io/x/lib is
github.com/vanadium/go.lib.

So if you're importing, say `"v.io/x/lib/toposort"` from somewhere under the prefix `v.io`, we should check the index first, and if the library's not in the index, we *must* fall back to the external / vendored resolver. 

I'm still concerned that this will break when `-index=false`, since nothing will be in the index.

@linzhp, could you update this PR to support this?

* You'll need to move and rename the `updateConfig.useIndex` flag from cmd/gazelle/fix-update.go to `Config.IndexLibraries` in config/config.go. The flag will also need to be registered in `CommonConfigurer` instead of `updateConfigurer`.
* In language/go/resolve.go, the logic should be:
  * If the import can be resolved using the index, use that. There may be cases in the future where we add things to index even if `Config.IndexLibraries` is false.
  * If `Config.IndexLibraries` is false and the import path is under the current prefix, generate a local label.
  * Otherwise, use the external or vendored resolver.

linzhp:
Updated according to @jayconrod's suggestion

nugget:
Is there any way to avoid this behavior?  It broke my workflow with generated files which are created by a Bazel extension and do not exist on disk when gazelle is run.

jayconrod:
@nugget You can use `# gazelle:resolve` directives to map specific imports to labels, and you can use `# keep` comments to preserve manual edits to `deps`.

PR #410, opened 2018-12-31
populating VCS field when go_repository's remote is not null

linzhp:
`bazel-gazelle/cmd/fetch_repo` command requires `--vcs` to be present when  `--remote` is present.  As a result, when there is `source` field in dep lock file, the `vcs` field should be populated for the resulting `go_repository` rules as well. Otherwise the subsequent bazel build will fail when it is calling gazelle to fetch external libraries.


linzhp:
Thanks for accepting the PR. Looking forward to the better resolution of `source` field. This will unblock us for now.

PR #414, opened 2019-01-04
Do not check if go_prefix is a prefix of import path

siddharthab:
There are legitimate cases when an import path can have the current
repository's go_prefix as a prefix and yet can belong to a different
repository.

For example, v.io is github.com/vanadium/core, and v.io/x/lib is
github.com/vanadium/go.lib.

When doing external dependency resolution, we should rely on the result
from vcs.RepoRootForImportPath if the import path is not in the
resolver's rule index. For vendored dependency resolution, we can keep
the existing behavior.

### DIFF
--- a/language/go/resolve.go
+++ b/language/go/resolve.go
@@ -166,16 +166,16 @@ func resolveGo(c *config.Config, ix *resolve.RuleIndex, rc *repo.RemoteCache, r 
 		return label.New("bazel_gazelle", pkg, "go_default_library"), nil
 	}
 
+	if gc.depMode == externalMode {
+		return resolveExternal(rc, imp)
+	}
+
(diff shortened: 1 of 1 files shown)


language/go/resolve.go
```diff
@@ -166,16 +166,16 @@ func resolveGo(c *config.Config, ix *resolve.RuleIndex, rc *repo.RemoteCache, r
 		return label.New("bazel_gazelle", pkg, "go_default_library"), nil
 	}
 
+	if gc.depMode == externalMode {
+		return resolveExternal(rc, imp)
```
siddharthab: I can also fall back to current behavior if external dependency resolution fails for some reason. Let me know.

PR #424, opened 2019-01-17
Update known proto imports for Google APIs

jayconrod:


jayconrod:
cc @vam-google, FYI I'm getting ready to update `@go_googleapis` one last time before migration.

It looks like there's a bunch of new stuff in `google/ads/googleads/v0`. There's already generated code for it in `google.golang.org/genproto`, so I think they need to stay in sync.

language/proto/gen/gen_known_imports.go
```diff
@@ -131,5 +132,18 @@ func run(args []string) (err error) {
...
 		return err
 	}
-	return ioutil.WriteFile(knownImportsPath, knownImportsBuf.Bytes(), 0666)
+
+	if err := ioutil.WriteFile(knownImportsPath, knownImportsBuf.Bytes(), 0666); err != nil {
+		return err
+	}
+
+	if gofmtPath, err := exec.LookPath("gofmt"); err == nil {
+		// gofmt won't be available when run by Bazel, but it doesn't matter;
```
vam-google: `gofmt` is now available (added in one of the recent PRs). The path may be different in cases when it is executed with/wothout bazel, but it still should be available (so maybe put 2 conditions here, one trying to get regular gofmt, another one get the one available via bazel.

jayconrod: Even better: we can just use `go/format.Source` and not shell out to `gofmt` in either case.

PR #426, opened 2019-01-18
Add ability to update proto.csv by results of the bazel query.

vam-google:
This is a prerequisite to `go_googleapis` -> `com_google_googleapis` repository migration.

Since `com_google_googleapis` repository is Bazel-aware and declares its own `proto_library` and `go_proto_library` targets, we cannot just traverse the file tree of the repository and assume that there will be a corresponding `proto_library` and `go_proto_library` target for each found .proto file in a predefined format (the existing `BUILD.bazel` files in `com_google_googleapis` decide what the actual targets are and if they exist or no).

This PR provides an ability to construct `protos.csv` file based on the results of bazel query command, which outputs the existing `proto_library` and `go_proto_library` targets in `com_google_googleapis` repository.

To run the bazel query (and generate `googelapis.xml`) switch to `googleapis` repo directory and run:
```bash
bazel query 'kind("proto_library rule", //google/...)' --output xml --experimental_enable_repo_mapping > /path/to/googleapis/googleapis.xml
```
To generate `protos.csv` from the generated `googleapis.xml`, switch to `language/proto/gen` directory (in gazelle workspace) and run:
```bash
go run update_proto_csv.go update_proto_csv_from_query.go -proto_csv=../proto.csv -googleapis_xml=/path/to/googleapis/googleapis.xml
```
To generate `protos.csv` in the "old way", switch to `language/proto/gen` directory (in gazelle workspace) and run:
```bash
go run update_proto_csv.go update_proto_csv_from_query.go -proto_csv=../proto.csv -go_googleapis=/path/to/googleapis/repo
```


language/proto/gen/update_proto_csv_from_query.go
```diff
@@ -0,0 +1,212 @@
...
+		label := rule.Labels[j]
+		if "proto" == label.Name {
+			protoLabel = label.Value
+			break
+		}
+	}
+
+	goLabel := repoPrefix + rule.Name
+
+	if len(protoLabel) > 0 {
```
jayconrod: `if protoLabel != "" {`

language/proto/gen/update_proto_csv_from_query.go
```diff
@@ -0,0 +1,212 @@
...
+	xmlVersion := byteValue[:19]
+	if string(xmlVersion) == `<?xml version="1.1"` {
+		copy(xmlVersion, `<?xml version="1.0"`)
+	}
+
+	err = xml.Unmarshal(byteValue, &query)
+	return query, err
+}
+
+func processQuery(query Query, repoPrefix string) (relPathList []string, err error) {
```
jayconrod: I'm a bit unsure of the format of the data in `relPathList` and `protoLabelMap`.

The final table should have one row per .proto file. So I think you need two maps:

* A map from .proto source files to labels of the `proto_library` targets that import them.
* A map from labels of `proto_library` targets to the labels of the `go_proto_library` targets that depend on them.

Those maps can be constructed by separate functions, independently.

Once you have them, you can build a list of keys from the first map, sort it, then print each row of the final table (effectively inner-joining the maps on proto label).

vam-google: I don't think we need a second map here. We can construct both, but after we constructed the second map, the only thing we would do with it is to iterate over it. Thus no need to construct the second map, we can just process the items of the "second map" on the fly, without storing them.

The generated final table has one row per `.proto` file. The only way it can have multiple rows per proto file if there is more than one `go_proto_library` rule declared for that file. Note, we do not control how the targets are created, so some of the older assumptions do not apply anymore. I actually do not know what to do in this case, as I think it is a non-supported scenario for gazelle as of now. It is also very unlikely to happen to be honest. The code above for such case will output multiple rows per same proto file (because it had more than one `go_proto_library` declared for it, so it would be a correct representation of what's inside the googleapis repo).

The generation strategy assumes that we need to handle only those protos, which have both targets declared for them `proto_library` and, transitively `go_proto_library`.  Thus, if you try to run this command against the current googleapis repo, it will output empty protos.csv (no `go_proto_library` rules are there yet).

The generation algorithm does the following: 
1. On first pass, process all `go_proto_library` rules. A `.proto` has a chance to be mentioned in `proto.csv` only in case if it has a corresponding `go_proto_library`. The protos without `go_proto_library` rules will be ignored. During this pass we are able to construct three of the four components of each line in proto.csv (2nd, 3rd and 4th). 
    1. 2nd - `proto` (or one for each element in `protos`) argument value (this is also the key in the constructed map)
    2. 3rd - `importpath` argument value
    3. 4th - the name of this `go_proto_library`

    The only missing piece is the path to the actual .proto file (1st component, the only "unique" in items in the `proto.csv` file). 

2. To find the path to the .proto files, on second pass process all `srcs` arguments of all `proto_library` targets and and map them to a corresponding `go_proto_library` targets (by using the map, constructed on step 1). If no corresponding `go_proto_library` target was found, ignore the .proto file, otherwise use the value of the `srcs` item as the 1st element in proto.csv row. At this point we have all 4 components of proto.csv row and ready to output. 
3) At the end of everythin we also sort the result (not sure if it is necessary, but it makes proto.csv look more like it was created from file tree traversal and more organized in general).

vam-google:
@jayconrod PTAL (also please check [my response](https://github.com/bazelbuild/bazel-gazelle/pull/426#discussion_r249220522) for more details about what is happening here)

language/proto/gen/update_proto_csv.go
```diff
@@ -66,25 +72,225 @@ func main() {
...
+	if bytes.HasPrefix(xmlData, []byte(`<?xml version="1.1"`)) {
+		copy(xmlData, []byte(`<?xml version="1.0"`))
+	}
+	query := Query{}
+	err := xml.Unmarshal(xmlData, &query)
+	return query, err
+}
+
+func processQuery(query Query) ([]string, error) {
+	protoLabelMap := make(map[string]string)
```
jayconrod: I'm a little uncomfortable with storing partially rendered information in here. Could you define a new type to use as the value? I think that would make this more readable. Something like:

```
type goProtoLibraryInfo struct {
  label, importpath string
}
```

Also, please add a comment explaining this maps `proto_library` labels to information about `go_proto_library` rules. I don't think that's clear from the name or the type right now.

language/proto/gen/update_proto_csv.go
```diff
@@ -66,25 +72,225 @@ func main() {
...
+	Value   string   `xml:"value,attr"`
+}
+
+type String struct {
+	XMLName xml.Name `xml:"string"`
+	Name    string   `xml:"name,attr"`
+	Value   string   `xml:"value,attr"`
+}
+
+func generateFromQuery(bazelPath, workspacePath string, protoContent io.Writer) error {
```
jayconrod: Nit: rename `protoContent` to `w`, and make it the first parameter.

language/proto/gen/update_proto_csv_from_query.go
jayconrod: This general approach sounds fine to me. One question though: will there be any `proto_library` targets in the wfuture (after all the rules are in place) that don't have corresponding `go_proto_library` targets but need to be available as dependencies for outside `proto_libraries`? If so, we may want to consider a more flexible approach (but it doesn't need to be in this PR).

language/proto/gen/update_proto_csv.go
```diff
@@ -66,25 +72,225 @@ func main() {
...
+			relPathList = append(relPathList, paths...)
+		}
+	}
+
+	sort.Strings(relPathList)
+	return relPathList, nil
+}
+
+func processGoProtoLibraryRule(rule Rule) (map[string]string, error) {
+	protoLabelMap := make(map[string]string)
```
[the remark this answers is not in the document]
vam-google: I would prefer leaving this as is. Using output parameters (like in this case, where we would modify a map passed as a parameter) makes it harder to control state (and, for example, while it is very common for languages like C, it is undesirable in Java, probably is considered ok in Go). Creating the new map has lightly worse performance but performance is not important here, at the same time it makes the method easier to use/test as a standalone entity (it is not dependent on initialization of the map by the outside code and does not have implicit assumption that the output map values hava a specific format).

language/proto/gen/update_proto_csv_from_query.go
vam-google: I'm not sure what you mean by the outside `proto_libraries`, but I want to say "No" (in general I think it is a pretty safe approach). All the core libraries will have `go_proto_library` defined for them, as well as all API-specific protos which choose to have Golang gapic client (which should be ALL apis, except the cases when the lib configuration is till "in progress").

language/proto/gen/update_proto_csv.go
vam-google: Changed (and also changed the `generateFromPath` signature for consistency). In general, the naming for many of the params/variables duplicate the original version (i.e. the `generateByPath` version.

language/proto/gen/update_proto_csv.go
vam-google: Added

vam-google:
@jayconrod PTAL

PR #434, opened 2019-01-28
Add new directives: proto_strip_import_prefix and proto_import_prefix

blico:
Bazel's `proto_library` rule now supports the `strip_import_prefix` and `import_prefix` attributes:
https://docs.bazel.build/versions/master/be/protocol-buffer.html

This change adds new directives that enable Gazelle to set the `strip_import_prefix` and `import_prefix` attributes when it generates `proto_library` rules.


README.rst
```diff
@@ -335,6 +335,23 @@ The following flags are accepted:
...
+| :flag:`-proto_strip_import_prefix example.com/repo`          |                                   |
++--------------------------------------------------------------+-----------------------------------+
+| The prefix to strip from the paths of the .proto files in the proto_library rule.                |
+|                                                                                                  |
+| When set, .proto source files in the srcs attribute of the rule are accessible at their path     |
+| with this prefix cut off. If it's a relative path (not starting with a slash), it's taken as a   |
+| package-relative one. If it's an absolute one, it's understood as a repository-relative path.    |
+| The prefix in the import_prefix attribute is added after this prefix is stripped.                |
++--------------------------------------------------------------+-----------------------------------+
+| :flag:`import_prefix repo`                                   |                                   |
```
jayconrod: Should be `-proto_import_prefix`

language/proto/kinds.go
```diff
@@ -20,7 +20,11 @@ import "github.com/bazelbuild/bazel-gazelle/rule"
 var protoKinds = map[string]rule.KindInfo{
 	"proto_library": {
 		NonEmptyAttrs:  map[string]bool{"srcs": true},
-		MergeableAttrs: map[string]bool{"srcs": true},
+		MergeableAttrs: map[string]bool{
+			"srcs": 						 		true,
```
jayconrod: Something weird happened with the whitespace. Please run `gofmt`.

Also, I'm not entirely sure if these should be mergeable. That means Gazelle will override user values without `# keep` comments. It will delete existing attributes if the directives aren't set.

README.rst
```diff
@@ -335,6 +335,23 @@ The following flags are accepted:
 | Determines the proto option Gazelle uses to group .proto files into rules                        |
 | when in ``package`` mode. See details in `Directives`_ below.                                    |
 +--------------------------------------------------------------+-----------------------------------+
+| :flag:`-proto_strip_import_prefix example.com/repo`          |                                   |
```
jayconrod: Nit: change the value from `example.com/repo` to `vendor`. I think that will be the most common use for this flag.

README.rst
```diff
@@ -335,6 +335,23 @@ The following flags are accepted:
...
++--------------------------------------------------------------+-----------------------------------+
+| The prefix to strip from the paths of the .proto files in the proto_library rule.                |
```
jayconrod: In the text here and for the attribute below, just say something like "Sets the `import_prefix` attribute of generated `proto_library` rules. This is a prefix to add to import paths of .proto files.". Make sure to link `import_prefix` to `https://docs.bazel.build/versions/master/be/protocol-buffer.html#proto_library.import_prefix`. No need to give much more detail than that.

README.rst
```diff
@@ -335,6 +335,23 @@ The following flags are accepted:
...
+| The prefix in the import_prefix attribute is added after this prefix is stripped.                |
++--------------------------------------------------------------+-----------------------------------+
+| :flag:`import_prefix repo`                                   |                                   |
++--------------------------------------------------------------+-----------------------------------+
+| The prefix to add to the paths of the .proto files in the proto_library rule.                    |
+|                                                                                                  |
+| When set, the .proto source files in the srcs attribute of the rule are accessible as the value  |
+| of this attribute prepended to their repository-relative path.                                   |
+| The prefix in the strip_import_prefix attribute is removed before this prefix is added.          |
++--------------------------------------------------------------+-----------------------------------+
```
jayconrod: Please also describe the new directives in the Directives section.

cmd/gazelle/integration_test.go
```diff
@@ -751,6 +751,81 @@ go_library(
...
+		}, {
+			Path: "foo.proto",
+			Content: `syntax = "proto3";`,
+		},
+	}
+
+	dir, cleanup := testtools.CreateFiles(t, files)
+	defer cleanup()
+
+	args := []string{"update","-go_prefix", "example.com/repo",
```
jayconrod: Spacing doesn't look quite right here. Please run `gofmt`.

language/proto/config.go
```diff
@@ -46,6 +46,16 @@ type ProtoConfig struct {
 	// groupOption is an option name that Gazelle will use to group .proto
 	// files into proto_library rules. If unset, the proto package name is used.
 	groupOption string
+
+	// StripImportPrefix The prefix to strip from the paths of the .proto files.
```
jayconrod: Do these need to be exported?

blico: good catch, sorry I am new to Go

language/proto/kinds.go
blico: The downside I see to making these attributes mergeable is repos that are already using the attributes will either have to start using the directives, or add a lot of `# keep` comments. The benefit of the attributes being mergeable is it makes it easier to mass edit them.

blico:
Updated `Imports` to be aware of the added attributes and added tests.

README.rst
```diff
@@ -335,22 +337,15 @@ The following flags are accepted:
...
-|                                                                                                  |
-| When set, .proto source files in the srcs attribute of the rule are accessible at their path     |
-| with this prefix cut off. If it's a relative path (not starting with a slash), it's taken as a   |
-| package-relative one. If it's an absolute one, it's understood as a repository-relative path.    |
-| The prefix in the import_prefix attribute is added after this prefix is stripped.                |
+| Sets the `strip_import_prefix`_ attribute of generated ``proto_library`` rules. This is a        |
+| prefix to strip from the import paths of .proto files.                                           |
 +--------------------------------------------------------------+-----------------------------------+
-| :flag:`import_prefix repo`                                   |                                   |
+| :flag:`proto_import_prefix repo`                             |                                   |
```
jayconrod: Nit: there should be a `-` in front of the flag name.

language/proto/kinds.go
```diff
@@ -19,9 +19,13 @@ import "github.com/bazelbuild/bazel-gazelle/rule"
 
 var protoKinds = map[string]rule.KindInfo{
 	"proto_library": {
-		NonEmptyAttrs:  map[string]bool{"srcs": true},
-		MergeableAttrs: map[string]bool{"srcs": true},
-		ResolveAttrs:   map[string]bool{"deps": true},
+		NonEmptyAttrs: map[string]bool{"srcs": true},
+		MergeableAttrs: map[string]bool{
+			"srcs":                true,
```
jayconrod: Quoting the previous thread (I really wish GitHub would stop losing comments):

> The downside I see to making these attributes mergeable is repos that are already using the attributes will either have to start using the directives, or add a lot of # keep comments. The benefit of the attributes being mergeable is it makes it easier to mass edit them.

It's not clear to me which side of this we should be on. If these attributes already had widespread use, I'd say let's avoid modifying them if they're already set. But since they are new, maybe mergeable is okay. Still makes me nervous.

For reference, existing attributes can be bulk-deleted or modified with [buildozer](https://github.com/bazelbuild/buildtools/tree/master/buildozer). I don't think there's any way to bulk-add `# keep` comments though.

language/proto/resolve.go
```diff
@@ -34,7 +34,22 @@ func (_ *protoLang) Imports(c *config.Config, r *rule.Rule, f *rule.File) []reso
 	rel := f.Pkg
 	srcs := r.AttrStrings("srcs")
 	imports := make([]resolve.ImportSpec, len(srcs))
+	pc := GetProtoConfig(c)
 	for i, src := range srcs {
+		// If strip_import_prefix is a relative path (not starting with a slash), it's taken as a package-relative one.
```
jayconrod: `rel` shouldn't be modified. Instead, determine the prefix to strip and the prefix to add outside the of the loop. Within the loop, try to strip the old prefix and add the new prefix.

If `rel` doesn't start with the prefix to strip, skip indexing these files. An error should be reported by the configurer when the prefix is set.

language/proto/config.go
```diff
@@ -201,6 +213,10 @@ func (_ *protoLang) Configure(c *config.Config, rel string, f *rule.File) {
 				pc.ModeExplicit = true
 			case "proto_group":
 				pc.groupOption = d.Value
+			case "proto_strip_import_prefix":
+				pc.stripImportPrefix = d.Value
+			case "proto_import_prefix":
+				pc.importPrefix = d.Value
```
jayconrod: You mentioned validation in another comment. I think we should validate `proto_strip_import_prefix` but not `proto_import_prefix`.

Let's require that `proto_strip_import_prefix` is absolute. It should have a leading slash, and the path should be a prefix of the package name. I don't think relative paths make sense for Gazelle: if I understand correctly, they would only be useful if your .proto files are in a subdirectory of the directory where the build file is. Gazelle won't generate rules like that.

If that's not the case, we should print an error here in `Configure`.

language/proto/config.go
```diff
@@ -174,14 +184,16 @@ func (_ *protoLang) RegisterFlags(fs *flag.FlagSet, cmd string, c *config.Config
 	// this is set for compatibility with older versions.
 	fs.Var(&modeFlag{&pc.Mode}, "proto", "default: generates a proto_library rule for one package\n\tpackage: generates a proto_library rule for for each package\n\tdisable: does not touch proto rules\n\tdisable_global: does not touch proto rules and does not use special cases for protos in dependency resolution")
 	fs.StringVar(&pc.groupOption, "proto_group", "", "option name used to group .proto files into proto_library rules")
+	fs.StringVar(&pc.stripImportPrefix, "proto_strip_import_prefix", "", "When set, .proto source files in the srcs attribute of the rule are accessible at their path with this prefix cut off.")
```
jayconrod: If we require the strip prefix to be absolute, we should probably drop this flag and rely on the directive. There would be nothing to strip in the root directory where this would apply.

PR #436, opened 2019-01-31
(Start to) Add typescript support (hugely wip)

petemounce:
/wip

I'd like extremely early feedback on whether this is the right sort of direction to go to start adding typescript language support into gazelle?

### DIFF
--- a/language/typescript/config.go
+++ b/language/typescript/config.go
@@ -1,0 +1,44 @@
+package typescript
+
+import (
+	"flag"
+	"github.com/bazelbuild/bazel-gazelle/config"
+	"github.com/bazelbuild/bazel-gazelle/rule"
+)
+
+// typescriptConfig contains configuration values related to Typescript rules.
+type typescriptConfig struct {
+}
+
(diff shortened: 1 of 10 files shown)


petemounce:
@jayconrod cheers; I'll coordinate over there.

Would you welcome a PR that improves the documentation for how to add another language to gazelle? I think that would probably include:
* how to bootstrap - i.e. an illustration of a suggested set of files, and a suggested testing approach (I like the generate-and-verify-against-pristine method here and was planning that)
  * Expand the TODO on https://github.com/bazelbuild/bazel-gazelle/blob/master/extend.rst#example basically
* how to then consume that inside gazelle if it's present in a different repo (I agree, that makes the most sense)
  * probably a couple of examples on https://github.com/bazelbuild/bazel-gazelle/blob/master/extend.rst#gazelle_binary

petemounce:
@alexeagle it exists in some form already? That's great!

alexeagle:
the ts_auto_deps code is already in a top-level folder in rules_typescript so that it can be re-used as a library. it would be great to see how it can be driven by gazelle.

PR #446, opened 2019-02-09
Merge hasDefaultVisibility and checkInternalVisibility functions

linzhp:
Merge redundant implementations of `hasDefaultVisibility` and  `checkInternalVisibility  ` functions to reduce redundancy.


language/go/generate.go
```diff
@@ -371,9 +359,9 @@ type generator struct {
 	shouldSetVisibility bool
 }
 
-func newGenerator(c *config.Config, f *rule.File, rel string) *generator {
-	shouldSetVisibility := f == nil || !hasDefaultVisibility(f)
-	return &generator{c: c, rel: rel, shouldSetVisibility: shouldSetVisibility}
+func newGenerator(c *config.Config, args language.GenerateArgs) *generator {
```
jayconrod: Nit: Let's actually delete `newGenerator` and just construct it with a literal.

I was thinking `newGenerator` and `language.GenerateArgs` should not be coupled, but if we pass in everything as parameters, it's a trivial constructor.

language/lang.go
```diff
@@ -129,6 +129,21 @@ type GenerateArgs struct {
 	OtherEmpty, OtherGen []*rule.Rule
 }
 
+// HasDefaultVisibility returns whether oldFile contains a "package" rule with
+// a "default_visibility" attribute. Rules generated by Gazelle should not
+// have their own visibility attributes if this is the case.
+func (args GenerateArgs) HasDefaultVisibility() bool {
```
jayconrod: I think this should be an exported method of `rule.File` instead. That would make it more generally useful outside of the `Generate` method. I could imagine it being used in the `Resolve` method for extensions that want to consider visibility.

PR #447, opened 2019-02-13
Add build options to update-repos

### LINKED ISSUE #175: update-repos: accept -build_file_name, -build_file_generation and other options
When these options are provided to `update-repos`, it should generate or update `go_repository` rules with the corresponding attributes. For example:

```
$ gazelle update-repos -build_file_name=BUILD.bazel github.com/pkg/errors
```

should create the rule:

```
go_repository(
    name = "com_github_pkg_errors",
    importpath = "github.com/pkg/errors",
    build_file_name = "BUILD.bazel",
    commit = "123456",
)
```

If the rule exists, the `build_file_name` attribute should be changed to match the command line.

This should work for `-from_file` as well.

blico:
Resolves #175 

Added all of `go_repository`'s attributes used for automatic build file generation as `update-repos` options.

Setting these options via `update-repos` will generate or update `go_repository` rules with the corresponding attributes.


jayconrod:
I think this is a valuable feature and a sound implementation, but I'm not sure I want to expand the language-specific command line interface to `update-repos`. I need to think about it a bit more.

`update-repos` was added before the language-specific logic was pulled out into separate extensions behind the `language` interface. `update-repos` is basically all Go-specific logic at the moment, but ideally the `language` interface should be expanded to allow `update-repos` to work with any language. That work has been on hold, mainly because I want to see what Bazel's eventual solution is for recursive WORKSPACE evaluation, since that may change everything.

So these flags would eventually need to move to the Go extension, but ideally in a way that doesn't break the command line interface. Give me a couple days to think about what this will look like. I'll get back to you soon.

cmd/gazelle/update-repos.go
```diff
@@ -32,9 +32,39 @@ import (
...
+	importPaths             []string
+	buildExternalAttr       string
+	buildFileNamesAttr      string
+	buildFileGenerationAttr string
+	buildTagsAttr           string
+	buildFileProtoModeAttr  string
+	buildExtraArgsAttr      string
+}
+
+func checkValidBuildExternalAttr(attr string) bool {
```
jayconrod: Could you extract these functions into an implementation of the `flag.Value` interface and move into the `//flag` package? Something like this would be good:

```
type AllowedStringFlag struct {
  Allowed []string
  Value *string
}

func (f *AllowedStringFlag) Set(v string) error {
  for _, a := range allowed {
    if v == a {
      *f.Value = v
      return nil
    }
  }
  return fmt.Errorf("invalid argument %q. Possible values are: %s", v, strings.Join(f.Allowed, ", "))
}
```

PR #448, opened 2019-02-15
gazelle:map_kind directive

robfig:
Well, I'm not sure this is the cleanest code ever, but it was the smallest change I could come up with to implement the use case described in https://github.com/bazelbuild/rules_go/issues/1942. 

I'm not sure if it's ok to omit `go_binary` from `Kinds()` for indexing purposes, but it seems to work. That is, it does not modify existing `go_binary` rules, but it does create new ones as `go_deployable` with the appropriate `load`

Tested manually by running this on our codebase:

```
gazelle(
    name = "gazelle",
    extra_args = [
        "-go_binary=go_deployable", 
        "-go_binary_load=//tools/go/def.bzl", 
        "gocode/src/corp"],
)
```

If this functionality seems like something you'd like to support, let me know and I can clean it up. If not, no worries.


jayconrod:
Let's decide how this should work before getting to implementation details.

From my perspective, I think there should be a directive like:

```
# gazelle:map_kind go_binary go_deployable //tools/go/def.bzl
```

The first value (`go_binary`) would be the kind to replace. The second value (`go_deployable`) would be the kind to replace it with. The third value (`//tools/go:def.bzl`) would be the file to load it from. There would be no way to configure this on the command line; it would only be a directive.

This mapping should be language and kind-agnostic. It should be possible for fix-update.go to call `Generate` and `Kinds` methods and fix up the results using these directives.

I think this would fix #99.

robfig:
I have an implementation of that idea that seems to work. 

It still requires some more work:
- A test in integration_test.go
- Documentation
- I haven't figured out why this caused TestSubdirectoryPrefixExternal to start failing

but I'd like some feedback before continuing, to be sure this is on the right track.

language/proto/generate.go
```diff
@@ -35,23 +35,11 @@ func (_ *protoLang) GenerateRules(args language.GenerateArgs) language.GenerateR
...
-	var regularProtoFiles []string
-	for _, name := range args.RegularFiles {
-		if strings.HasSuffix(name, ".proto") {
-			regularProtoFiles = append(regularProtoFiles, name)
-		}
-	}
-	var genProtoFiles []string
-	for _, name := range args.GenFiles {
-		if strings.HasSuffix(name, ".proto") {
-			genProtoFiles = append(args.GenFiles, name)
```
robfig: @jayconrod this looks like it might be a bug. It should be 

```go
genProtoFiles = append(genProtoFiles, name) 
```

right?

robfig:
RE TestSubdirectoryPrefixExternal, I figured out that I accidentally dropped the `c -> v.c` fix while merging. That's sorted now.

language/go/generate.go
```diff
@@ -41,7 +41,8 @@ func (gl *goLang) GenerateRules(args language.GenerateArgs) language.GenerateRes
 	protoPackages := make(map[string]proto.Package)
 	protoFileInfo := make(map[string]proto.FileInfo)
 	for _, r := range args.OtherGen {
-		if r.Kind() != "proto_library" {
+		// TODO: Apply kind mapping here (and similar tests below)?
```
jayconrod: I think this change should be made in a way that doesn't require any collaboration from the language extensions. Go shouldn't need to know the mapped kind for `proto_library`, and a bug in Go shouldn't cause it to emit `go_binary` instead of the mapped kind.

Instead, maybe something in fix-update.go could call `Rule.SetKind` on the result of `GenerateRules`, before the rules are merged into files? That would allow language extensions to generate rules normally, and they would see the unmapped rules generated by other extensions.

language/proto/generate.go
jayconrod: Yep, looks like a bug. Could you open a separate issue for this?

resolve/index.go
```diff
@@ -109,9 +112,16 @@ func NewRuleIndex(kindToResolver map[string]Resolver) *RuleIndex {
...
 // AddRule may only be called before Finish.
 func (ix *RuleIndex) AddRule(c *config.Config, r *rule.Rule, f *rule.File) {
-	var imps []ImportSpec
-	if rslv, ok := ix.kindToResolver[r.Kind()]; ok {
+	var (
+		imps []ImportSpec
+		rslv Resolver
+		ok   bool
+	)
+	if rslv, ok = ix.kindToResolver[r.Kind()]; ok {
```
jayconrod: I think we should replace `RuleIndex.kindToResolver` with a callback of type `func(r *rule.Rule, f *rule.File) Resolver` implemented in fix-update.go. It will use `r.Kind()` and `f.Pkg`.

This should make the `rule.resolver` field and `RuleIndex.Resolver` method unnecessary.

cmd/gazelle/fix-update.go
```diff
@@ -195,8 +195,15 @@ func runFixUpdate(cmd command, args []string) error {
 
 	// Visit all directories in the repository.
 	var visits []visitRecord
+	var replKinds = map[config.ReplacementKind]struct{}{} // record all known ReplacementKinds
```
jayconrod: This doesn't have enough information. `map_kind` directives should not be global; they should only be effective in the directory where they are set and in subdirectories.

language/proto/generate.go
robfig: Opened https://github.com/bazelbuild/bazel-gazelle/pull/455

language/go/generate.go
robfig: Great point, very much agree with that. Will do.

cmd/gazelle/fix-update.go
robfig: Yeah, good point. I'll add it to the visitRecord.

resolve/index.go
robfig: OK - Note that FindRulesByImport does not have a relevant *rule.File to pass to kindToResolver, but that can work out if we specify that a nil *rule.File is permitted but ignores any replacements.

go.sum
```diff
@@ -6,6 +6,7 @@ github.com/bazelbuild/buildtools v0.0.0-20190122172533-1f3678932acd h1:hxq17QUlh
 github.com/bazelbuild/buildtools v0.0.0-20190122172533-1f3678932acd/go.mod h1:5JP0TXzWDHXv8qvxRC4InIazwdyDseBDbzESUMKk1yU=
 github.com/bazelbuild/buildtools v0.0.0-20190202002759-027686e28d67 h1:zS8p6ZRbNVa7QfK3tpoIRDqGzCA2J0uJffaMTWoneac=
 github.com/bazelbuild/buildtools v0.0.0-20190202002759-027686e28d67/go.mod h1:5JP0TXzWDHXv8qvxRC4InIazwdyDseBDbzESUMKk1yU=
+github.com/bazelbuild/rules_go v0.0.0-20190219171851-bb5873bebd9f h1:jLEQofdYNm02DaBJCnfMlo9gDrmHwWxjJJFeCubwtxg=
```
robfig: (I'm not sure what caused this to be added)

resolve/index.go
robfig: I implemented a version that used a regular func, but in the course of cleaning it up I thought it was better to have a type for it instead. Otherwise it was passing in a method of type, which is a little unusual, and I didn't perceive a big advantage of having RuleIndex avoid the explicit dependency on that type.

robfig:
Refactor should be good, working on tests and doc updates now.

robfig:
I believe this is good to go, please have a look.  The CLA is still working its way through our legal department, but hopefully not much longer.

Out of curiosity, do you generate that neat ASCII table in the README.rst by hand, or is there some tool that makes it less demanding on the spacebar?

config/config.go
```diff
@@ -74,13 +75,23 @@ type Config struct {
 	// libraries in the workspace for dependency resolution
 	IndexLibraries bool
 
+	// KindMap maps from a kind name to its replacement. It provides a way for
+	// users to customize the kind of rules created by Gazelle, via
+	// # gazelle:map_kind.
+	KindMap map[string]ReplacementKind
```
jayconrod: (No change needed, just a comment):

I was thinking about whether it makes sense for this to be here or in `updateConfig`. I think it makes sense here. We may want to make `update-repos` observe the directive within `WORKSPACE`. That can be a TODO though.

config/config.go
```diff
@@ -201,16 +216,33 @@ func (cc *CommonConfigurer) CheckFlags(fs *flag.FlagSet, c *Config) error {
...
+		case "map_kind":
+			vals := strings.Fields(d.Value)
+			if len(vals) != 3 {
+				log.Printf("expected three arguments (gazelle:map_kind from_kind to_kind load_file), got %v", vals)
+				continue
+			}
+			if c.KindMap == nil {
+				c.KindMap = make(map[string]ReplacementKind)
+			}
+			var (
```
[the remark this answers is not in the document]
robfig: Done!

resolve/metaresolver.go
```diff
@@ -0,0 +1,48 @@
+package resolve
```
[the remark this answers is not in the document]
robfig: Done

cmd/gazelle/integration_test.go
```diff
@@ -1754,6 +1754,129 @@ go_library(
 	})
 }
 
+// TestMapKind tests these gazelle:map_kind directive scenarios:
+// 1. Disabled: // and //sub2
```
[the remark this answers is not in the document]
robfig: Even better! Done

cmd/gazelle/fix-update.go
```diff
@@ -146,6 +146,10 @@ type visitRecord struct {
 
 	// file is the build file being processed.
 	file *rule.File
+
+	// replKinds are mapped kinds used during this visit.
```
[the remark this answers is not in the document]
robfig: 👍

config/config.go
```diff
@@ -74,13 +75,23 @@ type Config struct {
...
+	KindMap map[string]ReplacementKind
+
 	// Exts is a set of configurable extensions. Generally, each language
 	// has its own set of extensions, but other modules may provide their own
 	// extensions as well. Values in here may be populated by command line
 	// arguments, directives in build files, or other mechanisms.
 	Exts map[string]interface{}
 }
 
+// ReplacementKind describes a replacement to use for a built-in kind.
```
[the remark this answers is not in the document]
robfig: Agreed, will do

README.rst
```diff
@@ -543,6 +543,20 @@ The following directives are recognized:
...
+|                                                                                            |
+| As a separate step after generating rules, any new rules of kind ``from_kind`` have their  |
+| kind replaced with ``to_kind``. This means that ``to_kind`` must accept the same           |
+| parameters and behave similarly.                                                           |
+|                                                                                            |
+| Most commonly, this would be used to replace the rules provided by ``rules_go`` with       |
+| custom macros. For example,                                                                |
+| ``gazelle:map_kind go_binary go_deployable //tools/go:def.bzl`` would configure Gazelle to |
+| produce rules of kind ``go_deployable`` as loaded from ``//tools/go:def.bzl`` instead of   |
+| ``go_binary``, for this directory or within.                                               |
```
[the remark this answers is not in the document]
robfig: Yah, good point. Updated.

Thanks for the color on ReST.

resolve/metaresolver.go
```diff
@@ -0,0 +1,48 @@
...
+
+import (
+	"github.com/bazelbuild/bazel-gazelle/config"
+	"github.com/bazelbuild/bazel-gazelle/rule"
+)
+
+// MetaResolver provides a rule.Resolver for any rule.Rule.
+type MetaResolver struct {
```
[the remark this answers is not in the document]
robfig: Makes sense, and it's easy enough. Will do

config/config.go
robfig: Yeah, the decision didn't seem super obvious to me either. SGTM.

Do you want me to add the TODO here in the code, or did you just mean as a mental note? (I usually don't add TODOs unless there are near-term plans to implement, since they tend to build up.)

cmd/gazelle/integration_test.go
```diff
@@ -1754,6 +1754,129 @@ go_library(
...
+// 2. Enabled: //sub1
+// 3. Inherited to a sub-directory: //sub1/sub11
+// 4. Overridden by a sub-directory: //sub1/sub12
+//
```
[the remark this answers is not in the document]
robfig: I added these tests, and I noticed that case 5 does not actually work -- a kind mapped from go_library does not have its srcs updated based on source files present.  I'll diagnose.. 

BTW, it's great how easy these tests are to write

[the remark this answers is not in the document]
robfig: It identified a bug, so good call on those test cases.

robfig:
Thanks for the review. I believe I've addressed everything. 

I'm not sure if I'm doing something wrong; I've been bugging the CLA bot on the other PR to avoid polluting this one:
https://github.com/bazelbuild/bazel-gazelle/pull/455

Maybe it's silent because the company signature is not yet recorded or something. I'll give it another few days, but then I think I'll need to escalate.

achew22:
FYI, the CLA bot for corporate CLAs is actually a human who has to physically validate the signature and company line up. It might take a few days to get the CLA verified, but once that happens your entire domain will be good to go. Terribly sorry it's taking so long!

jayconrod:
Hmm, I've verified that Yext has signed the Corp CLA and that the commit author email is `@yext.com`. @robfig Could you confirm:

* The same email is associated with your GitHub account (may need to be public, at least visible to the bot)
* It's part of the "authorized contributor" Google group associated with the CLA

If all that's good, I'll ping the CLA folks and see what's up.

Sorry to be taking so long on this! Usually it goes more smoothly.

robfig:
No worries at all! I figured there was a manual process involved.

I signed it!

jayconrod:
Hooray! :tada: Sorry to make you jump through all the hoops. Thank you again for working on this.

We do try to keep our legal friends happy though. :woman_judge:

robfig:
Woo hoo! Thanks for the thorough review and incorporating this stuff!

PR #450, opened 2019-02-16
go_repository: support modules

### LINKED ISSUE #400: go_repository: support modules
When I try to build the example program [hdfscat](https://github.com/EvyBongers/hdfscat), I get an error that prevents a successful build.

When attempting to build the project, I get this error:
```
ERROR: Analysis of target '//cmd/hdfscat:hdfscat' failed; build aborted: no such package '@com_github_colinmarc_hdfs//v2/internal/protocol/hadoop_hdfs': BUILD file not found on package path
```

Reviewing the code, I found that the hdfs library uses [semantic import versioning](https://research.swtch.com/vgo-import). In my `WORKSPACE` file, I do specify to use version two of the library:
```
go_repository(
    name = "com_github_colinmarc_hdfs",
    importpath = "github.com/colinmarc/hdfs",
    sha256 = "36a039f827742b49ef96eab1906a5b5f35aa3715a4448608224193eada35ffd8",
    strip_prefix = "hdfs-2.0.0",
    urls = ["https://github.com/colinmarc/hdfs/archive/v2.0.0.tar.gz"],
)
```

However, I can't find any way to tell bazel that it should expose the library as package `@com_github_colinmarc_hdfs//v2/internal/protocol/hadoop_hdfs`.

Is there currently any way to solve this problem with bazel/gazelle?

jayconrod:
go_repository has two new attributes: version and sum. If version is
specified, go_repository will run in module mode (distinct from
repository mode and http mode). version and sum must be specified
together.

In module mode, go_repository invokes fetch_repo with -version and
-sum arguments. fetch_repo will run "go mod download" to retrieve the
module.

Fixes #400

### DIFF
--- a/cmd/fetch_repo/fetch_repo.go
+++ b/cmd/fetch_repo/fetch_repo.go
@@ -13,78 +13,88 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-// Command fetch_repo is similar to "go get -d" but it works even if the given
-// repository path is not a buildable Go package and it checks out a specific
-// revision rather than the latest revision.
+// Command fetch_repo downloads a Go module or repository at a specific
+// version or commit.
 //
-// The difference between fetch_repo and "git clone" or {new_,}git_repository is
-// that fetch_repo recognizes import redirection of Go and it supports other
-// version control systems than git.
+// In module mode, fetch_repo downloads a module using "go mod download",
+// verifies the contents against a sum, then copies the contents of the module
+// to a target directory. fetch_repo respects GOPATH and GOPROXY.
 //
-// These differences help us to manage external Go repositories in the manner of
-// Bazel.
+// In repository mode, fetch_repo clones a repository using a VCS tool.
+// fetch_repo performs import path redirection in this mode.
 package main
 
 import (
 	"flag"
-	"fmt"
 	"log"
 
 	"golang.org/x/tools/go/vcs"
 )
 
 var (
-	remote     = flag.String("remote", "", "The URI of the remote repository. Must be used with the --vcs flag.")
-	cmd        = flag.String("vcs", "", "Version control system to use to fetch the repository. Should be one of: git,hg,svn,bzr. Must be used with the --remote flag.")
-	rev        = flag.String("rev", "", "target revision")
-	dest       = flag.String("dest", "", "destination directory")
+	// Common flags
 	importpath = flag.String("importpath", "", "Go importpath to the repository fetch")
+	dest       = flag.String("dest", "", "destination directory")
 
-	// Used for overriding in tests to disable network calls.
-	repoRootForImportPath = vcs.RepoRootForImportPath
+	// Repository flags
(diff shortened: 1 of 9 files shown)


cmd/fetch_repo/fetch_repo.go
```diff
@@ -13,78 +13,88 @@ See the License for the specific language governing permissions and
...
-// repository path is not a buildable Go package and it checks out a specific
-// revision rather than the latest revision.
+// Command fetch_repo downloads a Go module or repository at a specific
+// version or commit.
 //
-// The difference between fetch_repo and "git clone" or {new_,}git_repository is
-// that fetch_repo recognizes import redirection of Go and it supports other
-// version control systems than git.
+// In module mode, fetch_repo downloads a module using "go mod download",
+// verifies the contents against a sum, then copies the contents of the module
```
Globegitter: I wonder if rather than copying we could just do a hardlink to save disk space? Bazel itself offers a flag for this for the repo cache: https://docs.bazel.build/versions/master/command-line-reference.html#flag--experimental_repository_cache_hardlinks & https://github.com/bazelbuild/bazel/issues/5568 for background.

Globegitter:
The other question is more generic and might just come from the fact that I have not worked with go modules yet, but how would one be able to provide a sum? Is that something that will be logged on first download and can then be added?

cmd/fetch_repo/fetch_repo.go
jayconrod: Hard links and symbolic links both cause compatibility issues. They aren't universally supported across operating systems and file systems. Windows especially has problems, and they cause problems with sandboxfs, too.

jayconrod:
You can get a sum by reading `go.sum` (if you already have a module that depends on that module) or by running `go mod download -json <mod>@<version>`. `fetch_repo` does the latter and just compares the sum reported by the toolchains against the given value.

cmd/fetch_repo/fetch_repo.go
Globegitter: Yeah I was thinking as much - I wonder if it is worth having some option for this? Maybe configurable via an env var, something like `BAZEL_GO_MODULE_LINK`? So that people can choose tradeoffs specifically to their setup consciously? But yeah also totally fair if you want to leave it as is for now.

jayconrod: It's more a question of cost of implementation and maintenance. I'm sure it would work if we were careful, but implementing it, testing on all platforms, and fixing bugs afterward would take time. I'd rather keep it simple, at least for now.

Globegitter:
Anything still missing here to get this in? I would love to start using this. Also does this support the `update-repos` command?

jayconrod:
@Globegitter Caching is the important missing piece. I'd like to ensure that:

1. When gazelle or fetch_repo sources are modified, those binaries are rebuilt.
2. When gazelle or fetch_repo sources are modified, we don't need to download all `go_repository` rules again. The rules may need to be re-executed, but we should download to a shared cache somewhere else.

`update-repos` is not supported yet. That will need to be another PR. I don't expect it will be too complicated though.

PR #479, opened 2019-03-09
Clarifies "ambiguous import" message

### LINKED ISSUE #478: Unclear error message: "multiple rules (%s and %s) may be imported with %q from %s"
At: https://github.com/bazelbuild/bazel-gazelle/blob/19b1a50291025cd8e5285cc7c510610e7aef10cb/language/go/resolve.go#L230

The error message "multiple rules (%s and %s) may be imported with %q from %s" is very unclearly formulated. It is not obvious where the problem is nor how to handle it.

It would be nice if gazelle explained what the options are in this case.

filmil:
The old error message is not very useful: it is difficult to scan
visually, once you scan it it is not clear what it means, it offers no
suggestions at all what to do to fix the issue, and is very difficult to
google a solution for.

The new error is somewhat more verbose, but uses clearer language
(where the issue is comes first, details later) and offers clear ways
to resolve the issue.

Handles #478 in part.

### DIFF
--- a/language/go/resolve.go
+++ b/language/go/resolve.go
@@ -227,7 +227,8 @@ func resolveWithIndexGo(ix *resolve.RuleIndex, imp string, from label.Label) (la
 			// Current match is worse
 		} else {
 			// Match is ambiguous
-			matchError = fmt.Errorf("multiple rules (%s and %s) may be imported with %q from %s", bestMatch.Label, m.Label, imp, from)
+			// TODO: consider listing all the ambiguous rules here.
+			matchError = fmt.Errorf("rule %s imports %q which matches multiple rules: %s and %s. # gazelle:resolve may be used to disambiguate", from, imp, bestMatch.Label, m.Label)
 		}
 	}
 	if matchError != nil {


language/go/resolve.go
```diff
@@ -228,6 +228,11 @@ func resolveWithIndexGo(ix *resolve.RuleIndex, imp string, from label.Label) (la
 		} else {
 			// Match is ambiguous
 			matchError = fmt.Errorf("multiple rules (%s and %s) may be imported with %q from %s", bestMatch.Label, m.Label, imp, from)
+			matchError = fmt.Errorf(`code in build rule %s contains import %q which matches multiple rules: %s and %s
+Use # gazelle:resolve to disambiguate with one of the following:
+	# gazelle:resolve go %[2]s %[3]s
+	# gazelle:resolve go %[2]s %[4]s
+`, from, imp, bestMatch.Label, m.Label)
```
jayconrod: A few nits about wording:

* Just replace "code in build rule" with "rule".
* Replace "contains import" with "imports".
* Replace "Use # gazelle:resolve" with "# gazelle:resolve may be used".

Also, if this is going to be a multi-line error message now, lets report all of the matching rules instead of just two. `bestMatch` can be a list, and we'll append to it in this case. At the end of the loop, if it contains two or more elements, we'll report the error.

language/go/resolve.go
```diff
@@ -228,6 +228,11 @@ func resolveWithIndexGo(ix *resolve.RuleIndex, imp string, from label.Label) (la
 		} else {
 			// Match is ambiguous
 			matchError = fmt.Errorf("multiple rules (%s and %s) may be imported with %q from %s", bestMatch.Label, m.Label, imp, from)
```
jayconrod: Make sure to remove the previous assignment. This is now dead code.

language/go/resolve.go
filmil: Replaced, and made the text into a single line error message.
Since I don't really have the time to rewrite the code to be a full-blown recommendation, I put a TODO in for the time being.  A clearer error message and a recommendation what to do is IMHO already an improvement.

language/go/resolve.go
filmil: Done.

language/go/resolve.go
chickenandpork: TBH, as a new user, `Use # gazelle:resolve to disambiguate with one of the following:` *with an indication of what file this text goes into*, would have been easier to immediately leverage.  Your example would have been so much better.

<--- This new user is randomly guessing which `BUILD.bazel` file (at import, or where the targets are defined), or `WORKSPACE`.  Maybe my syntax is wrong.  Maybe spaces matter.  Maybe `@thing//`, or `//`.  The matrix of possibilities would be alleviated with the text of the example above.

filmil: Hm, once the amount of needed prerequisite knowledge crosses some threshold, it is hard to put all the needed advice into a single error message.  Perhaps adding a URL to a documentation page would work better?  Sadly, few choose to go that route.

chickenandpork: I get that -- too much hand-holding makes an error too verbose for most users, and I do appreciate the general effort you have to help others, plus specifically this situation you're improving by these PRs.

The use of a URL given right there, I know I would follow it, and it's not as rare as new users (gasp!) reading documentation.  Stick with me a sec: when you're stuck, and this error message is there, and really "gazelle:resolve" doesn't google any working examples, and I cannot find a working example in the `bazelbuild/bazel-gazelle` unit tests that show where to put the config hint, then the only path forward is the URL in the error message.  I mean, as a last resort, it could get used ;)

filmil: (Note I'm in the same boat as you, as I'm also user of bazel, not a contributor.)  FWIW, I write (other) software for a living, and *anywhere* we introduced URLs to documentation into error messages, or error codes and such, it has always improved user satisfaction.

Issue #301, opened 2018-08-23
Support converting Gopkg.lock into a .bzl file instead of mutating WORKSPACE


jmillikin-stripe:
Current Gazelle's support for converting Gopkg.lock into `go_repository()` calls will actually append those calls to the global `WORKSPACE` file, which makes it difficult to manage. It'd be nice if it could be told to write to a `.bzl` file instead with a wrapper function, which the user could load into `WORKSPACE` at the proper location.

robfig:
@jayconrod would you be open to a contribution for this feature? If so, do you have guidance on how you would like it approached? I'm not currently familiar with the Gazelle codebase, but I could probably get something together in the next month and a half.

jayconrod:
@robfig The hard part about implementing this would be updating the rule generation and merging logic.

[rule.go](https://github.com/bazelbuild/bazel-gazelle/tree/master/internal/rule) contains code for creating, updating, and deleting rules in build files. This pretty strongly assumes that rules looks like top-level function calls, but maybe that assumption can be relaxed.

[merger.go](https://github.com/bazelbuild/bazel-gazelle/blob/master/internal/merger/merger.go) is where we merge a generated file with an existing file. This relies heavily on logic in the `rule` package.

So maybe it should work like this:

* User runs `gazelle update-repos -from_file=... -to_macro=repositories.bzl%go_repositories`. The last argument is new. That would tell Gazelle to modify the body of the `go_repositories` function in `repositories.bzl`.
* We generate a list of `go_repository` rules. [`ImportRepoRules`](https://github.com/bazelbuild/bazel-gazelle/blob/e70995a016131c86f4be93978c7624e6a2760414/internal/repos/repo.go#L74) does this.
* [`merger.MergeFile`](https://github.com/bazelbuild/bazel-gazelle/blob/e70995a016131c86f4be93978c7624e6a2760414/internal/merger/merger.go#L47) needs to be updated to be able to merge rules into the body of a function instead of the top-level of the file.  This may require changes in the `rule` package or the `merger` package. I'm not sure which would be cleanest. Likely `rule`.
* [update-repos.go](https://github.com/bazelbuild/bazel-gazelle/blob/master/cmd/gazelle/update-repos.go) needs to support the new flag, and it needs to update the `load` statements in the target .bzl file instead of WORKSPACE. It probably shouldn't call `merger.FixWorkspace` or `merger.CheckGazelleLoaded` if the flag is given.

Issue #132, opened 2018-02-15
Directive to declare repository rules defined in macros


jayconrod:
Gazelle should understand directives in WORKSPACE that describe repositories declared using a mechanism other than `go_repository` (for example, a `local_repository` or an `http_archive` or something declared inside a macro). The directive should tell Gazelle the name of the repository and it's prefix.

```
# gazelle:repository <repo_name> <prefix>
```

Example:

```
# gazelle:repository com_github_pkg_errors github.com/pkg/errors
```

jayconrod:
This directive should be written so that it affects import resolution for a specific language.

```
# gazelle:repository go com_github_pkg_errors github.com/pkg/errors
```

jayconrod:
After #647, there is now a `Repos` field in `config.Config` which lists repositories read from the configuration file. `# gazelle:repository` should add fake rules to that list. That means the syntax should be rich enough to specify a rule kind and attributes. So something like this?

```
# gazelle:repository go_repository name=com_github_pkg_errors importpath=github.com/pkg/errors
```

It should be possible to override an actual repository with this rule. For example, if someone has a `git_repository` rule for `org_golang_x_tools`, they should be able to write:

```
# gazelle:repository go_repository name=org_golang_x_tools importpath=golang.org/x/tools
```

And Gazelle would proceed as if that rule were declared as a `go_repository` rule.

PR #480, opened 2019-03-13
Add -to_macro option to update-repos

blico:
Addresses #301

The `to_macro` option tells Gazelle to write repository rules into a `.bzl` macro function rather than the WORKSPACE file. The user can then load the macro into their WORKSPACE and call it from there. This helps with the maintainability of the WORKSPACE file, especially for projects with many `go_repository` rules.

The format of the option is `-to_macro=macroFile%defName`


README.rst
```diff
@@ -409,6 +413,10 @@ The following flags are accepted:
 |                                                                                                                                                         |
 | Gazelle will not process packages outside this directory.                                                                                               |
 +----------------------------------------------------------------------------------------------------------+----------------------------------------------+
+| :flag:`-to_macro macroFile%defName`                                                                      |                                              |
++----------------------------------------------------------------------------------------------------------+----------------------------------------------+
+| Tells Gazelle to write repository rules into a .bzl macro function rather than the WORKSPACE file.                                                      |
```
jayconrod: It should probably be pointed out here that Gazelle won't see repositories declared in the macro when updating build files. That means that it won't use custom repository names, and it will go out to the network when trying to determine repository roots.

This is a solvable problem of course, but let's leave it for another time.

rule/rule.go
```diff
@@ -97,6 +102,16 @@ func LoadWorkspaceFile(path, pkg string) (*File, error) {
 	return LoadWorkspaceData(path, pkg, data)
 }
 
+// LoadMacroFile is similar to LoadFile but parses the file as a bzl
```
jayconrod: Please expand the doc here. It's important to say that this parses rules out of a Starlark function. `Sync` and `Save` calls will update the body of the function.

rule/rule.go
```diff
@@ -115,24 +130,40 @@ func LoadWorkspaceData(path, pkg string, data []byte) (*File, error) {
...
+	}
+	f := ScanAST(pkg, ast, defStmt.Body)
+	f.function = defStmt
+	return f, nil
 }
 
 // ScanAST creates a File wrapped around the given syntax tree. This tree
 // will be modified by editing methods.
-func ScanAST(pkg string, bzlFile *bzl.File) *File {
+func ScanAST(pkg string, bzlFile *bzl.File, stmt []bzl.Expr) *File {
```
jayconrod: I'd rather not make a breaking API change if we don't have to. Let's add a function, `ScanASTBody` with the new signature. `ScanAST` can call that with `f.File.Stmt`.

rule/rule.go
```diff
@@ -115,24 +130,40 @@ func LoadWorkspaceData(path, pkg string, data []byte) (*File, error) {
...
+// LoadMacroData is similar to LoadData but parses the data as a
+// bzl file.
+func LoadMacroData(path, pkg, defName string, data []byte) (*File, error) {
+	ast, err := bzl.ParseDefault(path, data)
+	if err != nil {
+		return nil, err
+	}
+	defStmt := scanForDef(ast.Stmt, defName)
+	if defStmt == nil {
+		return nil, fmt.Errorf("Could not find function name in .bzl file %s", defName)
```
jayconrod: I don't actually think this should be an error. If the function doesn't already exist, `Sync` should create it.

rule/rule.go
```diff
@@ -97,6 +102,16 @@ func LoadWorkspaceFile(path, pkg string) (*File, error) {
...
+// LoadMacroFile is similar to LoadFile but parses the file as a bzl
+// file.
+func LoadMacroFile(path, pkg, function string) (*File, error) {
+	data, err := ioutil.ReadFile(path)
+	if err != nil {
+		return nil, err
+	}
+	return LoadMacroData(path, pkg, function, data)
+}
+
```
jayconrod: We may also want an `EmptyMacroFile` to handle the case where the file doesn't already exist.

cmd/gazelle/update-repos.go
```diff
@@ -108,20 +131,34 @@ func updateRepos(args []string) error {
...
-	workspacePath := filepath.Join(c.RepoRoot, "WORKSPACE")
-	f, err := rule.LoadWorkspaceFile(workspacePath, "")
+	var f *rule.File
+	var path string
+	if uc.macroFileName == "" {
+		path = filepath.Join(c.RepoRoot, "WORKSPACE")
+		f, err = rule.LoadWorkspaceFile(path, "")
+	} else {
+		path = uc.macroFileName
+		f, err = rule.LoadMacroFile(path, "", uc.macroDefName)
```
jayconrod: If the file doesn't exist, we should create a new one.

cmd/gazelle/update-repos.go
```diff
@@ -108,20 +131,34 @@ func updateRepos(args []string) error {
...
 
 	if err := uc.fn(uc, f, kinds); err != nil {
 		return err
 	}
-	merger.FixLoads(f, loads)
-	if err := merger.CheckGazelleLoaded(f); err != nil {
-		return err
+
+	if uc.macroFileName == "" {
+		merger.FixLoads(f, loads)
```
jayconrod: `FixLoads` should still work for macro files. That will require some changes in `File.Sync`, since the loads need to be outside of the macro.

cmd/gazelle/integration_test.go
```diff
@@ -1253,6 +1253,82 @@ go_repository(
...
+	}
+
+	testtools.CheckFiles(t, dir, []testtools.FileSpec{
+		{
+			Path: "repositories.bzl",
+			Content: `
+load("@bazel_gazelle//:deps.bzl", "go_repository")
+
+def go_repositories():
+    go_repository(name = "org_golang_x_net", build_file_generation = "off", commit = "66aacef3dd8a676686c7ae3716979581e8b03c47", importpath = "golang.org/x/net")
```
jayconrod: I think the output is pretty difficult to read with all of the attributes on one line. In `Rule.sync`, could you set the `ForceMultiLine` flag on rules with more than one attribute?

blico: It looks like it is not currently possible to print out the attributes on one line :/. I created a PR to fix this issue here: https://github.com/bazelbuild/buildtools/pull/580

cmd/gazelle/update-repos.go
blico: Fixed and tested with TestImportReposFromDepToEmptyMacro

cmd/gazelle/update-repos.go
blico: Fixed and tested with TestImportReposFromDepToEmptyMacro

rule/rule.go
blico: Fixed and tested with TestImportReposFromDepToEmptyMacro

rule/rule.go
blico: Fixed and tested with TestImportReposFromDepToEmptyMacro

rule/rule.go
blico: Fixed by creating `ScanASTBody` and updating `LoadMacroData` and

README.rst
blico: Hmmm, this was an oversight on my part. We are pretty dependent on the functionality you mentioned, so I would like to get this done sooner than later.

From a quick glance, it looks like one way to get this done would be to add a new flag named something like `-go_repositories_macro=path`.  If the flag is set we call `LoadMacroFile` from `newFixUpdateConfiguration`. We then call `FixWorkspace` and `FixLoads` for both the `macro` and `workspace` files, and also pass both files to `repo.ListRepositories`.

WDYT?

jayconrod: #132 describes a `# gazelle:repository` directive that would declare this information in WORKSPACE without requiring command line arguments. That would be a useful escape hatch. I just opened #483, describing a more complete solution. In WORKSPACE, we could have a directive that tells Gazelle to read a macro in a separate .bzl file. Bonus points if we can update existing rules in that macro without needing the `-to_macro` flag.

Until those are implemented, we should mention this problem here, since the problem won't be obvious to people trying this out.

rule/rule.go
jayconrod: It looks like `gofmt` might be needed here. GH is showing one space indent instead of tabs.

rule/rule.go
```diff
@@ -118,6 +147,34 @@ func LoadWorkspaceData(path, pkg string, data []byte) (*File, error) {
 	return ScanAST(pkg, ast), nil
 }
 
+// LoadMacroData parses a bzl file from a byte slice and scans for the load
+// statements and the rules called from the given Starlark function. If there is
+// no matching function name, then a new function with that name will be created.
+// The function's syntax tree will be returned within File and can be modified by
+// Sync and Save calls.
+func LoadMacroData(path, pkg, defName string, data []byte, isEmpty bool) (*File, error) {
```
jayconrod: I don't think the `isEmpty` parameter is needed. If there isn't any function named `defName`, we should create one and add it to `ast.Stmt` next time `Sync()` is called. It's okay if the file is empty or if there's other stuff in there already (`load` statements).

Ideally, `LoadMacroData` should just parse the file, then return `ScanASTBody(pkg, defName, ast)`, similarly to how `LoadData` and `ScanAST` are related.

rule/rule.go
```diff
@@ -126,19 +183,31 @@ func ScanAST(pkg string, bzlFile *bzl.File) *File {
...
-		switch stmt := stmt.(type) {
+	f.Rules, f.Loads, _ = ScanASTBody("", bzlFile.Stmt)
+	f.Directives = ParseDirectives(bzlFile)
+	return f
+}
+
+// ScanASTBody scans a list of Starlark expressions and returns the rules,
+// loads and if a defName is supplied it will try to return a function
+// matching that name. This data can be wrapped by a File and later modified.
+func ScanASTBody(defName string, stmt []bzl.Expr) (rules []*Rule, loads []*Load, fn *bzl.Function) {
```
jayconrod: I think the signature should be `ScanASTBody(pkg, defName string, bzlFile *bzl.File) *File`.

If `defName == ""`, it should do what `ScanAST` did previously. `ScanAST` can just return `ScanASTBody(pkg, "", bzlFile)`.

If `defName != ""`, then it should take care of the stuff `LoadMacroData` is doing now. `LoadMacroData` can just parse the file, then call this function.

rule/rule.go
```diff
@@ -118,6 +147,34 @@ func LoadWorkspaceData(path, pkg string, data []byte) (*File, error) {
...
+		File: ast,
+		Pkg:  pkg,
+		Path: ast.Path,
+	}
+	if !isEmpty {
+		_, f.Loads, f.function = ScanASTBody(defName, ast.Stmt)
+	}
+	if f.function == nil {
+		defStmt := bzl.DefStmt{Name: defName}
+		ast.Stmt = append(ast.Stmt, &defStmt)
```
jayconrod: Let's not add the function until `Sync`. We'll need a way to track whether it's part of the AST in case `Sync` is called multiple times.

Currently, if we don't end up adding any rules to the function, we'll have an empty body, and the formatter will complain. `Sync` should add a `pass` statement if the body would be empty.

We should also be able to replace a `pass` statement if the function already exists and that's the only thing there.

cmd/gazelle/integration_test.go
jayconrod: Thanks for looking into this. I saw you set the `ForceMultiline` flag. So that's getting ignored? Hmm, that's unfortunate.

README.rst
linzhp: I think we can simplify this problem a bit to start with. We can have a default bzl file like `//go_repositories.bzl` unless it's overridden by some the Gazelle directive proposed in #483. Then `--to_macro` becomes a boolean flag, and write `go_repository` rules to `//go_repositories.bzl` when the flag is set. `update` and `fix` will first check if there is a `//go_repositories.bzl`. If so, read those rules first. Then we should allow people to override rules in `//go_repositories.bzl` in WORKSPACE files. For example, if they want to apply a patch to some `go_repository`, they can redefine the rule with the same `name` in WORKSPACE. Then `//go_repositories.bzl` contains no manual edits.

blico:
Sorry for the delay, I was out sick most of last week.

cmd/gazelle/integration_test.go
blico: This change has been merged into buildtools. So we will need to update buildtools within Gazelle sometime soon.

jayconrod:
It looks like there have been some breaking changes in buildtools since we last updated, and updating to the latest version will be non-trivial. I don't have time to do it right now, maybe this weekend. Filed #489 to track the update.

PR #485, opened 2019-03-18
-from-file=go.mod: support replace directive

QuentinPerez:
I've realised after the fact that this is a duplicate of https://github.com/bazelbuild/bazel-gazelle/pull/350; although now that I'm looking at it, that one seems to have been abandonned?

### DIFF
--- a/repo/modules.go
+++ b/repo/modules.go
@@ -46,6 +46,9 @@ func importRepoRulesModules(filename string, _ *RemoteCache) (repos []Repo, err 
 	type module struct {
 		Path, Version, Sum string
 		Main               bool
+		Replace            *struct {
+			Path, Version string
+		}
 	}
 	pathToModule := map[string]*module{}
 	data, err := goListModules(tempDir)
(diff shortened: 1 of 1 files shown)


repo/modules.go
```diff
@@ -116,12 +119,17 @@ func importRepoRulesModules(filename string, _ *RemoteCache) (repos []Repo, err
...
-		repos = append(repos, Repo{
+		r := Repo{
 			Name:     label.ImportPathToBazelRepoName(mod.Path),
 			GoPrefix: mod.Path,
 			Version:  mod.Version,
 			Sum:      mod.Sum,
-		})
+		}
+		if mod.Replace != nil {
+			r.Remote = mod.Replace.Path
```
jayconrod: `Remote` can't be set in module mode. It only applies to VCS tools.

I think we'll need a new `replace` attribute in `go_repository` and `fetch_repo` to support this. We shouldn't need to specify the `version` and `sum` of the original module, so we can reuse those attributes for the replacement.

That should be done in a separate PR, merged before this one. Sorry, I imagine that's more complicated than you were hoping for.

repo/modules.go
```diff
@@ -46,6 +46,9 @@ func importRepoRulesModules(filename string, _ *RemoteCache) (repos []Repo, err
 	type module struct {
 		Path, Version, Sum string
 		Main               bool
+		Replace            *struct {
+			Path, Version string
```
jayconrod: We'll also need to read `Dir` and report an error if it's set.

repo/modules.go
jayconrod: I've opened an #488 to track this.

Globegitter:
Gentle ping @QuentinPerez any chance to update this PR?

ChinmayR:
This is blocking use of go.mod with gazelle as replace directives have become common. Any chance this could be updated?

+@QuentinPerez for initial PR

Globegitter:
@jayconrod has this now been implemented with https://github.com/bazelbuild/bazel-gazelle/commit/99f7bcae18d0c84524eca529384723979ce473bc?

jayconrod:
#512 makes this one obsolete.

@Globegitter #502 was only the `replace` support in `go_repository`. That was a prerequisite.

PR #492, opened 2019-03-30
Upgrade github.com/bazelbuild/buildtools to latest

### LINKED ISSUE #489: Update github.com/bazelbuild/buildtools
We want to pick up bazelbuild/buildtools#580 so that repository rules declared in macros have multi-line formatting.

Unfortunately, it looks like there have been some breaking changes since we last updated, so this will be a non-trivial operation.

jayconrod:
Fixes #489

### DIFF
--- a/cmd/gazelle/integration_test.go
+++ b/cmd/gazelle/integration_test.go
@@ -1353,7 +1353,12 @@ def go_repositories():
         urls = ["https://example.com/yaml.tar.gz"],
         sha256 = "1234",
     )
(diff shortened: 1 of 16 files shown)


blico:
We should probably update https://github.com/bazelbuild/bazel-gazelle/blob/master/rule/rule.go#L156 from `ParseDefault` to the newly added `ParseBzl`

jayconrod:
Good catch, done.

Issue #483, opened 2019-03-15
Directive to reference macros from WORKSPACE


jayconrod:
#480 will add the ability for `update-repos` to write repository rules to a macro in a separate .bzl file.

Gazelle currently reads repository rules declared in WORKSPACE in order to discover custom repository names and known prefixes. This information is used for dependency resolution. If repositories are declared in a macro, this information will no longer be available (though #132 would be a useful escape hatch).

We should have a directive that tells Gazelle to look for repository rules in a macro in a .bzl file. It should be possible to repeat this directive, since there may be multiple macros.

```
# gazelle:repository_macro repos.bzl%go_repositories
```

This is mainly useful for the `fix` and `update` commands, but bonus points if `update-repos` can update repository rules in the corresponding macros without additional command line arguments instead of writing copies to WORKSPACE.

ifoox:
Hi Jay, do you intend to make a release of Gazelle with this included any time soon? (I see the last release is from Feb).

jayconrod:
Soon, yes. #477 is the main release blocker. I've gotten sidetracked with some toolchain compatibility issues in new versions of Bazel, but I expect to come back to this soon.

PR #493, opened 2019-04-02
repository_macro directive for referencing macros from WORKSPACE

blico:
Addresses https://github.com/bazelbuild/bazel-gazelle/issues/483

Added the `repository_macro` directive with the following format:
`# gazelle:repository_macro repos.bzl%go_repositories`

Gazelle only recognizes this directive when reading from the WORKSPACE file.

`fix-update` uses the directive to know about rules defined in a macro file, so they can be used during dependency resolution.

`update-repos` can now update repository rules in the corresponding macros without additional command line arguments instead of writing copies to WORKSPACE. 

The `to-macro` flag functionality has been modified to only write new repository rules to the corresponding macro, and to update existing rules in macros and the WORKSPACE file.  However, duplicates will still be written to the macro if the repository rule is not a `go_repository` rule.


jayconrod:
Thanks for working on this! I'll review as soon as I can, but it might be a couple days. Sorry for the latency -- a lot is happening right now.

repo/repo.go
```diff
@@ -55,6 +55,9 @@ type Repo struct {
 
 	// Sum is the hash of the module to be verified after download.
 	Sum string
+
+	// File is a pointer to the file where the repository rule was defined.
+	File *rule.File
```
jayconrod: It took me a long time to understand what `ListRepositories` and `FilterRulesByFile` were doing, and I think this is what was confusing me. I don't think this field belongs here. `Repo` should be describing a repository, not a declaration.

We do need a way for `ListRepositories` to return the parsed files and to indicate which repos were declared where. Maybe it could return a `map[*File][]string` where the strings are the repo names, and then `FilterRulesByFile` could do something with that?

cmd/gazelle/update-repos.go
```diff
@@ -245,23 +250,37 @@ func updateImportPaths(c *updateReposConfig, f *rule.File, kinds map[string]rule
 			return err
 		}
 	}
-	merger.MergeFile(f, nil, genRules, merger.PreResolve, kinds)
+	rulesByFile := repo.FilterRulesByFile(repos, genRules, destFile)
+	for f, rules := range rulesByFile {
+		merger.MergeFile(f, nil, rules, merger.PreResolve, kinds)
+		if err := f.Save(f.Path); err != nil {
+			return fmt.Errorf("error writing %q: %v", f.Path, err)
```
jayconrod: You can just `return err` here. `f.Save` is responsible for reporting the path.

cmd/gazelle/update-repos.go
```diff
@@ -245,23 +250,37 @@ func updateImportPaths(c *updateReposConfig, f *rule.File, kinds map[string]rule
...
 	}
-	for i := range genRules {
-		applyBuildAttributes(c, genRules[i])
+	rulesByFile := repo.FilterRulesByFile(repos, genRules, destFile)
+	for f, rules := range rulesByFile {
+		for i := range rules {
+			applyBuildAttributes(c, rules[i])
+		}
+		merger.MergeFile(f, nil, rules, merger.PreResolve, kinds)
+		if err := f.Save(f.Path); err != nil {
```
jayconrod: What happens if the same file contains multiple macros? It seems like earlier changes would be overwritten here.

repo/repo.go
```diff
@@ -100,6 +109,40 @@ func ImportRepoRules(filename string, repoCache *RemoteCache) ([]*rule.Rule, err
 	return rules, nil
 }
 
+// FilterRulesByFile filters a list of generated repo rules by the file where
+// the rule should be written. This is done by comparing the generated
+// rules with the current list of repo rules declared in the repository. If
+// the generated rule matches a rule already in the repo, then it inherits the
+// file where the rule was defined. Else its file is set as the destFile parameter.
+func FilterRulesByFile(before []Repo, after []*rule.Rule, destFile *rule.File) map[*rule.File][]*rule.Rule {
```
jayconrod: `before` and `after` are a little misleading: it makes it sound like `after` is an out parameter that will be written into. `repos` and `genRules` would be fine. These are the names the caller uses.

Also, this function should not be exported. It's only needed within this package.

cmd/gazelle/update-repos.go
```diff
@@ -131,40 +131,42 @@ func updateRepos(args []string) error {
...
+		merger.FixWorkspace(destFile)
 	}
 
-	if err := uc.fn(uc, f, kinds); err != nil {
+	if err := uc.fn(uc, workspace, destFile, kinds); err != nil {
 		return err
 	}
 
-	merger.FixLoads(f, loads)
+	merger.FixLoads(destFile, loads)
```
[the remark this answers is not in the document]
jayconrod: That's a good question. Yes, I think so.

repo/repo.go
blico: `FilterRulesByFile` is called outside of this package by `update-repos.go`. Are you saying that the code should be refactored in a way that the function does not need to be exported?

repo/repo.go
blico: Sorry about that, I had a feeling this part may be confusing. `ListRepositories` new signature will look like this:
`func ListRepositories(workspace *rule.File) (repos []Repo, reposByFile map[*rule.File][]string, err error)`

repo/repo.go
jayconrod: Oops, you're right.

Let's replace this with `func MergeRules(genRules []*rule.Rule, existingRules map[*rule.File][]string, destFile *rule.File) error` in the `repo` package. The second argument would be what `ListRepositories` would return. This would merge the generated rules into the appropriate files (or into `destFile` if they're new rules). With this function, we'd have common logic for writing the rules out.

This function probably belongs in the `merger` package, but I'm not confident enough about API stability to put it there. I consider `repo` less stable than `merger`, since most of the `repo` functionality should eventually be moved into `language/go`.

repo/repo.go
jayconrod: 👍

cmd/gazelle/update-repos.go
blico: My latest change handles this case in `MergeRules` by calling `rule.SyncMacroFile`. It is not the greatest solution, so I am open to other ideas.

jayconrod:
(Once again, I apologize for review latency. I'll look at the updates as soon as I can.)

repo/repo.go
```diff
@@ -100,6 +107,55 @@ func ImportRepoRules(filename string, repoCache *RemoteCache) ([]*rule.Rule, err
...
 
+// MergeRules merges a list of generated repo rules with the already defined repo rules,
+// and then updates each rule's underlying file. If the generated rule matches an existing
+// one, then it inherits the file where the existing rule was defined. If the rule is new then
+// its file is set as the destFile parameter. A list of the updated files is returned.
+func MergeRules(genRules []*rule.Rule, existingRules map[*rule.File][]string, destFile *rule.File, kinds map[string]rule.KindInfo) []*rule.File {
+	sort.Stable(byRuleName(genRules))
+
+	rulesByFile := make(map[*rule.File][]*rule.Rule)
+	for file, repoNames := range existingRules {
```
jayconrod: I'm having trouble understanding how this loop works.

It's building `rulesByFile`, so for each generated rule, it needs to check whether there is a file in `existingRules`: if so, the generated rule should go there; otherwise, it should go in `destFile`.

I think it would be simpler to build a `map[string]*rule.File` from `existingRules`. Then building `rulesByFile` would be simpler.

repo/repo.go
```diff
@@ -167,13 +223,33 @@ func FindExternalRepo(repoRoot, name string) (string, error) {
...
+	repos, repoNamesByFile[workspace] = getRepos(workspace.Rules)
+
+	for _, d := range workspace.Directives {
+		switch d.Key {
+		case "repository_macro":
+			vals := strings.Split(d.Value, "%")
+			if len(vals) != 2 {
+				return nil, nil, fmt.Errorf("Failure parsing repository_macro: %s, expected format is macroFile%%defName", d.Value)
+			}
+			macroFile, err := rule.LoadMacroFile(vals[0], "", vals[1])
```
jayconrod: The macro file name should be a relative path from `filepath.Dir(workspace.Path)`.

Make sure to run `filepath.Clean` on it before joining it, and make sure it doesn't start with `".."`.

rule/rule.go
```diff
@@ -238,6 +238,25 @@ func MatchBuildFileName(dir string, names []string, files []os.FileInfo) string
 	return ""
 }
 
+// SyncMacroFile syncs the file's syntax tree with another file's. This is
+// useful for keeping multiple macro definitions from the same .bzl file in sync.
+func (f *File) SyncMacroFile(from *File) {
```
jayconrod: Let's be very careful here. Having multiple files share an AST means there are multiple mutable references to the same data structure. That could result in some subtle bugs if other edits are made and `Sync` is called later.

Ideally, I think this method would copy the macro from one `*File` to another, replacing the function of the same name that was already there. I'm not sure if there's a way to copy ASTs, so adding or replacing the macro function in `f.File.Stmt` is probably good enough.

repo/repo.go
```diff
@@ -100,6 +107,55 @@ func ImportRepoRules(filename string, repoCache *RemoteCache) ([]*rule.Rule, err
...
+	}
+
+	updatedFiles := make(map[string]*rule.File)
+	for f, rules := range rulesByFile {
+		if uf, ok := updatedFiles[f.Path]; ok {
+			f.SyncMacroFile(uf)
+		}
+		merger.MergeFile(f, nil, rules, merger.PreResolve, kinds)
+		f.Sync()
+		updatedFiles[f.Path] = f
```
jayconrod: Let's make sure not to return more than one file with the same path. If a file contains many macros, we should choose one main file, copy the other synced macros there, then return that.

blico:
> A few more comments. Sorry for the latency and thanks for working on this.
> 
> Also it looks like the linker in panicking in CI? Does it do that locally with `bazel test ...` or `go test ./...`? Very strange.

Looks like it was an import cycle in `merger_test.go` caused by `repo` adding `merger` as a dependency. I fixed this by making `merger_test` a different package. As an fyi one side effect is that I exported `merger.Match`

Issue #407, opened 2018-12-21
update-repos -from_file=go.mod should handle replace with relative paths


siddharthab:
In update-repos, go.mod is copied to a temporary directory before being read. This is a problem when the mod file references relative paths, for example, when using the [replace directive](https://github.com/golang/go/wiki/Modules#when-should-i-use-the-replace-directive).

I think it is OK even if `go list -m` changes the go.mod file in some cases. This operation is run manually by the user, and is not part of a build.

The alternative could be to restore an original copy of the go.mod file.

jayconrod:
I'm changing the issue title a bit: it sounds like the problem is that `update-repos` isn't handling relative paths, not specifically that `update-repos` uses a temp directory. 

I don't think `update-repos` should be able to modify the `go.mod` file, even temporarily. That would be pretty surprising behavior for me as a user (actually, I think go list's behavior is pretty surprising, but that's probably not going to change).

We might want to use `-mod=readonly` instead of making a temporary copy. Or we could adjust relative paths as part of the copy. In any case, `replace` directives with local paths should probably be imported as `local_repository` rules.

siddharthab:
Thank you. Sounds good. Note that some of these relative paths could be resolve within the workspace itself, in which case things will be handled correctly by the prefix directive to gazelle.

PR #512, opened 2019-04-26
Add support for the replace directive in go.mod 

bigkraig:
I think this may also partially address the issue raised in #407, that the replace directive in go.mod is not handled properly.

I noticed this at first when trying to add [Pulumi](https://github.com/pulumi/pulumi/) as a third_party dependency.

```
gazelle: error parsing "go.mod": exit status 1
```

With this patch I am able to directly run an `update-reps` and generate a proper `go_repository`.

For a given `go.mod`:
```
require (
        github.com/Nvveen/Gotty v0.0.0-20170406111628-a8b993ba6abd
)

replace (
        github.com/Nvveen/Gotty => github.com/ijc25/Gotty v0.0.0-20170406111628-a8b993ba6abd
)
```

Gazelle will produce:
```
    go_repository(
        name = "com_github_nvveen_gotty",
        importpath = "github.com/Nvveen/Gotty",
        replace = "github.com/ijc25/Gotty",
        sum = "h1:a0nQYN5aGC2AiAbRXGtbhFXVFEYhI8m0qm0rS4BZrlo=",
        version = "v0.0.0-20170406111628-a8b993ba6abd",
    )
```

### DIFF
--- a/repo/modules.go
+++ b/repo/modules.go
@@ -51,7 +51,6 @@ func importRepoRulesModules(filename string, _ *RemoteCache) (repos []Repo, err 
 		}
 	}
 	pathToModule := map[string]*module{}
-	replacePathToModule := map[string]*module{}
 	data, err := goListModules(tempDir)
 	if err != nil {
 		return nil, err
@@ -65,9 +64,10 @@ func importRepoRulesModules(filename string, _ *RemoteCache) (repos []Repo, err 
 		if mod.Main {
 			continue
 		}
-		pathToModule[mod.Path] = mod
 		if mod.Replace != nil {
-			replacePathToModule[mod.Replace.Path] = mod
+			pathToModule[mod.Replace.Path] = mod
+		} else {
+			pathToModule[mod.Path] = mod
 		}
 	}
 
@@ -86,9 +86,6 @@ func importRepoRulesModules(filename string, _ *RemoteCache) (repos []Repo, err 
 			continue
 		}
 		if mod, ok := pathToModule[path]; ok {
-			mod.Sum = sum
-		}
-		if mod, ok := replacePathToModule[path]; ok {
 			mod.Sum = sum
 		}
 	}
--- a/repo/modules.go
+++ b/repo/modules.go
@@ -19,6 +19,7 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"go/build"
 	"io"
 	"io/ioutil"
 	"log"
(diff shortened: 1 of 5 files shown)


repo/modules.go
```diff
@@ -81,6 +88,9 @@ func importRepoRulesModules(filename string, _ *RemoteCache) (repos []Repo, err
 		if mod, ok := pathToModule[path]; ok {
 			mod.Sum = sum
 		}
+		if mod, ok := replacePathToModule[path]; ok {
+			mod.Sum = sum
+		}
 	}
 
 	// If sums are missing, run go mod download to get them.
```
blico: If the sum is missing for a module with a replace directive, then we should use `replace.Path` and `replace.Version` to retrieve the sum via `goModDownload`.

repo/modules.go
```diff
@@ -62,6 +66,9 @@ func importRepoRulesModules(filename string, _ *RemoteCache) (repos []Repo, err
 			continue
 		}
 		pathToModule[mod.Path] = mod
+		if mod.Replace != nil {
```
blico: We probably don't need `replacePathToModule`. Whenever the `replace` directive is set, it should be ok to use `mod.Replace.Path` as the key for `pathToModule`.

repo/modules.go
bigkraig: Yep makes sense and this sort of comes about naturally once I removed replacePathToModule

repo/modules.go
bigkraig: :+1:

bigkraig:
@blico LMK how that looks

repo/modules.go
blico: We also need to modify `missingSumArgs` to contain `mod.Replace.Path` and `mod.Replace.Version` if `mod.Replace != nil`

bigkraig: :+1:

repo/modules.go
```diff
@@ -61,7 +64,11 @@ func importRepoRulesModules(filename string, _ *RemoteCache) (repos []Repo, err
 		if mod.Main {
 			continue
 		}
-		pathToModule[mod.Path] = mod
+		if mod.Replace != nil {
```
jayconrod: Check that `mod.Replace.Path` is not a file path. Both `path/filepath.IsAbs` and `go/build.IsLocalImport` should return false.

If it is a local path, warn that `go_repository` doesn't support file path replacements and skip the module.

repo/repo.go
```diff
@@ -56,6 +56,10 @@ type Repo struct {
 
 	// Sum is the hash of the module to be verified after download.
 	Sum string
+
+	// Replace is the Go import path of the module configured by the replace
+	// directive in go.mod.
+	Replace string
```
jayconrod: Please also add `replace` to `go_repository` `MergeableAttrs` in `language/go/kinds.go`. This will let Gazelle replace or remove existing `replace` attributes.

bigkraig: all set

repo/modules.go
bigkraig: LMK how that looks. I wasn't sure what a good warning message would look like

Issue #213, opened 2018-05-29
gazelle update-repos should support pruning unneeded dependencies


jayconrod:
For example, when importing external repository rules from a lock file, if the `-prune` flag (or something similar) is given, Gazelle should remove `go_repository` rules that no longer have equivalent repos in the lock file.

```
gazelle update-repos -from_file=Gopkg.lock -prune
```

Later, we should also be able to remove repositories that aren't imported by the main repository or transitive dependencies.

linzhp:
> Later, we should also be able to remove repositories that aren't imported by the main repository or transitive dependencies.

To do this as part of `update-repos`, can we assume Bazel rules exist in the main repo? If so, we can run `bazel query deps(//...)` and prune all `go_repository` rules not in the transitive dependency closure.

jayconrod:
> To do this as part of update-repos, can we assume Bazel rules exist in the main repo? If so, we can run bazel query deps(//...) and prune all go_repository rules not in the transitive dependency closure.

Yes we can assume Bazel rules exist in the main repo.

I don't think Gazelle should run `bazel query` or any other Bazel command though. Earlier, it was not possible since Gazelle is commonly invoked by `bazel run`, and that used to hold a global lock on the work space. But still I think it's better not to recursively invoke Bazel. This may interfere with autogazelle, and it will be slow. I also don't want to create the expectation that Gazelle can evaluate Starlark code: we can't update files that require evaluation (e.g., targets declared in macros).

It may be better to ask users to run `bazel fetch //...` first, then do a best-effort emulation of `bazel query`.

martin-randall-tanium:
It look like `-prune` was implemented in #514. Can this issue be closed?

jayconrod:
Yeah this can be closed. I don't think `update-repos` will support listing imports in source files, so `-prune` can only be used with `-from_file`.

PR #514, opened 2019-04-29
Add support for pruning unneeded rules to update-repos

blico:
Addresses #213 

Adds the `prune` flag to update-repos. When enabled, Gazelle will remove `go_repository` rules that no longer have equivalent repos in the `Gopkg.lock`/`go.mod` file.


README.rst
```diff
@@ -419,6 +419,10 @@ The following flags are accepted:
 |                                                                                                                                                         |
 | The ``repository_macro`` directive should be added to the WORKSPACE in order for future Gazelle calls to recognize the repos defined in the macro file. |
 +----------------------------------------------------------------------------------------------------------+----------------------------------------------+
+| :flag:`-prune true|false`                                                                                | :value:`false`                               |
++----------------------------------------------------------------------------------------------------------+----------------------------------------------+
+| When true Gazelle will remove `go_repository`_ rules that no longer have equivalent repos in the ``Gopkg.lock``/``go.mod`` file.                        |
```
jayconrod: Add comma: "When true, Gazelle..."

Mention that this only works with `-from_file`.

cmd/gazelle/update-repos.go
```diff
@@ -92,6 +93,7 @@ func (_ *updateReposConfigurer) RegisterFlags(fs *flag.FlagSet, cmd string, c *c
 	fs.Var(&gzflag.AllowedStringFlag{Value: &uc.buildFileProtoModeAttr, Allowed: validBuildFileProtoModeAttr}, "build_file_proto_mode", "Sets the build_file_proto_mode attribute for the generated go_repository rule(s).")
 	fs.StringVar(&uc.buildExtraArgsAttr, "build_extra_args", "", "Sets the build_extra_args attribute for the generated go_repository rule(s).")
 	fs.Var(macroFlag{macroFileName: &uc.macroFileName, macroDefName: &uc.macroDefName}, "to_macro", "Tells Gazelle to write repository rules into a .bzl macro function rather than the WORKSPACE file. . The expected format is: macroFile%defName")
+	fs.BoolVar(&uc.pruneRules, "prune", false, "When enabled, Gazelle will remove rules that no longer have equivalent repos in the Gopkg.lock/go.mod file.")
```
jayconrod: In `CheckFlags`, make sure `-from_file` was also set.

cmd/gazelle/update-repos.go
```diff
@@ -253,7 +255,11 @@ func updateImportPaths(c *updateReposConfig, workspace *rule.File, destFile *rul
 			return nil, err
 		}
 	}
-	files := repo.MergeRules(genRules, reposByFile, destFile, kinds)
+	var emptyRules []*rule.Rule
```
jayconrod: Remove this. If I understand this correctly, it would delete rules not explicitly mentioned on the command line.

I don't think a complete implementation of this is feasible right now, so just report an error in `CheckFlags` if `-prune` is set without `-from_file`. Ideally we would remove repositories that aren't needed by any libraries in the current repository or in other repositories. That requires a much more fleshed out `update-repos` in general (which unfortunately I don't have time to work on or even review right now).

repo/repo.go
```diff
@@ -107,15 +107,34 @@ func ImportRepoRules(filename string, repoCache *RemoteCache) ([]*rule.Rule, err
 	return rules, nil
 }
 
+// GetRulesToPrune returns a list of empty repo rules that can be pruned from the repo.
+// The list is determined by finding repo rules defined in the repository that no longer
+// have an equivalent repo defined in the Gopkg.lock/go.mod file.
+func GetRulesToPrune(rules []*rule.Rule, repos []Repo) []*rule.Rule {
```
jayconrod: I think this logic could be integrated into `MergeRules`. Maybe pass a `pruneRules bool` parameter instead of `emptyRules`? It seems like that function has all the information it needs to build this list for each file.

blico:
Thanks for the review, Jay! I really appreciate all of your help.

jayconrod:
@Globegitter That would be great. And it would be quite a bigger task. :) #213 will remain open for this reason.

I don't expect that to happen soon. Some other changes are needed first: indexing external repositories and integrating `update-repos` with the extension architecture. Unfortunately I don't have bandwidth to work on that right now or even to review the amount of code that would be needed to make that change.

PR #516, opened 2019-05-02
Expand use of the replace attribute

bnordbo:
This PR allows the "replace" attribute of go_repository to be used in conjunction with "tag" or "commit", in addition to "version".

As far as I can tell this works fine; the tests are passing, and I can successfully build projects using forked Go dependencies at specific commits.


bnordbo:
I signed it!

bnordbo:
I'm closing this PR: As far as I can tell, the same result can be achieved by setting "remote".

PR #524, opened 2019-05-14
Don't override modules with replace directives

### LINKED ISSUE #521: update-repos is overriding modules that have replace directives
In our `go.mod` file we have something like this:
```
require (
        github.com/bar v0.0.0-00010101000000-000000000000
	github.com/foo v1.4.0
)
replace github.com/bar/v0.0.0-00010101000000-000000000000 => github.com/foo v1.4.0
```
Gazelle calls `go list -m -json all` in order to discover relevant modules. When I call `go list -m all` with the `go.mod` file above I see the following output:
```
github.com/bar v0.0.0-00010101000000-000000000000 => github.com/foo v1.4.0
github.com/foo v1.4.0
```

With the current logic in `modules.go`, the first module listed (with the replace directive) will be overridden by the second module because they have the same import path. This results in update-repos generating the `go_repository` rule without a `replace` attribute. 

Before fixing this in Gazelle, I just want to make sure that it is expected for `go list -m all` to list the same import path twice.

blico:
`go list -m -json all` can list the same importpath twice, one with a replace directive and one without. This change fixes `update-repos` to properly handle the case when multiple modules are listed with the same importpath.

Fixes #521

### DIFF
--- a/repo/modules.go
+++ b/repo/modules.go
@@ -123,11 +123,12 @@ func importRepoRulesModules(filename string, _ *RemoteCache) (repos []Repo, err 
 			if err := dec.Decode(&dl); err != nil {
 				return nil, err
 			}
-			mod := pathToModule[dl.Path]
-			if mod == nil {
-				continue
+			if mod, ok := pathToModule[dl.Path]; ok {
+				mod.Sum = dl.Sum
 			}
-			mod.Sum = dl.Sum
+			if mod, ok := replacePathToModule[dl.Path]; ok {
+				mod.Sum = dl.Sum
+			}
 		}
 	}
 
--- a/repo/modules.go
+++ b/repo/modules.go
@@ -52,6 +52,7 @@ func importRepoRulesModules(filename string, _ *RemoteCache) (repos []Repo, err 
 		}
 	}
 	pathToModule := map[string]*module{}
+	replacePathToModule := map[string]*module{}
 	data, err := goListModules(tempDir)
 	if err != nil {
 		return nil, err
@@ -71,7 +72,7 @@ func importRepoRulesModules(filename string, _ *RemoteCache) (repos []Repo, err 
 					mod.Replace.Path)
 				continue
 			}
-			pathToModule[mod.Replace.Path] = mod
+			replacePathToModule[mod.Replace.Path] = mod
 		} else {
 			pathToModule[mod.Path] = mod
 		}
@@ -94,17 +95,21 @@ func importRepoRulesModules(filename string, _ *RemoteCache) (repos []Repo, err 
 		if mod, ok := pathToModule[path]; ok {
 			mod.Sum = sum
 		}
+		if mod, ok := replacePathToModule[path]; ok {
+			mod.Sum = sum
+		}
 	}
 
 	// If sums are missing, run go mod download to get them.
 	var missingSumArgs []string
 	for _, mod := range pathToModule {
 		if mod.Sum == "" {
-			if mod.Replace != nil {
-				missingSumArgs = append(missingSumArgs, fmt.Sprintf("%s@%s", mod.Replace.Path, mod.Replace.Version))
-			} else {
-				missingSumArgs = append(missingSumArgs, fmt.Sprintf("%s@%s", mod.Path, mod.Version))
-			}
+			missingSumArgs = append(missingSumArgs, fmt.Sprintf("%s@%s", mod.Path, mod.Version))
+		}
+	}
+	for _, mod := range replacePathToModule {
+		if mod.Sum == "" {
+			missingSumArgs = append(missingSumArgs, fmt.Sprintf("%s@%s", mod.Replace.Path, mod.Replace.Version))
 		}
 	}
 	if len(missingSumArgs) > 0 {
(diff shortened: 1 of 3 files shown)


repo/modules.go
```diff
@@ -94,17 +95,21 @@ func importRepoRulesModules(filename string, _ *RemoteCache) (repos []Repo, err
 		if mod, ok := pathToModule[path]; ok {
 			mod.Sum = sum
 		}
+		if mod, ok := replacePathToModule[path]; ok {
```
blico: Before setting `mod.Sum`, should we also check that the version of the module read from the `go.sum` file matches our module's version?

repo/modules.go
```diff
@@ -52,6 +52,7 @@ func importRepoRulesModules(filename string, _ *RemoteCache) (repos []Repo, err
 		}
 	}
 	pathToModule := map[string]*module{}
+	replacePathToModule := map[string]*module{}
```
jayconrod: We seem to be duplicating a lot of logic below. At the end of this loop, let's make `pathToModule` map importable module paths to `*module`. So for example, say we have

```
module M

require (
  A v1.0.0
  B v1.0.0
)

replace A v1.0.0 => B v1.0.0
```

`pathToModule` should map `A` to a module with `m.Path == "A"` and `m.Replace.Path == "B"`. `pathToModule` should not contain a separate entry for `B`.

If we see `B` first in the list, we should delete it from `pathToModule` when we see `A`. If we see `A` first, we should avoid adding `B`.

repo/modules.go
jayconrod: Yes, we should do that. We should skip entries if the versions don't match.

repo/modules.go
```diff
@@ -94,17 +95,21 @@ func importRepoRulesModules(filename string, _ *RemoteCache) (repos []Repo, err
...
 	// If sums are missing, run go mod download to get them.
 	var missingSumArgs []string
 	for _, mod := range pathToModule {
 		if mod.Sum == "" {
-			if mod.Replace != nil {
-				missingSumArgs = append(missingSumArgs, fmt.Sprintf("%s@%s", mod.Replace.Path, mod.Replace.Version))
-			} else {
-				missingSumArgs = append(missingSumArgs, fmt.Sprintf("%s@%s", mod.Path, mod.Version))
-			}
+			missingSumArgs = append(missingSumArgs, fmt.Sprintf("%s@%s", mod.Path, mod.Version))
```
jayconrod: If `pathToModule` is fixed as above, we shouldn't need the change below this point. We don't need sums for modules that were replaced. We especially don't need `go_repository` rules for both replaced and replacement modules.

repo/modules.go
blico: The example I mention in #521 (and the one you do above) will not be built by go: https://go-review.googlesource.com/c/go/+/126156/. 

When you call `go mod why` you get
```
go: a@v1.0.0 used for two different module paths (a and b)
```
With this in mind, I think both modules are equally valid/invalid, so it doesn't make sense to only generate a rule for the module with the replace directive. We should either generate rules for both modules or print a warning and not generate rules for either.

Also, it is a valid case when a single module has two different paths, as long as the paths don't have the same version:
```
module M

require (
  A v1.0.0
  B v1.0.0
)

replace A v1.0.0 => B v1.2.0
```

`go list -m all` returns
```
A v1.0.0 => B v1.2.0
B v1.0.0
```
And we will want two `go_repository` rules:
```
go_repository(
  importpath = "B",
  version = "1.0.0",
  sum = <sum_from_B@1.0.0>,
)
go_repository(
  importpath = "A",
  version = "1.2.0",
  replace = "B",
  sum = <sum_from_B@1.2.0>,
)
```

In this case, we will need both modules. I have thought of a way to consolidate the logic a little bit (always use `mod.Path` as `pathToModule`'s index, create an importpath=>sum map of the `go.sum` file and then iterate through `pathToModule` searching for matching sums using `mod.Replace.Path` if there is a replace directive.)

jayconrod: > With this in mind, I think both modules are equally valid/invalid, so it doesn't make sense to only generate a rule for the module with the replace directive. We should either generate rules for both modules or print a warning and not generate rules for either.

You're right, I just tried this with my fork of bazel-gazelle as a concrete example, and `go list -m -json all` reports an error. If that happens, we shouldn't generate any rules.

> Also, it is a valid case when a single module has two different paths, as long as the paths don't have the same version

Agreed, we should generate two `go_repository` rules for this. I mainly want to avoid doing anything different than what the Go command does. We shouldn't create two `go_repository` rules with different names for the same bits, but it should be okay if they're at different versions.

blico: > You're right, I just tried this with my fork of bazel-gazelle as a concrete example, and go list -m -json all reports an error. If that happens, we shouldn't generate any rules.

Interesting, I tried reproing with go 1.12.5 and did not get an error when calling `go list -m -json all`, only with `go mod why`... For older versions of go where `go list` does not throw an error in this scenario, should Gazelle treat it as an error or generate both of the rules?  

Going off of your desire to not sway from the Go command, I guess we should ignore the error. The user will get an error as soon as they try to build, and will then proceed to fix their `go.mod` file and re-run `update-repos`.

jayconrod: I was seeing this error in 1.12.5, too. Here's my go.mod file:

```
module m

require (
	github.com/bazelbuild/bazel-gazelle v0.0.0-20190513151847-b16e182bf090 // indirect
	github.com/jayconrod/bazel-gazelle v0.0.0-20190513151847-b16e182bf090
)

replace github.com/bazelbuild/bazel-gazelle => github.com/jayconrod/bazel-gazelle v0.0.0-20190513151847-b16e182bf090
```

In any case though, I think treating `go list -m -json all` as the source of truth is our best option here. The semantics and errors will change over time, and we'll need to adjust.

PR #530, opened 2019-05-17
[go] Respect go_proto_library rules generated by another plugin

mh-park:
This commit helps the gazelle go plugin to behave well with other
plugins that create `go_proto_library` rules.

`proto_library` rules will be matched with pregenerated `go_proto_library`
rules, and skip the go extension's `go_proto_library` rule generation if
there is a match. The go extension treats `proto_library` rules with a 
matching `go_proto_library` rule as if they do not exist.


language/go/generate.go
```diff
@@ -112,6 +122,11 @@ func (gl *goLang) GenerateRules(args language.GenerateArgs) language.GenerateRes
 						importPath: goProtoImportPath(gc, ppkg, args.Rel),
 						proto:      protoTargetFromProtoPackage(name, ppkg),
 					}
+					if r, ok := goProtoRules[":"+name]; ok {
+						// if a go_proto_library rule already exists for this
```
mh-park: This means if the directory:
  1. `# gazelle:proto default` or `"-proto=default"`
  2. has no go files, and
  3. only one proto package,
the `go_library` rule generated will use the `go_proto_library` rule's import path.

I'm not sure if this is the ideal behavior, and I'm marking this one up for discussion. Let me know if you have any other thoughts/ideas. @jayconrod

language/go/generate.go
```diff
@@ -37,10 +37,20 @@ func (gl *goLang) GenerateRules(args language.GenerateArgs) language.GenerateRes
...
+	//
+	//   key: proto_library rule name (colon-prefixed)
+	//   value: go_proto_library rule
+	goProtoRules := make(map[string]*rule.Rule)
 	var protoRuleNames []string
 	protoPackages := make(map[string]proto.Package)
 	protoFileInfo := make(map[string]proto.FileInfo)
 	for _, r := range args.OtherGen {
+		if r.Kind() == "go_proto_library" {
+			goProtoRules[r.AttrString("proto")] = r
```
jayconrod: Let's be very careful here.

* `proto` may not be set (`AttrString` may return `""`).
* `protos` may be set instead (it should be a list of strings).
* Either may be set to a list of labels that aren't necessarily in the same directory.
* More than one `go_proto_library` may reference the same `proto_library`.

language/go/generate.go
jayconrod: Integrating `go_library` with `go_proto_library` generated by another extension seems too hard. There are just too many corner cases and too many things that could go wrong.
 
If a `proto_library` is covered by a `go_proto_library` generated by another extension, I think we should simply proceed as if neither rule exists. The Go extension should not generate a `go_proto_library` for that proto, and it shouldn't generate a `go_library` if there aren't any other .go files in the directory.

mh-park: SGTM.

I assume this means that if we have colliding import paths with a `go_proto_library` and a `go_library`, we'll just have separate rules with the same import path, rather than embedding the `go_proto_library` rule in the `go_library` rule, correct?

mh-park:
@jayconrod Mind giving it another scan? 

> If a proto_library is covered by a go_proto_library generated by another extension, I think we should simply proceed as if neither rule exists.

I'm not too sure what to do for `go_proto_library` rules that are referencing protos that are not in the same directory with this approach. We have to maintain some global state, and we would have to walk back up directories. For now, I have it only for rules within the same directory: if a `go_proto_library` rule matches a `proto_library` rule _in the same directory_, we treat them as if they don't exist. If they're rules that match but are in different directories, nothing happens.

mh-park:
All good! Thanks for the reviews :)

Issue #490, opened 2019-03-27
Use external dependencies instead of vendor directory


jayconrod:
When built by Bazel, Gazelle should depend on packages in other modules via `go_repository` rules, not the vendor directory. This is a problem for extensions needing to reference the same dependencies, for example, `github.com/bazelbuild/buildtools`. They currently have to reference `@bazel_gazelle//vendor/...` or they pull in two copies of these packages, risking conflicts.

The vendor directory is still needed for the implementation of `go_repository`, but it shouldn't be needed outside of that.

PR #534, opened 2019-05-21
Use external dependencies instead of vendor directory

blico:
Addresses #490 

Generate `go_repository` rules for Gazelle's external dependencies, and update Gazelle to use these dependencies rather than the vendor directory.
 
Implemented with the following two commands:
```
bazel run //:gazelle -- update-repos -from_file=go.mod -to_macro=third_party/repos.bzl%go_repositories
bazel run //:gazelle -- -exclude=vendor
```

The vendor directory is still needed because the `go_repository` rule uses go build when creating `go_repository_tools`.


WORKSPACE
```diff
@@ -21,3 +21,8 @@ gazelle_dependencies()
 load("@io_bazel_rules_go//tests:bazel_tests.bzl", "test_environment")
 
 test_environment()
+
+load("//third_party:repos.bzl", "go_repositories")
```
jayconrod: I'd rather not introduce a new macro for this. Repos that `gazelle`, `fetch_repo`, and other binaries depend on directly should be in `gazelle_dependencies`. This will make those repositories available for anyone that uses Gazelle, so the set should be pretty minimal, and we should keep using `_maybe`.

It's fine to declare test dependencies directly in WORKSPACE, since other repositories don't need these.

third_party/repos.bzl
```diff
@@ -0,0 +1,81 @@
...
+        sum = "h1:qIbj1fsPNlZgppZ+VLlY7N33q108Sa+fhmuc+sWQYwY=",
+        version = "v1.0.0-20180628173108-788fd7840127",
+    )
+    go_repository(
+        name = "in_gopkg_yaml_v2",
+        importpath = "gopkg.in/yaml.v2",
+        sum = "h1:ZCJp+EgiOT7lHqUV2J862kp8Qj64Jo6az82+3Td9dZw=",
+        version = "v2.2.2",
+    )
+    go_repository(
```
jayconrod: We shouldn't replace `org_golang_x_sys` or `org_golang_x_tools`. `io_bazel_rules_go` declares these already. It provides additional build rules needed for nogo and other infrastructure to work.

Depending on when these get resolved, I don't think these declarations will actually override the other ones, but all the more reason not to include them.

WORKSPACE
blico: Ahh makes sense, my bad for not catching this.

Looks like we there are no unique test dependencies, so we don't need anything new in WORKSPACE.

PR #542, opened 2019-06-07
gazelle:exclude: use doublestar.Match for glob support

robfig:
Not sure if you're interested in this one, but I found it helpful. No pressure to merge if not


jayconrod:
Sorry, I'd rather not add a dependency unless it's really really needed. `path.Match` would probably be good enough to implement globbing excludes.

Issue #203, opened 2018-05-17
Read flags from pkg-config files


abergmeier:
The following does not work with _rules_go 0.12_ and _bazel 0.13_:

```
go_repository(
    name = "foo",
    importpath = "github.com/keroserene/go-webrtc",
    commit = "a1272c08ab1d5ca154c6794ddc5f73d2e576fe1b",
)
```

First off - I am quite impressed that it works as far as it does.
It seems to not find `_cgo_export.h`, though.

abergmeier:
Using the following it does at least compile:

```
load("@io_bazel_rules_go//go:def.bzl", "go_library", "go_test")

go_library(
    name = "go_default_library",
    srcs = [
        "configuration.go",
        "ctestenums.cc",
        "ctestenums.h",
        "datachannel.cc",
        "datachannel.go",
        "datachannel.h",
        "datachannel.hpp",
        "ice.go",
        "logging.go",
        "peerconnection.cc",
        "peerconnection.go",
        "peerconnection.h",
        "sdp.go",
        "utils.go",
    ] + glob(["include/**/*.h"]),
    cgo = True,
    clinkopts = ["-L/lib"],
    cxxopts = [
        "-std=c++0x",
        "-Iexternal/com_github_keroserene_go_webrtc/include",
        "-Ibazel-out/k8-fastbuild/bin/external/com_github_keroserene_go_webrtc/linux_amd64_stripped/go_default_library.cgo_codegen~",
    ],
    importpath = "github.com/keroserene/go-webrtc",
    visibility = ["//visibility:public"],
)

go_test(
    name = "go_default_test",
    srcs = [
        "configuration_test.go",
        "datachannel_test.go",
        "ice_test.go",
        "peerconnection_test.go",
        "sdp_test.go",
        "utils_test.go",
    ],
    embed = [":go_default_library"],
    deps = ["@com_github_smartystreets_goconvey//convey:go_default_library"],
)
```

jayconrod:
I'll look into this as soon as I can, but due to conference and travel schedules, it will be at least a few days before I can look into this.

A few things I notice from the build files:

* `glob(["include/**/*.h"])` - Gazelle will only ever consider files in the same directory, so it won't emit globs like this. Also note: `glob` only reaches down as far as the next build file.
[...]

PR #556, opened 2019-06-20
[WIP] Add support for cgo pkg-config

odinuge:
This fixes https://github.com/bazelbuild/bazel-gazelle/issues/203

The tests for pkg-config require libpng and libjpeg to be installed, and skips without failing if one of them isn't installed. 

Signed-off-by: Odin Ugedal <odin@ugedal.com>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,8 @@ addons:
     packages:
       - wget
       - pkg-config
+      - libjpeg-dev
+      - libpng-dev
 
 before_install:
   - |
--- a/language/go/fileinfo.go
+++ b/language/go/fileinfo.go
@@ -35,6 +35,7 @@ import (
 	"github.com/bazelbuild/bazel-gazelle/config"
 	"github.com/bazelbuild/bazel-gazelle/language/proto"
 	"github.com/bazelbuild/bazel-gazelle/rule"
+	"os/exec"
 )
 
 // fileInfo holds information used to decide how to build a file. This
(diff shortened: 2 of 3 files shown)


jayconrod:
Sorry, I don't think pkg-config can be reasonably supported. It fundamentally relies on system configuration outside the Bazel workspace. Flags appropriate for one system may not be appropriate for another. Different versions of a library may be installed (or not installed at all). Cross compilation and remote execution can't work.

A better way to depend on these libraries is to pull them in as `git_repository` or `http_archive` rules with build files that provide `cc_library` targets. Unfortunately, there's not an easy way to automate that.

BenTheElder:
Can you elaborate on the better way @jayconrod? This is blocking Kubernetes from using a non-ancient version of libseccomp bindings.

jayconrod:
@BenTheElder Sure, the key thing is that there needs to be a `cc_library` for the C/C++ dependency and `go_library` should depend on that via `cdeps`. `cdeps` must be added manually, but Gazelle will leave it alone.

Preparing the `cc_library` of course is the tricky part. There are a lot of ways to do that. I don't know enough about this library to know the best approach, but after a quick look at its configure.ac, it looks okay.

* You could check in precompiled static or dynamic libraries for target platforms, then wrap them with a [`cc_import`](https://docs.bazel.build/versions/master/be/c-cpp.html#cc_import). `cc_import` can be used in place of a `cc_library`.
* You could run `./configure && make -j` in a `genrule`, then wrap the outputs in `cc_import`
* It might be a simple enough library to write a `cc_library` or a handful of related `cc_library` rules.

BenTheElder:
Thanks for the response, that sounds pretty involved :(

Fom some testing the library involved just needs a link flag to link the cgo to the library, `pkg-config` is not giving any other flags. Previously the go library configured that with a cgo directive but they switched to pkg-config (presumably for correctness?)

Kubernetes still checks all of its dependency code into vendor/ so I don't think this work around will work because it will cause gazelle to error out on parsing the code in vendor.

I'd be happy to manually patch in that link flag to `clinkopts` in the `BUILD` for the platforms we build the dependent binary (`kubelet`) on (Linux only..) and just ignore the pkg-config when building with bazel... in order to update the dependency but continue building as we did before.

Could we get a mode in gazelle to warn / skip / ignore pkg-config directives instead of failing?

jayconrod:
Gazelle should just print a warning for `pkg-config` directives. From the code, it looks like the error is just passed to `log.Printf`. If it's failing and not generating anything, please open an issue.

After a warning though, manually setting `clinkopts` is probably okay but "at your own risk" :) Just make sure the library is there on the execution platform.

BenTheElder:
thanks! I'm not sure what I was seeing before, this seems to be working I
think ...

PR #565, opened 2019-06-29
Pass args to Configure with config.ConfigureArgs

alexshtin:
In my extension I need list of files and directories for current `rel` in `Configure` func. I found out that `Walk` already has this data and I just need to pass it to `Configure`. Because it is braking change I also decided to convert all arguments to `ConfigureArgs` (same way as `GenerateRule` already has) and at least make it extensible for future.

I also modified `go/config.go` to use the `RegularFiles` slice instead of `os.Stat`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ bazel-bin
 bazel-genfiles
 bazel-out
 bazel-testlogs
+.idea
--- a/cmd/gazelle/fix-update.go
+++ b/cmd/gazelle/fix-update.go
@@ -155,7 +155,7 @@ func (ucr *updateConfigurer) CheckFlags(fs *flag.FlagSet, c *config.Config) erro
 
 func (ucr *updateConfigurer) KnownDirectives() []string { return nil }
 
-func (ucr *updateConfigurer) Configure(c *config.Config, rel string, f *rule.File) {}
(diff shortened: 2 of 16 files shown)


jayconrod:
Sorry, I'm not accepting breaking changes to this API right now.

We may need to make some changes at some point in the future in order to make `update-repos` language-agnostic, but the migration will need to be carefully planned, and we'll need to issue a v2 so the old API continues working. If you'd like a change to be included in the API, please file an issue proposing a change.

Maybe there's another way to fix your problem though? Can the work you're doing in `Configure` be deferred until `GenerateRules`, or is it important that it happens in pre-order instead of post-order? Can you read files in the current directory using `ioutil.ReadDir` or something similar?

Also, if you're working on an extension you'd like to publish externally, I'd be happy to merge a PR adding it to the list in [Supported languages](https://github.com/bazelbuild/bazel-gazelle/blob/master/extend.rst#supported-languages) when it's ready.

alexshtin:
Ok, I understand.
Sure, I have the solution for my particular problem. I don't even need `ReadDir` there. I know what I am looking for, so simple `os.Stat` is enough. I just wanted to make it more elegant and perform since these file lists are already there and save on file system I/O calls. And yes, order is important.

I don't think that my extension would help the community. It is for some test executor specific to our company.

PR #572, opened 2019-07-09
Fix go_repository rule when using urls

siddharthab:
When urls attribute is set, go_repository rule currently throws this error:
```
ERROR: local variable 'generate' is referenced before assignment.
```

This change moves the definition of generate to just before its first use.

### DIFF
--- a/internal/go_repository.bzl
+++ b/internal/go_repository.bzl
@@ -84,6 +84,7 @@ def _go_repository_impl(ctx):
     else:
         fail("one of urls, commit, tag, or importpath must be specified")
 
+    generate = ctx.attr.build_file_generation == "on"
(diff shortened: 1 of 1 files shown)


internal/go_repository.bzl
```diff
@@ -84,6 +84,7 @@ def _go_repository_impl(ctx):
     else:
         fail("one of urls, commit, tag, or importpath must be specified")
 
+    generate = ctx.attr.build_file_generation == "on"
```
jayconrod: Thanks for looking at this, but this fix isn't sufficient. The whole block where `generate` is defined should move up.

siddharthab: Thanks; I should have noticed. Done. PTAL.

PR #590, opened 2019-07-17
Suppress stderr from fetchrepo if successful

siddharthab:
Without this change, most repos while fetching will print messages like
this on stderr:
  .../go_repository.bzl:127:13: fetch_repo: go: finding google.golang.org/api v0.1.0

### DIFF
--- a/cmd/fetch_repo/module.go
+++ b/cmd/fetch_repo/module.go
@@ -62,13 +62,15 @@ func fetchModule(dest, importpath, version, sum string) error {
 	}
 
 	buf := &bytes.Buffer{}
+	bufErr := &bytes.Buffer{}
 	cmd := exec.Command(goPath, "mod", "download", "-json", importpath+"@"+version)
 	cmd.Stdout = buf
-	cmd.Stderr = os.Stderr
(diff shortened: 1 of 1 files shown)


cmd/fetch_repo/module.go
```diff
@@ -61,21 +60,19 @@ func fetchModule(dest, importpath, version, sum string) error {
 		return fmt.Errorf("error closing temporary go.mod: %v", err)
 	}
 
-	buf := &bytes.Buffer{}
 	cmd := exec.Command(goPath, "mod", "download", "-json", importpath+"@"+version)
-	cmd.Stdout = buf
-	cmd.Stderr = os.Stderr
-	dlErr := cmd.Run()
+	out, dlErr := cmd.Output()
```
jayconrod: `cmd.Output` may truncate stderr. Instead, let's set `cmd.Stderr` to a separate `*bytes.Buffer`.

siddharthab: Thanks. Done. Happy to follow any styling, variable renaming suggestions, etc. as well.

PR #595, opened 2019-07-26
repo: add version/commit select for update-repos

weixiao-huang:
make `bazel run //:gazelle update-repos` can support version select as follow:

```
bazel run //:gazelle update-repos github.com/sirupsen/logrus@v1.4.0
bazel run //:gazelle update-repos github.com/sirupsen/logrus@dae0fa8
```

### DIFF
--- a/cmd/gazelle/integration_test.go
+++ b/cmd/gazelle/integration_test.go
@@ -1205,8 +1205,6 @@ gazelle_dependencies()
 	dir, cleanup := testtools.CreateFiles(t, files)
 	defer cleanup()
 
-	os.Setenv("GOCACHE", os.TempDir())
-	os.Setenv("GOPATH", os.TempDir())
 	args := []string{"update-repos", "github.com/sirupsen/logrus@v1.3.0"}
 	if err := runGazelle(dir, args); err != nil {
 		t.Fatal(err)
--- a/cmd/gazelle/integration_test.go
+++ b/cmd/gazelle/integration_test.go
@@ -1164,6 +1164,98 @@ go_library(
 	})
 }
 
+func TestUpdateReposWithQueryToWorkspace(t *testing.T) {
+	files := []testtools.FileSpec{
+		{
+			Path: "WORKSPACE",
+			Content: `
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+
+http_archive(
+    name = "io_bazel_rules_go",
+    urls = [
+        "https://storage.googleapis.com/bazel-mirror/github.com/bazelbuild/rules_go/releases/download/0.19.1/rules_go-0.19.1.tar.gz",
+        "https://github.com/bazelbuild/rules_go/releases/download/0.19.1/rules_go-0.19.1.tar.gz",
+    ],
+    sha256 = "8df59f11fb697743cbb3f26cfb8750395f30471e9eabde0d174c3aebc7a1cd39",
+)
+
+http_archive(
+    name = "bazel_gazelle",
+    urls = [
+        "https://storage.googleapis.com/bazel-mirror/github.com/bazelbuild/bazel-gazelle/releases/download/0.18.1/bazel-gazelle-0.18.1.tar.gz",
+        "https://github.com/bazelbuild/bazel-gazelle/releases/download/0.18.1/bazel-gazelle-0.18.1.tar.gz",
+    ],
+    sha256 = "be9296bfd64882e3c08e3283c58fcb461fa6dd3c171764fcc4cf322f60615a9b",
+)
+
+load("@io_bazel_rules_go//go:deps.bzl", "go_register_toolchains", "go_rules_dependencies")
+
+go_rules_dependencies()
+
+go_register_toolchains(nogo = "@bazel_gazelle//:nogo")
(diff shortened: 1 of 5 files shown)


jayconrod:
Could you open an issue with the feature request first? I'd prefer to discuss that before reviewing any implementation. `update-repos` should some day support multiple languages, so please think about whether this interface is suitable for Java or Python for example.

repo/remote.go
```diff
@@ -41,7 +41,15 @@ import (
 // information may already be locally available. Frequently though, information
 // will be fetched over the network, so this function may be slow.
 func UpdateRepo(rc *RemoteCache, modPath string) (Repo, error) {
-	name, version, sum, err := rc.ModVersion(modPath, "latest")
+	arg := strings.SplitN(modPath, "@", 2)
```
jayconrod: It's more clear to parse it like this:

```
query := "latest"
if i := strings.IndexByte(modPath, '@'); i >= 0 {
  query = modPath[i+1:]
  modPath = modPath[:i]
}
```

weixiao-huang:
I try to add test for `bazel run //:gazelle update-repos github.com/sirupsen/logrus@v1.3.0`. I can pass test by using
```
go test -timeout 30s github.com/bazelbuild/bazel-gazelle/cmd/gazelle -run '^(TestUpdateReposWithQueryToWorkspace)$'
```
However, while using bazel test as
```
bazel test --test_output=errors --action_env="GTEST_COLOR=1" cmd/gazelle:go_default_test
```
I got a lot of errors as shown in the ci. The typical tip for the errors was:
```
integration_test.go:1210: finding module version and sum for github.com/sirupsen/logrus@v1.3.0: exit status 1: build cache is required, but could not be located: GOCACHE is not defined and $HOME is not defined
```
Also, in `cmd/gazelle/integration_test.go`, I haven't see any tests for `bazel run //:gazelle update-repos example.com/new/repo`. So is there some problems while testing it in bazel's sandbox environment?

jayconrod:
Sounds like some environment variables may not be passed to `go mod download` during that test then? `findGoTool` should be used to locate the Go command; `GOROOT` should not be set explicitly. `GO111MODULE=on` should be set. `GOPATH` should be set but not overridden if it's not set already and `HOME` isn't set. Probably to a temporary directory

weixiao-huang:
Sorry for delay. I try to add this two lines for fixing the error:
```golang
	os.Setenv("GOCACHE", os.TempDir())
	os.Setenv("GOPATH", os.TempDir()
```
it worked in MacOS and Ubuntu, but it seems not working in buildkite/bazel-gazelle/pr/windows-openjdk-8 and other 3 ci. I don't know why. And I try to use `dir` variable defined before instead of `os.TempDir()`, also not work.
So how can I pass all of the ci? I feel confused

cmd/gazelle/integration_test.go
```diff
@@ -1164,6 +1164,98 @@ go_library(
...
+gazelle_dependencies()
+`,
+		},
+	}
+
+	dir, cleanup := testtools.CreateFiles(t, files)
+	defer cleanup()
+
+	os.Setenv("GOCACHE", os.TempDir())
+	os.Setenv("GOPATH", os.TempDir())
```
jayconrod: `os.TempDir()` will return something like `/tmp`. We can't dump files there without deleting them. Create directories with `io/ioutil.TestDir` instead and delete them with `os.RemoveAll`.

Also, these variables are global and will affect other tests. That's probably okay: other tests will need these in the future. However, these should be set in a `TestMain` function. Make sure to clean up the directories, even if `m.Run` panics.

One more thing: the windows tests are failing because `go.exe` is not in `PATH`. Gazelle will look in `GOROOT/bin`, so let's so GOROOT. Add `data = ["@go_sdk//:files"],` to the `go_test` rule `//cmd/gazelle:go_default_test` to make sure the go tool and all the related files are visible in the sandbox. Add `args = ["-goroot=$(location @go_sdk//:ROOT)"],` to the same rule. Define a `flag.String` for `goroot` in the test. In the `TestMain`, call `flag.Parse` and set the `GOROOT` environment variable to the `filepath.Dir` of that flag.

weixiao-huang:
Thanks for your advice. According to your tips, I add `args = ["-goroot", "$(location @go_sdk//:ROOT)"],` into `//cmd/gazelle:go_default_test`. However, I got the error below:
```
/workdir/cmd/gazelle/BUILD.bazel:58:1: in args attribute of go_test rule //cmd/gazelle:go_default_test: label '@go_sdk//:ROOT' in $(location) expression is not a declared prerequisite of this rule. Since this rule was created by the macro 'go_test_macro', the error might have been caused by the macro implementation in /workdir/cmd/gazelle/BUILD.bazel:58:1
```
This also confuses me a lot

jayconrod:
I pushed some fixes rather than doing another round of code review. It looks like the test is still failing on Windows. My Windows machine isn't working right now, but I'll take a look at the failure when I get home.

weixiao-huang:
Finally, it can be passed! Thank you for your help. Need I rebase and squash these commits?

jayconrod:
Nope, all good. Thanks for working on this.

PR #598, opened 2019-07-29
Expose "resolve" directive as a flag

linzhp:
...so `go_repository` rules can tell Gazelle how to resolve some dependencies


jayconrod:
Superceded by #603. Thanks for working on this!

PR #597, opened 2019-07-27
Add proto_strip_import_prefix to flags

linzhp:
`proto_strip_import_prefix` is a Gazelle directive but not a flag. This PR adds it to the flags so it can be passed down from `go_repository` rule to generate proto rules properly in third-party libraries.

PR #603, opened 2019-07-30
go_repository takes Gazelle directives

linzhp:
As discussed in #598, this PR make `go_repository` take a list of Gazelle directives to be written to the root level build file.

This PR supercedes #597 and #598

### DIFF
--- a/internal/go_repository.bzl
+++ b/internal/go_repository.bzl
@@ -88,7 +88,7 @@ def _go_repository_impl(ctx):
     generate = ctx.attr.build_file_generation == "on"
     if ctx.attr.build_file_generation == "auto":
         generate = True
-        for name in set(["BUILD", "BUILD.bazel"] + ctx.attr.build_file_name.split(",")):
+        for name in depset(["BUILD", "BUILD.bazel"] + ctx.attr.build_file_name.split(",")).to_list():
             path = ctx.path(name)
             if path.exists and not env_execute(ctx, ["test", "-f", path]).return_code:
                 generate = False
--- a/internal/go_repository.bzl
+++ b/internal/go_repository.bzl
@@ -88,7 +88,7 @@ def _go_repository_impl(ctx):
     generate = ctx.attr.build_file_generation == "on"
     if ctx.attr.build_file_generation == "auto":
         generate = True
-        for name in ["BUILD", "BUILD.bazel", ctx.attr.build_file_name]:
+        for name in set(["BUILD", "BUILD.bazel"] + ctx.attr.build_file_name.split(",")):
             path = ctx.path(name)
             if path.exists and not env_execute(ctx, ["test", "-f", path]).return_code:
                 generate = False
@@ -147,7 +147,13 @@ def _go_repository_impl(ctx):
             print("fetch_repo: " + result.stderr)
 
     if generate:
-        # Build file generation is needed
+        # Build file generation is needed. Populate Gazelle directive at root build file
+        ctx.file(
+            ctx.attr.build_file_name.split(",")[0],
+            "\n".join(["# gazelle:" + d for d in ctx.attr.gazelle_directives]),
+        )
+
+        # Run Gazelle
         _gazelle = "@bazel_gazelle_go_repository_tools//:bin/gazelle{}".format(executable_extension(ctx))
         gazelle = ctx.path(Label(_gazelle))
         cmd = [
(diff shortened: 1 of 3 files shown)


internal/go_repository.bzl
```diff
@@ -88,7 +88,7 @@ def _go_repository_impl(ctx):
     generate = ctx.attr.build_file_generation == "on"
     if ctx.attr.build_file_generation == "auto":
         generate = True
-        for name in ["BUILD", "BUILD.bazel", ctx.attr.build_file_name]:
+        for name in depset(["BUILD", "BUILD.bazel"] + ctx.attr.build_file_name.split(",")).to_list():
```
jayconrod: Depsets are suprisingly complicated, and deduplication isn't really important here. Maybe just:

```
["BUILD", "BUILD.bazel"] + ctx.attr.build_file_name.split(",")
```

internal/go_repository.bzl
```diff
@@ -147,7 +147,13 @@ def _go_repository_impl(ctx):
             print("fetch_repo: " + result.stderr)
 
     if generate:
-        # Build file generation is needed
+        # Build file generation is needed. Populate Gazelle directive at root build file
+        ctx.file(
```
jayconrod: We should only write this if `ctx.attr.gazelle_directives` is set.

Ideally we would append to the file rather than replacing it if it already exists. I don't think there's a simple way to do that though. Perhaps setting this attribute should be an error if the file exists.

internal/go_repository.bzl
```diff
@@ -147,7 +147,13 @@ def _go_repository_impl(ctx):
...
+            ctx.attr.build_file_name.split(",")[0],
```
jayconrod: It's possible that a build file exists with a different name, for example, if `build_file_generation = "on"` and a `BUILD` file is present. We should check this in the loop at L91, which will need to be unconditional.

internal/go_repository.bzl
```diff
@@ -147,7 +147,13 @@ def _go_repository_impl(ctx):
...
+            "\n".join(["# gazelle:" + d for d in ctx.attr.gazelle_directives]),
```
jayconrod: Let's call this attribute `directives`, and let's not force elements to start with `gazelle:`. I think buildifier has some of its own directives, and Gazelle integrates part of that. It's possible other tools with different prefixes may be used in the future.

internal/go_repository.bzl
linzhp: If `build_file_generation = "auto"` and build files exist in the repo, Gazelle will not run and no Gazelle directive is needed. We can just ignore them. If `build_file_generation = "on"`, `go_repository` will override the existing build files if there is any. Overriding the root build file with Gazelle directives feels consistent with the behavior. Do you think it necessary to report an error for this case?

internal/go_repository.bzl
```diff
@@ -85,14 +85,17 @@ def _go_repository_impl(ctx):
...
-    generate = ctx.attr.build_file_generation == "on"
-    if ctx.attr.build_file_generation == "auto":
+    existing_build_file = ""
+    for name in ["BUILD", "BUILD.bazel"] + ctx.attr.build_file_name.split(","):
+        path = ctx.path(name)
+        if path.exists and not env_execute(ctx, ["test", "-f", path]).return_code:
+            existing_build_file = name
+            break
+
+    if ctx.attr.build_file_generation == "on" or (not existing_build_file and ctx.attr.build_file_generation == "auto"):
```
jayconrod: Nit: simplify to

```
generate = (ctx.attr.build_file_generation == "on" or
    (not existing_build_file and ctx.attr.build_file_generation == "auto"))
```

internal/go_repository_test.go
```diff
@@ -44,6 +44,20 @@ go_repository(
...
+    directives = [
+        "gazelle:proto_strip_import_prefix /include/",
+        "gazelle:exclude include/mesos/attributes.hpp",
+        "gazelle:exclude include/mesos/executor.hpp",
+        "gazelle:exclude include/mesos/hook.hpp",
+    ],
+    importpath = "github.com/apache/mesos",
+    sum = "h1:65atl5ZUZ7fzMM/oIVFV5VdO0plKO0hye9taZR6ssHw=",
+    version = "v0.0.0-20190710134757-4ae064484664",
+)
```
jayconrod: I don't think the test will actually build this yet. Add a new test function `TestDirectives` that builds a target that wouldn't otherwise build.

If you can avoid using protos in the test, that would be ideal. protoc takes a long, long time to compile if it's not cached.

The simplest way I can think of would be to `# gazelle:exclude` some directory, then run `bazel query @repo//...` and verify that that rules were not generated for that directory.

PR #612, opened 2019-08-06
Add ios platform and OS aliases

### LINKED ISSUE #302: Linux dependencies should appear in android select cases
If a package is imported by a source file with the `linux` constraint, the dependency for that package should be listed in both `linux` and `android` cases in `select` expressions. Same for cgo options.

### LINKED ISSUE #566: language/go: generate iOS cases in select expressions
The `config_setting` `@io_bazel_rules_go//go/platform:darwin` is activated if the target operating system is `@bazel_tools//platforms:osx`. Unfortunately, there's no way to match both `osx` and `ios`, which is what the `darwin` tag requires.

bazelbuild/rules_go#2118 adds `@io_bazel_rules_go//go/platform:ios`. Gazelle should generate additional `select` cases with this when a `darwin` tag is found.

steeve:
Anything that matches `darwin` should match `ios`, but not the other way
around. Same for `android` & `linux`.

Introduce aliases and extend the matching to them.

I haven't tested it in situ but the build files look good. Also, not sure about the `android -> linux` alias.

### DIFF
--- a/language/go/testdata/platforms/BUILD.want
+++ b/language/go/testdata/platforms/BUILD.want
@@ -21,7 +21,13 @@ go_library(
     _gazelle_imports = [
         "example.com/repo/platforms/generic",
     ] + select({
+        "@io_bazel_rules_go//go/platform:android": [
+            "example.com/repo/platforms/linux",
+        ],
         "@io_bazel_rules_go//go/platform:darwin": [
+            "example.com/repo/platforms/darwin",
+        ],
+        "@io_bazel_rules_go//go/platform:ios": [
             "example.com/repo/platforms/darwin",
         ],
         "@io_bazel_rules_go//go/platform:linux": [
@@ -33,6 +39,9 @@ go_library(
     copts = [
         "-DGENERIC",
     ] + select({
+        "@io_bazel_rules_go//go/platform:android": [
+            "-DLINUX",
+        ],
         "@io_bazel_rules_go//go/platform:linux": [
             "-DLINUX",
         ],
--- a/cmd/gazelle/integration_test.go
+++ b/cmd/gazelle/integration_test.go
@@ -149,6 +149,28 @@ import (
 `,
 		},
 		{
+			Path: "foo_android_build_tag.go",
+			Content: `
+// +build android
+
+package foo
+
+import (
+    _ "example.com/foo/outer_android_build_tag"
+)
+`,
+		},
+		{
+			Path: "foo_android.go",
+			Content: `
+package foo
+
+import (
+    _ "example.com/foo/outer_android_suffix"
+)
+`,
+		},
+		{
 			Path: "bar.go",
 			Content: `// +build linux
 
@@ -156,6 +178,8 @@ package foo
 `,
 		},
 		{Path: "outer/outer.go", Content: "package outer"},
+		{Path: "outer_android_build_tag/outer.go", Content: "package outer_android_build_tag"},
+		{Path: "outer_android_suffix/outer.go", Content: "package outer_android_suffix"},
 		{Path: "outer/inner/inner.go", Content: "package inner"},
 	})
 	want := `load("@io_bazel_rules_go//go:def.bzl", "go_library")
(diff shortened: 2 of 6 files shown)


jayconrod:
This is marked as a draft PR. Are you looking for more feedback?

It looks like `//cmd/gazelle:go_default_test` needs to be updated.

In the Go generate test, it would be good to confirm that a file with the `android` tag or suffix is in the list for `@io_bazel_rules_go//go/platform:android` but not `linux`.

It's unfortunate that `ios` is not an officially recognized tag in the Go command, but I think this usage of it makes sense. We'll need a corresponding change in rules_go source filtering though to make this work end to end.

steeve:
Just coming back on this

> This is marked as a draft PR. Are you looking for more feedback?

I was! I'm gonna mark it ready for review.

> It looks like `//cmd/gazelle:go_default_test` needs to be updated.

Will do.

> In the Go generate test, it would be good to confirm that a file with the `android` tag or suffix is in the list for `@io_bazel_rules_go//go/platform:android` but not `linux`.

Will do.

> We'll need a corresponding change in rules_go source filtering though to make this work end to end.

I'm not sure I get that ?

steeve:
Updated the test and added an android only package in the test

language/go/testdata/platforms/BUILD.want
```diff
@@ -21,9 +21,15 @@ go_library(
     _gazelle_imports = [
         "example.com/repo/platforms/generic",
     ] + select({
+        "@io_bazel_rules_go//go/platform:android": [
```
jayconrod: Let's also test that a file with the `_android.go` suffix and no `+build` tag gets included.

I think that makes sense in this test case, but it could be anywhere in the generator test.

jayconrod:
> We'll need a corresponding change in rules_go source filtering though to make this work end to end.

What I meant by this is that there is no special handling for the `ios` tag in rules_go. As with the Go command, it has no special meaning to the toolchain. To avoid filtering out these files, the `ios` toolchains should switch that tag on.

language/go/testdata/platforms/BUILD.want
steeve: done!

steeve:
> What I meant by this is that there is no special handling for the `ios` tag in rules_go. As with the Go command, it has no special meaning to the toolchain. To avoid filtering out these files, the `ios` toolchains should switch that tag on.

Can you tell me where it's done ? I can't find it with a quick look.
I though `rules_go` filtered files based on they extension, not their suffix ?

jayconrod:
> Can you tell me where it's done ? I can't find it with a quick look.

Ah, sorry, should have been more detailed. We need to make sure the `ios` build tag is enabled without requiring the user to explicitly write `--define gotags=ios` on the command line. Probably the right way to do this is to add a `go_tags` attribute to `go_toolchain`, a list of strings that are on by default. We could add a column to the `PLATFORMS` table for this, then declare those in `declare_toolchains`. It would only need to be set for the `ios` platforms. `_go_context_data_impl` retrieves the toolchain and also `ctx.var["gotags"]`, so that's the right place to combine the two lists. Everything after that should work without modification.

> I though rules_go filtered files based on they extension, not their suffix ?

It does both. We emulate what `go build` does by calling `go/build.Context.MatchFile` in `go/tools/builders/filter.go`. Gazelle uses the same logic for building platform-specific `deps`. However, rules_go still needs to do the filtering at execution time because we don't generate platform-specific `srcs`, and for generated files, there may be no way to tell ahead of time whether a file should be included or not.

steeve:
> Ah, sorry, should have been more detailed. We need to make sure the `ios` build tag is enabled without requiring the user to explicitly write `--define gotags=ios` on the command line. Probably the right way to do this is to add a `go_tags` attribute to `go_toolchain`, a list of strings that are on by default. We could add a column to the `PLATFORMS` table for this, then declare those in `declare_toolchains`. It would only need to be set for the `ios` platforms. `_go_context_data_impl` retrieves the toolchain and also `ctx.var["gotags"]`, so that's the right place to combine the two lists. Everything after that should work without modification.

Ah gotcha. The `ios` tag is automatically added by the apple platform detection at https://github.com/bazelbuild/rules_go/blob/master/go/platform/apple.bzl#L44, so it does work for now. Perhaps putting it in the toolchain would be cleaner.

> It does both. We emulate what `go build` does by calling `go/build.Context.MatchFile` in `go/tools/builders/filter.go`. Gazelle uses the same logic for building platform-specific `deps`. However, rules_go still needs to do the filtering at execution time because we don't generate platform-specific `srcs`, and for generated files, there may be no way to tell ahead of time whether a file should be included or not.

Gotcha, I'll take a look

PR #624, opened 2019-08-21
Fix update-repos -build_tags documentation [ci skip]

asv:
To redefine `gotags`, the `define` flag should be used, not the `features`.


asv:
@googlebot I signed it!

asv:
Oh my fault. Fixed.

PR #627, opened 2019-08-22
autogazelle: use default syntax for expanding $1

deadNightTiger:
When calling `bazel` without parameters, the script was giving an error `$1: unbound variable`.
This fixes it by using an explicit default value of an empty string.


deadNightTiger:
@googlebot I signed it!

Issue #619, opened 2019-08-13
When an external library depends on an internal package of another dependency, Gazelle creates a rule that doesn't compile.


draftcode:
# What version of gazelle are you using?

0.18.1

### What version of rules_go are you using?

0.19.1

### What version of Bazel are you using?

0.28.1

### Does this issue reproduce with the latest releases of all the above?

Yes

### What operating system and processor architecture are you using?

Linux AMD64

### What did you do?

Create a workspace with the following files.

https://gist.github.com/draftcode/8b60bb587c3739d1a64bd4f61343e8b9

This is basically creating a minimum workspace that depends on cloud.google.com/go/logging.

Build with `bazel build ...`.

### What did you expect to see?

Compile without an error.

### What did you see instead?

Build failure like this:

```
ERROR: $HOME/.cache/bazel/_bazel_masayasuzuki/15220e3a80e87c9c56aecefb5d0ff5e7/external/com_google_cloud_go_logging/BUILD.bazel:3:1: in go_library rule @com_google_cloud_go_logging//:go_default_library: target '@com_google_cloud_go//internal/version:go_default_library' is not visible from target '@com_google_cloud_go_logging//:go_default_library'. Check the visibility declaration of the former target if you think the dependency is legitimate
ERROR: Analysis of target '//:go_default_library' failed; build aborted: Analysis of target '@com_google_cloud_go_logging//:go_default_library' failed; build aborted
INFO: Elapsed time: 0.161s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (0 packages loaded, 0 targets configured)
zsh: exit 1     bazel build ...
```

---

It seems cloud.google.com/go/logging depends on cloud.google.com/go's internal package (https://github.com/googleapis/google-cloud-go/blob/master/logging/logging.go#L43). Because Gazelle uses a heuristics on creating a visibility, the internal package is not visible from cloud.google.com/go/logging.

Perhaps Gazelle can generate a package_group that includes all external libraries and use that to let external libraries depend on other external libraries. This way, the code in the workspace still cannot depend on the internal packages of external libraries, but the external libraries can depend on other external libraries' internal packages.

jayconrod:
Oh, fun, cross-module use of `internal` directories.

For now, I'd suggest working around this with `patches`.

1. Copy `$(bazel info output_base)/external/com_google_cloud_go)` to a directory `a`.
2. `cp -r a b`
[...]

PR #641, opened 2019-09-22
Revert "language/go, go_repository: submodule visibility (#637)"

jayconrod:
This reverts commit d55bfe02e49d7224fa6bf0a9d908f640837bf76a.

Updates #619. Parts of this PR will be added back after some refactoring work is done, but for now, it's easier to proceed without it.


linzhp:
Are you going to merge this? #637 broke some of our existing builds. If you are going to revert it and refactor, we will hold our fix until the refactoring is done.

jayconrod:
I was going to wait until the refactor was ready to merge this, but if it's breaking builds for you, I'll merge it now.

What exactly broke for you?

jayconrod:
Just to give a heads up on what the refactoring is doing: I'm moving some Go-specific stuff out of `repo` and `rule` into `language/go`, and I'm adding interfaces to `language` that will let `update-repos` be language-agnostic. It's a breaking change in terms of API compatibility, but I don't expect that packages other than `language/go` are relying on that functionality.

But in #642, you mentioned you're working on another extension that dealt with Go imports. Does that need anything from `repo`, other than `RemoteCache`?

achew22:
@jayconrod since you're merging breaking changes, could we take this opportunity to merge something like https://github.com/bazelbuild/bazel-gazelle/pull/565?

linzhp:
@jayconrod I am working on a Gazelle extension that generate gomock rules and go_library rules that compile the mock. There are some changes I need (so far) from Gazelle:

1. The Gazelle extension needs to know what Go package it needs to create mocks for. I designed a Gazelle directive to configure the package path, which will need to translate into Bazel label to assign to `library` parameter of `gomock` and transform into the `importpath` of the go_library according to the convention we decided. This is done with #642.
2. The generated mock file should be only consumed by the go_library rule the extension generates, not to be mixed with regular Go files in the `go_default_library` that the Go extension generates. This is done with #643 
3. The generated go_library sometimes needs to depend on the internal package of a third-party library. For example, if we mock the `Client` interface in `go.uber.org/cadence/client`, the generated mock file will have imports like this:

```
import (
        context "context"
        gomock "github.com/golang/mock/gomock"
        shared "go.uber.org/cadence/.gen/go/shared"
        internal "go.uber.org/cadence/internal"
        reflect "reflect"
)
```

Before #637, all packages in third-party libraries are public, but now `go.uber.org/cadence/internal` is only visible to subpackages of `go.uber.org/cadence`, and the generated mock is not one of them. breaking our build.

I was planning to add a Gazelle directive to provide additional packages that internal packages should be visible to. Since you are reverting #637, I will just wait until your refactoring is done.

jayconrod:
> since you're merging breaking changes, could we take this opportunity to merge something like #565?

@achew22 I'd rather not ever break the `language.Language` interface or interfaces embedded there. If we need breaking changes there, we'd need to introduce new interfaces instead (`Language2`, `Configurer2`, ...), and old version would continue to be supported. I'd rather not deal with that much complication though.

jayconrod:
> Before #637, all packages in third-party libraries are public, but now go.uber.org/cadence/internal is only visible to subpackages of go.uber.org/cadence, and the generated mock is not one of them, breaking our build.

This sounds like it was not working as intended? Packages under `go.uber.org/cadence/internal` should have had visibility `//:__subpackages__` already, which would prevent packages outside that module from importing them. The problem in #619 (which #637 was trying to solve) was that `cloud.google.com/go` did the same thing: a non-root module was referencing internal packages in its parent module, and the visibility declaration was preventing that.

> I was planning to add a Gazelle directive to provide additional packages that internal packages should be visible to. Since you are reverting #637, I will just wait until your refactoring is done.

Seems reasonable. Maybe file an issue to discuss? I was thinking about something similar instead of declaring submodules, but it seemed like it would inevitably be language specific: for example, you probably don't want to add visibility for test rules.

PR #643, opened 2019-09-22
Excluding consumed generated files from go_library

linzhp:
Some generated files may have been consumed by other rules, either handwritten or generated by other extensions. The `go` extension should not add them into the `go_default_library` rule in such case.

### DIFF
--- a/language/go/generate.go
+++ b/language/go/generate.go
@@ -236,6 +236,9 @@ func (gl *goLang) GenerateRules(args language.GenerateArgs) language.GenerateRes
 			for _, f := range r.AttrStrings("srcs") {
 				consumedFileSet[f] = true
 			}
+			if f := r.AttrString("src"); f != "" {
+				consumedFileSet[f] = true
+			}
 		}
 		for _, f := range genFiles {
 			if regularFileSet[f] || consumedFileSet[f] {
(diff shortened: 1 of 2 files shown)


language/go/generate_test.go
```diff
@@ -231,6 +231,30 @@ func TestGenerateRulesPrebuiltGoProtoRules(t *testing.T) {
 	}
 }
 
+func TestConsumedGenFiles(t *testing.T)  {
```
jayconrod: Please add a quick comment explaining what is being tested here.

Also, it looks like some spaces are missing in the `GenerateArgs` literal below. Please run gofmt.

language/go/generate.go
```diff
@@ -230,8 +230,15 @@ func (gl *goLang) GenerateRules(args language.GenerateArgs) language.GenerateRes
 		for _, f := range regularFiles {
 			regularFileSet[f] = true
 		}
+		// Some of the generated files may have been consumed by other rules
+		consumedFileSet := make(map[string]bool)
+		for _, r := range args.OtherGen {
+			for _, f := range r.AttrStrings("srcs") {
+				consumedFileSet[f] = true
+			}
+		}
```
jayconrod: Also:

```
if f := r.AttrString("src"); f != "" {
  consumedFileSet[f] = true
}
```

Issue #602, opened 2019-07-29
go_repository: statically resolve external imports


blico:
After 76bba8ba862f0a0c4ca9b92bc72057043eccec6e, if we pass `-repo_config=@//:WORKSPACE` to a `go_repository` rule then we can assume all of the main repository's known external dependencies are being passed to every `go_repository` rule. (as long as all dependencies are specified in WORKSPACE or with `# gazelle:repository_macro`.)

What this implies is that we should no longer need to determine repository roots from `go_repository` rules, because we pass all of them in already with `repo_config`. This will eliminate unnecessary calls to `go list` when an external dependency contains `importpath`s that are not used within the main repository (and therefore not needed).

We should probably make this feature optional, because `repo_config` is also optional and may not contain all of the main repositories's external dependencies.

jayconrod:
Sounds good to me. This would mean Gazelle wouldn't need to go out to the network in `go_repository`, which would make it much faster.

The new mode should probably be a new value in `language/go.dependencyMode` (`staticMode` maybe?). It would be like `externalMode` when modules are disabled, except if no known repositories match, and none of the special cases apply, we should just assume the domain name is the repo root.

I think this mode should actually be the default for `go_repository`. #132 should be done first though.

jayconrod:
Now that #647 is done and there's a clear path to #132, I think it's clear how to do this. If `-go_repository_mode` is set, then when we resolve a path externally, instead of going to `RemoteCache`, we should look for modules with paths that are prefixes in `config.Config.Repos`. If more than one module matches, the longest path should win.

I'd eventually like to refactor `RemoteCache` to bring all the Go-specific code into `language/go`, but that shouldn't interfere with this change, since we shouldn't call anything on `RemoteCache` in this mode.

PR #647, opened 2019-09-29
repo, language/go: refactor Go-specific functionality out of repo

jayconrod:
Added experimental interfaces RepoUpdater and RepoImporter to
language.Language. update-repos now iterates over the list of
languages, finding a Language that also implements one of these
interfaces. update-repos still contains language-agnostic rule
merging, pruning, and macro handling logic.

Most Go-specific code is moved from repo to language/go. There is
still a lot of Go-specific code in repo.RemoteCache. This will be
moved eventually. The plan is to add a Do method that takes a function
and stores an arbitrary key/value pair, much like
cmd/go/internal/par.Cache.

Updates #619


jayconrod:
cc @linzhp @blico Please let me know if this breaks anything for you. Ideally, I'd like to get all Go-specific functionality into `language/go`, and that may mean breaking some things in the `repo` package.

blico:
@jayconrod  I just verified that this doesn't break us. Maybe now that `repo` has been refactored, it would be a good time for me to implement #132 and #602.

jayconrod:
@blico 

> I just verified that this doesn't break us.

Great! Thanks for checking.

> Maybe now that repo has been refactored, it would be a good time for me to implement #132 and #602.

Sure, that'd be awesome if you have bandwidth. Nothing should be blocking #132. I'd like to do some refactoring in `RemoteCache`, but I don't think it will interfere with #602 (and it probably won't be done soon).

I'll add notes to both issues.

PR #648, opened 2019-09-29
language/go: simpler submodule visibility

### LINKED ISSUE #619: When an external library depends on an internal package of another dependency, Gazelle creates a rule that doesn't compile.

jayconrod:
This change supercedes #637, which was reverted.

In #647, Gazelle now reads the repository configuration file (set with
-repo_config; WORKSPACE by default) and adds repository rules to the
Repos list, which all languages may access.

When language/go is configured, it reads Repos and makes a list of
submodules by looking for go_repository rules with importpaths that
have the current module path as a prefix. This is simpler than the
previous approach, which required explicitly adding build_submodule
attributes to go_repository.

When Go libraries and binaries are generated in internal directories,
the submodule list is used to add additional visibility entries.

Fixes #619.

### DIFF
--- a/cmd/gazelle/integration_test.go
+++ b/cmd/gazelle/integration_test.go
@@ -2822,6 +2822,56 @@ go_library(
 	})
 }
 
-// TODO(jayconrod): more tests
-//   run in fix mode in testdata directories to create new files
-//   run in diff mode in testdata directories to update existing files (no change)
+// TestGoImportVisibility checks that submodules implicitly declared with
+// go_repository rules in the repo config file (WORKSPACE) have visibility
+// for rules generated in internal directories where appropriate.
+// Verifies #619.
+func TestGoImportVisibility(t *testing.T) {
+	files := []testtools.FileSpec{
+		{
+			Path: "WORKSPACE",
+			Content: `
+go_repository(
+		name = "com_example_m_logging",
+    importpath = "example.com/m/logging",		
+)
+`,
+		}, {
+			Path:    "internal/version/version.go",
+			Content: "package version",
+		}, {
+			Path:    "internal/version/version_test.go",
+			Content: "package version",
+		},
+	}
+	dir, cleanup := testtools.CreateFiles(t, files)
+	defer cleanup()
+
+	args := []string{"update", "-go_prefix", "example.com/m"}
+	if err := runGazelle(dir, args); err != nil {
+		t.Fatal(err)
+	}
+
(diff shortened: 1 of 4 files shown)


jayconrod:
@linzhp This is the replacement for #637. It makes the same visibility change in a simpler way. I don't think this should break anything: it's only adding new visibility labels, not restricting existing visibility.

If it does break something though, please file an issue with the details.

pcasaretto:
How can we test this? Not that familiar with bazel.
Can we use something like `git_repository` pointing to this commit?

jayconrod:
@pcasaretto Yes, exactly that. Replace the `http_archive` rule that declares `bazel_gazelle` with a `git_repository` rule pointing to `fe0ec4f9369d78ff80a1f3f519c63d715f6cbd75`.

PR #651, opened 2019-10-03
repo: repository directive

blico:
Implements #132 

A repository directive specifies a repository rule that Gazelle should know about. The directive can be repeated multiple times and can be declared from within a macro definition that Gazelle knows about. At the very least the  directive must define a rule kind and a name attribute, but it can define extra attributes after that


repo/repo.go
```diff
@@ -89,17 +99,54 @@ func ListRepositories(workspace *rule.File) (repos []*rule.Rule, repoFileMap map
...
 	}
 	return repos, repoFileMap, nil
 }
 
+func parseRepositoryDirectives(directives []rule.Directive) (repoMap map[string]*rule.Rule, err error) {
+	repoMap = make(map[string]*rule.Rule)
+	for _, d := range directives {
+		switch d.Key {
+		case "repository":
+			vals := strings.Split(d.Value, " ")
```
jayconrod: Use `strings.Fields` instead.

repo/repo.go
```diff
@@ -89,17 +99,54 @@ func ListRepositories(workspace *rule.File) (repos []*rule.Rule, repoFileMap map
...
+			if len(vals) < 2 {
+				return nil, fmt.Errorf("Failure parsing repository: %s, you must define at least a rule kind and name attribute", d.Value)
```
jayconrod: Nit: errors should start with lowercase letters (golint nags me about this) and should not address the user directly. 

Also, no need to say the name attribute is missing because we check for that below. So maybe if `len(vals) == 0`, then `parsing repository directive: expected repository kind and attributes`.

repo/repo.go
```diff
@@ -89,17 +99,54 @@ func ListRepositories(workspace *rule.File) (repos []*rule.Rule, repoFileMap map
...
+			}
+			kind := vals[0]
+			r := rule.NewRule(kind, "")
+			for _, val := range vals[1:] {
+				kv := strings.SplitN(val, "=", 2)
```
jayconrod: Let's just look for the first `=` character. The value could conceivable have its own `=` characters.

repo/repo.go
```diff
@@ -89,17 +99,54 @@ func ListRepositories(workspace *rule.File) (repos []*rule.Rule, repoFileMap map
...
+				if len(kv) != 2 {
+					return nil, fmt.Errorf("Failure parsing repository: %s, expected format for attributes is <attr1_name=attr1_value>", d.Value)
```
jayconrod: Nit: drop the `<` and `>` here. It would be easy to misinterpret that, thinking those characters are needed literally.

repo/repo.go
blico: I'm pretty sure I am already doing that with `SplitN`... N, in this case, is the number of substrings to split into

jayconrod: Oops, you're right.

Issue #654, opened 2019-10-06
go_repository rules are not using the correct build_file_name 


ClareCat:
I think I found a few issues with how build_file_name is determined and if code generation should happen while trying to get the build_directives options working for an external go repository that was being included via git commit. The repo I was adding as a dep has a BUILD file in its root dir so I was surprised that BUILD.bazel files were getting generated. I tried using the build_file_name option, but still only got BUILD.bazel.

After some digging I think the code causing the issues is here: https://github.com/bazelbuild/bazel-gazelle/blob/295335e68d09596b430b9ad48e162a57f3bb2160/internal/go_repository.bzl#L78-L85

It seems that the gazelle go_repository rule is checking if build files exist in the repo's root dir before the repository is fetched which means it always returns False because the directory is empty. Thus, the existing_build_file variable is never set so gazelle's build file generation will always happen unless it is explicitly set to off. The build_file_name that is used in generation is dependent on the existing_build_file variable. Since that doesn't get set the build_file_name is always set to BUILD.bazel and ignores the user's input. Also, the loop that checks for build file existence always checks/prefers ['BUILD', 'BUILD.bazel'] before any user input. If there is no user input the list of names it checks is ['BUILD', 'BUILD.bazel', 'BUILD.bazel', 'BUILD']. Is one of those the right behaviour? From the docs it seems like the list should only be the user input.

### What version of gazelle are you using?

This issue is present on current master (295335e68d09596b430b9ad48e162a57f3bb2160)

### What version of rules_go are you using?

Tried on 0.19.4 and 0.19.5

### What version of Bazel are you using?

0.29.1

### Does this issue reproduce with the latest releases of all the above?

Yes

### What operating system and processor architecture are you using?

Macos darwin amd64

PR #655, opened 2019-10-06
go_repository should correctly identify build files and respect build_file_name

ClareCat:
Fixes for: https://github.com/bazelbuild/bazel-gazelle/issues/654

### DIFF
--- a/internal/go_repository.bzl
+++ b/internal/go_repository.bzl
@@ -74,49 +74,7 @@ def _go_repository_impl(ctx):
     else:
         fail("one of urls, commit, tag, or importpath must be specified")
 
-    # Generate build files if needed.
-    existing_build_file = ""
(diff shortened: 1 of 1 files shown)


internal/go_repository.bzl
```diff
@@ -260,6 +226,42 @@ go_repository = repository_rule(
 )
 """See repository.rst#go-repository for full documentation."""
 
+def get_env(ctx):
```
jayconrod: Let's keep this code inlined in `_go_repository_impl`. I don't think it needs to be a separate function. It's fine to build the environment unconditionally.

ClareCat:
@googlebot I signed it!

internal/go_repository.bzl
ClareCat: Done!

PR #666, opened 2019-10-27
Propagate standard SSL certificate environment variables

andrew-d:
This passes through the environment variables `SSL_CERT_FILE` and `SSL_CERT_DIR`, which the Go standard library [will use](https://golang.org/pkg/crypto/x509/) to override the system default locations for SSL root certificates.

This is helpful in cases where either (a) the files aren't at the expected location, or (b) you want to fetch from a server that's using TLS with a non-public root certificate.

### DIFF
--- a/internal/go_repository.bzl
+++ b/internal/go_repository.bzl
@@ -95,6 +95,8 @@ def _go_repository_impl(ctx):
 
         # Settings below are used by vcs tools.
         "SSH_AUTH_SOCK",
+        "SSL_CERT_FILE",
+        "SSL_CERT_DIR",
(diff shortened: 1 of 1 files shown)


andrew-d:
@jayconrod - Thanks!

andrew-d:
@jayconrod - Apologies for the bother, but would you mind creating a release of Gazelle? I'd like to submit a patch to bump the version of Gazelle used in https://github.com/google/gvisor (currently on v0.19.0; a v0.19.1 would be lovely, if you don't mind).

jayconrod:
@andrew-d Sure, it's out now.

PR #674, opened 2019-11-21
Prevent Gazelle from generating duplicate repository rules

linzhp:
When a `go.mod` file has:

```
require (
	github.com/Selvatico/go-mocket v1.0.7
	github.com/selvatico/go-mocket v1.0.7
)
```

`gazelle update-repos` would generate two `go_repository` rules with the same name, causing indeterminism in the order of resulting `go_repository` rules. In most cases, users should not import both `github.com/Selvatico/go-mocket` and `github.com/selvatico/go-mocket` in the same module, so `gazelle update-repos` should report errors. However, if that's what users really want, they should to add a `replace` statement in the `go.mod` file like this:

```
require (
	github.com/Selvatico/go-mocket v1.0.7
	github.com/selvatico/go-mocket v0.0.0-00010101000000-000000000000
)

replace github.com/selvatico/go-mocket => github.com/Selvatico/go-mocket v1.0.7
```

In this case, only one `go_repository` rule is generated.

### DIFF
--- a/cmd/gazelle/integration_test.go
+++ b/cmd/gazelle/integration_test.go
@@ -22,6 +22,7 @@ package main
 import (
 	"bytes"
 	"flag"
+	"fmt"
 	"io/ioutil"
 	"log"
 	"os"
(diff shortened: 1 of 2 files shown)


cmd/gazelle/integration_test.go
```diff
@@ -2875,3 +2876,65 @@ go_test(
...
+		},
+	}
+	dir, cleanup := testtools.CreateFiles(t, files)
+	defer cleanup()
+
+	args := []string{"update-repos", "--from_file=go.mod"}
+	if err := runGazelle(dir, args); err != nil {
+		t.Fatal(err)
+	}
+}
```
jayconrod: Let's check that a rule for `com_github_selvatico_go_mocket` gets generated with version `v1.0.7`.

cmd/gazelle/update-repos.go
```diff
@@ -175,7 +175,16 @@ func updateRepos(args []string) (err error) {
 	var newGen []*rule.Rule
 	genForFiles := make(map[*rule.File][]*rule.Rule)
 	emptyForFiles := make(map[*rule.File][]*rule.Rule)
+	ruleSet := make(map[string]*rule.Rule)
```
jayconrod: Nit: rename to `genNames`

cmd/gazelle/update-repos.go
```diff
@@ -175,7 +175,16 @@ func updateRepos(args []string) (err error) {
...
 	for _, r := range gen {
+		if existingRule, exist := ruleSet[r.Name()]; exist {
```
jayconrod: Nit: rename `exist` to `ok`. It's a bit more idiomatic. Alternatively, just check whether `existingRule` is nil.

PR #686, opened 2019-12-23
Replace 'cat' tool call with repository_ctx.read() call.

tomaszstrejczek:
Newer versions of bazel provide repository_ctx.read() (https://docs.bazel.build/versions/master/skylark/lib/repository_ctx.html#read). 
On the other hand there is no standard 'cat' utility on Windows.

### DIFF
--- a/internal/go_repository_cache.bzl
+++ b/internal/go_repository_cache.bzl
@@ -76,11 +76,8 @@ go_repository_cache = repository_rule(
 )
 
 def read_cache_env(ctx, path):
-    result = ctx.execute(["cat", path])
-    if result.return_code:
-        fail("failed to read cache environment: " + result.stderr)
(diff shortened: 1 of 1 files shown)


jayconrod:
Seems reasonable. I tried to make a change like this when `ctx.read` was introduced, but at the time, it reported an error when reading files in different repositories. I can't reproduce that anymore though.

Please update `go_repository_config.bzl`, which uses `cat`, too.

Also, consider changing calls to `read_cache_env` to pass a `path` or `Label` instead of a string.

Note that msys2 (which includes `cat`) is a [dependency](https://github.com/bazelbuild/rules_go/blob/master/windows.rst) of rules_go. It's needed for the shell, the C compiler (MSVC is not supported by cgo), and a few utilities like `patch`.

jayconrod:
Superceded by #722

PR #693, opened 2020-01-13
Sorting repo rules in macros by name

linzhp:
Developers working on the same repo may add external repos simultaneously. Appending new repos to the end often results in merge conflict. This PR sort repo rules by name to reduce the chance of conflict.

Note the it only sorts rules in macros by not in WORKSPACE, as the latter may have some repo rules that are better to keep at the top because of their importance, e.g., loading rules_go, gazelle, setting up go toolchain.

cc @blico

### DIFF
--- a/rule/rule.go
+++ b/rule/rule.go
@@ -369,14 +369,15 @@ func (f *File) Format() []byte {
 	return bzl.Format(f.File)
 }
 
-// Sort rules in the macro of this File. It doesn't sort the rules if
+// SortMacros sorts rules in the macro of this File. It doesn't sort the rules if
 // this File does not have a macro, e.g., WORKSPACE.
 // This method calls Sync internally.
-func (f *File) Sort() {
+func (f *File) SortMacro() {
 	f.Sync()
-	//TODO: sort load statements too
 	if f.function != nil {
 		sort.Stable(byName{f.Rules, f.function.stmt.Body})
+	} else {
+		panic(fmt.Sprintf("no macros found in %s", f.Path))
 	}
 }
 
--- a/rule/rule.go
+++ b/rule/rule.go
@@ -369,7 +369,7 @@ func (f *File) Format() []byte {
 	return bzl.Format(f.File)
 }
 
-// SortMacros sorts rules in the macro of this File. It doesn't sort the rules if
+// SortMacro sorts rules in the macro of this File. It doesn't sort the rules if
 // this File does not have a macro, e.g., WORKSPACE.
 // This method calls Sync internally.
 func (f *File) SortMacro() {
@@ -377,7 +377,7 @@ func (f *File) SortMacro() {
 	if f.function != nil {
 		sort.Stable(byName{f.Rules, f.function.stmt.Body})
 	} else {
-		panic(fmt.Sprintf("no macros found in %s", f.Path))
+		panic(fmt.Sprintf("%s: not loaded as macro file", f.Path))
 	}
 }
 
--- a/rule/rule.go
+++ b/rule/rule.go
@@ -781,8 +781,7 @@ func (r *Rule) Args() []bzl.Expr {
 }
 
 // Insert marks this statement for insertion at the end of the file. Multiple
-// statements will be inserted in the order Insert is called. Inserted rules
-// will be sorted together with other rules at File.Save()
+// statements will be inserted in the order Insert is called.
 func (r *Rule) Insert(f *File) {
 	var stmt []bzl.Expr
 	if f.function == nil {
--- a/rule/rule.go
+++ b/rule/rule.go
@@ -369,8 +369,11 @@ func (f *File) Format() []byte {
 	return bzl.Format(f.File)
 }
 
-// Sort rules in this File
+// Sort rules in the macro of this File. It doesn't sort the rules if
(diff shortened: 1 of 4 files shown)


rule/rule.go
```diff
@@ -750,15 +750,20 @@ func (r *Rule) Args() []bzl.Expr {
...
-	// TODO(jayconrod): should rules always be inserted at the end? Should there
-	// be some sort order?
 	var stmt []bzl.Expr
 	if f.function == nil {
 		stmt = f.File.Stmt
+		r.index = len(stmt)
 	} else {
 		stmt = f.function.stmt.Body
+		var i int
+		for i = range f.Rules {
```
jayconrod: `sort.Search` would be better than looping over the list.

This assumes the rules are sorted already though, right? I don't think they necessarily are.

Also what if the inserted rules aren't sorted? They're not required to be, in general.

linzhp:
Updated the PR according to https://github.com/bazelbuild/bazel-gazelle/issues/696#issuecomment-574290212, except:

1. `rule.File.Sort` does not call `Sync()`, but instead to have `Save()` to call `Sync()` before calling `Sort()`. I can also call `Sync()` in `Sort()` as suggested, and remove the `Sync()` call from `Save()`. Let me know which you like
2. This PR does not sort load statements. I can do it in a separate PR.

rule/rule.go
```diff
@@ -369,9 +369,18 @@ func (f *File) Format() []byte {
...
+	//TODO: sort load statements too
+	if f.function != nil {
+		sort.Stable(byName{f.Rules, f.function.stmt.Body})
+	}
+}
+
 // Save writes the build file to disk. This method calls Sync internally.
 func (f *File) Save(path string) error {
 	f.Sync()
+	f.Sort()
```
jayconrod, line 383: If `f.Sort()` is called unconditionally here, won't that sort `WORKSPACE` and `BUILD.bazel` files? I think only dependency macros should be sorted.

linzhp: `f.Sort()` is a no-op if `f.function == nil`

linzhp:
@jayconrod Any more feedback on this?

rule/rule.go
jayconrod: Ah, sorry, I missed that.

Based on the method signatures and documentation, this behavior seems surprising to me. I think that:

* `Sort` should work unconditionally.
* The caller in `update-repos.go` should be responsible for calling `Sort`. This package doesn't know whether it's appropriate to call `Sort`; it may be controlled by directives or something else later on.
* `Sort` should internally call `Sync` (and the doc should mention that). It's okay if `Sync` is called again later in `Save`.

linzhp:
Addressed comments. Please take another look

rule/rule.go
```diff
@@ -369,6 +369,17 @@ func (f *File) Format() []byte {
 	return bzl.Format(f.File)
 }
 
+// Sort rules in the macro of this File. It doesn't sort the rules if
+// this File does not have a macro, e.g., WORKSPACE.
+// This method calls Sync internally.
+func (f *File) Sort() {
+	f.Sync()
+	//TODO: sort load statements too
+	if f.function != nil {
```
jayconrod, line 378: I still think this condition should be removed. `update-repos.go` can check `f.DefName != ""` before deciding whether to call `Sort`.

linzhp: This method needs to use `f.function.stmt.Body`. If this condition is removed, what should it do if it is called with `function == nil`? Or should we assume the caller knows the precondition of this method?

jayconrod: I was going to suggest it should be able to sort targets in a regular build files... but that's probably more complicated than necessary, since it would need to handle load statements and other expressions.

I think my objection is that if I'm reading code that calls `Sort`, I'd really expect it to be able to sort targets in any file. The name gives no hint it's limited to macros.

Perhaps instead, this should be renamed to `SortMacro` to make the limitation more obvious, and it should panic if it's called on something that's not a macro file.

linzhp: yeah, let's limit the scope of this PR to macros. It's less clear whether we should sort other files

rule/rule.go
```diff
@@ -369,14 +369,15 @@ func (f *File) Format() []byte {
 	return bzl.Format(f.File)
 }
 
-// Sort rules in the macro of this File. It doesn't sort the rules if
+// SortMacros sorts rules in the macro of this File. It doesn't sort the rules if
```
jayconrod, line 372: Nit: `SortMacro`.

rule/rule.go
```diff
@@ -369,14 +369,15 @@ func (f *File) Format() []byte {
...
 // this File does not have a macro, e.g., WORKSPACE.
 // This method calls Sync internally.
-func (f *File) Sort() {
+func (f *File) SortMacro() {
 	f.Sync()
-	//TODO: sort load statements too
 	if f.function != nil {
 		sort.Stable(byName{f.Rules, f.function.stmt.Body})
+	} else {
+		panic(fmt.Sprintf("no macros found in %s", f.Path))
```
jayconrod, line 380: This message doesn't seem quite right: maybe instead `"%s: not loaded as macro file"`

Agreed that panicking is the right thing to do here.

PR #697, opened 2020-01-14
Ignore Go files consumed by existing rules

linzhp:
Related to #643

Users may want some generated file to be build in a different `go_library` rule than the `go_default_library`. So they may create a `go_library` rule manually with the generated Go files. In this case, we should not add those Go files into `go_default_library`


jayconrod:
Could you open an issue to discuss this? (In general, please open issues for anything more than narrow bug fixes).

This seems like it would cause problems for people who want a source file to appear in the `srcs` list for multiple rules. Gazelle itself has an extension (`//internal/language/test_filegroup`) that generates `filegroups`. I think this PR would break that extension.

PR #703, opened 2020-01-22
language/proto: Fixed service definition regexp

### LINKED ISSUE #702: gazelle add grpc compilers attributes on non grpc protos
0.19.1

0.19.9

1.2.1

Yes.

Linux asmirnov 5.0.0-25-generic #26~18.04.1-Ubuntu SMP Thu Aug 1 13:51:02 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux

### What did you do?

1. Create simple .proto:

```protobuf
syntax = proto3;

package bug;

message Foo {
  string service = 1;
}
```

2. Run bazel run //:gazelle

The `compilers` attribute should not be present in `go_proto_library` if the source .proto does not contain `service` definitions.

### What did you see instead?
BUILD.bazel file contains attribute `compilers = ["@io_bazel_rules_go//proto:go_grpc"]` in go_proto_library definition.

asv:
Fixes #702

### DIFF
--- a/cmd/gazelle/integration_test.go
+++ b/cmd/gazelle/integration_test.go
@@ -745,7 +745,7 @@ go_library(
 
 option go_package = "example.com/repo";
 
-service {}
+service TestService {}
 `,
 		},
 	}
@@ -2481,7 +2481,7 @@ syntax = "proto3";
 
 option go_package = "example.com/repo/service";
 
-service {}
+service TestService {}
 `,
 		},
 	}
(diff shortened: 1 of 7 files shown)


language/go/generate_test.go
```diff
@@ -90,6 +90,7 @@ func TestGenerateRules(t *testing.T) {
 			f.Sync()
 			got := string(bzl.Format(f.File))
 			wantPath := filepath.Join(dir, "BUILD.want")
+			t.Logf("Using %s file for test", wantPath)
```
jayconrod: Let's drop this log statement.

asv: OK, but without Log, it is not always obvious on which testdata the problem occurred...

jayconrod: Each subtest should have the testdata subdirectory as part of its name. Or am I missing something?

PR #720, opened 2020-02-28
proto_library for generated proto files

linzhp:
When a package only has generated proto files but no regular proto files, Gazelle should still generate `proto_library` for the generated proto files.


linzhp:
Thanks for the feedback. Added a test case

PR #722, opened 2020-03-03
Use repository_ctx.read() instead of "cat".

tomlu:
"cat" doesn't exist on Windows, and it seems
cleaner to use a built-in anyway.

Tested: Manually on Windows, and bazel test //...

The latter fails some tests, presumably because of
other Windows incompatibilities like autogazelle.


tomlu:
Sorry for the delay. Updated go_repository_config.bzl as requested.

PR #730, opened 2020-03-15
Allow MatchAttrs to be slices

### LINKED ISSUE #725: Matching proto_library with its srcs
`proto_library` targets are [resolved](https://github.com/bazelbuild/bazel-gazelle/blob/master/language/proto/resolve.go#L35) using its `srcs`. So in that sensee, `srcs` are like `importpath` of `go_library`. If there is an existing `proto_library` rule in the same BUILD.bazel file that has the same `srcs` as the one just generated, Gazelle should update the existing rule instead of adding a new `proto_library`.

92bc7b03d768a3581bd9bdd30d5e3635bbf854da

c5a0b28c495edc83ec80477e1c8addf631f156a6

2.0

It should

macOS, amd64

### What did you do?

* rename an existing `proto_library` previously generated by Gazelle
* run Gazelle

Gazelle should not generate a new `proto_library`

### What did you see instead?

Gazelle generates a new `proto_library`

linzhp:
Fixed #725


language/proto/kinds.go
```diff
@@ -19,6 +19,7 @@ import "github.com/bazelbuild/bazel-gazelle/rule"
 
 var protoKinds = map[string]rule.KindInfo{
 	"proto_library": {
+		MatchAttrs:[]string{"srcs"},
```
jayconrod: Looks like some space is missing here. Please run `gofmt`.

rule/rule.go
```diff
@@ -816,6 +816,27 @@ func (r *Rule) sync() {
 	r.updated = false
 }
 
+// AttrEqual decides if the rule and the other rule has the same value on the attribute
+func (r *Rule) AttrEqual(other *Rule, key string) bool {
```
jayconrod: Let's move this into `merger` as `func attrsMatch(x, y *Rule, key string) bool`. Everything in `rule` needs to be pretty general. This is pretty targeted toward matching and merging, so I think it should be in `merger`.

I'm not sure if it makes sense to require that the two lists are equal. If a source file is added or deleted, the rules will stop matching. Perhaps it would be better to check whether at least one element matches? If you want to leave it as-is for now though, we can always loosen it later.

rule/rule.go
```diff
@@ -816,6 +816,27 @@ func (r *Rule) sync() {
...
+	value := r.AttrString(key)
+	if len(value) > 0 && value == other.AttrString(key) {
```
jayconrod: Nit: `value != ""` instead of `len(value) > 0`

rule/rule.go
```diff
@@ -816,6 +816,27 @@ func (r *Rule) sync() {
...
+		return true
+	}
+	myValues := r.AttrStrings(key)
```
jayconrod: Rename `myValues` to `rValues` (or `xValues` and `yValues` when this is moved to `merger`).

rule/rule.go
```diff
@@ -816,6 +816,27 @@ func (r *Rule) sync() {
...
+	otherValues := other.AttrStrings(key)
+	if myValues != nil && otherValues != nil && len(myValues) == len(otherValues){
```
jayconrod: Small simplification:

```
if myValues == nil || otherValues == nil || len(myValues) != len(otherValues) {
  return false
}
sort.Strings(myValues)
sort.Strings(otherValues)
for i, v := range myValues {
  if v != otherValues[i] {
    return false
  }
}
return true
```
```

File: rule/rule_test.go [hunk 6]
```diff
@@ -346,3 +346,59 @@ func TestShouldKeepExpr(t *testing.T) {
...
+			attr: "importpath",
+			value1: "example.com/foo",
+			value2: "example.com/foo",
+			want: true,
+		},
+		{
+			desc: "different strings",
+			attr:"importpath",
+			value1:"example.com/foo",
+			value2:"example.com/bar",
```
jayconrod: Please run `gofmt` on this file, too.

rule/rule.go
linzhp: To handle added and delete source files, we can check if one list is a subset of the other, but I am not sure how to handle renames

language/proto/kinds.go
```diff
@@ -19,6 +19,7 @@ import "github.com/bazelbuild/bazel-gazelle/rule"
 
 var protoKinds = map[string]rule.KindInfo{
 	"proto_library": {
+		MatchAttrs:    []string{"srcs"},
```
jayconrod: Could you add a test that verifies this in `cmd/gazelle/integration_test.go`?

merger/merger.go
```diff
@@ -248,3 +245,23 @@ func Match(rules []*rule.Rule, x *rule.Rule, info rule.KindInfo) (*rule.Rule, er
 
 	return nil, nil
 }
+
+func AttrMatch(x, y *rule.Rule, key string) bool {
```
jayconrod: Let's leave this un-exported for now. Seems likely to change.

linzhp: Current merger_test.go is an external test. I have to create another internal test file to contain `TestAttrMatch`. How do I update `internal/go_repository_tools_srcs.bzl` after creating a new file?

jayconrod: Actually, could you add cases to `TestMatch` instead of creating `TestAttrMatch`? Better to test against the exported API.

If you need to add a file in the future, you can `go run internal/list_repository_tools_srcs.go -generate internal/go_repository_tools_srcs.bzl`.

PR #736, opened 2020-03-21
apply patches before build file generation

### LINKED ISSUE #735: go_repository should apply patches before executing gazelle
Problem appears if patch introduces new files and in that case new these files wouldn't be enlisted in generated build file.

lazdmx:
This commit forces patches be applied to just fetched repository before the
build file generation. This allows new files added by the patches be correctly
enlisted into build files.

Fixes #735

### DIFF
--- a/internal/go_repository.bzl
+++ b/internal/go_repository.bzl
@@ -127,6 +127,9 @@ def _go_repository_impl(ctx):
         if result.stderr:
             print("fetch_repo: " + result.stderr)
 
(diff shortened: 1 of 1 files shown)


lazdmx:
@googlebot  I signed it!

lazdmx:
Hmm. Obviously there is a problem if there is a need to patch a build file...

robfig:
This is an extremely-breaking change, and we can't merge it.

PR #738, opened 2020-03-21
Fix invalid output for globs with `exclude` arguments.

danielheller:
Currently, output is a KV pair as appropriate for a starlark dict,
where it should be a function argument. Parameter name is `exclude`.

Before:
    glob(
        ["*.star"],
        "excludes": ["*.bzl"],
    )

After:
    glob(
        ["*.star"],
        exclude = ["*.bzl"],
    )

### DIFF
--- a/rule/rule_test.go
+++ b/rule/rule_test.go
@@ -346,3 +346,16 @@ func TestShouldKeepExpr(t *testing.T) {
 		})
 	}
 }
+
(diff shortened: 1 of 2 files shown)


jayconrod:
Happy to take a look at this, but please sign the CLA first.

jayconrod:
Closing, sorry I can't review this without a CLA.

Issue #631, opened 2019-08-28
Stable resolution of importpath to repo name mappings


asuffield:
We've been chasing this cow around the field for a while with #529 and #132 and #602, and it just fell on me today in a way that more precisely illustrates the problem we are trying to solve. I believe I now have a workable design for solving the remaining cases.

I have a repo which provides a go library. This library depends on the kubernetes client libraries, which use a vanity domain: k8s.io/client-go is https://github.com/kubernetes/client-go. Types from these libraries form part of my interface, so users of my library will end up importing them for temporary variables. I'm also running without direct internet access so the network lookups to k8s.io fail, and gazelle's static heuristics interpret "import k8s.io/client-go/plugin/pkg/client/auth" as @io_k8s_client_go_plugin//pkg/client/auth, which is broken. In order for this to generate correct rules, gazelle needs to see an explicit mapping in the form of a go_repository rule. (The networking tangle is its own mess, but for the purposes of this discussion, let's just accept it as axiomatic that sometimes we're going to need gazelle to see the contents of go_repository rules in order to get name resolution right)

Users of this repo will call a foo_deps() function that I provide to get the dependencies they need, which includes go_repository rules for all the many kubernetes client libraries. If those users then import the library I provide and run gazelle themselves, it will generate incorrect dependencies on the k8s.io packages, because it can't see those go_repository rules.

There is currently nothing those users can do to get gazelle to read the go_repository rules, because they are in a foreign repo. Copying every such foo_deps function into a local .bzl file for gazelle:repository-macro file is combinatorially-hard and an intolerable layering violation.

(Description of the problem ends here. If this is too abstract to understand, I can attempt to create an instance of this problem on github for inspection)

So, what can we do about this? The only plausible way to extract all the go_repository rules in the repo rules is to run bazel query (which has its own bug: https://github.com/bazelbuild/bazel/issues/7855) or bazel sync --experimental_repository_resolved_file. We cannot reasonably do either of these from within any repo rule or build action. There is only one time that we could plausibly collect this data: during a "bazel run :gazelle" invocation.

This leads us to the first important idea of this design:

> The gazelle() rule should extract the complete set of repo name mappings that it discovers and write them to a file in the repo, so that it can be used later by other rules.

[...]

PR #748, opened 2020-04-04
[Feature] allow generated repo_configs in the `gazelle` rule

grahamjenson:
TLDR: I want to be able to pass the `gazelle` rule a generated repo_config so that I don't have to manually update dependencies from `go.mod`

Using the `update-repos` workflow to update repositories from a go.mod is pretty painful. To simplify this I want to be able to generate a repository config file automatically and have gazelle use it, without having to run `update-repos`. To do this I wrote a repository rule:

```
def _go_mod_impl(ctx):
    ctx.file("BUILD", content = 'exports_files(["repositories.bzl"])')
    ctx.file("WORKSPACE", content = "workspace(name = \"%s\")" % ctx.name)

    bzl = 'load("@bazel_gazelle//:deps.bzl", "go_repository")\n\ndef go_repositories():\n    pass\n\n# gazelle:repository_macro repositories.bzl%go_repositories'

    ctx.file("repositories.bzl", content = bzl)

    _gazelle = "@bazel_gazelle_go_repository_tools//:bin/gazelle"
    gazelle = ctx.path(Label(_gazelle))

    result = ctx.execute(
        [
            gazelle,
            "update-repos",
            "-from_file=%s" % ctx.path(ctx.attr.go_mod),
            "-to_macro=repositories.bzl%go_repositories",
        ],
    )

    if result.return_code:
        fail("failed go_mod %s: %s" % (ctx.name, result.stderr))
    if result.stderr:
        print("go_mod: " + result.stderr)


go_mod_download = repository_rule(
    implementation = _go_mod_impl,
    attrs = {
        "go_mod": attr.label(
            mandatory = True,
            allow_single_file = True,
        ),
    },
)
```

That is called like:

```
go_mod_download(
    name = "go_modules",
    go_mod = "//:go.mod",
)
load("@go_modules//:repositories.bzl", default_go_repositories = "go_repositories")
default_go_repositories()
```

This works great. Except because of the way gazelle optimizes resolution of repositories, it has made running gazelle **very slow**.

To solve this I want to be able to pass my generated `@go_modules//:repositories.bzl` as a repo config to gazelle, to tell it about all the repositories I already have defined. For example:

```
gazelle(
    name = "gazelle",
    repo_config = "@go_modules//:repositories.bzl",
)
```

This PR lets gazelle work with a generated repo_config. This is also very extensible, because now I can concat a bunch of files that have `go_repositories` in them, into a single repository config and have gazelle know what is what. 

With this approach, I can run `go get -u <package>` to update a package then run `bazel build <rule>` and it will have the new package immediately without having to ru `update-repos` AND if I run `bazel run gazelle` it doesn't take minutes.

### Other approach

I tried using something like  `# gazelle:repository_macro bazel-repo/external/repositories.bzl%go_repositories` this doesn't work because when running `bazel run gazelle` the file might not exist yet. So gazelle rule must know about this file before hand anyway.

### DIFF
--- a/def.bzl
+++ b/def.bzl
@@ -47,6 +47,10 @@ DEFAULT_LANGUAGES = [
 
 def _gazelle_runner_impl(ctx):
     go = _go_context(ctx)
+    runfiles = [
+        ctx.executable.gazelle,
+        go.go,
+    ]
     args = [
         ctx.attr.command,
         "-mode",
@@ -58,11 +62,19 @@ def _gazelle_runner_impl(ctx):
         args.extend(["-go_prefix", ctx.attr.prefix])
     if ctx.attr.build_tags:
         args.extend(["-build_tags", ",".join(ctx.attr.build_tags)])
+
+    # file_args absolute paths are found in the script
+    file_args = []
+    if ctx.attr.repo_config:
+        file_args.extend(["-repo_config", ctx.file.repo_config.path])
+        runfiles.append(ctx.file.repo_config)
+
     args.extend(ctx.attr.extra_args)
 
     out_file = ctx.actions.declare_file(ctx.label.name + ".bash")
     substitutions = {
         "@@ARGS@@": shell.array_literal(args),
+        "@@FILE_ARGS@@": shell.array_literal(file_args),
         "@@GAZELLE_LABEL@@": shell.quote(str(ctx.attr.gazelle.label)),
         "@@GAZELLE_SHORT_PATH@@": shell.quote(ctx.executable.gazelle.short_path),
         "@@GENERATED_MESSAGE@@": """
@@ -78,13 +90,11 @@ def _gazelle_runner_impl(ctx):
         substitutions = substitutions,
         is_executable = True,
     )
-    runfiles = ctx.runfiles(files = [
-        ctx.executable.gazelle,
-        go.go,
-    ])
+
+
     return [DefaultInfo(
         files = depset([out_file]),
-        runfiles = runfiles,
+        runfiles = ctx.runfiles(files = runfiles),
         executable = out_file,
     )]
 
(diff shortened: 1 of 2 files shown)


jayconrod:
Please open an issue to discuss this. I'd prefer not to review code until we've agreed on a design.

You may want to read through #631 first. Gazelle won't execute Starlark code, and it won't invoke Bazel, which means that Gazelle can't directly access configuration files in other workspaces. It's okay for rules and repository rules that invoke Gazelle to depend on files (possibly generated files) in other workspaces though and pass those in.

Also, take a look at the `gazelle_dependencies` function. It accepts a `config` parameter that defaults to `@//:WORKSPACE`, but could be set to a generated file in another repository. I think that was added by the Uber folks, who were doing something similar. If that works for you, then maybe it just need to be better documented.

grahamjenson:
Hey, 
I tried running my above solution with `go_repository_default_config = "@go_modules//:repositories.bzl"` and `gazelle_dependencies(go_repository_default_config = "@go_modules//:WORKSPACE")` with a variety of directives. It looks like that argument `go_repository_default_config` can take a file where the directive `# gazelle:repository_macro ` is a label which is super confusing, because I am unable to use a label and point to the function `go_repositories` which is needed. 

The core issue with this approach is that running `bazel run gazelle` is very slow compared because it is not be able create a list of known imports as the repository config at https://github.com/bazelbuild/bazel-gazelle/blob/960ddc5b3e5c4a7da610b27b7d122f40ec56a188/cmd/gazelle/fix-update.go#L131 always resolves to <my_repo>/WORKSPACE and not the workspace set in the `gazelle_dependencies` function. I am not sure if there is some trick to fi this, maybe make `@go_modeules` look like a pseudo go_repository or something?

I have created https://github.com/bazelbuild/bazel-gazelle/issues/752 to have further discussion.

robfig:
Cleaning up old PRs.. this one seems to have stalled, and the final design seems likely to be not related to this, so I'll close it for now. Thanks

PR #749, opened 2020-04-04
walk: convert gazelle:exclude to a doublestar.Match pattern

### LINKED ISSUE #750: walk: exclude directive should support globs
It should be possible to exclude all files matching a glob pattern, as defined by [`path.Match`](https://pkg.go.dev/path?tab=doc#Match).

This would make it possible to exclude files matching a given extension. The syntax should be familiar and intuitive.

```
# gazelle:exclude *.pb.go
# gazelle:exclude *_windows[._]*
```

kalbasit:
To make editors happy we generate the equivalent `.pb.go` for all `go_proto_library` targets, the `.gen.go` for all `go_embed_data` and then create a symbolic link to those files from Bazel's sandbox back to the source code. I'm pretty that's not the best approach, but it has worked for months. However, there's an issue with Gazelle where it tries to add these files to the target. It happens randomly as most of the time it ignores them, but I've noticed it happens mostly if the symlink got broken (bazel clean or other reason).

This pull request allows us to ignore files based on [path.Match](https://pkg.go.dev/path?tab=doc#Match) pattern relative location in the repository.

closes #750

### DIFF
--- a/walk/config.go
+++ b/walk/config.go
@@ -51,8 +51,11 @@ func (wc *walkConfig) isExcluded(rel, base string) bool {
 	for _, x := range wc.excludes {
 		matched, err := doublestar.Match(x, f)
 		if err != nil {
-			log.Printf("error running doublestar.PathMatch(%q, %q): %s", x, f, err)
-			continue
+			// doublestar.Match returns only one possible error, and only if the
+			// pattern is not valid. During the configuration of the walker (see
+			// Configure below), we discard any invalid pattern and thus an error
+			// here should not be possible.
+			log.Panicf("error during doublestar.Match. This should not happen, please file an issue https://github.com/bazelbuild/bazel-gazelle/issues/new: %s", err)
 		}
 		if matched {
 			return true
@@ -75,7 +78,7 @@ func (_ *Configurer) KnownDirectives() []string {
 	return []string{"exclude", "follow", "ignore"}
 }
 
-func (_ *Configurer) Configure(c *config.Config, rel string, f *rule.File) {
+func (cr *Configurer) Configure(c *config.Config, rel string, f *rule.File) {
 	wc := getWalkConfig(c)
 	wcCopy := &walkConfig{}
 	*wcCopy = *wc
@@ -85,6 +88,10 @@ func (_ *Configurer) Configure(c *config.Config, rel string, f *rule.File) {
 		for _, d := range f.Directives {
 			switch d.Key {
 			case "exclude":
+				if err := checkPathMatchPattern(path.Join(rel, d.Value)); err != nil {
+					log.Printf("the exclusion pattern is not valid %q: %s", path.Join(rel, d.Value), err)
+					continue
+				}
 				wcCopy.excludes = append(wcCopy.excludes, path.Join(rel, d.Value))
 			case "follow":
 				wcCopy.follow = append(wcCopy.follow, path.Join(rel, d.Value))
@@ -95,4 +102,9 @@ func (_ *Configurer) Configure(c *config.Config, rel string, f *rule.File) {
 	}
 
 	c.Exts[walkName] = wcCopy
+}
+
+func checkPathMatchPattern(pattern string) error {
+	_, err := doublestar.Match(pattern, "x")
+	return err
 }
--- a/walk/config.go
+++ b/walk/config.go
@@ -69,7 +69,7 @@ type Configurer struct{}
 func (_ *Configurer) RegisterFlags(fs *flag.FlagSet, cmd string, c *config.Config) {
 	wc := &walkConfig{}
 	c.Exts[walkName] = wc
-	fs.Var(&gzflag.MultiFlag{Values: &wc.excludes}, "exclude", "path.Match pattern that should be ignored (may be repeated)")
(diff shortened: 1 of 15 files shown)


jayconrod:
I've been thinking about this, too, but I'd rather Gazelle excluded files based on glob patterns than just extensions. Glob patterns are more powerful, and [`path.Match`](https://pkg.go.dev/path?tab=doc#Match) already handles them natively.

I've opened #750 to track this.

Please also add a test and update documentation for `# gazelle:exclude`.

kalbasit:
Oh, interesting. Yes, I like your approach better. I'll get back to you today with the updated pull request. Thanks @jayconrod!

kalbasit:
@jayconrod `path.Match` does not support `**` globbing.

For instance, Consider the glob pattern `**/*.pb.go` it matches `a/b.pb.go` but does not match `a.pb.go` nor `a/b/c.pb.go`
Simlarly the pattern `*.pb.go` matches only `a.pb.go` but does not matches anything sub-directory.

There's an open issue about this: https://github.com/golang/go/issues/11862

At this point, I'm going to look for a third-party library such as https://github.com/bmatcuk/doublestar

P.S: Are you active on Slack's BazelBuild? I PM'd you there.

kalbasit:
@jayconrod https://github.com/bmatcuk/doublestar is working well. Please take another look.

jayconrod:
> P.S: Are you active on Slack's BazelBuild? I PM'd you there.

I hang out on the Gophers Slack but not the Bazel Slack. I watch the #modules and #bazel channels.

I should probably keep a closer eye on the Bazel Slack, but my laptop only has so much RAM :P

kalbasit:
> I'm pretty hesitant to add a new dependency for this. Could you say why `doublestar` is needed specifically? I'd prefer to use `path.Match` if we can.

Of course.

The stdlib function `path.Match` does not currently support the `**` syntax (globstar) when globbing as explained in [this issue](https://github.com/golang/go/issues/11862). Say you want to ignore `a/b/c.pb.go`, and `root.pb.go` files.
- With globstar enabled, you can ignore `**/*.pb.go` and it will match both files.
- Without globstarr, and using the stdlib `path.Match`, you'll have to add two ignores: `*.pb.go` and `a/b/*.pb.go` (or `*/*/*.pb.go`).

Essentially, what the problem is that none of the terms that are supported by [path.Match](https://pkg.go.dev/path?tab=doc#Match) match a path separator (aka `/`) so you have to write a match for every sub-directory level you have.

> > P.S: Are you active on Slack's BazelBuild? I PM'd you there.
> 
> I hang out on the Gophers Slack but not the Bazel Slack. I watch the #modules and #bazel channels.
> 
> I should probably keep a closer eye on the Bazel Slack, but my laptop only has so much RAM :P

:rofl: I hear you!

Alright, I'll catch up with you on the Gopher Slack then.

walk/config.go
```diff
@@ -46,7 +49,12 @@ func (wc *walkConfig) isExcluded(rel, base string) bool {
 	}
 	f := path.Join(rel, base)
 	for _, x := range wc.excludes {
-		if f == x {
+		matched, err := doublestar.Match(x, f)
+		if err != nil {
```
jayconrod, line 53: This could get very chatty if there are a lot of files. Let's instead check for an error in `walk.Config.Configure` when we first read the pattern by calling `doublestar.Match` with a dummy file. Then we can ignore the error here.

walk/config.go
```diff
@@ -58,7 +66,7 @@ type Configurer struct{}
 func (_ *Configurer) RegisterFlags(fs *flag.FlagSet, cmd string, c *config.Config) {
 	wc := &walkConfig{}
 	c.Exts[walkName] = wc
-	fs.Var(&gzflag.MultiFlag{Values: &wc.excludes}, "exclude", "Path to file or directory that should be ignored (may be repeated)")
+	fs.Var(&gzflag.MultiFlag{Values: &wc.excludes}, "exclude", "path.Match pattern that should be ignored (may be repeated)")
```
jayconrod, line 69: Let's say `"pattern that should be ignored ...` without naming the specific implementation.

walk/walk_test.go
```diff
@@ -166,22 +166,31 @@ func TestExcludeFiles(t *testing.T) {
 		{
 			Path: "BUILD.bazel",
 			Content: `
+# gazelle:exclude **/*.pb.go
+# gazelle:exclude *.gen.go
```
jayconrod, line 170: Let's also test a directory.

```
# gazelle:exclude a/**/b
```

Should exclude `a/b/bad`, and `a/x/b/bad`.

kalbasit: done

walk/config.go
kalbasit: done

walk/config.go
kalbasit: I've addressed this by checking the validity of the pattern during the configure phase. I've noticed also that some invalid patterns do not trigger an error with some strings but do so with other strings. I'm not sure what the best course of action here, but I ended up replacing this log with a panic so the user can file a bug with us if the checking failed.

Issue #731, opened 2020-03-17
Handling proto files that are consumed by existing proto_library rules


linzhp:
I kind of shooting myself in the foot by #720. Now proto extension generates a proto_library for all generated proto files. After discussion in #727, we decided to write our own extension to generate a proto_library for each proto file because they need to compile into different Go libraries. So we no longer need Gazelles's proto extension to do that anymore.

Should I revert #720, or should I do something like #643 for proto?

linzhp:
Or should I have go extension to generate `go_proto_library` whenever there is a `proto_library`? Basically removing [this if statement](https://github.com/bazelbuild/bazel-gazelle/blob/master/language/go/generate.go#L395-L399)

jayconrod:
If you don't need `language/proto` to generate `proto_library` rules for generated .proto files anymore, it's probably best to roll back #720. Either behavior seems reasonable to me, but doing less is probably better, especially if it's getting in the way for you.

About `go_proto_library`, that `if` statement is necessary when proto rule generation is disabled. Not sure what you need here exactly. Do you want to generate `go_proto_library` for some `proto_library` rules but not others?

linzhp:
> About go_proto_library, that if statement is necessary when proto rule generation is disabled. Not sure what you need here exactly. Do you want to generate go_proto_library for some proto_library rules but not others?

An alternative is to disable proto_library generation in Gazelle's default proto extension. Our private extension doesn't honor `# gazelle:proto`, so it would continue to generate proto_library. The if that `if` statement is removed, Go extension can generate `go_proto_library` whenever there is a `proto_library`, whether it is handwritten or generated by our private extension.

Reverting #720 is probably the simplest solution here.

PR #758, opened 2020-04-14
Excluding consumed generated files from proto_library

ekleiner:
**What type of PR is this?**
Bug fix

language/proto

Some generated files may have been consumed by other rules, either handwritten or generated by other extensions. The go extension should not add them into the proto_library rule in such case.

https://github.com/bazelbuild/bazel-gazelle/issues/731

### DIFF
--- a/language/proto/generate.go
+++ b/language/proto/generate.go
@@ -41,13 +41,29 @@ func (_ *protoLang) GenerateRules(args language.GenerateArgs) language.GenerateR
 			regularProtoFiles = append(regularProtoFiles, name)
 		}
 	}
-	var genProtoFiles []string
+
+	// Some of the generated files may have been consumed by other rules
+	consumedFileSet := make(map[string]bool)
+	for _, r := range args.OtherGen {
+		for _, f := range r.AttrStrings("srcs") {
+			consumedFileSet[f] = true
+		}
+		if f := r.AttrString("src"); f != "" {
+			consumedFileSet[f] = true
+		}
+	}
+	var (
+		genProtoFiles, genProtoFilesNotConsumed []string
+	)
 	for _, name := range args.GenFiles {
 		if strings.HasSuffix(name, ".proto") {
 			genProtoFiles = append(genProtoFiles, name)
+			if !consumedFileSet[name] {
+				genProtoFilesNotConsumed = append(genProtoFilesNotConsumed, name)
+			}
 		}
 	}
-	pkgs := buildPackages(pc, args.Dir, args.Rel, regularProtoFiles, genProtoFiles)
+	pkgs := buildPackages(pc, args.Dir, args.Rel, regularProtoFiles, genProtoFilesNotConsumed)
 	shouldSetVisibility := args.File == nil || !args.File.HasDefaultVisibility()
 	var res language.GenerateResult
 	for _, pkg := range pkgs {
--- a/language/proto/generate.go
+++ b/language/proto/generate.go
@@ -52,13 +52,12 @@ func (_ *protoLang) GenerateRules(args language.GenerateArgs) language.GenerateR
 			consumedFileSet[f] = true
 		}
 	}
-	var (
-		// genProtoFilesNotConsumed represents only not consumed generetad files.
-		// genProtoFiles represents all generated files.
-		// This is required for not generating empty rules for consumed generated
-		// files.
-		genProtoFiles, genProtoFilesNotConsumed []string
-	)
+
+	// genProtoFilesNotConsumed represents only not consumed generetad files.
+	// genProtoFiles represents all generated files.
+	// This is required for not generating empty rules for consumed generated
+	// files.
+	var genProtoFiles, genProtoFilesNotConsumed []string
 	for _, name := range args.GenFiles {
 		if strings.HasSuffix(name, ".proto") {
 			genProtoFiles = append(genProtoFiles, name)
--- a/language/proto/generate.go
+++ b/language/proto/generate.go
@@ -45,6 +45,9 @@ func (_ *protoLang) GenerateRules(args language.GenerateArgs) language.GenerateR
 	// Some of the generated files may have been consumed by other rules
 	consumedFileSet := make(map[string]bool)
 	for _, r := range args.OtherGen {
+		if r.Kind() != "proto_library" {
+			continue
+		}
 		for _, f := range r.AttrStrings("srcs") {
 			consumedFileSet[f] = true
 		}
--- a/language/proto/generate.go
+++ b/language/proto/generate.go
@@ -51,11 +51,12 @@ func (_ *protoLang) GenerateRules(args language.GenerateArgs) language.GenerateR
 		for _, f := range r.AttrStrings("srcs") {
 			consumedFileSet[f] = true
 		}
-		if f := r.AttrString("src"); f != "" {
-			consumedFileSet[f] = true
-		}
 	}
 	var (
+		// genProtoFilesNotConsumed represents only not consumed generetad files.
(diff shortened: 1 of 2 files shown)


jayconrod:
This was implemented in #720 then rolled back in #733. I'm not really convinced this is a good idea.

Is there a compelling reason to reopen #731? I think in this situation it's better to maintain proto rules manually or to implement your own proto extension.

ekleiner:
Yes. Because of current logic the generated proto files are added into existing proto rules, so in case of mix of generated and real protos there is no other workaround. 
And this way seems to be more appropriate to fix it because it touches only already used generated proto files.

linzhp:
@jayconrod this is quite different than #720. #720 added generated proto files to the proto_library generated by Gazelle's proto extension, even though there is not regular proto file in that package. We later decided to write our own extension for generated proto files and we don't want the default proto extension to add generated protos to proto_library anymore. So #720 was rolled back as a quick fix. However, this PR should be the proper fix of #731 because when there is regular proto files in the same package as the generated ones, the default proto extension still adds the generated ones to the proto_library it generates. In that case, generated proto files were added to two proto_library rules: one generated by our extension, and the other generated by default proto extension.

jayconrod:
@linzhp Thanks for clarifying, that makes more sense.

@ekleiner It's still not really clear to me that when another rule consumes a generate `.proto` file, it should be excluded from the `proto_library` rules Gazelle generates. For example, suppose I add a `filegroup` or `pkg_tar` that packages up `.proto` files for an integration test or a release artifact. After I add that rule, Gazelle would stop updating my `proto_library` rules. I don't think that's the right behavior.

This seems like a pretty specific situation. Maybe instead you could generate files with a particular extension (say, `.gen.proto`), then add `# gazelle:exclude **/*.gen.proto` to the root build file (after #749)?

language/proto/generate.go
```diff
@@ -41,13 +41,29 @@ func (_ *protoLang) GenerateRules(args language.GenerateArgs) language.GenerateR
 			regularProtoFiles = append(regularProtoFiles, name)
 		}
 	}
-	var genProtoFiles []string
+
+	// Some of the generated files may have been consumed by other rules
+	consumedFileSet := make(map[string]bool)
+	for _, r := range args.OtherGen {
```
linzhp, line 47: @jayconrod I think this can be more specific by having:
```Go
if r.Kind() != "proto_library" {
  continue
}
```
It should be an invalid use case for two proto_library rules to compile the same proto file. Gazelle won't be able to resolve that. It would still allow filegroup or pkg_tar rules to consume the proto files

ekleiner:
@jayconrod 
In my case I have a bunch of proto_library rules with generated proto files which are generated by separate extension. They need to coexist with real proto fies in the same directory and the dependencies still need to be resolvable by proto extension. Excluding will not work because it will break proto dependencies calculation.
After a change proposed by @linzhp filegroup or pkg_tar should work fine. And there is no sense to have 2 proto_library rules in the same directory with the same proto file. So, I don't think that will break anything.

language/proto/generate.go
```diff
@@ -41,13 +41,32 @@ func (_ *protoLang) GenerateRules(args language.GenerateArgs) language.GenerateR
...
+		if r.Kind() != "proto_library" {
+			continue
+		}
+		for _, f := range r.AttrStrings("srcs") {
+			consumedFileSet[f] = true
+		}
+		if f := r.AttrString("src"); f != "" {
```
jayconrod, line 54: This `if` should no longer be needed, since `proto_library` doesn't have `src`.

language/proto/generate.go
```diff
@@ -41,13 +41,32 @@ func (_ *protoLang) GenerateRules(args language.GenerateArgs) language.GenerateR
...
+			consumedFileSet[f] = true
+		}
+	}
+	var (
```
jayconrod, line 58: `var genProtoFiles, genProtoFilesNotConsumed []string`

Also, please add a quick comment explaining these are separate so we don't generate empty rules for generated files that *are* consumed.

language/proto/generate_test.go
```diff
@@ -188,6 +188,75 @@ func TestGeneratePackage(t *testing.T) {
...
+		Config:       c,
+		Dir:          filepath.FromSlash("testdata/protos"),
+		File:         old,
+		Rel:          "protos",
+		RegularFiles: []string{"foo.proto"},
+		GenFiles:     []string{"gen.proto", "gen_not_consumed.proto"},
+		OtherGen:     []*rule.Rule{genRule1, genRule2},
+	}
+
+	args.Config = c
```
jayconrod, line 235: This line can be deleted.

Maybe just call `lang.GenerateRules` with a struct literal, as in `TestGeneratePackage`? That would be simpler.

language/proto/generate_test.go
```diff
@@ -188,6 +188,75 @@ func TestGeneratePackage(t *testing.T) {
...
+	res := lang.GenerateRules(args)
+	// Make sure that "gen.proto" is not added to existing foo_proto rule because it is consumed
```
jayconrod, line 237: This is a little unclear, please rephrase.

Also, add a blank line above and wrap at 80 columns.

language/proto/generate_test.go
```diff
@@ -188,6 +188,75 @@ func TestGeneratePackage(t *testing.T) {
...
+	// "gen_not_consumed.proto" is consumed not by "proto_library", this is why it is added
+	if len(res.Gen) != 1 {
```
jayconrod, line 239: As with `res.Empty` below, create an empty file, insert the rules, format, and compare with a string. That should make it more clear what's expected.

language/proto/generate_test.go
```diff
@@ -188,6 +188,75 @@ func TestGeneratePackage(t *testing.T) {
...
+	wantSrcs := []string{"foo.proto", "gen_not_consumed"}
+	if len(gotSrcs) != len(wantSrcs) || gotSrcs[0] != wantSrcs[0] {
+		t.Errorf("got:\n%s\nwant:\n%s", gotSrcs, wantSrcs)
+	}
+	// Make sure that gen.proto is not among empty because it is in GenFiles
+	f := rule.EmptyFile("test", "")
+	for _, r := range res.Empty {
+		r.Insert(f)
+	}
+	f.Sync()
```
jayconrod, line 252: `got := strings.TrimSpace(string(f.Format())`

No need to call `f.Sync` or format `f.File` manually.

language/proto/generate.go
jayconrod: Hmm, okay, I think I agree.

language/proto/generate_test.go
```diff
@@ -188,6 +188,75 @@ func TestGeneratePackage(t *testing.T) {
...
+	oldContent := []byte(`
+proto_library(
+    name = "existing_gen_proto",
+    srcs = ["gen.proto"],
+)
+proto_library(
+    name = "dead_proto",
+    srcs = ["dead.proto"],
+)
+`)
```
jayconrod, line 212: Add a `filegroup` that consumes both `gen.proto` and `gen_not_consumed.proto`. That should verify the new behavior.

language/proto/generate_test.go
```diff
@@ -188,6 +188,84 @@ func TestGeneratePackage(t *testing.T) {
 	}
 }
 
+// Test generated files that have been consumed by other rules should not be
+// added to the  rule
```
jayconrod, line 192: A doc comment should start with "TestConsumedGenFiles checks that ...". Remove the double space and put a period at the end of the sentence.

ekleiner:
> Ok, looks good.

Thank you

Issue #678, opened 2019-12-03
bazel build //:gazelle fails when workspace file is named WORKSPACE.bazel instead of WORKSPACE


cdelguercio:
### What version of gazelle are you using?

v0.19.1

### What version of rules_go are you using?

v0.20.2

### What version of Bazel are you using?

1.2.1

### Does this issue reproduce with the latest releases of all the above?

Yes

### What operating system and processor architecture are you using?

Ubuntu 19.10 x86_64

### What did you do?

I ran bazel build //:gazelle, and it fails only when my Bazel workspace file is named WORKSPACE.bazel. Renaming the file fixes the issue. From the Bazel docs: "Bazel also supports WORKSPACE.bazel file as an alias of WORKSPACE file. If both files exist, WORKSPACE.bazel will take the priority."

### What did you expect to see?

Extracting Bazel installation...
WARNING: Waiting for server process to terminate (waited 5 seconds, waiting at most 60)
Starting local Bazel server and connecting to it...
INFO: Writing tracer profile to '/home/xxx/.cache/bazel/_bazel_xxx/19290186472e4c15bae74e9b7d943605/command.profile.gz'
INFO: Analyzed target //:gazelle (58 packages loaded, 6671 targets configured).
INFO: Found 1 target...
Target //:gazelle up-to-date:
  bazel-bin/gazelle-runner.bash
  bazel-bin/gazelle
INFO: Elapsed time: 28.258s, Critical Path: 3.83s
INFO: 34 processes: 34 linux-sandbox.
INFO: Build completed successfully, 47 total actions

### What did you see instead?

Starting local Bazel server and connecting to it...
INFO: Call stack for the definition of repository 'bazel_gazelle_go_repository_config' which is a go_repository_config (rule definition at /home/xxx/.cache/bazel/_bazel_xxx/19290186472e4c15bae74e9b7d943605/external/bazel_gazelle/internal/go_repository_config.bzl:57:24):
 - /home/xxx/.cache/bazel/_bazel_xxx/19290186472e4c15bae74e9b7d943605/external/bazel_gazelle/deps.bzl:77:5
 - /home/xxx/xxx/WORKSPACE.bazel:26:1
ERROR: An error occurred during the fetch of repository 'bazel_gazelle_go_repository_config':
   Not a regular file: /home/xxx/xxx/poker/WORKSPACE
[...]

PR #769, opened 2020-04-27
Update WORKSPACE finder to support "WORKSPACE.bazel" file name

### LINKED ISSUE #764: Workspace root detection fails when "WORKSPACE.bazel" used
In a Bazel workspace using the `WORKSPACE.bazel` filename instead of plain `WORKSPACE`, Gazelle is unable to discover the root of the repository.

```
gazelle: -repo_root not specified, and WORKSPACE cannot be found: file does not exist
```

The issue seems to be in [`internal/wspace/finder.go`](https://github.com/bazelbuild/bazel-gazelle/blob/master/internal/wspace/finder.go), where it assumes that the `WORKSPACE` file must always be named exactly that.

Either this file could hard-code the two legal names for WORKSPACE files, or it could become a configurable option like BUILD file names are. 

I wanted to open this up for discussion before making a PR.

Note that this is similar to, but not the same as, #678, which possibly indicates there needs to be some testing for this scenario.

v0.19.1

v0.21.0

2.0.0

Yes

HALtheWise:
**What type of PR is this?**

Bug fix

internal/wspace

Makes WORKSPACE root detection work with `WORKSPACE.bazel` filename

Fixes #764

### DIFF
--- a/cmd/gazelle/update-repos.go
+++ b/cmd/gazelle/update-repos.go
@@ -25,6 +25,7 @@ import (
 	"strings"
 
 	"github.com/bazelbuild/bazel-gazelle/config"
+	"github.com/bazelbuild/bazel-gazelle/internal/wspace"
 	"github.com/bazelbuild/bazel-gazelle/language"
 	"github.com/bazelbuild/bazel-gazelle/merger"
 	"github.com/bazelbuild/bazel-gazelle/repo"
@@ -99,7 +100,10 @@ func (*updateReposConfigurer) CheckFlags(fs *flag.FlagSet, c *config.Config) err
 	}
 
 	var err error
-	workspacePath := filepath.Join(c.RepoRoot, "WORKSPACE")
+	workspacePath, err := wspace.FindWorkspaceFile(c.RepoRoot)
+	if err != nil {
+		return fmt.Errorf("loading WORKSPACE file: %v", err)
+	}
 	uc.workspace, err = rule.LoadWorkspaceFile(workspacePath, "")
 	if err != nil {
 		return fmt.Errorf("loading WORKSPACE file: %v", err)
@@ -207,7 +211,7 @@ func updateRepos(args []string) (err error) {
 		macroPath = filepath.Join(c.RepoRoot, filepath.Clean(uc.macroFileName))
 	}
 	for f := range genForFiles {
-		if macroPath == "" && filepath.Base(f.Path) == "WORKSPACE" ||
+		if macroPath == "" && strings.Contains(filepath.Base(f.Path), "WORKSPACE") ||
 			macroPath != "" && f.Path == macroPath && f.DefName == uc.macroDefName {
 			newGenFile = f
 			break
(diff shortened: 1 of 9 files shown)


HALtheWise:
@googlebot I fixed this

HALtheWise:
@googlebot I fixed it.

cmd/gazelle/update-repos.go
```diff
@@ -207,7 +208,7 @@ func updateRepos(args []string) (err error) {
 		macroPath = filepath.Join(c.RepoRoot, filepath.Clean(uc.macroFileName))
 	}
 	for f := range genForFiles {
-		if macroPath == "" && filepath.Base(f.Path) == "WORKSPACE" ||
+		if macroPath == "" && strings.Contains(filepath.Base(f.Path), "WORKSPACE") ||
```
jayconrod, line 211: `strings.Contains` isn't right. It shouldn't match `aaaWORKSPACEbb` for example.

Perhaps `wspace` should provide a function `func IsWORKSPACE(path) bool` that can be used here.

internal/wspace/finder.go
```diff
@@ -19,23 +19,39 @@ import (
 	"strings"
 )
 
-const workspaceFile = "WORKSPACE"
+var workspaceFiles = []string{"WORKSPACE.bazel", "WORKSPACE"}
 
-// Find searches from the given dir and up for the WORKSPACE file
+// FindWorkspaceFile returns a path to a file in the provided root directory,
+// either to an existing WORKSPACE or WORKSPACE.bazel file, or to root/WORKSPACE
+// by default. Note that this function does NOT recursively check parent directories
```
jayconrod, line 26: In stead of "by default", say "if neither file exists".

Also, add a period at the end of the last sentence.

internal/wspace/finder.go
```diff
@@ -19,23 +19,39 @@ import (
...
+func FindWorkspaceFile(root string) string {
```
jayconrod, line 27: Let's call this `FindWORKSPACEFile`. That should make it more clear that this is a `WORKSPACE` file, not a file in a workspace.

internal/wspace/finder.go
```diff
@@ -19,23 +19,39 @@ import (
...
+			return path
+		}
+	}
+	return filepath.Join(root, "WORKSPACE")
+}
+
+// FindRoot searches from the given dir and up for a directory containing a WORKSPACE file
 // returning the directory containing it, or an error if none found in the tree.
-func Find(dir string) (string, error) {
+func FindRoot(dir string) (string, error) {
```
jayconrod, line 39: Let's call this `FindRepoRoot`. That term is used pretty consistently throughout the code base.

HALtheWise:
Resolved

smhmayboudi:
This problem steel available on osx.

ERROR: An error occurred during the fetch of repository 'bazel_gazelle_go_repository_config':
   Not a regular file: /Users/xyz/Developer/a-b/a_b/WORKSPACE

HALtheWise:
I think you might be hitting #678 try the workaround there

PR #791, opened 2020-05-22
Adding go_proto directive to disable go_proto_library generation

linzhp:
**What type of PR is this?**

Feature

language/go

Some packages comes with pregenerated Go files. In some situations, people may need to import the proto files from those packages but use the pregenerated Go files when they are compiling their Go files. One of such situation is gogoproto. If we compile github.com/gogo/protobuf/gogoproto/gogo.proto with Google's go proto compiler, it will end up with type erros in https://github.com/bazelbuild/rules_go/issues/2512. If we configure the compiler to be one of gogoproto compiler, there will be circular dependency:

```
in go_library rule @com_github_gogo_protobuf//gogoproto:go_default_library: cycle in dependency graph:
.-> @com_github_gogo_protobuf//gogoproto:go_default_library
|   @com_github_gogo_protobuf//gogoproto:gogoproto_go_proto
|   @io_bazel_rules_go//proto:gogoslick_grpc
`-- @com_github_gogo_protobuf//gogoproto:go_default_library
```

So we have to use the pregenerated Go file in github.com/gogo/protobuf/gogoproto. At the same time, some other proto files may need to import github.com/gogo/protobuf/gogoproto/gogo.proto (e.g., github.com/coreos/etcd/raft/raftpb/raft.proto). To build these proto files, we need a `proto_library` rule for gogo.proto.

This PR adds `gazelle:go_proto` directive to disable generating `go_proto_library`. So we can generate a `proto_library` and a `go_library` in gogoproto, independent of each other.

This directive may also be useful for non-Go repos, where they still need `proto_library` but not `go_proto_library`.

Fixes https://github.com/bazelbuild/rules_go/issues/2512


language/go/config.go
```diff
@@ -234,6 +238,7 @@ func (*goLang) KnownDirectives() []string {
 		"go_visibility",
 		"importmap_prefix",
 		"prefix",
+		"go_proto",
 	}
```
jayconrod, line 242: Nit: keep sorted

language/go/config.go
```diff
@@ -369,7 +374,10 @@ func (*goLang) Configure(c *config.Config, rel string, f *rule.File) {
 				}
 				gc.preprocessTags()
 				gc.setBuildTags(d.Value)
-
+			case "go_proto":
```
jayconrod, line 377: I wonder if this should have a more explicit name like `go_generate_proto`?

language/go/config.go
```diff
@@ -369,7 +374,10 @@ func (*goLang) Configure(c *config.Config, rel string, f *rule.File) {
...
+				if d.Value == "disable" {
```
jayconrod, line 378: Let's `log.Printf` an error for any values we don't recognize.

Do you expect this will need to accept more than a Boolean value in the future? If not, lets use `strconv.ParseBool`. It seems like we'll only ever want to follow the current proto mode or disable rule generation, so a Boolean should be fine.

language/go/generate.go
```diff
@@ -79,7 +79,8 @@ func (gl *goLang) GenerateRules(args language.GenerateArgs) language.GenerateRes
 	// to any .proto files present.
 	regularFiles := append([]string{}, args.RegularFiles...)
 	genFiles := append([]string{}, args.GenFiles...)
-	if !pcMode.ShouldIncludePregeneratedFiles() {
+	gc := getGoConfig(c)
```
jayconrod, line 82: It might be simpler to change `getProtoMode` to return `proto.DisableMode` when `gc.goProto` is false. That would cause everything to behave as if proto rule generation were disabled, even though there are generated proto rules.

PR #801, opened 2020-06-02
Making a hard copy the overrides slice

### LINKED ISSUE #800: gazelle:resolve in sibling directories can override each other
`resolveConfig.overrides` is copied from parent directory using `[:]`:
https://github.com/bazelbuild/bazel-gazelle/blob/8a478f3feb7ce56185bd9da646cd34be640dbd50/resolve/config.go#L80

As a result, `rcCopy.overrides` and  `rc.overrides` shares the same backing array. When `rc.overrides` has `[a, b, c]` with capacity of 4 and length of 3, and `rcCopy.overrides` at dir1 appends `d` into it, the backing array has `[a, b, c, d]`. The length of `rcCopy.overrides` is 4 while that of `rc.overrides` is 3. When a sibling dir2 creates another `rcCopy.overrides`, the backing array is still `[a, b, c, d]` but the length is 3. So when `rcCopy.overrides` of dir2 appends `e`, the backing array becomes `[a, b, c, e]`. Unexpectedly, `rcCopy.overrides` of dir1 becomes `[a, b, c, e]` too.

Here is a simple example:

```go
package main

import (
	"fmt"
)

func main() {
	parent := make([]int, 3, 4)
	parent[0] = 0
	parent[1] = 1
	parent[2] = 2
	child1 := parent[:]
	child1 = append(child1, 3)
	fmt.Println(child1)
	child2 := parent[:]
	child2 = append(child2, 4)
	fmt.Println(child1)
}
```

8a478f3feb7ce56185bd9da646cd34be640dbd50

0.23.1

3.1.0

It should

Darwin amd 64

linzhp:
**What type of PR is this?**

Bug fix

resolve

Copying `resolveConfig.overrides` without sharing the backing arrary

Fixes #800

### DIFF
--- a/resolve/config.go
+++ b/resolve/config.go
@@ -77,8 +77,9 @@ func (_ *Configurer) KnownDirectives() []string {
 func (_ *Configurer) Configure(c *config.Config, rel string, f *rule.File) {
 	rc := getResolveConfig(c)
 	rcCopy := &resolveConfig{
-		overrides: rc.overrides[:],
+		overrides: make([]overrideSpec, len(rc.overrides)),
 	}
+	copy(rcCopy.overrides, rc.overrides)
 
 	if f != nil {
 		for _, d := range f.Directives {


resolve/config.go
```diff
@@ -77,8 +77,9 @@ func (_ *Configurer) KnownDirectives() []string {
 func (_ *Configurer) Configure(c *config.Config, rel string, f *rule.File) {
 	rc := getResolveConfig(c)
 	rcCopy := &resolveConfig{
-		overrides: rc.overrides[:],
+		overrides: make([]overrideSpec, len(rc.overrides)),
```
jayconrod, line 80: Instead of unconditionally allocating a new slice here, let's do this:

```
rcCopy := &resolveConfig{
  overrides: rc.overrides[:len(rc.overrides):len(rc.overrides)],
}
```

That way, we'll use the original backing array most of the time, but we'll allocate a new array if we append anything. Most build files won't have `# gazelle:resolve` directives, so this should save some memory.

PR #808, opened 2020-06-07
Go naming convention

### LINKED ISSUE #5: Use package name instead of go_default_library for rules
Originally bazelbuild/rules_go#265

We used `go_default_library` as the name for `go_library` rules generated by Gazelle to simplify dependency resolution. Both the Go rules and Gazelle relied on this, along with `go_prefix`.

Now, `go_prefix` is deprecated and Gazelle adds `importpath` attributes to all Go rules. It also uses an index for dependency resolution instead of transforming import path strings. There's no longer a technical need for `go_default_library`, so we'd like to migrate away from that name.

In general, we'd like the library name to match the last component of the Bazel package. For example, a library stored at `//foo/bar` should be named `bar`, so it can be compiled with `bazel build //foo/bar` instead of `bazel build //foo/bar:go_default_library`. Tests would be named `bar_test` or `bar_xtest`.

We may want this to work differently in packages with a `go_binary` (sources have `package main`). For these packages, the `go_binary` name should match the package (again, so it can be compiled with `//foo/bar`). The library will need a different name. We may want to put sources and dependencies in `go_binary` instead of embedding a library if there are no tests.

tomlu:
**What type of PR is this?**
Feature

 language/go

Adds a naming convention for Go target generation.

Fixes #5


language/go/fix.go
```diff
@@ -34,6 +35,110 @@ func (_ *goLang) Fix(c *config.Config, f *rule.File) {
...
+	for _, r := range f.Rules {
+		if r.Kind() == "go_binary" {
+			return r.Name()
+		}
+	}
+	return ""
+}
+
+// import path returns the existing import path from the first encountered Go rule with the attribute set.
+func importPath(f *rule.File) string {
```
jayconrod, line 111: Probably better to use `InferImportPath` (from `package.go`), which is used to set `importpath` in generated rules.

language/go/fix.go
```diff
@@ -34,6 +35,110 @@ func (_ *goLang) Fix(c *config.Config, f *rule.File) {
...
+		case "go_binary", "go_library", "go_test":
+			if ip, ok := r.Attr("importpath").(*bzl.StringExpr); ok && ip.Value != "" {
+				return ip.Value
+			}
+		}
+	}
+	return f.Pkg
+}
+
+func replaceInListAttr(r *rule.Rule, attr, old, new string) {
```
jayconrod, line 123: Can be simplified a bit:

```
list := r.AttrStrings(attr)
if list == nil {
  return
}
for i, v := range list {
  if v == old {
    list[i] = new
  }
}
r.SetAttr(attr, list)
```

language/go/fix.go
```diff
@@ -72,6 +177,9 @@ func migrateGrpcCompilers(c *config.Config, f *rule.File) {
 // MergeFile will remove unused values and attributes later.
 func squashCgoLibrary(c *config.Config, f *rule.File) {
 	// Find the default cgo_library and go_library rules.
+	binName := binName(f)
```
jayconrod, line 180: No change needed to any of these fixes. They just need to run first.

language/go/package.go
```diff
@@ -166,6 +169,54 @@ func (pkg *goPackage) inferImportPath(c *config.Config) error {
 	return nil
 }
 
+// Returns a a suitable go_library name based on the dir.
+// Major version suffixes (eg. "v1") are dropped.
+func libNameFromDir(dir string) string {
```
jayconrod, line 174: Change name and signature to `libNameFromImportPath(imp string) (string, bool)`. If we can't find the library name because the import path is empty or garbled, let's return false. Callers may want to handle that.

Maybe the regexp can just match the last component, dropping the version suffix? Haven't tested this, but:

```
^(?:.*/)([^/]+)(?:/v\d+)?$
```

That could simplify the function to:

```
m := goPackageRe.FindStringSubmatch(imp)
if m == nil {
  return "", false
}
return m[1], true
```

language/go/package.go
```diff
@@ -166,6 +169,54 @@ func (pkg *goPackage) inferImportPath(c *config.Config) error {
 	return nil
 }
 
+// Returns a a suitable go_library name based on the dir.
```
jayconrod, line 172: Nit: `libNameFromDir returns ...`

language/go/package.go
```diff
@@ -166,6 +169,54 @@ func (pkg *goPackage) inferImportPath(c *config.Config) error {
...
+		return defaultLibName
+	}
+	if binName != "" {
+		return binName + "_lib"
+	}
+	return libNameFromDir(imp)
+}
+
+// A simpler version of libNameByConvention for library dependencies.
+func resolvedLibName(nc namingConvention, imp string) string {
```
jayconrod, line 205: Nit: probably just inline this function.

language/go/resolve.go
```diff
@@ -255,7 +256,7 @@ func resolveWithIndexGo(ix *resolve.RuleIndex, imp string, from label.Label) (la
 
 var modMajorRex = regexp.MustCompile(`/v\d+(?:/|$)`)
 
-func resolveExternal(moduleMode bool, rc *repo.RemoteCache, imp string) (label.Label, error) {
+func resolveExternal(c *config.Config, gc *goConfig, rc *repo.RemoteCache, imp string) (label.Label, error) {
```
jayconrod, line 259: If both `c` and `gc` are needed, just pass in `c` and use `getGoConfig`.

language/go/resolve.go
```diff
@@ -288,11 +290,20 @@ func resolveExternal(moduleMode bool, rc *repo.RemoteCache, imp string) (label.L
 		pkg = pathtools.TrimPrefix(impWithoutSemver, prefix)
 	}
 
-	return label.New(repo, pkg, defaultLibName), nil
+	nc := gc.goNamingConvention
+	if gr, ok := c.GoRepos[repo]; ok {
```
jayconrod, line 294: Instead of `GoRepos`, find the matching rule in `c.Repos`, then `r.AttrString("build_naming_convention")`.

Optionally, create a `map[string]namingConvention` to speed up the lookup, and store it in `goConfig`. `Configure` is probably the right place to do that, when `rel == ""` (in the root directory).

language/go/resolve.go
```diff
@@ -288,11 +290,20 @@ func resolveExternal(moduleMode bool, rc *repo.RemoteCache, imp string) (label.L
 		pkg = pathtools.TrimPrefix(impWithoutSemver, prefix)
 	}
 
-	return label.New(repo, pkg, defaultLibName), nil
+	nc := gc.goNamingConvention
```
jayconrod, line 293: Add a comment:

```
// Determine what naming convention is used by the go_repository rule.
// If none is specified with "build_naming_convention", the either naming convention
// can be used, so we'll default to whatever's used in the current workspace to avoid churn.
```

repo/remote.go
```diff
@@ -117,7 +117,7 @@ type modVersionValue struct {
 // lookup logic should be moved to language/go. This means RemoteCache will
 // need to be initialized in a different way.
 type Repo struct {
-	Name, GoPrefix, Remote, VCS string
+	Name, GoPrefix, BuildNamingConvention, Remote, VCS string
```
jayconrod, line 120: Commented above: I think we should either read this from `c.Repos` or build a `map[string]namingConvention` in `language/go.goLang.Configure` with this info. No need to add more Go-specific logic to this package.

config/config.go
```diff
@@ -85,6 +86,9 @@ type Config struct {
 	// generation and dependency resolution.
 	Repos []*rule.Rule
 
+	// go_repository rules, mapped by name.
+	GoRepos map[string]repo.Repo
```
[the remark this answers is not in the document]
tomlu: This is where it would have been very nice to have the repo cache be Go specific instead of pretending to be generic. I just can't get information across to the resolver because it passes through a "generic" conduit.

I can do simply use the repos, but I think it's a linear pass through all the repo rules every time we want to resolve an external dep. That _sounds_ like a lot to me, we will end up having hundreds of repo rules and many thousand deps.

Any ideas?

language/go/package.go
tomlu: Yeah you don't get a lot of bang from this tiny function. But I thought it would be nice to have all the naming convention code in a single place so you can easily see the semantics of the naming convention by reading a single place.

If you still want it inlined I can do it.

language/go/config.go
```diff
@@ -372,6 +446,17 @@ Update io_bazel_rules_go to a newer version in your WORKSPACE file.`
 				log.Printf("Found RULES_GO_VERSION %s. Minimum compatible version is %s.\n%s", gc.rulesGoVersion, minimumRulesGoVersion, message)
 			}
 		}
+		repoNamingConvention := map[string]namingConvention{}
+		for _, repo := range c.Repos {
+			if repo.Kind() == "go_repository" {
+				if nc, err := namingConventionFromString(repo.AttrString("build_naming_convention")); err == nil {
+					repoNamingConvention[repo.Name()] = nc
+				} else {
+					log.Printf("%v\n", err)
```
jayconrod, line 455: More context needed here:

```
log.Printf("in go_repository named %q: %v", repo.Name(), err)
```

No need for a newline at the end of `log.Printf` messages: one is added automatically.

internal/go_repository.bzl
```diff
@@ -235,6 +234,14 @@ go_repository = repository_rule(
                 "off",
             ],
         ),
+        "build_naming_convention": attr.string(
+            values = [
```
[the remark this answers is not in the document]
jayconrod: GitHub wouldn't let me reply to the previous comment, so I'll reply here:

> Do we have to hard launch this, instead of making it opt-in?

We don't have to, but I'd argue that we should.

In general, I'd prefer not to give people too many configuration options. Each option adds complexity, both in usage and in implementation. In situations where we can figure out what the Right Thing is, we should do that. If we can't infer the Right Thing, or if there are several reasonable Right Things, only then should we ask the user.

I definitely think the new naming convention is the Right Thing; `go_default_library` is a historical accident. We need to support the old convention because people are still using it, but the new one should really be the default going forward.

language/go/package.go
```diff
@@ -166,6 +169,49 @@ func (pkg *goPackage) inferImportPath(c *config.Config) error {
 	return nil
 }
 
+// libNameFromDir returns a a suitable go_library name based on the import path.
```
jayconrod, line 172: libNameFromImportPath

language/go/resolve.go
```diff
@@ -288,11 +291,23 @@ func resolveExternal(moduleMode bool, rc *repo.RemoteCache, imp string) (label.L
 		pkg = pathtools.TrimPrefix(impWithoutSemver, prefix)
 	}
 
-	return label.New(repo, pkg, defaultLibName), nil
+	// Determine what naming convention is used by the go_repository rule.
+	// If none is specified with "build_naming_convention", the either naming convention
+	// can be used, so we'll default to whatever's used in the current workspace to avoid churn.
+	nc := gc.goNamingConvention
+	if gc.repoNamingConvention != nil {
```
jayconrod, line 298: Nit: I think `gc.repoNamingConvention` will always be non-nil at this point, so this condition can be removed.

`Configure` should be called in the root directory before any call to `Resolve`.

repository.rst
```diff
@@ -143,6 +143,14 @@ returned by ``go env GOPATH``.
 | A value for ``sum`` may be found in the ``go.sum`` file or by running                                                 |
 | ``go mod download -json <module>@<version>``.                                                                         |
 +--------------------------------+----------------------+---------------------------------------------------------------+
+| :param:`build_naming_convention`  | :type:`string`       | :value:`""`                                                |
```
jayconrod, line 146: Make sure the `|` separators line up with the `+` above and below.

Default value should be `"import_alias"`.

linzhp:
This broke Gazelle runs Uber's Go monorepo. I will debug a bit and come up with more details. cc: @blico

tomlu:
@linzhp Are you sure? This isn't in master yet.

language/go/config.go
```diff
@@ -236,6 +243,69 @@ func (f tagsFlag) String() string {
...
+func namingConventionFromString(s string) (namingConvention, error) {
+	switch s {
+	case "go_default_library":
+		return goDefaultLibraryNamingConvention, nil
+	case "import":
+		return importNamingConvention, nil
+	case "import_alias":
+		return importAliasNamingConvention, nil
+	default:
+		return goDefaultLibraryNamingConvention, fmt.Errorf("unknown naming convention %q", s)
```
linzhp, line 305: The default should be `goDefaultLibraryNamingConvention` to be backward compatible. Plus, the updated `README.md` says: "By default, library targets are named  ``go_default_library`` and test targets ``go_default_test``."

linzhp, line 305: This function should not return an error for the default case, otherise [this line](https://github.com/bazelbuild/bazel-gazelle/pull/808/files#diff-2c95a0b4a5b567eb1f16249785031a67R453) would fail

README.rst
```diff
@@ -575,6 +580,24 @@ The following directives are recognized:
...
 | ``@io_bazel_rules_go//proto:gogofaster_grpc``.                                             |
 +---------------------------------------------------+----------------------------------------+
+| :direc:`# gazelle:go_naming_convention`           | n/a                                    |
++---------------------------------------------------+----------------------------------------+
+| Controls the names of generated Go targets. By default, library targets are named          |
+| ``go_default_library`` and test targets ``go_default_test``.                               |
+|                                                                                            |
+| Valid values are:                                                                          |
+|                                                                                            |
+| * ``go_default_library``: Library targets are named ``go_default_library``, test targets   |
```
linzhp, line 590: Can we change this value to be "default" to cover both "go_default_library" and "go_default_test"?

linzhp:
@jayconrod The above mentioned breakage is due to `namingConventionFromString` function returns an error when the naming convention is not specified explicitly. `gazelle update-repos` doesn't set `build_naming_convention` attribute of `go_repository` rules, leaving it to empty by default. So all `go_repository` rules produced by `gazelle update-repos` broke.

tomlu:
@linzhp Thank you for taking a keen interest in this. You are looking at a mid-review pull request whose latest iteration breaks in this way. Don't worry, we are not going to break Uber or anyone else! You can see that the Gazelle buildkite CI is telling us the same thing, so I couldn't push even if I wanted to. You just happened to catch us in the middle of a half-baked review.

Hope that helps calm your nerves! :)

linzhp:
No worries. Jay notified us about this PR today. That was why I am testing it. There were previous issues due to major changes merged into master without testing in Uber. So I guess Jay is more cautious about this one.

Let me know when you are ready to merge. I can do another round of test.

language/go/config.go
```diff
@@ -372,6 +447,17 @@ Update io_bazel_rules_go to a newer version in your WORKSPACE file.`
 				log.Printf("Found RULES_GO_VERSION %s. Minimum compatible version is %s.\n%s", gc.rulesGoVersion, minimumRulesGoVersion, message)
 			}
 		}
+		repoNamingConvention := map[string]namingConvention{}
+		for _, repo := range c.Repos {
+			if repo.Kind() == "go_repository" {
+				if nc, err := namingConventionFromString(repo.AttrString("build_naming_convention")); err == nil {
```
jayconrod, line 453: Based on @linzhp's comment, if `build_naming_convention` isn't explicitly set, let's supply our own. We know the default in `go_repository` may be different than the default for Gazelle.

```
if repo.Kind() == "go_repository" {
  if attr := repo.AttrString("build_naming_convention"); attr == "" {
    repoNamingConvention = importAliasNamingConvention // default for go_repository
  } else if nc, err := namingConventionFromString(attr); err != nil {
    log.Printf("in go_repository named %q: %v", repo.Name(), err)
  } else {
    repoNamingConvention[repo.Name()] = nc
  }
}
```

language/go/config.go
jayconrod: I think it's correct for `namingConventionFromString` to return an error when it's passed the empty string. However, we should avoid calling it with the empty string and should set a default value instead.

I've added a comment at the call site.

tomlu:
Fixed some comments (and build breakages, I had accidentally deleted some important code so no major dramas, there's nothing fundamentally wrong with the PR).

There are four open issues as far as I'm concerned:
* Where alias generation occurs (that is, not during `fix`). I don't disagree (or know enough to disagree), I just haven't got around to it.
* What should the default convention be.
* What should the default naming convention be for externals. It is currently consistently `go_default_library`, you want `import_alias` I believe.
* Should we add inference code to scan externals that do not have a naming convention.

jayconrod:
> Where alias generation occurs (that is, not during fix). I don't disagree (or know enough to disagree), I just haven't got around to it. Does this need to be fixed to approve the PR?

We could land this on a development branch, then continue work there. That might be a good idea since large GitHub reviews get a bit unwieldy.

> What should the default convention be when one isn't specified?

For now, `goDefaultLibraryNamingConvention` is okay.

Before this is merged to the main branch, we should have an `unknownNamingConvention` that indicates we should try to detect the convention in use.

> What should the default naming convention be for externals? It is currently consistently go_default_library, you want import_alias I believe.

Yep, `import_alias` will be good for compatibility, since either set of names can be used.

> Should we add inference code to scan externals that do not have a naming convention?

Do you mean external `go_repository` repos or something else like `http_archive`? In the former case, we'll know the default, so there's no need to scan. In the latter case, there's not much we can do: Gazelle won't run Bazel, so it can't scan external repos. Users can use `# gazelle:repository` directives to Gazelle to act as if these were `go_repository` repos with certain attributes. Other than that, we should probably assume `go_default_library`.

tomlu:
I agree landing it on a dev branch could be good. That way we can iterate on it and keep it up-to-date with master. How do we do that?

language/go/config.go
```diff
@@ -236,6 +248,59 @@ func (f tagsFlag) String() string {
...
+}
+
+func (f goNamingConventionFlag) String() string {
+	return ""
+}
+
+// namingConvention determines how go targets are named.
+type namingConvention int
+
+const (
```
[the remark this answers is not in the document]
jayconrod: Resolving for now: this can be done in a later PR.

language/go/config.go
```diff
@@ -408,12 +478,15 @@ Update io_bazel_rules_go to a newer version in your WORKSPACE file.`
...
-
 				if goGenerateProto, err := strconv.ParseBool(d.Value); err == nil {
 					gc.goGenerateProto = goGenerateProto
 				} else {
 					log.Printf("parsing go_generate_proto: %v", err)
 				}
+			case "go_naming_convention":
+				if err := gc.setNamingConvention(d.Value); err != nil {
+					log.Print(err)
+				}
```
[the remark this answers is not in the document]
jayconrod: Resolving for now: this can be done in a later PR.

jayconrod:
I just created a new branch with the same base commit and changed the PR to go there when it's merged.

There are still a number of open comments. I've resolved the ones related to auto-detection, but I think the others should be fixed. GitHub doesn't make them super easy to find: make sure to go back to previous reviews and click through to hidden conversations (overly aggressive pagination).

jayconrod:
@tomlu Gentle ping. If you resolve the remaining comments, I'll land this on a dev branch and add auto-detection logic. Should be ready to merge after that.

tomlu:
There are a number of unanswered questions among the comments awaiting the gentleman's pleasure :)

jayconrod:
I may be missing something through the code review UI, but I'm not seeing any unresolved threads with open questions.

You mentioned a number of open issues in an [earlier comment](https://github.com/bazelbuild/bazel-gazelle/pull/808#issuecomment-644993167), but I think I covered everything in my answer [here](https://github.com/bazelbuild/bazel-gazelle/pull/808#issuecomment-645433885).

What else is open?

tomlu:
I'm probably using GitHub wrong or something, or my rebasing messed everything up?

Anyway, if you go to the "files changed" view you can see quite a few questions. I'm getting some crashes that I don't understand for when I try to change the default to `import_alias`, for instance.

language/go/fix.go
jayconrod: Which tests fail exactly? I'm assuming just stuff in fix_test.go.

If the old tests are failing, I think it's fine to lock those into the old naming convention.

If the new tests are failing, some change may be needed. The fixes are transformations run in sequence. Earlier transformations should work without knowledge of newer changes, but maybe they just need to be fixed to ignore files already written with the new naming convention?

language/go/fix.go
tomlu: I don't see how I'm able to use InferImportPath here. I don't have any information about the Go packages by just looking at the BUILD file. What am I supposed to pass as "rel"? The file's package is no good.

jayconrod: I'll withdraw my comment actually. For `migrateNamingConvention` specifically, it would be better to find the `importpath` attribute being used instead of trying to infer one based on the configuration and file location. If no `importpath` is found, it would be better to skip the transformation by returning `""` here and then returning early from `migrateNamingConvention`.

I don't think this function should be called from other transformations. They shouldn't need to support new naming conventions (for example, no one should be using `cgo_library` with the new naming convention).

For reference, the file's `Pkg` field should be the same as `rel`. Both are slash-separated paths relative to the workspace root (with `""` meaning the root itself). There may be some exotic where that's not true, but it should be fine here.

tomlu:
`//internal:go_repository_test` fails with `error: could not load default workspace name: C:\users\tomlu\_bazel_tomlu\vhinxgwb\execroot\bazel_gazelle\bazel-o
ut\x64_windows-fastbuild\bin\internal\go_repository_test_\go_repository_test.exe.runfiles\bazel_gazelle\WORKSPACE
: workspace name not set`.

Re `importpath` the problem is that `f.Pkg` is in fact empty at root, so there is no sensible name to give anything when running fix. I there is a library there I have to use the existing importpath I think. I just don't have the information that I have later in `generate`.

language/go/fix.go
```diff
@@ -34,6 +35,110 @@ func (_ *goLang) Fix(c *config.Config, f *rule.File) {
...
+			}
+		case "go_test":
+			if r.Name() == migrateTestName {
+				r.SetName(testName)
+				replaceInListAttr(r, "embed", ":"+migrateLibName, ":"+libName)
+			}
+		}
+	}
+
+	// Alias migration
```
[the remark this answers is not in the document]
tomlu: OK, I think I've done it.

language/go/fix.go
tomlu: Same response as yesterday but in the thread.

I don't think I can use `f.Pkg` because if there's a lib at the root then I can't infer its import path.

language/go/fix.go
tomlu: What's the harm though in updating these?

language/go/fix.go
jayconrod: `InfertImportPath` considers both the current prefix and `rel`, so if `rel == ""`, the import path is just the prefix. But this is probably moot: it would be better to use the current `importpath`, regardless of what we would infer.

I think the only change needed here is if we don't find any matching rules, return `""` and have `migrateNamingConvention` return early.

language/go/fix.go
jayconrod: It adds complexity and requires additional testing.

The `Fix` method is meant to be like `go fix`. In the early days of Gazelle and rules_go, there were some conventions that we wanted to migrate away from. For example, a separate `cgo_library` target was needed for cgo code, and separate `go_test` targets were needed for internal and external test packages.

Gazelle won't generate targets in those conventions anymore, but we wanted to help people migrate away from them automatically. So `Fix` applies a stack of transformations. 

For the most part, these transformations interact very little, but the stack is loosely ordered, so a transformation may assume that if it needs to be applied, then later transformations haven't been applied yet. For example, `squashCgoLibrary` doesn't need to worry about naming conventions since `migrateNamingConvention` will still be applied later.

Ideally, we should be able to add a new transformation on this stack without modifying any existing transformation. Transformations shouldn't need to support being applied in an arbitrary order.

language/go/generate.go
```diff
@@ -459,9 +468,24 @@ func (g *generator) generateLib(pkg *goPackage, embed string) *rule.Rule {
...
 }
 
-func (g *generator) generateBin(pkg *goPackage, library string) *rule.Rule {
-	name := pathtools.RelBaseName(pkg.rel, getGoConfig(g.c).prefix, g.c.RepoRoot)
-	goBinary := rule.NewRule("go_binary", name)
+func (g *generator) generateAlias(pkg *goPackage, libName string) *rule.Rule {
+	if pkg.isCommand() || libName == "" {
+		return nil
+	}
+	c := getGoConfig(g.c)
```
jayconrod, line 475: Nit: use the abbreviation `gc` for `*goConfig` to distinguish it from `*config.Config`.

tomlu: Done.

tomlu:
I think all comments should be resolved now.

Yes, I am still getting `error: could not load default workspace name: C:\users\tomlu\_bazel_tomlu\vhinxgwb\execroot\bazel_gazelle\bazel-o
ut\x64_windows-fastbuild\bin\internal\go_repository_test_\go_repository_test.exe.runfiles\bazel_gazelle\WORKSPACE
: workspace name not set`

It's coming from here: https://github.com/bazelbuild/rules_go/blob/master/go/tools/bazel_testing/bazel_testing.go#L464

tomlu:
I note that the corresponding test in master (though it has been renamed and consolidated) seems to fail in the same manner for me. Consequently I think this should probably be addressed separately to this pull request.

jayconrod:
Ok, cool, I think this is in a good state to be merged onto the dev branch. Thanks again for working on this!

It does look like `//internal:go_repository_test` doesn't work on Windows. It's disabled on Windows in CI. Not sure I remember why, but there are a lot of issues with the test framework due to lack of symlinks and sandboxing. Doesn't seem to be related to this PR in any case.

I played around with switching the Gazelle repo to use this. I ran into a couple issues, but I think they'll be easier to fix in a follow-up PR. I'll set aside some time on Friday to work on this and to add mode auto-detection.

* Test and alias actuals seem to be named based on the package name instead of the last component of the import path. That causes a duplicate test in `//language/go` (which uses the package name `golang` because `go` is a keyword). May cause dependency resolution issues elsewhere.
* Repositories that don't use `go_repository` should still use the old names by default. For example, an import of `golang.org/x/tools/go/vcs` should resolve to `@org_golang_x_tools//go/vcs:go_default_library` since that comes from an `http_archive` that hasn't been updated yet.

tomlu:
> Test and alias actuals seem to be named based on the package name instead of the last component of the import path

Yeah that is wrong and wasn't my intention. We happen to have google3-like Glaze compatible rules in our little internal monorepo, so I wouldn't have noticed if I got this wrong.

> Repositories that don't use go_repository should still use the old names by default

We vendor our dependencies and use `local_repository` rules rather than `go_repository`. I suppose we would need to put Gazelle directives into our `WORKSPACE` file to simulate having a `go_repository` with the naming convention set?

tomlu:
> Test and alias actuals seem to be named based on the package name instead of the last component of the import path

I tried to repro this with a test, but I don't think I quite understand what you mean. The test does fail, but I don't think it's related to what you are talking about:

https://github.com/tomlu/bazel-gazelle/commit/81411c794cbf22979de54710b5e498ba0a43a7da

jayconrod:
> We vendor our dependencies and use local_repository rules rather than go_repository. I suppose we would need to put Gazelle directives into our WORKSPACE file to simulate having a go_repository with the naming convention set?

Probably yes. We won't know what naming convention is used in any repository not declared with `go_repository`, unless there's some explicit hint. By default, at least initially, we'll need to assume folks haven't updated their build files, so we'll use `go_default_library`.

> I tried to repro this with a test, but I don't think I quite understand what you mean. The test does fail, but I don't think it's related to what you are talking about:

I noticed when updating the Gazelle repo. It happened in `//language/go`. I don't have the diff easily available right now, but I have some time set aside this afternoon to work on it.

PR #811, opened 2020-06-09
Use go-cmp to provide file differences

achew22:
Instead of printing both files out and asking the user to visually diff,
perform the diff for the user and print it out.

**What type of PR is this?**

>
Other

Usability improvement

* `testtools`

This adds a dependency on, and then uses the `go-cmp` which is released by Google. This package is widely used both inside and out of Google for comparison testing and has good performance characteristics (it won't allocate 500mb of ram just for fun).

None, this is simple enough that you should be able to accept/reject on the face of it. No problems if you want to reject, this just made building the `skylib` plugin A LOT easier.

**Other notes for review**

Thanks for all your hard work on Gazelle!

### DIFF
--- a/deps.bzl
+++ b/deps.bzl
@@ -142,6 +142,14 @@ def gazelle_dependencies(
         version = "v1.2.2",
     )
 
+    _maybe(
+        go_repository,
+        name = "com_github_google_go_cmp",
+        importpath = "github.com/google/go-cmp",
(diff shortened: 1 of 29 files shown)


achew22:
@jayconrod I believe I've addressed your concerns. PTAL

swsnider:
This change introduces both go-cmp and xerrors, but accidentally names both `go_repository()` invocations the same. Pretty sure this is only working because `rules_go` also pulls in xerrors?

PR #818, opened 2020-06-16
Update bazelbuild/buildtools to 3.2.1

prashantv:
* Dict elements are now KeyValueExpr, no casts are needed
  See https://github.com/bazelbuild/buildtools/pull/856
* Update tests for new expected field order.

**What type of PR is this?**

Dependency upgrade, and changes required to compile with new version.

all

It's needed to compile gazelle with the latest buildtools release.


prashantv:
> Please run `go mod tidy` and `go mod vendor` and include the updated `go.sum` and `vendor` directory.

Done.

> Also, please drop the parts of this PR not related to updating the dependency (buildifier on the test data I think?). That makes it more difficult than necessary to review.

Unfortunately this seems required by the update too. Without those changes, tests fail, e.g.,
```
--- FAIL: TestMoveLabels (0.00s)
    --- FAIL: TestMoveLabels/move (0.00s)
        move_labels_test.go:213: new/a/BUILD diff (-want,+got):
              strings.Join({
                ... // 2 identical lines
                "x_binary(",
                `    name = "a",`,
            +   "    locs = [",
            +   `        "abc $(location  //new/b) $(locations //new/b:b_lib) xyz",`,
            +   "    ],",
                "    deps = [",
                `        ":a_lib",`,
                `        "//new/b",`,
                `        "//new/b:b_lib",`,
                `        "@c//old:c_lib",`,
            -   "    ],",
            -   "    locs = [",
            -   `        "abc $(location  //new/b) $(locations //new/b:b_lib) xyz",`,
                "    ],",
                ")",
              }, "\n")
```

I can move them into a separate PR, since the older version doesn't seem like it reformats the order.

prashantv:
Sorry for the delay, I didn't have time to investigate the test failures, it turns out I just needed to delete the `BUILD.bazel` files in vendor.

deps.bzl
```diff
@@ -106,8 +106,8 @@ def gazelle_dependencies(
         go_repository,
         name = "com_github_bazelbuild_buildtools",
         importpath = "github.com/bazelbuild/buildtools",
-        sum = "h1:OfyUN/Msd8yqJww6deQ9vayJWw+Jrbe6Qp9giv51QQI=",
-        version = "v0.0.0-20190731111112-f720930ceb60",
+        sum = "h1:1I4g46mO4vmg8ThkDgnsDVhpK+58ob6rfweZCLQ8HQg=",
+        version = "v0.0.0-20200612171844-a60df6e5d134",
```
jayconrod, line 110: Please set this to the same version as in `go.mod`. I don't think it makes sense for these to be at different versions in Go / Bazel builds.

PR #822, opened 2020-06-25
Setting GO111MODULE=on in fetch_repo

linzhp:
**What type of PR is this?**

Bug fix

cmd/fetch_repo

When `GO111MODULE=off` externally, fetch_repo fails with errors like:

```
ERROR: An error occurred during the fetch of repository 'com_github_bazelbuild_buildtools':
  failed to fetch com_github_bazelbuild_buildtools: fetch_repo: unexpected end of JSON input
ERROR: Skipping '//:gazelle': no such package '@com_github_bazelbuild_buildtools//buildifier': failed to fetch com_github_bazelbuild_buildtools: fetch_repo: unexpected end of JSON input
```

Which give no information about the root cause. This PR print out more information from stderr of `go mod` command when there are errors and explicit turn on `GO111MODULE` before running `go mod` to be immune of external env.

### DIFF
--- a/cmd/fetch_repo/module.go
+++ b/cmd/fetch_repo/module.go
@@ -79,6 +79,7 @@ func fetchModule(dest, importpath, version, sum string) error {
 		cmd.Args = append(cmd.Args, "-modcacherw")
 	}
 	cmd.Args = append(cmd.Args, importpath+"@"+version)
+	cmd.Env = append(os.Environ(), "GO111MODULE=on")
 	cmd.Stdout = buf
 	cmd.Stderr = bufErr
 	dlErr := cmd.Run()
@@ -93,9 +94,11 @@ func fetchModule(dest, importpath, version, sum string) error {
 	// Parse the JSON output.
 	var dl struct{ Dir, Sum, Error string }
 	if err := json.Unmarshal(buf.Bytes(), &dl); err != nil {
+		_, _ = os.Stderr.Write(bufErr.Bytes())
 		return err
 	}
 	if dl.Error != "" {
+		_, _ = os.Stderr.Write(bufErr.Bytes())
 		return errors.New(dl.Error)
 	}
 	if dlErr != nil {
--- a/cmd/fetch_repo/module.go
+++ b/cmd/fetch_repo/module.go
@@ -26,6 +26,7 @@ import (
 	"os/exec"
 	"path/filepath"
 	"runtime"
+	"strings"
 )
 
 func fetchModule(dest, importpath, version, sum string) error {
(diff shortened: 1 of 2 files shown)


cmd/fetch_repo/module.go
```diff
@@ -93,9 +94,11 @@ func fetchModule(dest, importpath, version, sum string) error {
 	// Parse the JSON output.
 	var dl struct{ Dir, Sum, Error string }
 	if err := json.Unmarshal(buf.Bytes(), &dl); err != nil {
+		_, _ = os.Stderr.Write(bufErr.Bytes())
```
jayconrod, line 97: This function shouldn't directly write to stdout or stderr; we should avoid doing that above with `dlErr` (no need to make that change in this CL though).

Instead, `err` should contain the error text. For this case:

```
if err := json.Unmarshal(buf.Bytes(), &dl); err != nil {
  if bufErr.Len() > 0 {
    return fmt.Errorf("%s %s: %s", cmd.Path, strings.Join(cmd.Args, " "), bufErr.Bytes())
  } else {
    return fmt.Errorf("%s %s: %v", cmd.Path, strings.Join(cmd.Args, " "), err)
  }
}
```

cmd/fetch_repo/module.go
```diff
@@ -93,9 +94,11 @@ func fetchModule(dest, importpath, version, sum string) error {
...
 		return err
 	}
 	if dl.Error != "" {
+		_, _ = os.Stderr.Write(bufErr.Bytes())
```
jayconrod, line 101: If `go mod download` writes out JSON, it won't write anything useful on stderr (maybe download chatter). So let's drop this line.

cmd/fetch_repo/module.go
```diff
@@ -79,6 +79,7 @@ func fetchModule(dest, importpath, version, sum string) error {
 		cmd.Args = append(cmd.Args, "-modcacherw")
 	}
 	cmd.Args = append(cmd.Args, importpath+"@"+version)
+	cmd.Env = append(os.Environ(), "GO111MODULE=on")
```
jayconrod, line 82: Let's move this to `internal/go_repository.bzl` next to where `GOSUMDB` is set to `off`. All the environment variable settings should be together.

internal/go_repository.bzl
```diff
@@ -110,6 +110,8 @@ def _go_repository_impl(ctx):
     env.update({k: ctx.os.environ[k] for k in env_keys if k in ctx.os.environ})
 
     if fetch_repo_args:
+        # Override external GO111MODULE, because it is needed by module mode, no-op in repository mode
+        fetch_repo_env["GO111MODULE"] = "on"
```
jayconrod, line 114: This line needs to be after L117, since `fetch_repo_env` hasn't been defined yet.

PR #823, opened 2020-06-25
resolve: support cross-language dependency resolution, and enable go to proto resolution

### LINKED ISSUE #771: resolve: support cross-language dependency resolution in extensions
Currently, Autogazelle always passes `-index=false` to Gazelle as a way to keep Gazelle runs fast. Therefore for the dependency resolution of custom-named rules, you must use `# gazelle:resolve` directives within the repo's root bazel file.

Due to the large size of our repo and the number of rules we have that require an index to be resolved, it would not be feasible for us to maintain a top-level bazel file with all of the needed `# gazelle:resolve` directives.

I believe we could get around this issue if Gazelle was able to take in a pre-calculated rule index to use for dependency resolution. Autogazelle could generate the index during its initial run on the entire repo and then pass in the generated index to subsequent gazelle runs. Each run would also need to update the index to reflect any changes within the supplied directories.

Before digging too deep, I was wondering if this idea sounds good to you and if there are any gotchas you may be aware of?

blico:
**What type of PR is this?**

Feature

- Defines the `CrossResolver` interface, which can be implemented by language extensions in order to provide custom cross-language dependency resolution logic
- Creates `resolve.NewRuleIndexCrslv` and `resolve.FindRulesByImportCrslv` to enable the usage of cross resolvers during dependency resolution
- Implements a `language/proto` `CrossResolver`, which provides dependency resolution for the go extension using well-known type information
- Refactor `language/go`'s `Resolve` logic to depend on the proto cross resolver

Fixes #771


jayconrod:
Thanks for working on this. Just to give a heads up, this may take a little longer to review than usual. I should be able to do at least one pass early next week, but I won't have much bandwidth until the following week due to the holiday.

resolve/index.go
```diff
@@ -61,13 +61,23 @@ type Resolver interface {
 	Resolve(c *config.Config, ix *RuleIndex, rc *repo.RemoteCache, r *rule.Rule, imports interface{}, from label.Label)
 }
 
+// CrossResolver is an interface that language extensions can implement to provide
+// custom dependency resolution logic for other languages.
+type CrossResolver interface {
+	// CrossResolve attempts to resolve an import string to a rule for languages
+	// other than the implementing extension. lang is the langauge of the rule
+	// with the dependency.
+	CrossResolve(c *config.Config, ix *RuleIndex, rc *repo.RemoteCache, imp ImportSpec, lang string) []FindResult
```
jayconrod, line 70: Let's drop `rc` from here. It seems to require a `*repo.RemoteCache` in a lot of places without really providing value.

I was thinking of generalizing `repo.RemoteCache` (it's very Go-specific right now), but now I think it would be better for extensions to just keep a cache like that themselves instead of passing around a shared cache.

resolve/index.go
```diff
@@ -61,13 +61,23 @@ type Resolver interface {
...
+}
+
 // RuleIndex is a table of rules in a workspace, indexed by label and by
 // import path. Used by Resolver to map import paths to labels.
 type RuleIndex struct {
 	rules     []*ruleRecord
 	labelMap  map[label.Label]*ruleRecord
 	importMap map[ImportSpec][]*ruleRecord
 	mrslv     func(r *rule.Rule, pkgRel string) Resolver
+	crslvs    []CrossResolver
```
jayconrod, line 80: Rename to `crossResolvers`. Too much of an abbreviation I think. (`mrslv` is an unfortunate precedent).

resolve/index.go
```diff
@@ -104,6 +114,15 @@ func NewRuleIndex(mrslv func(r *rule.Rule, pkgRel string) Resolver) *RuleIndex {
 	}
 }
 
+// NewRuleIndexCrslv creates a new index with CrossResolver capabilities.
+func NewRuleIndexCrslv(mrslv func(r *rule.Rule, pkgRel string) Resolver, crslvs []CrossResolver) *RuleIndex {
```
jayconrod, line 118: I'm a little worried we'd need to introduce another new constructor if more interfaces like `CrossResolver` are added in the future.

How about instead we add a parameter `exts ...interface{}` to the original constructor, then just pass the full list of languages to that? It can pick out anything that satisfies `CrossResolver` instead of making `fix-update.go` do that.

Technically this is still a breaking change, since the function can't be assigned to a variable of the old type. Calls should still work though, so I think I'm comfortable with that.

resolve/index.go
```diff
@@ -229,6 +248,19 @@ func (ix *RuleIndex) FindRulesByImport(imp ImportSpec, lang string) []FindResult
 	return results
 }
 
+// FindRulesByImportCrslv attempts to resolve an import to a rule first by
+// checking the rule index, then if no matches are found any registered
+// CrossResolve implementations are called.
+func (ix *RuleIndex) FindRulesByImportCrslv(c *config.Config, rc *repo.RemoteCache, imp ImportSpec, lang string) []FindResult {
```
jayconrod, line 254: Let's call this `FindRulesByImportWithConfig` instead. It shouldn't need `rc` if that's dropped from `CrossResolve`.

Let's add `// Deprecated: use FindRulesByImportWithConfig instead` to the doc comment for `FindRulesByImport`. It's unfortunate that we can't consolidate them, but the `*config.Config` seems necessary for this to work.

resolve/index.go
blico: Unfortunately, `[]language.Language` cannot be passed into `resolve.NewRuleIndex` as an `interface{}` without causing an import cycle. 

Instead, I added `exts ...interace{}` to the original constructor, but `fix-update.go` is still picking out any `lang` that satisfies `CrossResolver`.

resolve/index.go
```diff
@@ -97,10 +107,17 @@ type ruleRecord struct {
 //
 // kindToResolver is a map from rule kinds (for example, "go_library") to
 // Resolvers that support those kinds.
-func NewRuleIndex(mrslv func(r *rule.Rule, pkgRel string) Resolver) *RuleIndex {
+func NewRuleIndex(mrslv func(r *rule.Rule, pkgRel string) Resolver, exts ...interface{}) *RuleIndex {
+	var crossResolvers []CrossResolver
+	for _, e := range exts {
```
jayconrod, line 112: Let's have callers pass in a slice of extensions rather than a slice of slices:

```
for _, e := range exts {
  if cr, ok := e.(CrossResolver); ok {
    crossResolvers = append(crossResolvers, cr)
  }
}
```

resolve/index.go
```diff
@@ -229,6 +248,19 @@ func (ix *RuleIndex) FindRulesByImport(imp ImportSpec, lang string) []FindResult
 	return results
 }
 
+// FindRulesByImportWithConfig attempts to resolve an import to a rule first by
+// checking the rule index, then if no matches are found any registered
+// CrossResolve implementations are called.
+func (ix *RuleIndex) FindRulesByImportWithConfig(c *config.Config, imp ImportSpec, lang string) []FindResult {
+	results := ix.FindRulesByImport(imp, lang)
+	if len(results) == 0 {
```
jayconrod, line 256: Let's flip this condition and return early so we don't have to indent the loop:

```
if len(results) > 0 {
  return results
}
for _, cr := ...
```

cmd/gazelle/fix-update.go
```diff
@@ -243,15 +243,19 @@ func runFixUpdate(cmd command, args []string) (err error) {
...
 	for _, lang := range languages {
 		cexts = append(cexts, lang)
 		for kind, info := range lang.Kinds() {
 			mrslv.AddBuiltin(kind, lang)
 			kinds[kind] = info
 		}
 		loads = append(loads, lang.Loads()...)
+		if c, ok := lang.(resolve.CrossResolver); ok {
+			crossResolvers = append(crossResolvers, c)
+		}
```
jayconrod, line 256: We shouldn't need to check the `CrossResolver` interface here. Let's just put all extensions in a `[]interface{}`:

```
exts := make([]interface{}, 0, len(languages))
  ...
  exts = append(exts, lang)
}
```

In hindsight, I wish it were possible for authors to write extensions that don't implement `language.Language`; they could register extensions that implement any subset of supported interfaces like `config.Configurer` or `resolve.CrossResolver`.

At least in this case, we can pass extensions to `resolve` without worrying what interfaces are implemented or might be added in the future.

PR #825, opened 2020-06-29
fix-update: only write changed build files to disk 

### LINKED ISSUE #807: fix-update: only write changed build files to disk
Currently, when `mode=fix` Gazelle unconditionally writes all build files that it visits to disk. Given the size of our repository (~30,000 build files), if Gazelle only writes changed build files back to disk, it would result in a significant performance improvement. A quick prototype shows this would speed up Gazelle runs from the root of our repository by ~20%.

Gazelle already reads every build file it visits into memory, so to implement we could simply cache the original contents of each file and then do a comparison with the final contents before `fixFile` writes out to disk.

If not all users want this feature enabled by default, we could potentially make a new `mode` for it (e.g. `fixOnChange`).

blico:
**What type of PR is this?**

Feature

Adds the `Content` field to `rule.File`, which contains the file's underlying disk content. `file.Content` is then used by `gazelle/fix` and `gazelle/diff` to avoid unnecessary file operations.

Fixes #807

### DIFF
--- a/cmd/gazelle/fix.go
+++ b/cmd/gazelle/fix.go
@@ -19,15 +19,24 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"reflect"
 
 	"github.com/bazelbuild/bazel-gazelle/config"
 	"github.com/bazelbuild/bazel-gazelle/rule"
 )
 
 func fixFile(c *config.Config, f *rule.File) error {
+	newContent := f.Format()
+	if reflect.DeepEqual(f.Content, newContent) {
+		return nil
+	}
 	outPath := findOutputPath(c, f)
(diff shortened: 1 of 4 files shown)


cmd/gazelle/fix.go
```diff
@@ -19,15 +19,24 @@ import (
...
 	"path/filepath"
+	"reflect"
 
 	"github.com/bazelbuild/bazel-gazelle/config"
 	"github.com/bazelbuild/bazel-gazelle/rule"
 )
 
 func fixFile(c *config.Config, f *rule.File) error {
+	newContent := f.Format()
+	if reflect.DeepEqual(f.Content, newContent) {
```
jayconrod, line 30: Use `bytes.Equal` instead of `reflect.DeepEqual` for `[]byte`.

Issue #827, opened 2020-07-03
Support transitive runfiles from language plugins


greenboxal:
### What version of gazelle are you using?

```
http_archive(
    name = "bazel_gazelle",
    sha256 = "bfd86b3cbe855d6c16c6fce60d76bd51f5c8dbc9cfcaef7a2bb5c1aafd0710e8",
    urls = [
        "https://mirror.bazel.build/github.com/bazelbuild/bazel-gazelle/releases/download/v0.21.0/bazel-gazelle-v0.21.0.tar.gz",
        "https://github.com/bazelbuild/bazel-gazelle/releases/download/v0.21.0/bazel-gazelle-v0.21.0.tar.gz",
    ],
)
```

### What version of rules_go are you using?

```
http_archive(
    name = "io_bazel_rules_go",
    sha256 = "87f0fb9747854cb76a0a82430adccb6269f7d394237104a4523b51061c469171",
    urls = [
        "https://mirror.bazel.build/github.com/bazelbuild/rules_go/releases/download/v0.23.1/rules_go-v0.23.1.tar.gz",
        "https://github.com/bazelbuild/rules_go/releases/download/v0.23.1/rules_go-v0.23.1.tar.gz",
    ],
)
```

### What version of Bazel are you using?

```
Build label: 3.3.0
Build target: bazel-out/darwin-opt/bin/src/main/java/com/google/devtools/build/lib/bazel/BazelServer_deploy.jar
Build time: Wed Jun 17 12:59:56 2020 (1592398796)
Build timestamp: 1592398796
Build timestamp as int: 1592398796
```

### Does this issue reproduce with the latest releases of all the above?

Yes.

### What did you do?

Custom languages should be able to depend on other tools, for example, a parser that can extract dependencies from source code. This is especially useful as it's unlikely a Go parser would exist, not mentioning compatibility issues.

Specifying `data` in `go_library` should be enough so runfiles are propagated to the gazelle runner:

```
# language/js/BUILD
go_library(
    name = "go_default_library",
    srcs = [
        "builtin.go",
        "config.go",
        "fileinfo.go",
        "generate.go",
        "generator.go",
        "language.go",
        "resolve.go",
        "resolver.go",
        "utils.go",
        "worker.go",
    ],
    data = [
        "@brex_rules_js//packages/import-parser",
    ],
    importpath = "github.com/brexhq/brex_rules_js/gazelle/language/js",
    visibility = ["//visibility:public"],
    deps = [
[...]

PR #830, opened 2020-07-14
gazelle rule: simplify runner script

jayconrod:
This change removes several parts of the bash script generated by the
gazelle rule. The script can no longer be run without 'bazel
run'. It's now simple enough that it can probably be translated into a
batch script on Windows. Unfortunately, it's not possible to eliminate
it completely, since we still need to pass in arguments to a Gazelle
binary built by another rule.

The runner script no longer changes directories, and Gazelle no longer
absolutizes its arguments based on its working directory when invoked
by 'bazel run'. Instead, Gazelle joins the environment variable
BUILD_WORKSPACE_DIRECTORY (when set by 'bazel run') with relative path
arguments. This should allow Gazelle extensions to find runfiles.

For #827


jayconrod:
cc @greenboxal

rohansingh:
I think this might have broken the ability to invoke `update-repos` from the Gazelle rule:

```
$ bazel run //:gazelle -- update-repos -from_file=go.mod
INFO: Analyzed target //:gazelle (0 packages loaded, 0 targets configured).
INFO: Found 1 target...
Target //:gazelle up-to-date:
  bazel-bin/gazelle-runner.bash
  bazel-bin/gazelle
INFO: Elapsed time: 0.132s, Critical Path: 0.00s
INFO: 0 processes.
INFO: Build completed successfully, 1 total action
INFO: Build completed successfully, 1 total action
gazelle: open go.mod: no such file or directory
```

jayconrod:
@rohansingh Thanks for reporting. I'll revert this PR for now. Won't have time to fix it for a while unfortunately.

PR #832, opened 2020-07-15
language/go: export GetGoConfig, GoConfig, GoConfig.Prefix, GoConfig.PrefixRel

blico:
Export `GoConfig` and related fields/methods so that they can be used by other language extensions. 

Has to do with https://github.com/bazelbuild/bazel-gazelle/issues/771#issuecomment-642127471, about having an extension which automatically updates a list of `resolve` directives.

### DIFF
--- a/language/go/config.go
+++ b/language/go/config.go
@@ -40,8 +40,8 @@ import (
 
 var minimumRulesGoVersion = version.Version{0, 20, 0}
 
-// goConfig contains configuration values related to Go rules.
-type goConfig struct {
+// GoConfig contains configuration values related to Go rules.
+type GoConfig struct {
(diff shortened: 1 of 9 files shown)


jayconrod:
Could you say more about how this will be used?

In general, `GoConfig` is something that could easily change in incompatible ways. Exporting it will make changes more difficult. The `Prefix` and `PrefixRel` fields are less likely to change, so I don't think this change is totally off the table, but there must be a good reason.

It might be better instead to just keep track of the `prefix` directive in your extension's `Configure` method. A little bit of code copying may be better than tight coupling.

blico:
Gah, it didn't even cross my mind that I can just keep track of the `prefix` directive in my extension's `Configure` method. Sorry about the added noise. Closing this.

PR #842, opened 2020-07-17
language/go: several naming convention fixes plus auto-detection

jayconrod:
* For external dependency resolution, if we don't know what naming
  convention the external repository uses (for example, because
  there's no known repository), we'll now use
  goDefaultLibraryNamingConvention. This avoids assuming that
  repositories with build files fetched with http_archive have been
  updated.
* In migrateNamingConvention, print a warning and avoid renaming if
  there's already a target with the new name.
* Library, test, and alias actuals are now generated based on import
  path instead of package name.
* Added unknownNamingConvention, a new zero value.
* Added detectNamingConvention. It reads the root build file and build
  files in subdirectories one level deep to infer the naming
  convention used if one is not specified in the root build
  file. Defaults to importNamingConvention.
* Fixed tests.

For #5

### DIFF
--- a/cmd/gazelle/diff_test.go
+++ b/cmd/gazelle/diff_test.go
@@ -55,7 +55,7 @@ func TestDiffExisting(t *testing.T) {
  # gazelle:prefix example.com/hello
  
 +go_library(
-+    name = "go_default_library",
++    name = "hello",
 +    srcs = ["hello.go"],
 +    importpath = "example.com/hello",
 +    visibility = ["//visibility:public"],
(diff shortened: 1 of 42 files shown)


jayconrod:
cc @tomlu 

I made some progress today, but this isn't quite ready yet. Automatic detection breaks most of the tests, and I still need to fix the integration tests. Will pick it up next week.

Still to do after this (not in this PR):

* Convert Gazelle repo to new convention.
* Make sure we can smoothly switch back to the old convention.
* Test that library targets are deleted after a main package is removed. I think only `foo` will be removed, not `foo_lib`.

Issue #847, opened 2020-07-27
Add `netrc` (like that of `http_archive`) to `go_repository`?


qzmfranklin:
There are cases where I need to use `go_repository` to instantiate repositories from servers that require login credentials.

With plain `http_archive`, there is [`netrc`](https://docs.bazel.build/versions/master/repo/http.html#http_archive-netrc). I'd like something similar to that with `go_repository`.

PR #848, opened 2020-07-28
feat: go_repository supports netrc

qzmfranklin:
**What type of PR is this?**
Feature

go_repository

This PR adds the `netrc` and `auth_patterns` attributes to the `go_repository` rule.

This allows `go_repository` to be used with downloading servers that require login  credentials.

This also makes `go_repository` more closely resemble the interface of `http_archive`.

Fixes https://github.com/bazelbuild/bazel-gazelle/issues/847

**Other notes for review**

This patch was largely the result of directly pirating from the source code of `http_archive`. It may not be the most beautiful code ever.

This patch has been tested on my local machine with a custom http server that requires login. I successfully re-used the same netrc file for `go_repository` and curl (with --netrc-file).

If more tests are needed, please let me know.

### DIFF
--- a/internal/go_repository.bzl
+++ b/internal/go_repository.bzl
@@ -12,11 +12,36 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+load(
+    "@bazel_tools//tools/build_defs/repo:utils.bzl",
+    "read_netrc",
+    "use_netrc",
+)
 load("@io_bazel_rules_go//go/private:common.bzl", "env_execute", "executable_extension")
 load("@bazel_gazelle//internal:go_repository_cache.bzl", "read_cache_env")
 
 # We can't disable timeouts on Bazel, but we can set them to large values.
 _GO_REPOSITORY_TIMEOUT = 86400
+
+def _get_auth(ctx, urls):
+    """Given the list of URLs obtain the correct auth dict."""
+    if ctx.attr.netrc:
+        netrc = read_netrc(ctx, ctx.attr.netrc)
+        return use_netrc(netrc, urls, ctx.attr.auth_patterns)
+
+    if "HOME" in ctx.os.environ and not ctx.os.name.startswith("windows"):
+        netrcfile = "%s/.netrc" % (ctx.os.environ["HOME"])
+        if ctx.execute(["test", "-f", netrcfile]).return_code == 0:
+            netrc = read_netrc(ctx, netrcfile)
+            return use_netrc(netrc, urls, ctx.attr.auth_patterns)
+
+    if "USERPROFILE" in ctx.os.environ and ctx.os.name.startswith("windows"):
+        netrcfile = "%s/.netrc" % (ctx.os.environ["USERPROFILE"])
+        if ctx.path(netrcfile).exists:
+            netrc = read_netrc(ctx, netrcfile)
+            return use_netrc(netrc, urls, ctx.attr.auth_patterns)
+
+    return {}
 
 def _go_repository_impl(ctx):
     # TODO(#549): vcs repositories are not cached and still need to be fetched.
@@ -28,11 +53,13 @@ def _go_repository_impl(ctx):
         for key in ("commit", "tag", "vcs", "remote", "version", "sum", "replace"):
             if getattr(ctx.attr, key):
                 fail("cannot specifiy both urls and %s" % key, key)
+        auth = _get_auth(ctx, ctx.attr.urls)
         ctx.download_and_extract(
             url = ctx.attr.urls,
             sha256 = ctx.attr.sha256,
             stripPrefix = ctx.attr.strip_prefix,
             type = ctx.attr.type,
+            auth = auth,
         )
     elif ctx.attr.commit or ctx.attr.tag:
         # repository mode
@@ -214,6 +241,8 @@ go_repository = repository_rule(
         "strip_prefix": attr.string(),
         "type": attr.string(),
         "sha256": attr.string(),
+        "netrc": attr.string(),
+        "auth_patterns": attr.string_dict(),
 
         # Attributes for a module that should be downloaded with the Go toolchain.
         "version": attr.string(),


qzmfranklin:
@jayconrod 

I'm a bit surprised that this use case has not been covered yet.

I ran into this problem because I had a personal git server that requires http login-password credentials for downloading archives.

The `netrc` file I used is like this:
```
machine git.corp.internaldomain.com
login myusername
password mypassword
```

I used this netrc file with `curl --netrc-file`, the `requests` python library, the `http_archive` rule, and this patched `go_repository` rule. They all yield the same and expected results.

internal/go_repository.bzl
```diff
@@ -12,12 +12,37 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+load(
+    "@bazel_tools//tools/build_defs/repo:utils.bzl",
+    "read_netrc",
+    "use_netrc",
```
jayconrod, line 18: Could you confirm that Bazel 1.2.0 provides these functions, and if not, what is the oldest version of Bazel that does? I'd like to make sure we're not breaking compatibility here. If these were added recently, it would be better to copy them here.

internal/go_repository.bzl
```diff
@@ -12,12 +12,37 @@
...
 
+def _get_auth(ctx, urls):
+    """Given the list of URLs obtain the correct auth dict."""
+    if ctx.attr.netrc:
+        netrc = read_netrc(ctx, ctx.attr.netrc)
+        return use_netrc(netrc, urls, ctx.attr.auth_patterns)
+
+    if "HOME" in ctx.os.environ and not ctx.os.name.startswith("windows"):
+        netrcfile = "%s/.netrc" % (ctx.os.environ["HOME"])
+        if ctx.execute(["test", "-f", netrcfile]).return_code == 0:
```
jayconrod, line 34: Why use `test` here but `ctx.path(netrcfile).exists` below? Seems like the latter should work in both cases.

internal/go_repository.bzl
```diff
@@ -12,12 +12,37 @@
...
 load("@io_bazel_rules_go//go/private:common.bzl", "env_execute", "executable_extension")
 load("@bazel_gazelle//internal:go_repository_cache.bzl", "read_cache_env")
 
 # We can't disable timeouts on Bazel, but we can set them to large values.
 _GO_REPOSITORY_TIMEOUT = 86400
 
+def _get_auth(ctx, urls):
+    """Given the list of URLs obtain the correct auth dict."""
+    if ctx.attr.netrc:
+        netrc = read_netrc(ctx, ctx.attr.netrc)
```
jayconrod, line 29: To avoid repetition, please separate the code that decides which file to read from the calls to `read_netrc` and `use_netrc`.

internal/go_repository.bzl
```diff
@@ -12,12 +12,37 @@
...
+        return use_netrc(netrc, urls, ctx.attr.auth_patterns)
+
+    if "HOME" in ctx.os.environ and not ctx.os.name.startswith("windows"):
```
jayconrod, line 32: Before looking for a `.netrc` file in the home directory, check the `NETRC` environment variable.

internal/go_repository.bzl
```diff
@@ -12,12 +12,37 @@
...
+        netrcfile = "%s/.netrc" % (ctx.os.environ["HOME"])
+        if ctx.execute(["test", "-f", netrcfile]).return_code == 0:
+            netrc = read_netrc(ctx, netrcfile)
+            return use_netrc(netrc, urls, ctx.attr.auth_patterns)
+
+    if "USERPROFILE" in ctx.os.environ and ctx.os.name.startswith("windows"):
+        netrcfile = "%s/.netrc" % (ctx.os.environ["USERPROFILE"])
```
jayconrod, line 39: On Windows, it should be `_netrc`, not `.netrc`.

internal/go_repository.bzl
```diff
@@ -214,6 +241,8 @@ go_repository = repository_rule(
         "strip_prefix": attr.string(),
         "type": attr.string(),
         "sha256": attr.string(),
+        "netrc": attr.string(),
+        "auth_patterns": attr.string_dict(),
```
jayconrod, line 245: Please add documentation to `go_repository.bzl` for these attributes.

internal/go_repository.bzl
```diff
@@ -12,12 +12,37 @@
...
+    "read_netrc",
+    "use_netrc",
+)
 load("@io_bazel_rules_go//go/private:common.bzl", "env_execute", "executable_extension")
 load("@bazel_gazelle//internal:go_repository_cache.bzl", "read_cache_env")
 
 # We can't disable timeouts on Bazel, but we can set them to large values.
 _GO_REPOSITORY_TIMEOUT = 86400
 
+def _get_auth(ctx, urls):
```
jayconrod, line 26: I expect folks will want to have just one `netrc` file shared by all rules, so let's add these attributes to `go_repository_config` as well and make sure they can be read by `go_repository` if they're set there.

internal/go_repository.bzl
```diff
@@ -28,11 +53,13 @@ def _go_repository_impl(ctx):
         for key in ("commit", "tag", "vcs", "remote", "version", "sum", "replace"):
             if getattr(ctx.attr, key):
                 fail("cannot specifiy both urls and %s" % key, key)
+        auth = _get_auth(ctx, ctx.attr.urls)
         ctx.download_and_extract(
             url = ctx.attr.urls,
             sha256 = ctx.attr.sha256,
             stripPrefix = ctx.attr.strip_prefix,
             type = ctx.attr.type,
+            auth = auth,
```
jayconrod, line 62: It looks like this will only work for HTTP mode. It also needs to work for module mode though (the `ctx.attr.version` branch).

I think that just means making sure `NETRC` is set to the correct file in `env_keys`, but please verify.

dmivankov:
Does seem to work with Bazel 3.3.1 on github private repos like this
```py
go_repository(
    name = "...",
    auth_patterns = {
        "api.github.com": "Bearer <password>",
    },
    importpath = "github.com/org/repo",
    strip_prefix = "org-repo-commit_here",
    type = "zip", # zipball -> tarball and zip -> tar should also work
    urls = ["https://api.github.com/repos/org/repo/zipball/commit_here"],
)
```
plus ~/.netrc
```
machine api.github.com
  password github_personal_token
  login anything
```

linzhp:
Superseded by #1090

PR #855, opened 2020-08-03
Report error when multiple rules with the same name

### LINKED ISSUE #850: merger: report error when generated rule matches multiple rules with the same name
Gazelle [detected the duplicate names](https://sourcegraph.com/github.com/bazelbuild/bazel-gazelle@30aa1d7c9406aaf0838781e907eefc4bc2d1eddc/-/blob/merger/merger.go#L220-L221) in the old file while merging rules, but the error was [ignored](https://sourcegraph.com/github.com/bazelbuild/bazel-gazelle@30aa1d7c9406aaf0838781e907eefc4bc2d1eddc/-/blob/merger/merger.go#L130). The new version of the rule is subsequently thrown away. Gazelle still finishes successfully. This is undesirable behavior because no two rules in the same file should have the same name. For regular build rules, Bazel doesn't not allow that. Although Bazel allows repository rules with the same name, Gazelle should not allow it for files/macros its updating.

If it's too chatty to report all merge errors, as the comment suggested, we can create a special error type for duplicate name, of which we are more certain about the validity, and properly handle this error type.

4d8b5cf7fbc9271f9fbb6882a3d19df408c71299

v0.23.3

3.3.1

yes

Darwin amd64

### What did you do?

* Create a copy of a `go_repository` rule and put it in the same file as the original rule
* Update the version of that Go repository in `go.mod` and run `gazelle update-repos`

At least one of the `go_repository` rule is updated

### What did you see instead?

Nothing updated

linzhp:
**What type of PR is this?**

Bug fix

merger

Created a new error type for duplicate rule names in the same file. Checking that specific error type when merging files

Fixes #850

**Other notes for review**

### DIFF
--- a/merger/merger.go
+++ b/merger/merger.go
@@ -37,6 +37,7 @@ limitations under the License.
 package merger
 
 import (
+	"errors"
 	"fmt"
 	"sort"
 	"strings"
@@ -96,7 +97,7 @@ const (
 // If an attribute is marked with a "# keep" comment, it will not be merged.
 // If a rule is marked with a "# keep" comment, the whole rule will not
 // be modified.
-func MergeFile(oldFile *rule.File, emptyRules, genRules []*rule.Rule, phase Phase, kinds map[string]rule.KindInfo) {
+func MergeFile(oldFile *rule.File, emptyRules, genRules []*rule.Rule, phase Phase, kinds map[string]rule.KindInfo) error {
 	getMergeAttrs := func(r *rule.Rule) map[string]bool {
 		if phase == PreResolve {
 			return kinds[r.Kind()].MergeableAttrs
@@ -126,6 +127,9 @@ func MergeFile(oldFile *rule.File, emptyRules, genRules []*rule.Rule, phase Phas
 	substitutions := make(map[string]string)
 	for i, genRule := range genRules {
 		oldRule, err := Match(oldFile.Rules, genRule, kinds[genRule.Kind()])
+		if err == duplicateErr {
+			return fmt.Errorf("could not match %s(%s) in %s: %v", genRule.Kind(), genRule.Name(), oldFile.Path, err)
+		}
 		if err != nil {
 			// TODO(jayconrod): add a verbose mode and log errors. They are too chatty
 			// to print by default.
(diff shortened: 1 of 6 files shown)


merger/merger.go
```diff
@@ -96,7 +97,7 @@ const (
 // If an attribute is marked with a "# keep" comment, it will not be merged.
 // If a rule is marked with a "# keep" comment, the whole rule will not
 // be modified.
-func MergeFile(oldFile *rule.File, emptyRules, genRules []*rule.Rule, phase Phase, kinds map[string]rule.KindInfo) {
+func MergeFile(oldFile *rule.File, emptyRules, genRules []*rule.Rule, phase Phase, kinds map[string]rule.KindInfo) error {
```
jayconrod, line 100: This would be an incompatible change to an exported function. Let's avoid that if we can.

Perhaps instead, `rule.LoadData`, `rule.LoadWorkspaceData`, and `rule.LoadMacroData` can call some common function, `func checkFile(f *File) error` that returns an error if there are multiple rules with the same name (we could conceivably check more invariants later).

I think this makes more sense than checking for duplicates in `MergeFile` anyway; arguably this is a pre-condition that `MergeFile` shouldn't have to worry about.

rule/rule.go
```diff
@@ -145,6 +145,9 @@ func LoadData(path, pkg string, data []byte) (*File, error) {
 		return nil, err
 	}
 	f := ScanAST(pkg, ast)
+	if err = checkFile(f); err != nil {
```
jayconrod, line 148: Nit: `if err := checkFile(f); err != nil`, here and below.

Using `=` makes it look like there's a subtle reason not to shadow `err`, but since that's not the case, it's better to use `:=`.

rule/rule.go
```diff
@@ -903,3 +912,17 @@ func (s byAttrName) Less(i, j int) bool {
 func (s byAttrName) Swap(i, j int) {
 	s[i], s[j] = s[j], s[i]
 }
+
+func checkFile(f *File) error {
+	names := make(map[string]bool)
+	for _, r := range f.Rules {
+		if r.Name() == "" {
```
jayconrod, line 919: Let's extract `r.Name()` into a variable. `rule.Name` is not a simple getter, so we should avoid calling it multiple times if we can.

rule/rule.go
```diff
@@ -903,3 +912,17 @@ func (s byAttrName) Less(i, j int) bool {
...
+			continue
+		}
+		if names[r.Name()] {
+			return fmt.Errorf("multiple rules have the name %q in %s", r.Name(), f.Path)
```
jayconrod, line 923: Nit: reorder to `%s: multiple rules have the same name %q`

Issue #499, opened 2019-04-18
strict mode: exit non-zero for multiple packages and other errors


jasharpe:
Being used to manual Bazel BUILD files, I wasn't aware of the Gazelle restriction that you couldn't have multiple packages in the same directory.

If you have this, you get a message like this from Gazelle:

```
gazelle: found packages foo (foo.go) and bar (bar.go) in <some directory>
```

Nowhere does it state that this is a problem (i.e. with words like "error" or "failure"), so I assumed this was just a helpful info message, and was confused why my BUILD.bazel file was not being changed.

This is further confused by the fact that when you have this failure, the exit code of Gazelle is 0, so it appears that everything is fine:

```
$ bazel run //:gazelle && echo "everything is fine!"
...
gazelle: found packages foo (foo.go) and bar (bar.go) in <some directory>
everything is fine!
```

If some part of the tree couldn't be updated it would be nice for this not to be treated as a success.

At the very least including some language in the message indicating that it's a problem would be really nice.

I'd send a PR myself but I can't find the "found packages" message anywhere in the code of this repository so I'm not sure where this message is actually generated. :woman_shrugging:

jasharpe:
Ah, it's here: https://golang.org/src/go/build/build.go?s=15775:15809#L473

jayconrod:
This is kind of working as intended, but there's room for improvement.

Gazelle needs to run in a lot of situations with very little intervention from the user. `go_repository` is a major case. Treating errors like this as fatal in those situations is rarely desirable since these come up a lot, even in the `golang.org/x/` repos. We frequently see directories with tools or source files that are intended to be built manually (`go build foo.go bar.go`) or not built at all (`testdata`).

There's been some discussion in the past about a "strict mode" where we'd treat problems like this as hard errors. It would probably be enabled by default when Gazelle is run in a working repository but not in `go_repository`. Would that solve this for you?

(Note: I won't have bandwidth to work on this or even review large PRs for a while. It would be good to have a planned solution on the record though).

jasharpe:
I think some kind of strict mode would work, enabled by default for working directory.

I think this wouldn't be a problem for anyone who is versed in external Go, but coming from inside Google (in this case trying to open source something) where multiple packages per directory is common, it's very confusing for a newbie. :)

jayconrod:
Per #700 and #622, in strict mode, Gazelle should fail with an error instead of going out to the network to resolve unknown imports.

[...]

PR #859, opened 2020-08-06
Add flag to suppress non-error output.

amscanne:
**What type of PR is this?**

> Other

^^ Minor usability fix.

> all

This PR adds a "quiet" option which eliminates the resolve spam. For an example of this issue, see:

https://source.cloud.google.com/results/invocations/77d52c85-2987-4701-b301-1e341c11fddd/log

N/A - minor usability fix, can be discussed in the request.

### DIFF
--- a/cmd/gazelle/fix-update.go
+++ b/cmd/gazelle/fix-update.go
@@ -83,6 +83,7 @@ func (ucr *updateConfigurer) RegisterFlags(fs *flag.FlagSet, cmd string, c *conf
 	fs.StringVar(&uc.patchPath, "patch", "", "when set with -mode=diff, gazelle will write to a file instead of stdout")
 	fs.Var(&gzflag.MultiFlag{Values: &ucr.knownImports}, "known_import", "import path for which external resolution is skipped (can specify multiple times)")
 	fs.StringVar(&ucr.repoConfigPath, "repo_config", "", "file where Gazelle should load repository configuration. Defaults to WORKSPACE.")
+	fs.BoolVar(&c.Quiet, "quiet", false, "when true, additional debug output during resolution is suppressed")
 }
 
 func (ucr *updateConfigurer) CheckFlags(fs *flag.FlagSet, c *config.Config) error {
--- a/config/config.go
+++ b/config/config.go
@@ -94,6 +94,10 @@ type Config struct {
 	// extensions as well. Values in here may be populated by command line
 	// arguments, directives in build files, or other mechanisms.
 	Exts map[string]interface{}
+
+	// Quiet indicates that non-error output should be suppressed; this does
+	// not change any underlying operation.
+	Quiet bool
 }
 
 // MappedKind describes a replacement to use for a built-in kind.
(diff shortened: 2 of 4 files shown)


jayconrod:
Sorry, I'd rather not add a flag for this. I agree there's a lot of unhelpful output here though, and it would be better to fix that unconditionally.

#499 is a suggestion to change how dependency resolution works inside `go_repository`. That seems to be the main thing in this log. Specifically, Gazelle would only resolve external dependencies based on a given configuration (normally based on `WORKSPACE`) and would not go out to the network to resolve dependencies.

amscanne:
I'm not sure that will work. Gazelle is generating a BUILD file for the entire repository, even though we may only be using a small portion of the tree. Consider this:

github.com/org/megapkg/api/... <- just stdlib dependencies.
github.com/org/megapkg/cmd/... <- has 10,000 indirect dependencies.

We need to declare a go_repository rule for this package. At the go_repository level, gazelle is not aware that we only intend to use /api/. If it can't build the BUILD file for megapkg/cmd does it fail the entire go_repository rule? It's would be a non-starter for us to enumerate the huge number of indirect (and constantly shifting) dependencies from /cmd/. Unfortunately there's a decent number of packages that look like this (example [1]).

If you do make it an error, then I think that gazelle will need to accept a specific set of packages that it would generate BUILD files for. If that can happen, then the plan would work fine for us.

Alternately, can gazelle just pull in go.mod dependencies by farming out to the tool directly? :) The reason for this spam is that gazelle is not properly supporting the go.mod "replace" directive when traversing recursively (I believe that it is only applying on the actual go_repository rule?).

[1] https://github.com/containerd/cri

jayconrod:
> Gazelle is generating a BUILD file for the entire repository, even though we may only be using a small portion of the tree.

That's fair. The main thing is that Gazelle shouldn't go out to the network in `go_repository`. It should either resolve external dependencies locally, or it should fail. The exactly functionality (and whether there will be a flag) is TBD though.

> If you do make it an error, then I think that gazelle will need to accept a specific set of packages that it would generate BUILD files for. If that can happen, then the plan would work fine for us.

I'd rather not support that, since it requires too much configuration. In cases that require a lot of customization, I usually recommend generating build files offline, then providing them to `go_repository` or `http_archive` as a patch. Writing a custom repository rule is also an option.

> Alternately, can gazelle just pull in go.mod dependencies by farming out to the tool directly? :) The reason for this spam is that gazelle is not properly supporting the go.mod "replace" directive when traversing recursively (I believe that it is only applying on the actual go_repository rule?).

The way this is intended to work is that before any `go_repository` rule runs, Gazelle generates a configuration file based on `WORKSPACE` (or files loaded from `WORKSPACE`, declared with `# gazelle:repository_macro file.bzl%function` that contains all the `go_repository` declarations. That provides a mapping between module paths and Bazel workspace names. When Gazelle runs within a `go_repository`, if an imported package is covered by any module in that mapping, Gazelle doesn't need to look it up.

`WORKSPACE` should declare everything needed to build packages in the main workspace. As you point out though, it may not always contain all dependencies of other workspaces. Currently, it should contain everything if you import `go.mod`, but that may not always be the case in the future.

PR #864, opened 2020-08-10
fastwalk: use godirwalk to speed up filesystem walk

robfig:
This change addresses this note in the //walk/walk.go:

		// TODO: OPT: ReadDir stats all the files, which is slow. We just care about
		// names and modes, so we should use something like
		// golang.org/x/tools/internal/fastwalk to speed this up.
		files, err := ioutil.ReadDir(dir)

godirwalk provides a drop-in replacement to ioutil.ReadDir that skips calling
stat on every file in the directory. Since we only need the mode bits and the
path, this is an easy update.

To finalize this change, I performed the following steps:

	go mod vendor
	go run internal/list_repository_tools_srcs.go -dir `pwd` -generate internal/go_repository_tools_srcs.bzl

(I just guessed that those steps were necessary by mimicking what I saw in the repo -- I didn't see an official process documented anywhere)

**What type of PR is this?**

Optimization

walk

Addressing an optimization note in the code

No issue filed, since the change was small & simple.

**Other notes for review**

New dependency

x/tools/internal/fastwalk was the starting point, but it is not usable as-is (being an internal package), and Gazelle's usage does not easily fit into the Walk function, since it expects to process a list of files in a directory at a time. godirwalk appeared to be the most popular option for fastwalk-like functionality and provided a drop-in replacement for ioutil.ReadDir.

Performance improvement measurement

I'm not sure how to accurately measure changes to filesystem code -- there seems to be a lot of variation from a cold run to a repeat run, even after running `purge` to purge filesystem caches. (Tested on Darwin)

With those caveats, I did *not* see a noticeable speed improvement on a 15-25s Gazelle run. I'm not sure why, perhaps my test case does not have enough files. 

It's worth considering whether this optimization is worthwhile or not if it doesn't make a noticeable difference.

Walk Ordering

I chose to sort the output of `godirwalk.ReadDirents` to mimic the order that it would have been returned by ioutil.ReadDir, since I did not completely understand the ordering requirements after glancing at the unit tests. This may be worth revisiting if this PR is deemed worth pursuing.

### DIFF
--- a/cmd/gazelle/BUILD.bazel
+++ b/cmd/gazelle/BUILD.bazel
@@ -51,6 +51,7 @@ go_library(
         "//rule:go_default_library",
         "//walk:go_default_library",
         "@com_github_bazelbuild_buildtools//build:go_default_library",
+        "@com_github_karrick_godirwalk//:go_default_library",
         "@com_github_pmezard_go_difflib//difflib:go_default_library",
     ],
 )
--- a/cmd/gazelle/fix-update.go
+++ b/cmd/gazelle/fix-update.go
@@ -36,6 +36,7 @@ import (
 	"github.com/bazelbuild/bazel-gazelle/resolve"
 	"github.com/bazelbuild/bazel-gazelle/rule"
 	"github.com/bazelbuild/bazel-gazelle/walk"
+	"github.com/karrick/godirwalk"
 )
 
 // updateConfig holds configuration information needed to run the fix and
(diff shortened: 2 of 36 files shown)


blico:
Just to add, I also considered this same optimization but similarly, my prototype showed no noticeable speed improvements on Gazelle runs within our repo (which is fairly large).

jayconrod:
I'd rather not take on a new dependency for this.

Adding `doublestar` worked out pretty poorly, so at this point before adding a new dependency, I'd like to audit the code and make sure it efficiently does what we need. `godirwalk` seems very large for what it is. `golang.org/x/tools/internal/fastwalk` is a lot smaller, so I think copying `fastwalk` into an internal directory would likely be a better solution.

We should make sure there is a significant performance improvement here, too, before doing anything. Running `filepath.Walk` on one thread is nice and simple.

robfig:
`fastwalk.Walk` is not really helpful -- we need a `fastwalk.ReadDir()`. I'm not sure how difficult it would be to add that. 

Regardless, it sounds like there are two data points that suggest this is not a significant improvement, so I'm inclined to abandon this change.

PR #866, opened 2020-08-10
walk/config: use path.Match unless exclude pattern contains **

### LINKED ISSUE #786: Possible optimization: doublestar.Match memory usage
Gazelle takes about 20s to run on our repository, and I was curious if there were some low-hanging optimizations. I'm just sharing this tidbit; I'm not sure if it's worth acting upon.

Gazelle allocates ~900MB of memory, and about half of that  ( > 400MB) is done by the single line:
https://github.com/bazelbuild/bazel-gazelle/blob/master/walk/config.go#L52

Drilling down, it's this one line:
https://github.com/bmatcuk/doublestar/blob/master/doublestar.go#L51

I've attached screenshots from the profile at the bottom.

It feels like the sort of thing that could be improved, although I'm not sure how much it really matters to the wall time. Rewriting the doublestar match code to avoid that looks difficult, but maybe a quick fix would be to only use doublestar if a "**" is detected in the exclude? 

I can put up a PR for that if you'd like.

Screen Shot 2020-05-18 at 9 04 32 PM
Screen Shot 2020-05-18 at 8 55 04 PM

robfig:
Exclude matching is called very often and doublestar.Match allocates about 400 MB
(~50% of the total) in our work repository.

This is the simplest change possible to use path.Match unless the pattern makes
use of the "**" glob.

Fixes #786 

I looked at making matchers at configuration time, e.g. 
```go
type matcher func(filename string) bool 

type walkConfig struct {
	excludes []matcher
	ignore   bool
	follow   []string
}
```

.. but I didn't see a good way of updating this code for that approach:

```go
func (_ *Configurer) RegisterFlags(fs *flag.FlagSet, cmd string, c *config.Config) {
	wc := &walkConfig{}
	c.Exts[walkName] = wc
	fs.Var(&gzflag.MultiFlag{Values: &wc.excludes}, "exclude", "pattern that should be ignored (may be repeated)")
}
```

.. and the current PR is not pretty, but not terrible either.


robfig:
Sorry, I'm not following your reasoning. 

Gazelle recently changed from always using path.Match to always using doublestar.Match, which changed how all exclude patterns were processed. I presume no issue was raised because doublestar's intention is to exactly mimic path.Match while adding recursive glob. That gives me confidence they are compatible. 

Even if they were bugs lurking, this change doesn't seem to make them any worse or more confusing, since many (most?) excludes were written during the reign of path.Match, and the rule for selecting which matcher to use is consistent: it's the original behavior Gazelle had, unless a "**" is used in the pattern. One could argue that this is a good fix to any subtle/confusing incompatibilities that were introduced by moving to doublestar, since only doublestar-specific patterns which would failed to process under path.Match are processed with doublestar. 

On the flip side, composing an equivalent Regexp for exclude patterns seems like it carries a high risk for incompatibilities with path.Match. What is the motivation for that change? It sounded like "no allocations" was one, but AFAIK path.Match already performs no allocations. It doesn't seem like executing a regex would be more efficient than custom-coded version of the same pattern, so it doesn't seem like speed could be a benefit either. I do agree that the processing of recursive globs should be improved vs doublestar.Match, though.

jayconrod:
I don't mean to argue this isn't a good fix: it mitigates a problem, and the risk of bugs is fairly low.

My objection is more architectural: we would now have two different packages used to match patterns, one of which is a superset of the other with an expensive implementation. I'd rather switch to a single implementation that handles all the uses cases we care about without the extra cost.

In [my comment] on #786, I said I'd be happy to change like this as a short term solution. It's been a while since then though, and I think it would be better to focus on a permanent fix.

robfig:
FYI, saw this other globbing library, which has benchmarks and compile-once semantics.
https://github.com/gobwas/glob

I'm not sure if it's drop-in compatible for file.Match like doublestar is, though.

robfig:
Any guidance on how to move forward with an improvement here? I don't think I have the bandwidth to write a new, more efficient version of doublestar that isn't just delegating to doublestar itself or filepath.Match. 

If not, no worries, I'll close this out.

achew22:
@robfig out of curiosity, have you looked into using `regexp`s like [Globexp](https://github.com/go-godo/godo/blob/v2.0.9/glob/glob.go#L51)? I believe that would prevent having to create a performant version of a glob evaluator. That link appears to be MIT Licensed, @jayconrod What are your thoughts on extracting a single function from a 3rd party repo that is compatibly licensed for the purpose of dramatically reducing memory?

jayconrod:
`Globexp` is pretty much what I had in mind: something that converts a `doublestar` pattern into a `*regexp.Regexp`.

It doesn't look like that function recognizes the same syntax, but maybe the same idea can be implemented in an `internal/glob` package in Gazelle.

wolfd:
I was actually about to make a similar change and just found this. I think this would help my current performance woes, right now the way gazelle excludes work, combined with the recent `.bazelignore` change makes gazelle so slow that I ~can't~ _don't want to_ update it to master.

#1026 has a log of my woes/ideas.

I suspect that an alternative to using two different path matching algorithms is to do more upfront reasoning about the string we've been passed:
- Assume that the presence of any of doublestar's special characters means that we have to use the expensive operation.
- If not, we might be able to immediately check for the file/directory's presence on disk.
- If the exclude doesn't have any `/`s either, then (I think) it doesn't need to be checked outside the current directory, which would save a ton of time that's currently spent checking things that can't possibly match in subdirectories.

I might take a stab at this soon, otherwise I'll probably PR a change that allows me to ignore the `.bazelignore` file :stuck_out_tongue:

PR #868, opened 2020-08-11
language/go: add go_naming_convention_extern flag and directive

jayconrod:
Allows the user to set the default Go naming convention used when
resolving imports of packages in external repositories with unknown
naming conventions.

For #5


jayconrod:
cc @tomlu Does this work for you? If it does, I'll add this commit to #863 before merging.

tomlu:
I verified that this does the trick, thanks.

Issue #803, opened 2020-06-02
gazelle_binary: deprecate and remove mode attributes


jayconrod:
`gazelle_binary` currently supports the attributes `pure`, `static`, `msan`, `race`, `goos`, and `goarch`.

Supporting these attributes requires either `go_archive_aspect` or `go_transition_binary`, both of which are internal to rules_go. Using them in Gazelle makes it difficult to change `go_transition_binary` safely.

None of these attributes should be necessary. `gazelle_binary` should almost always be built for the host platform in the default configuration. If it's necessary to build for a different configuration, that can be done by setting the relevant build settings on the command line (for example, `--@io_bazel_rules_go//go/config:static`) or by targeting an appropriate platform.

dududko:
I think that it would not be an issue if `gazelle_binary` was split on two parts:
1) go_library
2) go_binary

(1) contains all the magic with `go_context(ctx)`; (2) is a direct use of `go_binary` rule (with a proper use of goos/goarch).

Please correct me if I am wrong.

jayconrod:
Maybe... whatever depends on the `go_binary` would need to reset all the build settings to default values though, and that causes problems like bazelbuild/rules_go#2538. In any case, I'd rather not wire this together with genrules and macros though.

Why not remove these attributes? `gazelle_binary` should really only be built for the host configuration. I can't think of why it would need race instrumentation for example.

dududko:
In out team we use executable rule to push build artifacts to s3 (with executable binary written in go) and we build artifacts for both linux and darwin. `goarch/goos` allow us to both build and push all artifacts with a single `bazel run` command, which is very convenient because id does not require to have some side scripts for this job.

Since we extended gazelle with some custom logic for typescript, we also do the release management for gazelle binary for both linux and darwin.

jayconrod:
That seems like a pretty customized situation. I don't think that justifies keeping these attributes for everyone, since they're causing a maintenance problem.

jayconrod:
This is blocking bazelbuild/rules_go#2621.

achew22:
@jayconrod would you like me to take a look into this? aka is this a blocker AI for me? It's going to take me quite a while to get context on it unfortunately. I've never poked the transitions part of rules_go before.

[...]

PR #879, opened 2020-08-24
Depend on rules_go internals in their new location

achew22:
**What type of PR is this?**

Bug fix

cmd/gazelle

Fix for https://github.com/bazelbuild/rules_go/pull/2621

Issue https://github.com/bazelbuild/rules_go/issues/2619

### DIFF
--- a/internal/gazelle_binary.bzl
+++ b/internal/gazelle_binary.bzl
@@ -18,7 +18,7 @@ load(
     "go_context",
 )
 load(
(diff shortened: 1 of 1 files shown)


jayconrod:
Sorry, I can't take this one. Gazelle shouldn't be depending on anything in the private directory. rules_go and Gazelle versions are decoupled, so there's no safe way to move private files in rules_go while this dependency exists. Fixing #803 should remove the dependency entirely.

achew22:
That's an even better solution. Thanks for looking into this!

PR #882, opened 2020-08-26
gazelle_binary: remove mode attributes

### LINKED ISSUE #803: gazelle_binary: deprecate and remove mode attributes

jayconrod:
This PR removes mode attributes (goos, goarch, race, etc.) from the
gazelle_binary rule. These attributes were implemented using private
files in rules_go, so supporting them is blocking changes in the
rules_go implementation.

These attributes should probably never be used. gazelle_binary should
generally be built for the host configuration (by 'bazel run' or
'bazel build' without configuration flags). If a binary is still
needed in a different configuration, that may still be accomplished
with command-line flags like --platforms or
--@io_bazel_rules_go//go/config:race.

Fixes #803

### DIFF
--- a/cmd/gazelle/BUILD.bazel
+++ b/cmd/gazelle/BUILD.bazel
@@ -4,21 +4,6 @@ load("@bazel_gazelle//:def.bzl", "DEFAULT_LANGUAGES", "gazelle_binary")
 gazelle_binary(
     name = "gazelle",
     languages = DEFAULT_LANGUAGES,
-    msan = "off",
-    pure = "off",
-    race = "off",
-    static = "off",
-    visibility = ["//visibility:public"],
-)
-
-gazelle_binary(
-    name = "gazelle_pure",
-    languages = DEFAULT_LANGUAGES,
-    msan = "off",
-    pure = "on",
-    race = "off",
-    static = "off",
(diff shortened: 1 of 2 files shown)


achew22:
With this submitted (and thank you so much for doing it!) I believe we can land https://github.com/bazelbuild/rules_go/pull/2621 if you tag a new release. I will update the PR to reflect that change if you're willing to do a new Gazelle release.

jayconrod:
I'm planning to tag rules_go and Gazelle releases today. That should unblock the `bzl_library` change. I won't bandwidth to review anything more this week though. I'm falling behind on other things.

PR #898, opened 2020-09-05
Avoid embedding go_library when there is no internal test

linzhp:
**What type of PR is this?**

Other

language/go

`go_test` may [recompile](https://github.com/bazelbuild/rules_go/blob/c69d159acddfeb3a6849e105ec7a57542cac8987/go/private/rules/test.bzl#L245) external test archive where there embedded libraries. For packages with only external tests, this is not necessary and may slow down the build. This PR parse the test files to see if there are any internal tests. If not, the generated `go_test` will not embed `go_default_library`.

### DIFF
--- a/cmd/gazelle/integration_test.go
+++ b/cmd/gazelle/integration_test.go
@@ -3807,7 +3807,7 @@ go_library(
 go_test(
     name = "foo_test",
     srcs = ["foo_test.go"],
-	embed = [":foo"],
+    embed = [":foo"],
 )`,
 		},
 	}
--- a/cmd/gazelle/integration_test.go
+++ b/cmd/gazelle/integration_test.go
@@ -3772,3 +3772,59 @@ gazelle_dependencies()
 		},
 	})
 }
+
+func TestExternalOnly(t *testing.T) {
+	files := []testtools.FileSpec{
+		{
+			Path: "WORKSPACE",
+		},
+		{
+			Path: "foo/foo.go",
+			Content: `package foo
+import _ "golang.org/x/baz"
+`,
+		},
+		{
+			Path: "foo/foo_test.go",
+			Content: `package foo_test
+import _ "golang.org/x/baz"
+import _ "example.com/foo"
+`,
+		},
+		{
+			Path:    "BUILD.bazel",
+			Content: "# gazelle:prefix example.com",
+		},
+	}
+	dir, cleanup := testtools.CreateFiles(t, files)
+	defer cleanup()
+
+	var args []string
+	if err := runGazelle(dir, args); err != nil {
+		t.Fatal(err)
+	}
+
+	testtools.CheckFiles(t, dir, []testtools.FileSpec{
+		{
+			Path: "foo/BUILD.bazel",
+			Content: `load("@io_bazel_rules_go//go:def.bzl", "go_library", "go_test")
+
+go_library(
+    name = "foo",
+    srcs = ["foo.go"],
+    importpath = "example.com/foo",
+    visibility = ["//visibility:public"],
+    deps = ["@org_golang_x_baz//:go_default_library"],
+)
+
+go_test(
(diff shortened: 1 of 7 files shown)


language/go/generate.go
```diff
@@ -517,7 +519,16 @@ func (g *generator) generateTest(pkg *goPackage, library string) *rule.Rule {
 	if !pkg.test.sources.hasGo() {
 		return goTest // empty
 	}
-	g.setCommonAttrs(goTest, pkg.rel, nil, pkg.test, library)
+	var embed string
+	for src := range pkg.test.sources.strs {
+		f, err := parser.ParseFile(token.NewFileSet(), filepath.Join(pkg.dir, src), nil, parser.PackageClauseOnly)
```
jayconrod, line 524: Let's not re-open and re-parse these files. Instead, it would be better to set a `hasInternalTest bool` flag in the `goTarget` for the test.

language/go/testdata/platforms/BUILD.want
```diff
@@ -57,6 +57,5 @@ go_test(
         "generic_test.go",
         "suffix_linux_test.go",
     ],
-    _gazelle_imports = [],
-    embed = [":platforms"],
+    _gazelle_imports = ["example.com/repo/platforms"],
```
jayconrod, line 60: Could you also add a test in `cmd/gazelle/integration_test.go`? It should confirm that if there are only external test sources, the `embed` attribute is removed from an existing `go_test`, and the same label is added to `deps`, assuming the external test actually imports the library under test.

Also please manually confirm that builds. IIRC, everything in `embed` is effectively also in `deps` for the external test archive, so if `embed` is removed, that will cause a missing import error unless the same dependency is also added to `deps`.

jayconrod (commented):
Thanks for working on this.

Do you have a sense of what kind of speedup this gives when building tests with a clean cache? I guess it depends on the tests involved (still not sure how common this pattern is).

There is a drawback to doing this: changing a test file from internal to external or vice versa would now require re-running Gazelle. That's probably okay, but let's make sure there's a measurable performance improvement.

linzhp:
Building from clean cache would spend most of the time building the dependencies. So I instead did the following:

1. building an external-only test target embedding `go_default_library`
2. adding `go_default_library` to the `deps` of the `go_test`, remove the `embed`, getting the json trace profile
3. reverting the change back to using `embed`, getting the json trace profile.

With 2, it took 1.5 seconds before `GoLink` action started. With 3, it took 11 seconds before `GoLink`. Among others on the critical path, there was a `GoCompilePkg go_default_test.recompile2854.a` action taking 3.7 seconds, a `GoCompilePkg go_default_test.recompile2855.a` action taking 0.73 seconds, and a `GoCompilePkg go_default_test.recompile2856.a` action taking 2.5 seconds.

language/go/testdata/platforms/BUILD.want
linzhp: Manually confirmed, with both the test and library imports a package, Gazelle add the package to the deps of both `go_test` and `go_library`, even though the former embeds the latter.

cmd/gazelle/integration_test.go
```diff
@@ -3772,3 +3772,75 @@ gazelle_dependencies()
...
+    srcs = ["foo.go"],
+    importpath = "example.com/foo",
+    visibility = ["//visibility:public"],
+    deps = ["@org_golang_x_baz//:go_default_library"],
+)
+
+go_test(
+    name = "foo_test",
+    srcs = ["foo_test.go"],
+	embed = [":foo"],
```
jayconrod, line 3810: Nit: fix indentation

jayconrod (commented):
One small spacing nit. Looks good otherwise.

Thanks for working on this. The improvement in build time definitely seems like it's worth pursuing.

jayconrod (approved):

PR #912, opened 2020-09-21
update-repos can work with a go.mod file in place

### LINKED ISSUE #650: Gazelle erroring on relative path replace directives

0.18.1

0.19.0

0.28.1

I believe so.

macOS / x64

### What did you do?

With a go.mod file like the following: 

```
module github.com/foo/bar

go 1.12

require (
	baz.com v0.0.0
)

replace baz.com => ./build-support/baz
```

I was hoping to see the same behavior as I do with absolute paths.  Namely:

`gazelle: go_repository does not support file path replacements for compass.com -> /Users/cwaeland/development/foo/build-support/baz`

### What did you see instead?

I get the following output:

```
go: parsing build-support/baz/go.mod: open /private/var/folders/s6/tpqlfq494czcv34zgzwm4vy40000gp/T/gazelle-temp-gomod900090332/build-support/baz/go.mod: no such file or directory
go: error loading module requirements
gazelle: error parsing "go.mod": exit status 1
```

I thought this would have been covered by https://github.com/bazelbuild/bazel-gazelle/issues/407 but it doesn't seem to be.

dragonsinth:
**What type of PR is this?**

Bug fix

> language/go
gazelle update-repos

Allow update-repos to operate in-place rather than through a temp copy.  Previously, any go.mod files containing relative on-disk replace directives would completely break update-repos, since the temp copy's relative paths would all be broken.

Fixes #650

### DIFF
--- a/cmd/gazelle/integration_test.go
+++ b/cmd/gazelle/integration_test.go
@@ -2958,6 +2958,13 @@ require (
 )
 `,
 		},
+		{
+			Path: "go.sum",
+			Content: `
+github.com/Selvatico/go-mocket v1.0.7/go.mod h1:4gO2v+uQmsL+jzQgLANy3tyEFzaEzHlymVbZ3GP2Oes=
+github.com/selvatico/go-mocket v1.0.7/go.mod h1:7bSWzuNieCdUlanCVu3w0ppS0LvDtPAZmKBIlhoTcp8=
+`,
+		},
 	}
 	dir, cleanup := testtools.CreateFiles(t, files)
 	defer cleanup()
@@ -2990,6 +2997,12 @@ require (
 )
 
 replace github.com/selvatico/go-mocket => github.com/Selvatico/go-mocket v1.0.7
+`,
+		},
+		{
+			Path: "go.sum",
+			Content: `
+github.com/Selvatico/go-mocket v1.0.7/go.mod h1:4gO2v+uQmsL+jzQgLANy3tyEFzaEzHlymVbZ3GP2Oes=
 `,
 		},
 	}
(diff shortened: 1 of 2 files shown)


jayconrod (commented):
Thanks for working on this.

The change looks good, but it looks like some tests are failing because they lack `go.sum` files. `go list -m -mod=readonly` should fail in that case, so I think those tests should be fixed by adding `go.sum` files.

`go.sum` files can be a bit unwieldy, so rather than specifying them literally, I'd suggest adding a function in `cmd/gazelle/integration_test.go` that runs `go mod tidy` or `go get -d ./...` to populate `go.sum`. That function could be called before `runGazelle` in the failing tests.

dragonsinth:
> The change looks good, but it looks like some tests are failing because they lack `go.sum` files. `go list -m -mod=readonly` should fail in that case, so I think those tests should be fixed by adding `go.sum` files.
> 
> `go.sum` files can be a bit unwieldy, so rather than specifying them literally, I'd suggest adding a function in `cmd/gazelle/integration_test.go` that runs `go mod tidy` or `go get -d ./...` to populate `go.sum`. That function could be called before `runGazelle` in the failing tests.

Makes sense... `language/go.runGoCommandForOutput()` is ideally what I'd like to reuse for the tests, but it's not visible, and I assume we wouldn't want that as part of the public api of that package.  Is there somewhere else that could go, like `internal/language/go` so it could be shared?

dragonsinth:
@jayconrod after playing with this a bit, adding go.sum files seems easier than doing anything else.  While your suggestion to run `go mod tidy` or `go get...` worked fine for `TestImportCollision`, it didn't work for `TestImportCollisionWithReplace` or `TestUpdateReposWithGlobalBuildTags` -- those tests make go unhappy with the case-conflicting requires, causing those go commands to fail.

jayconrod (approved):
That works. Looks good, thanks!

PR #916, opened 2020-09-25
Allow Label.Name to be set as the empty string

vpanta:
**What type of PR is this?**

Bug fix (or Documentation-truthiness?)

label

Documentation for `Label.Name` is:
> Name is the name of the target the label refers to. If omitted, Name is assumed to be the same as Pkg.

However, if you use `label.New` and leave `name` empty, you get a label that ends in `:` when you stringify it, as `String()` doesn't handle an empty name. Given `Name` is technically a "public" field, the correct fix seemed like ensuring it creates the expected string label when `Name` is empty.

There is not a filed issue, but I can make one if you want it.

**Other notes for review**

### DIFF
--- a/label/label.go
+++ b/label/label.go
@@ -134,7 +134,7 @@ func (l Label) String() string {
 		repo = fmt.Sprintf("@%s", l.Repo)
 	}
 
-	if path.Base(l.Pkg) == l.Name {
+	if path.Base(l.Pkg) == l.Name || l.Name == "" {
 		return fmt.Sprintf("%s//%s", repo, l.Pkg)
 	}
 	return fmt.Sprintf("%s//%s:%s", repo, l.Pkg, l.Name)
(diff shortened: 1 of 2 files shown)


jayconrod:
Sorry, I don't agree with this change.

The documentation for the `Name` field seems misleading. I think it's referring to the behavior of `label.Parse`. That should probably be fixed.

`Name` shouldn't be empty though. `label.Parse` returns an error for a string like `//x:`. `label.New` doesn't check any constraints; it doesn't return an error, so it can't do much. There are a lot of things that just check the `Name` field without also checking `Pkg` if `Name` is empty.

vpanta:
+1 no worries, fixing the documentation works as an acceptable alternative. I'm good either way, so long as it's not considered an "acceptable" use case as the old docs implied.

Thanks!

PR #922, opened 2020-09-30
Merge private attrs in MergeRules

HALtheWise:
**What type of PR is this?**

Other

rule.MergeRules

I am working on a custom language extension, and was trying to use SetPrivateAttr(...) to pass data between the Generate and Resolve functions. In this particular case, Resolve needs information about the parsed contents of the source file that is not apparent from just the generated rule, and we want to avoid paying the cost of parsing the files twice.

This works fine for rules that are newly-generated by this invocation of Bazel, but when there is an existing rule in the BUILD file with the same name, it inexplicably fails due to MergeRules not propagating private attributes.

This seemed small enough to not need a dedicated issue, but I'm happy to open one if more discussion is needed.

**Other notes for review**

### DIFF
--- a/rule/merge.go
+++ b/rule/merge.go
@@ -46,6 +46,20 @@ func MergeRules(src, dst *Rule, mergeable map[string]bool, filename string) {
 		return
 	}
 
+	// If src has private attrs, copy them into dst
+	for k, v := range src.private {
+		dst.private[k] = v
+	}
+
+	// Merge the private attrs from src into dest
+	for k, v := range src.private {
+		if _, ok := dst.private[k]; ok {
+			start, _ := dst.expr.Span()
+			log.Printf("%s:%d: Private attribute \"%s\" overridden during merge", filename, start.Line, k)
+		}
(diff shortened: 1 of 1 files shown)


jayconrod:
Sorry, this is working as intended. Private attributes exist in memory only. They aren't written to build files, so `merger` shouldn't deal with them. It's likely not even possible to merge them since they could have any type. Instead, `Resolve` should translate private attributes into non-private attributes that `merger` can manage.

If you're calling `SetPrivateAttr` on a rule generated by `Generate`, and then the private attribute isn't visible later when `Resolve` is called on the same rule, it sounds like something else might be wrong. Feel free to open an issue to discuss and debug further.

In general, please open an issue first before a PR. PRs that fix typos or trivial bugs are okay without issues, but for anything larger, I'd like to agree on a solution before reviewing code.

HALtheWise:
OK, opening #923 to debug and discuss. In making a limited test case, it seems to actually affect only Imports(), not Resolve()

Issue #925, opened 2020-10-01
mode=diff does not respect missing/empty files


vpanta:
### What version of gazelle are you using?

v0.22.0

### What version of rules_go are you using?

v0.24.2

### What version of Bazel are you using?

3.1.0

### Does this issue reproduce with the latest releases of all the above?

yes

### What operating system and processor architecture are you using?

macOS Catalina, 10.15.7, x64

### What did you do?

I ran `bazel run :gazelle` followed by `bazel run :gazelle -- -mode=diff`

Or alternatively, I changed the `gazelle()` target to include `args = ["-mode=diff"]` and re-ran for the second command. I attempted both ways.

### What did you expect to see?

No output from diff and a success code, as I'd just fixed the BUILD files according to gazelle.

### What did you see instead?

Diffs, including empty `BUILD.bazel` files showing a diff of just a newline, and an error code.
I had hoped to use `-mode=diff` as a precommit, but as it returns unexpected diffs, I cannot.

vpanta:
So I've been looking further, and it actually seems like all the diffs are files it decided need to exist with a single newline in them, but don't currently exist.

I think it's a result of https://github.com/bazelbuild/bazel-gazelle/commit/70f8ccb87fe70445ca036e31ab9b892e03f56081 removing
```go
diff.A = difflib.SplitLines(string(oldContent))
```
 when the file is not found, but still using
```go
diff.B = difflib.SplitLines(string(newContent))
```
[As `SplitLines` seems to turn an empty string into `[]string{"\n"}`](https://github.com/pmezard/go-difflib/blob/5d4384ee4fb2527b0a1256a821ebfc92f91efefc/difflib/difflib.go#L768)

vpanta:
I just put that out as a prospective fix. Let me know if it's worthwhile or if you'd rather we take this a different direction.

jayconrod:
I think this is a duplicate of #853.

vpanta:
Ah, it is, the title there is misleading, because these files are missing and diff-mode is actually saying they should exist but be empty.

Issue #853, opened 2020-08-02
Diff mode wants me to delete on-purpose empty BUILD files


gertvdijk:
### What version of gazelle are you using?

Master, commit 30aa1d7c9406aaf0838781e907eefc4bc2d1eddc.

### What version of rules_go are you using?

0.23.6

### What version of Bazel are you using?

3.4.1

### Does this issue reproduce with the latest releases of all the above?

Yes.

### What operating system and processor architecture are you using?

Linux, Ubuntu 18.04, amd64.

### What did you do?

Have some on-purpose empty `BUILD.bazel` files in some directories with sibling files there. They allow me to reference sibling files there as `//mydir/mysubdir:otherfile.xyx` rather than `//:mydir/mysubdir/otherfile.xyx`.

1. Run Gazelle in `-mode diff`.
2. Run Gazelle in `-mode fix`.

### What did you expect to see?

1. Gazelle not to touch the empty BUILD.bazel files which are unrelated to what I use Gazelle for.
2. diff and fix to be consistent with each other.

### What did you see instead?

Bug part 1: , I get LOTS of:

```diff
--- /dev/null   1970-01-01 00:00:00.000000000 +0000
+++ mydir/mysubdir/BUILD.bazel     1970-01-01 00:00:00.000000000 +0000        
@@ -0,0 +1 @@                                            
+
```
Bug part 2: in `-mode fix`, this isn't fixed, making diff and fix inconsistent with each other.

This behavior was not part of Gazelle 0.20.0 which I used before. Solely updating Gazelle to the master version triggers this change in behavior.

PR #926, opened 2020-10-01
diff: make a non-existing file match an empty generation

vpanta:
**What type of PR is this?**

Bug fix

cmd/gazelle

`-mode=diff` doesn't handle the generation of empty files the same way as `-mode=fix` does.
Fix won't actually print out empty files, but diff says the empty file should contain a single newline (due to how `difflib.SplitLines` handles the empty string).

#925

**Other notes for review**

The test I added actually passes without this change, so it's not actually helpful. I'd love assistance turning it into something which requires this change to pass, and would fail otherwise.

### DIFF
--- a/cmd/gazelle/diff.go
+++ b/cmd/gazelle/diff.go
@@ -16,6 +16,7 @@ limitations under the License.
 package main
 
 import (
+	"bytes"
 	"fmt"
 	"io"
 	"os"
@@ -42,22 +43,30 @@ func diffFile(c *config.Config, f *rule.File) error {
 		ToDate:   date,
 	}
 
-	if len(f.Content) == 0 {
-		diff.FromFile = "/dev/null"
-	} else {
-		diff.A = difflib.SplitLines(string(f.Content))
-		if c.ReadBuildFilesDir == "" {
-			path, err := filepath.Rel(c.RepoRoot, f.Path)
-			if err != nil {
-				return fmt.Errorf("error getting old path for file %q: %v", f.Path, err)
-			}
-			diff.FromFile = filepath.ToSlash(path)
-		} else {
-			diff.FromFile = f.Path
-		}
+	newContent := f.Format()
+	if bytes.Equal(newContent, f.Content) {
+		// No change.
+		return nil
 	}
 
-	newContent := f.Format()
+	if _, err := os.Stat(f.Path); os.IsNotExist(err) {
+		diff.FromFile = "/dev/null"
+	} else if err != nil {
+		return fmt.Errorf("error reading original file: %v", err)
+	} else if c.ReadBuildFilesDir == "" {
+		path, err := filepath.Rel(c.RepoRoot, f.Path)
+		if err != nil {
+			return fmt.Errorf("error getting old path for file %q: %v", f.Path, err)
+		}
+		diff.FromFile = filepath.ToSlash(path)
+	} else {
+		diff.FromFile = f.Path
+	}
+
+	if len(f.Content) != 0 {
+    		diff.A = difflib.SplitLines(string(f.Content))
+	}
+
 	diff.B = difflib.SplitLines(string(newContent))
 	outPath := findOutputPath(c, f)
 	if c.WriteBuildFilesDir == "" {
--- a/cmd/gazelle/diff_test.go
+++ b/cmd/gazelle/diff_test.go
@@ -102,6 +102,33 @@ func TestDiffNew(t *testing.T) {
 	testtools.CheckFiles(t, dir, want)
 }
 
+func TestDiffMissingAndNoChange(t *testing.T) {
+	files := []testtools.FileSpec{
+		{Path: "WORKSPACE"},
+	}
+	dir, cleanup := testtools.CreateFiles(t, files)
+	defer cleanup()
+
+	if err := runGazelle(dir, []string{"-go_prefix=example.com/hello", "-mode=diff", "-patch=p"}); err != nil {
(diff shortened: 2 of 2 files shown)


cmd/gazelle/diff.go
```diff
@@ -44,6 +44,8 @@ func diffFile(c *config.Config, f *rule.File) error {
 
 	if len(f.Content) == 0 {
 		diff.FromFile = "/dev/null"
+		// difflib.SplitLines will turn the empty string into "\n"
+		diff.A = "\n"
```
jayconrod, line 48: Seems like the main issue here is that `len(f.Content) == 0` regardless of whether the file was empty or non-existent. It would be better to pass in more information to distinguish those two cases.

If the file doesn't exist, and `f.Content` is empty after formatting, then we don't have to print anything (here or in `printFile`).

dragonsinth:
https://github.com/bazelbuild/bazel-gazelle/issues/853

cmd/gazelle/diff.go
dragonsinth: Could use `[]byte(nil)` vs. `[]byte{}` to differentiate

vpanta: So I updated it to check if the file was explicitly non-existent for now, but I'm open to refining it further.

cmd/gazelle/diff_test.go
```diff
@@ -102,6 +102,18 @@ func TestDiffNew(t *testing.T) {
...
+func TestDiffEmpty(t *testing.T) {
+	files := []testtools.FileSpec{
+		{Path: "WORKSPACE"},
+	}
+	dir, cleanup := testtools.CreateFiles(t, files)
+	defer cleanup()
+
+	if err := runGazelle(dir, []string{"-go_prefix=example.com/hello", "-mode=diff", "-patch=p"}); err != nil {
+		t.Fatalf("Expected no error, but got %q", err)
+	}
```
jayconrod, line 114: The test should check the content of the patch file `p`. You can use `testtools.CheckFiles` like in the other tests.

jayconrod (commented):
Looks like tests are failing. Please fix.

cmd/gazelle/diff_test.go
vpanta: Yeah, np, I actually still need help with this test, as it doesn't fail when the change is reverted. That's why I haven't updated it yet.

I can't figure out a way to verify that it'd fail without the proposed change. Any thoughts are appreciated.

vpanta: nevermind, figured it out.

vpanta:
Ok, found my issue, I thought it was that gazelle wanted to create empty files, but no the issue is that a completely empty file is expected to contain a single `\n` character. This fixes that, and updates so that:

1. If it's an empty file, the diff sees it as empty.
2. If it's a missing file, the diff uses `/dev/null` as it's name.

These two cases are no longer conflated.

cmd/gazelle/diff.go
```diff
@@ -42,7 +42,12 @@ func diffFile(c *config.Config, f *rule.File) error {
 		ToDate:   date,
 	}
 
-	diff.A = difflib.SplitLines(string(f.Content))
+	newContent := f.Format()
+	if string(newContent) == string(f.Content) {
```
jayconrod, line 46: Use `bytes.Equal` to compare `[]byte`. No need to allocate and copy to strings.

cmd/gazelle/diff_test.go
```diff
@@ -105,12 +105,17 @@ func TestDiffNew(t *testing.T) {
 func TestDiffEmpty(t *testing.T) {
 	files := []testtools.FileSpec{
 		{Path: "WORKSPACE"},
+		{Path: "BUILD.bazel"},
```
jayconrod, line 108: This will create an empty `BUILD.bazel` file before running Gazelle. We want to verify the diff is empty if the file is *not* present though, so let's take this out.

cmd/gazelle/diff_test.go
```diff
@@ -105,12 +105,17 @@ func TestDiffNew(t *testing.T) {
...
 	}
 	dir, cleanup := testtools.CreateFiles(t, files)
 	defer cleanup()
 
 	if err := runGazelle(dir, []string{"-go_prefix=example.com/hello", "-mode=diff", "-patch=p"}); err != nil {
-		t.Fatalf("Expected no error, but got %q", err)
+		t.Error("Expected no diff, but got a diff.")
+		want := append(files, testtools.FileSpec{
```
jayconrod, line 115: This code only checks that `p` is empty if `runGazelle` returns an error, but we want no error *and* and empty patch.

```
if err := runGazelle(...); err != nil {
  t.Fatalf("Expected no error, but got %q", err)
}
testtools.CheckFiles(t, dir, []testtools.FileSpec{{Path: "p"}})
```

cmd/gazelle/diff_test.go
vpanta: So, I originally had this backwards with what was erroring. I thought I'd checked my local code well enough, but apparently the issues raised in #925 were exactly the same as those in #853: the BUILD files existed and were empty.

That's why my initial test wouldn't fail with the original code, because the missing-file case was actually fine.

I've added two cases now, `TestDiffMissingAndNoChange` and `TestDiffEmptyAndNoChange`. The first passes both with the original code and my changes, and the second fails with the original code but passes with my changes.

jayconrod (approved):
Cool, looks good. Thanks for fixing this.

PR #936, opened 2020-10-17
Added bzl_library targets for use in external repos using stardoc

UebelAndre:
**What type of PR is this?**

Feature

other

This PR exposes `bzl_library` targets.

This fixes the ability to use `stardoc` in other projects that depend on `bazel-gazelle`

**Other notes for review**
This implements the same thing as #760 but with a fix for the visibility of the top level `//:bzl_srcs` target so it can **actually** be used in other projects. 

Also, I'm unsure if https://github.com/bazelbuild/rules_go/pull/2621 can be applied to this repo as well.

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -1,3 +1,4 @@
+load("@bazel_skylib//:bzl_library.bzl", "bzl_library")
 load("@io_bazel_rules_go//go:def.bzl", "nogo")
 load("//:def.bzl", "gazelle", "gazelle_binary")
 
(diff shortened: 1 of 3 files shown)


jayconrod:
Sorry, I'd rather not merge this. It's a duplicate of #760. I'll comment on that one.

UebelAndre:
@jayconrod Why would you close this PR (one that works) in favor of an older one that's had no activity for months?

jayconrod:
This one isn't correct either. At a quick glance, it adds a duplicate dependency in `WORKSPACE`, and it doesn't follow conventions from `@bazel_skylib//gazelle/bzl`. I don't want to spend time reviewing duplicate PRs.

PR #939, opened 2020-10-20
attempt to explain why replace + commit/tag fail

Mythra:
**What type of PR is this?**

Documentation

go_repository

Attempt to explain why commit/tag can be used in conjuction with replace. Not needed per say, but I ran into this and because I wasn't exactly thinking it through, it caused me to think I needed this to work getting me more and more frustrated :sweat:

### DIFF
--- a/repository.rst
+++ b/repository.rst
@@ -156,6 +156,12 @@ returned by ``go env GOPATH``.
 | A replacement for the module named by ``importpath``. The module named by                                                 |
 | ``replace`` will be downloaded at ``version`` and verified with ``sum``.                                                  |
 |                                                                                                                           |
+| NOTE: Replace does not work with `commit` or `tag`, or other git selection params                                         |
+| since in this mode, bazel doesn't need to resolve dependencies like go does                                               |
+| where replace would make a difference. To get replace behaviour simply change                                             |
+| the importpath in a rule downloading the dependency without a replace                                                     |
+| replace attribute. See the "Download from git fork" example above.                                                        |
+|                                                                                                                           |
 | NOTE: There is no ``go_repository`` equivalent to file path ``replace``                                                   |
 | directives. Use ``local_repository`` instead.                                                                             |
 +------------------------------------+----------------------+---------------------------------------------------------------+


achew22:
@Mythra thanks so much for following up on this and for taking action to help the community. Without people who step up and help, we will never get to a better place. I hope your day improves!

repository.rst
```diff
@@ -156,6 +156,12 @@ returned by ``go env GOPATH``.
 | A replacement for the module named by ``importpath``. The module named by                                                 |
 | ``replace`` will be downloaded at ``version`` and verified with ``sum``.                                                  |
 |                                                                                                                           |
+| NOTE: Replace does not work with `commit` or `tag`, or other git selection params                                         |
+| since in this mode, bazel doesn't need to resolve dependencies like go does                                               |
+| where replace would make a difference. To get replace behaviour simply change                                             |
+| the importpath in a rule downloading the dependency without a replace                                                     |
+| replace attribute. See the "Download from git fork" example above.                                                        |
```
jayconrod, line 163: Rephrase and format:

```
NOTE: ``replace`` only works in module mode and cannot be used with ``commit``,
``tag``, or other VCS attributes. To download directly from a Git repository,
see the "Download from git fork" example above.
```

It is possible to download from a forked repo in module mode though. So it might be better to add another example as well (and mention it here):

```
# Download from a module fork
go_repository(
    name = "com_github_pkg_errors",
    importpath = "github.com/pkg/errors",
    replace = "github.com/fork/errors",
    sum = "h1:iURUrRGxPUNPdy5/HRSm+Yj6okJ6UtLINN0Q9M4+h3I=",
    version = "v0.8.1-forkedversion",
)
```

jayconrod (commented):
Thanks for working on this! A couple suggestions below.

repository.rst
Mythra: First off sorry for the delay in this, I don't see it in my Github Notifcation box at all :/ not sure what happened there.

I don't like this particular rephrasing though, the particular brunt of my frustration that brought around this PR is the lack of really thinking about _why_ replace doesn't actually even need to work, and why it doesn't work. This rephrasing simply says: "This doesn't work" which isn't actually helpful if you're not familiar with why it doesn't work (and why it really in effect doesn't need too):

```
NOTE: `replace` only affects things in module mode and cannot be used with ``commit``,
``tag``, or other VCS attributes. If using VCS attributes you should simply specify a seperate
importpath.
```

jayconrod: I'd change the last sentence to 

```
When using VCS attributes, you can use the ``remote`` attribute instead.
See the "Download from git fork" example above.
```

`importpath` is the main attribute of `go_repository`. It is a prefix for the `importpath` of every generated `go_library` target in the repo. Changing it means you're fetching a different module.

`replace` and `remote` are used to change where the module is downloaded from (in module mode and VCS mode, respectively). Normally, `importpath` determines that, but `replace` and `remote` let you download the same module (or a modified version with the same import paths) from a different location.

achew22:
@Mythra friendly ping on this 😄

Mythra:
@achew22 thanks for the reminder in this! It got lost between all the stuff I was up too.

repository.rst
Mythra: Should be fixed in the latest patch!

jayconrod: Wording looks good. Please format code with two quotes:

```
``like this`` not `like this`
```

robfig:
Hi @Mythra , sorry for the delay in re-reviewing. It looks like there's one remaining review item to take care of, and then I'll merge it. Jay mentioned it, but I think it got lost in the UI:

Please format code with two quotes:

  ``like this`` not `like this`

PR #949, opened 2020-10-29
walk: UpdateSubdirsMode

### LINKED ISSUE #945: walk: add UpdateSubDirsMode
Currently when `--index=false` and `--r=true` [Gazelle visits](https://github.com/bazelbuild/bazel-gazelle/blob/master/cmd/gazelle/fix-update.go#L123-L129) every directory in the repository and updates the passed in directories plus their subdirectories ([`walk.VisitAllUpdateSubdirsMode`](https://github.com/bazelbuild/bazel-gazelle/blob/master/walk/walk.go#L38-L41)).

In this scenario because `--index=false` there is no need to visit the entire repository, which is very expensive in large repos. Gazelle could support a new `walkMode`, `UpdateSubDirsMode` which behaves similarly to [`UpdateDirsMode`](https://github.com/bazelbuild/bazel-gazelle/blob/master/walk/walk.go#L47-L50) except it also updates the subdirectories of the dirs passed into Gazelle. This mode can be enabled when `--index-false` and `--r=true`, or alternatively a new flag could be added to enable this mode (e.g. `--updateSubDirs`).

blico:
Introduces `walk.UpdateSubdirsMode` set by `fix-update` when `--index=false` and `--r=true`. 
In this mode, Gazelle only visits and updates the directories given to it and their subdirectories.
Before, with this configuration Gazelle would visit all of the directories in the repository which is
unnecessary because `--index=false`.

Fixes #945

### DIFF
--- a/walk/walk.go
+++ b/walk/walk.go
@@ -48,6 +48,12 @@ const (
 	// Build files in parent directories are read in order to produce a complete
 	// configuration, but the callback is not called for parent directories.
 	UpdateDirsMode
+
+	// In UpdateSubdirsMode, Walk visits and updates the directories given to Walk
+	// and their subdirectories. Build files in parent directories are read in
+	// order to produce a complete configuration, but the callback is not called
+	// for parent directories.
+	UpdateSubdirsMode
 )
 
 // WalkFunc is a callback called by Walk in each visited directory.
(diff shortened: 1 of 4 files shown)


walk/walk.go
```diff
@@ -208,23 +214,35 @@ func buildUpdateRelMap(root string, dirs []string) map[string]bool {
 
 // shouldCall returns true if Walk should call the callback in the
 // directory rel.
-func shouldCall(rel string, mode Mode, updateRels map[string]bool) bool {
-	return mode != UpdateDirsMode || updateRels[rel]
+func shouldCall(rel string, mode Mode, updateParent bool, updateRels map[string]bool) bool {
+	if mode != UpdateDirsMode && mode != UpdateSubdirsMode {
```
jayconrod, line 218: Having trouble following the logic. I think changing it to a `switch` would be clearer.

```
switch mode {
case VisitAllUpdateSubdirsMode, VisitAllUpdateDirsMode:
  return true
case UpdateSubdirsMode:
  return updateParent || updateRels[rel]
default: // UpdateDirsMode
  return updateRels[rel]
}
```

jayconrod (commented):
A few small comments, but this looks good. Thanks for working on this.

Sorry for the slow review. I took last week off.

jayconrod (approved):
Looks good. Thanks again!

PR #950, opened 2020-11-01
merger: support dict attributes

### LINKED ISSUE #937: Support for merging dicts
0.20.0

0.24.2

1.4.0

Yes

Linux x86

### What did you do?
I was writing a Gazelle extension and looking to update a rule which has a `string_dict` attribute. From what I can tell, Gazelle doesn't support merging plain key-value pairs? so that setting a map attribute panics
```go
myrule.SetAttr('a', map[string]string { "k": "v" })
```
Is this use case maybe not supposed to be supported (I could see a `mergeDict` which seemed to only be used for `select`s at the moment)? is there a workaround for this? e.g a way to replace the attribute rather than merge?

An expression in `myrule` with attribute `a = { "k": "v" }`

### What did you see instead?
// gazelle: type not supported: <nil>
// panic: type not supported: <nil>

iffyio:
**What type of PR is this?**
> Feature

> language/go

Adds support for merging dict attributes - Makes `platformStringExprs` hold generic `bzl.Expr`s as a result
and updates code paths to support Dicts as well.

Fixes #937

**Other notes for review**
Happy to change anything that isn't right!

### DIFF
--- a/rule/expr.go
+++ b/rule/expr.go
@@ -18,6 +18,7 @@ package rule
 import (
 	"fmt"
 	"log"
+	"reflect"
 	"strings"
 
 	"github.com/bazelbuild/bazel-gazelle/label"
@@ -147,7 +148,7 @@ func FlattenExpr(e bzl.Expr) bzl.Expr {
 		return true
 	}
 
-	if ps.generic != nil {
+	if ps.generic != nil && !reflect.ValueOf(ps.generic).IsNil() {
 		if !addList(ps.generic) {
 			return e
 		}
@@ -173,7 +174,7 @@ func isScalar(e bzl.Expr) bool {
 	}
 }
 
-func dictEntryKeyValue(e bzl.Expr) (string, *bzl.ListExpr, error) {
+func dictEntryKeyValue(e bzl.Expr) (string, bzl.Expr, error) {
 	kv, ok := e.(*bzl.KeyValueExpr)
 	if !ok {
 		return "", nil, fmt.Errorf("dict entry was not a key-value pair: %#v", e)
@@ -182,11 +183,7 @@ func dictEntryKeyValue(e bzl.Expr) (string, *bzl.ListExpr, error) {
 	if !ok {
 		return "", nil, fmt.Errorf("dict key was not string: %#v", kv.Key)
 	}
-	v, ok := kv.Value.(*bzl.ListExpr)
-	if !ok {
-		return "", nil, fmt.Errorf("dict value was not list: %#v", kv.Value)
-	}
-	return k.Value, v, nil
+	return k.Value, kv.Value, nil
 }
 
 func stringValue(e bzl.Expr) string {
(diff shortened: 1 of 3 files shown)


rule/expr.go
```diff
@@ -147,7 +148,7 @@ func FlattenExpr(e bzl.Expr) bzl.Expr {
 		return true
 	}
 
-	if ps.generic != nil {
+	if ps.generic != nil && !reflect.ValueOf(ps.generic).IsNil() {
```
jayconrod, line 151: Interface values should almost always be non-nil or untyped nil. This check implies a typed nil is a valid value, and I'm worried that's very error-prone.

rule/expr.go
```diff
@@ -173,7 +174,7 @@ func isScalar(e bzl.Expr) bool {
 	}
 }
 
-func dictEntryKeyValue(e bzl.Expr) (string, *bzl.ListExpr, error) {
+func dictEntryKeyValue(e bzl.Expr) (string, bzl.Expr, error) {
```
jayconrod, line 177: It's fine to rename this function (maybe `selectDictEntryKeyValue`), but its semantics shouldn't change. It's matching entries in a dict passed to a `select`, and it needs to keep doing that.

rule/expr.go
```diff
@@ -209,7 +206,7 @@ func stringValue(e bzl.Expr) string {
 // The four collections may appear in any order, and some or all of them may
 // be omitted (all fields are nil for a nil expression).
 type platformStringsExprs struct {
-	generic            *bzl.ListExpr
+	generic            bzl.Expr
```
jayconrod, line 209: This needs to stay `*bzl.ListExpr`. `platformStringsExprs` is an expression matching a specific pattern (described in the doc comment). The list expression cannot be anything other than a list.

rule/expr.go
```diff
@@ -241,7 +238,13 @@ func extractPlatformStringsExprs(expr bzl.Expr) (platformStringsExprs, error) {
 		switch part := part.(type) {
 		case *bzl.ListExpr:
 			if ps.generic != nil {
-				return platformStringsExprs{}, fmt.Errorf("expression could not be matched: multiple list expressions")
+				return platformStringsExprs{}, fmt.Errorf("expression could not be matched: multiple generic expressions")
+			}
+			ps.generic = part
+
+		case *bzl.DictExpr:
```
jayconrod, line 245: A bare dict cannot be part of `platformStringsExpr`. This function matches a specific pattern, and that shouldn't change.

jayconrod (commented):
Thanks for working on this. This needs some changes though.

`rule.PlatformStrings` and `rule.platformStringsExpr` are meant to handle a specific pattern that's useful for the Go extension. They pre-date Gazelle's extension mechanism (Gazelle used to only support Go), and they shouldn't really be used for anything other than Go.

It's fine to add support for merging other expressions that are not `PlatformStrings`, but `PlatformStrings` itself should not be changed.

iffyio:
Thanks for the review @jayconrod  I'll close this PR as I haven't found time to have another look into it

PR #959, opened 2020-11-24
support dependencies on mapped kinds

### LINKED ISSUE #943: Gazelle not working on honnef.co/go/tools
### What versions are you using?

Bazel 3.5.0
rules_go-v0.24.4
bazel-gazelle-v0.22.2

### What did you do?

I was attempting to compose some go_tool_library rules for staticcheck for the purpose of #578 and found that Gazelle did not appear to run correctly on the repo. 

Repo with WORKSPACE / BUILD.bazel adding Gazelle rule:
https://github.com/robfig/go-tools

To reproduce:
```
$ git clone git@github.com:robfig/go-tools.git
$ cd go-tools
$ bazel run :gazelle
```

Running Gazelle produces error messages like this:
```
module honnef.co/go/tools@latest found (v0.0.1-2020.1.6), but does not contain package honnef.co/go/tools/go/ir
```

It seems like the `go_prefix` must be set incorrectly, but I have it configured like this in //BUILD.bazel:
```
# gazelle:prefix honnef.co/go/tools
```

and I also tried passing it as a command line argument.

The result of this error is that the rules incorrectly treat internal packages as external packages, e.g. 
```
go_tool_library(
    name = "staticcheck",
    srcs = [
        "analysis.go",
        "buildtag.go",
        "doc.go",
        "lint.go",
        "rules.go",
        "structtag.go",
    ],
    importpath = "honnef.co/go/tools/staticcheck",
    deps = [
        "@co_honnef_go_tools//go/types/typeutil:go_default_library",
        "@co_honnef_go_tools//internal/passes/buildir:go_default_library",
        "@co_honnef_go_tools//internal/sharedcheck:go_default_library",
        "@co_honnef_go_tools//pattern:go_default_library",
        "@co_honnef_go_tools//printf:go_default_library",
    ...
```

I feel like I must be missing something obvious, but I'm not seeing it.. any tips? 

Thank you

robfig:
Previously, rules with a mapped kind added to the RuleIndex would not
be importable.

The MetaResolver records mapped kinds generated by recorded associated
with the package. When looking up an import, we do not have the
associated package. This change addresses that by storing the language
name on the ruleRecord itself, obviating the need to look up the
Resolver.

The additional test for this scenario fails prior to this change and
succeeds as a result of it.

Fixes #943

### DIFF
--- a/cmd/gazelle/metaresolver.go
+++ b/cmd/gazelle/metaresolver.go
@@ -17,6 +17,8 @@ package main
 
 import (
 	"github.com/bazelbuild/bazel-gazelle/config"
+	"github.com/bazelbuild/bazel-gazelle/label"
+	"github.com/bazelbuild/bazel-gazelle/repo"
 	"github.com/bazelbuild/bazel-gazelle/resolve"
 	"github.com/bazelbuild/bazel-gazelle/rule"
 )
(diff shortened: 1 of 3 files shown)


cmd/gazelle/metaresolver.go
```diff
@@ -51,11 +53,61 @@ func (mr *metaResolver) MappedKind(pkgRel string, kind config.MappedKind) {
...
+	imkr.inverseMapKind(r, func(r *rule.Rule) {
+		imkr.delegate.Resolve(c, ix, rc, r, imports, from)
+	})
+}
+
+func (imkr inverseMapKindResolver) inverseMapKind(r *rule.Rule, fn func(r *rule.Rule)) {
+	origKind := r.Kind()
+	r.SetKind(imkr.fromKind)
+	fn(r)
+	r.SetKind(origKind)
```
jayconrod, line 112: I'm a little nervous about mutating the original rule. It's safe for now, but I'd be a little worried about data races if this is ever parallelized. Maybe instead:

```
rCopy := *r
rCopy.SetKind(imkr.fromKind)
fn(&rCopy)
```

Also, this is more cosmetic, but instead of accepting `fn` as an argument, it would be more straightforward to return `&rCopy`, then let callers pass that to `Imports`, `Embeds`, and `Resolve`.

jayconrod (commented):
Looks good with one comment. Thanks for fixing this.

And sorry for the slow review. I went on vacation before Thanksgiving, and I'm almost caught up now.

cmd/gazelle/metaresolver.go
robfig: That makes a lot of sense, and returning the copy of the rule makes for much cleaner code. Updated, thanks!

robfig:
No worries at all, time off is important. I certainly appreciate your work on rules_go -- it really improves the day-to-day life of developers at my work. 

Thanks for the review, updated!

jayconrod (approved):
Looks good, thanks!

PR #967, opened 2020-12-14
gazelle_dependencies: add support to set project specific go env variables

### LINKED ISSUE #966: go_repository: support setting environment variables like GOPROXY
We are mirroring all our dependency to our company's local network and also use an on-premise
GOPROXY implementation.

We can enable GOPROXY by

1. setting a host environment variable before running any bazel command
2. setting --repo_env in .bazelrc

The problem we have with 1: our users should not have to change their host's environment to use our goproxy. If our users have a misconfigured environment they will start accessing remote locations (which they can't/shouldn't)

The problem with solution 2: --repo_env does not work with `bazel query` commands and our CI depends on those commands and running bazel query alternating with bazel build/run  the analysis cache is discarded every time with message

INFO: Build option --repo_env has changed, discarding analysis cache.

The current workaround is a patch that we apply when importing gazelle into our WORKSPACE.

The patch looks like this:
```
diff -urN b/deps.bzl c/deps.bzl
--- b/deps.bzl
+++ c/deps.bzl
@@ -38,11 +38,13 @@
 
 def gazelle_dependencies(
         go_sdk = "",
-        go_repository_default_config = "@//:WORKSPACE"):
+        go_repository_default_config = "@//:WORKSPACE",
+       go_proxy = ""):
     if go_sdk:
         go_repository_cache(
             name = "bazel_gazelle_go_repository_cache",
             go_sdk_name = go_sdk,
+           go_proxy = go_proxy,
         )
     else:
         go_sdk_info = {}
@@ -60,6 +62,7 @@
         go_repository_cache(
             name = "bazel_gazelle_go_repository_cache",
             go_sdk_info = go_sdk_info,
+           go_proxy = go_proxy,
         )
 
     go_repository_tools(
diff -urN b/internal/go_repository_cache.bzl c/internal/go_repository_cache.bzl
--- b/internal/go_repository_cache.bzl
+++ c/internal/go_repository_cache.bzl
@@ -64,6 +64,11 @@
         gopath = go_path,
         gocache = go_cache,
     )
+    if ctx.attr.go_proxy:
+       env_content += """
+GOPROXY='{goproxy}'
+""".format(goproxy = ctx.attr.go_proxy)
+
     ctx.file("go.env", env_content)
     ctx.file("BUILD.bazel", 'exports_files(["go.env"])')
 
@@ -72,6 +77,7 @@
     attrs = {
         "go_sdk_name": attr.string(),
         "go_sdk_info": attr.string_dict(),
+       "go_proxy": attr.string(),
     },
     # Don't put anything in environ. If we switch between the host cache
     # and Bazel's cache, it shouldn't actually invalidate Bazel's cache.
```

Using this patch it is possible to define a go_proxy as an argument to `gazelle_dependencies`.
This go_proxy will still be overridden by a env variavle GOPROXY from host or --repo_env.

Is this a solution worthwhile to create a pull request from?

dirkboye:
* add parameter `go_env` to rule `gazelle_dependencies`

* add documentation describing new parameter `go_env` in README.rst

* add unit test for inclusion of values in `external/bazel_gazelle_go_repository_cache/go.env`

**What type of PR is this?**

Feature

go_repository

Fixes #966 

Open questions:
1. Use a blacklist for GOROOT / GOPATH / GOCACHE or use same whitelist as in go_repository.bzl? ( see line 68 in internal/go_repository_cache.bzl )
2. How to test black/whitelist which will need different WORKSPACE for each test?

### DIFF
--- a/internal/go_repository_cache.bzl
+++ b/internal/go_repository_cache.bzl
@@ -54,16 +54,17 @@ def _go_repository_cache_impl(ctx):
             fail("GOCACHE must be set when GO_REPOSITORY_USE_HOST_CACHE is enabled.")
         go_cache = res.stdout.strip()
 
-    env_tpl = """
-GOROOT='{goroot}'
-GOPATH='{gopath}'
-GOCACHE='{gocache}'
-"""
-    env_content = env_tpl.format(
-        goroot = go_root,
-        gopath = go_path,
-        gocache = go_cache,
-    )
+    cache_env = {
+        "GOROOT": go_root,
+        "GOPATH": go_path,
+        "GOCACHE": go_cache,
+    }
+    for key, value in ctx.attr.go_env.items():
+        if key in cache_env:
+            fail("{} cannot be set in go_env".format(key))
+        cache_env[key] = value
+    env_content = "\n".join(["{k}='{v}'\n".format(k = k, v = v) for k, v in cache_env.items()])
+
     ctx.file("go.env", env_content)
     ctx.file("BUILD.bazel", 'exports_files(["go.env"])')
 
@@ -72,6 +73,7 @@ go_repository_cache = repository_rule(
     attrs = {
         "go_sdk_name": attr.string(),
         "go_sdk_info": attr.string_dict(),
+        "go_env": attr.string_dict(),
     },
     # Don't put anything in environ. If we switch between the host cache
     # and Bazel's cache, it shouldn't actually invalidate Bazel's cache.
--- a/internal/go_repository_test.go
+++ b/internal/go_repository_test.go
@@ -17,6 +17,7 @@ package bazel_test
 
 import (
 	"bytes"
+	"io/ioutil"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -57,7 +58,12 @@ func main() {}
 	WorkspaceSuffix: `
 load("@bazel_gazelle//:deps.bzl", "gazelle_dependencies", "go_repository")
 
-gazelle_dependencies()
+gazelle_dependencies(
+	go_env = {
+		"GOPRIVATE": "example.com/m",
+		"GOSUMDB": "off",
+	},
+)
 
 # gazelle:repo test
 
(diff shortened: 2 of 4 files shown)


internal/go_repository_cache.bzl
```diff
@@ -64,6 +64,11 @@ GOCACHE='{gocache}'
         gopath = go_path,
         gocache = go_cache,
     )
+    for name, value in ctx.attr.go_env.items():
```
jayconrod, line 67: It would be more straightforward if the given dict and the cache dict were combined before formatting the file:

Something like:

```
cache_env = {
    "GOROOT": go_root,
    "GOPATH": go_path,
    "GOCACHE": go_cache,
}
for key, value in go_env.items():
    if key in cache_env:
        fail("{} cannot be set in go_env".format(key))
    cache_env[key] = value
env_content = "\n".join("{k}='{v}'\n".format(k, v) for k, v in go_env.items()]
```

internal/go_repository_test.go
```diff
@@ -161,6 +167,25 @@ func TestModcacheRW(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	goRoot := filepath.Join(outputBase, "external/go_sdk")
+	goPath := filepath.Join(outputBase, "external/bazel_gazelle_go_repository_cache")
+	goCache := filepath.Join(outputBase, "external/bazel_gazelle_go_repository_cache/gocache")
+	testtools.CheckFiles(t, goPath, []testtools.FileSpec{
+		{
+			Path:    "go.env",
+			Content: fmt.Sprintf("GOROOT='%s'\nGOPATH='%s'\nGOCACHE='%s'\nGOPRIVATE='example.com/m'\nGOSUMDB='off'", goRoot, goPath, goCache),
```
jayconrod, line 184: This might be a little brittle. Instead, let's read the file and check whether it contains `GOPRIVATE='example.com/m'`.

jayconrod (commented):
Thanks for working on this. Couple comments below.

internal/go_repository_cache.bzl
dirkboye: done.

internal/go_repository_test.go
dirkboye: done.

dirkboye:
i forced push the change as it was a small change according to comments. do you usually want fixups and a squash in the end?

jayconrod (approved):
Looks good, thanks!

jayconrod:
> i force-pushed the change as it was a small change according to comments. do you usually want fixups and a squash in the end?

For reference, I always squash when merging PRs, so each commit on the main branch corresponds to a PR. I'd prefer people to push additional commits on PRs, since GitHub doesn't let me see differences or respond to comment threads after a force push.

PR #977, opened 2020-12-31
Allow setting proto_strip_import_prefix from root BUILD.bazel

linzhp:
**What type of PR is this?**
Bug fix

language/proto

In `go_repository` rules, all Gazelle directives are at the root level BUILD.bazel. We should allow setting `proto_strip_import_prefix` at the root BUILD.bazel without warnings

### DIFF
--- a/language/proto/config.go
+++ b/language/proto/config.go
@@ -282,7 +282,7 @@ func checkStripImportPrefix(prefix, rel string) error {
 	if !strings.HasPrefix(prefix, "/") {
 		return fmt.Errorf("proto_strip_import_prefix should start with '/' for a prefix relative to the repository root")
 	}
-	if !pathtools.HasPrefix(rel, prefix[1:]) {
+	if rel != "" && !pathtools.HasPrefix(rel, prefix[1:]) {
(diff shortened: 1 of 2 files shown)


achew22:
@linzhp not a maintainer of Gazelle, but you might consider adding some tests to demonstrate the old broken behavior and ensure that there are no regressions.

language/proto/config.go
```diff
@@ -282,7 +282,7 @@ func checkStripImportPrefix(prefix, rel string) error {
 	if !strings.HasPrefix(prefix, "/") {
 		return fmt.Errorf("proto_strip_import_prefix should start with '/' for a prefix relative to the repository root")
 	}
-	if !pathtools.HasPrefix(rel, prefix[1:]) {
+	if rel != "" && !pathtools.HasPrefix(rel, prefix[1:]) {
```
jayconrod, line 285: If this ends up being repeated anywhere else, it might make sense  for `pathtools.HasPrefix` and `TrimPrefix` to support empty paths.

jayconrod (approved):
Looks good. Thanks for fixing this.

PR #979, opened 2021-01-06
correctly handle duplicate definitions in generate_repo_config

### LINKED ISSUE #968: duplicate go_repository rules are now rejected

0.22.2

0.25.0

3.6

As of some recent gazelle version (probably whatever picked up https://github.com/bazelbuild/bazel-gazelle/pull/855), duplicate go_repository rules are rejected in at least some combinations that were previously accepted.

The error will be reported as "multiple rules have the name %s". There turns out to be a range of scenarios where this can occur which are arguably incorrect. I have only partially debugged it so far, so I'm going to be a bit vague about what those scenarios are, but it includes things like:

- The WORKSPACE file includes both repositories defined in external _deps() functions and in go_dependencies.bzl generated by gazelle update-repos
- Some of those repository names match, and the ordering in WORKSPACE has been arranged so that gazelle's definitions override the previous ones.
- gazelle is told to only update go_dependencies.bzl

I believe this scenario should be accepted by gazelle, and currently (at least in some combinations, details TBD) it is rejected. Another similar scenario is where a `# gazelle:repository ` line is overriding something generated in go_dependencies.bzl.

I suspect, but have not yet proved, that this is happening because of the collapsing of definitions into @bazel_gazelle_go_repository_config losing information. Certainly in the cases I'm working on debugging there are two entries in that config which are literal duplicates of each other. My working hypothesis is that generate_repo_config is misinterpreting these scenarios as duplicate rules.

I'll do some more debugging tomorrow, but I figured I'd dump my work so far into this issue in case the problem is immediately obvious to you.

asuffield:
**What type of PR is this?**

Bug fix

go_repository

ListRepositories currently attempts to merge duplicates, but has several bugs in this code. Refactored the whole thing to have no duplicates in the output and to match bazel's own logic for picking the winner. In scenarios where there are both gazelle directives and rules, we prefer the gazelle directive. When multiple gazelle directives are found, we pick the last one; to make this work, ListRepositories is adjusted to maintain the ordering of rules.

This notably fixes the way `generate_repo_config` interacts with overrides between multiple `# gazelle:repository_macro` definitions, which currently breaks. It likely also fixes other corner cases where there are multiple definitions with different importpath, which would confound both update-repos and fix-update, although I don't have a reproduction case.

The test added to generate_repo_config_test.go is a minimal scenario that fails on the latest release and works after this change.

Fixes #968


repo/repo.go
```diff
@@ -71,72 +72,102 @@ func FindExternalRepo(repoRoot, name string) (string, error) {
...
+
+func isFromDirective(repo *rule.Rule) bool {
+	b, ok := repo.PrivateAttr(config.GazelleFromDirectiveKey).(bool)
+	if !ok {
+		return false
+	}
+	return b
+}
+
+func parseRepositoryDirectives(file *rule.File, recursive bool) (repos []repoDef, err error) {
```
jayconrod, line 136: I'd rather not support recursion here. It's just more complexity than I'm willing to deal with.

jayconrod (commented):
Thanks for fixing this, and sorry I haven't been able to review it earlier.

I think this would be a good bug to fix, but I really want to avoid introducing additional complexity. Let's just do the simple thing: don't support recursion and let the last definition override earlier definitions. I think that should cut down this PR's size significantly.

midnightconman:
I just ran into this issue today, is there a work around?

asuffield:
I totally remembered to finish this off.

Some refactoring is needed for more recent changes, I'll take another pass at it.

repo/repo.go
asuffield: This code has changed beyond recognition

repo/repo.go
```diff
@@ -71,72 +72,102 @@ func FindExternalRepo(repoRoot, name string) (string, error) {
...
+			// This name has not been seen before, append it
 			repos = append(repos, repo)
+			repoFileMap[name] = file
+			repoIndexMap[name] = len(repos) - 1
+			continue
 		}
-		repoFileMap[repo.Name()] = workspace
+
+		// We have a duplicate declaration of this name. The rest of this logic determines which one we want to keep.
+		other := repos[i]
```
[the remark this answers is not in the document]
asuffield: Done

asuffield:
The implementation has changed radically, but the description of the PR remains accurate: much refactoring of this code fixes a collection of bugs which add up to this problem.

repo/repo.go
```diff
@@ -217,17 +248,27 @@ func parseRepositoryDirectives(directives []rule.Directive) (repos []*rule.Rule,
 	return repos, nil
 }
 
+type RepoMacroDirective struct {
```
tfrench-uber, line 251: I really like making this a struct:
nit: Calling this a "Directive" is a bit confusing, since this is very specific to a `gazelle:repository_macro statement`. I think calling the struct a `RepoMacro` would be more accurate.

repo/repo.go
```diff
@@ -217,17 +248,27 @@ func parseRepositoryDirectives(directives []rule.Directive) (repos []*rule.Rule,
...
+	File     string
```
tfrench-uber, line 252: optional nit: `Path` or `FilePath` might be more clear to not confuse with `rule.File` structs

tfrench-uber: https://github.com/bazelbuild/bazel-gazelle/blob/master/rule/rule.go#L57-L61 uses `Path` and `DefName`, which would probably be good naming conventions to mirror and stay consistent with.

tfrench-uber:
Hey @asuffield, do you plan merge this in the short term? If not - I would be happy to take over this PR and get it merged. Please let me know what you prefer!

asuffield:
> Hey @asuffield, do you plan merge this in the short term? If not - I would be happy to take over this PR and get it merged. Please let me know what you prefer!

Go ahead, I've been swamped

kmicklas:
@blico This can be closed now that #1146 is merged.

PR #981, opened 2021-01-08
Allow FixLoads to fix any CallExpr and support requirement("name") style dependencies

### LINKED ISSUE #978: [Question] Emitting a macro call as a dependency
v0.22.0

v0.24.2

3.5.0

I'm writing a custom gazelle language extension for Python, and one pattern that all the python pip rule systems use is to hide repository names of installed pip packages behind a call to a `requirement` macro.

```py
load("@rules_python//python:defs.bzl", "py_binary", "py_library")
load("@pip3//:requirements.bzl", "requirement")

py_library(
    name = "mylibrary",
    srcs = [
        "__init__.py",
        "fun.py",
    ],
    deps = [
        "//other/label",
        requirement("numpy"),
    ],
)
```
I was able to make the py_library look correct using the `BzlExprValue` interface:
```go
// RequirementMacroExpr implements gazelle's BzlExprValue interface so that rule.SetAttr can
// write out requirement macros
type RequirementMacroExpr struct {
	pypiName string
}

// BzlExpr gets the starlark AST node for a requirement macro (e.g. `requirement("pypiName")`)
func (req *RequirementMacroExpr) BzlExpr() bzl.Expr {
	return &bzl.CallExpr{
		X:    &bzl.Ident{Name: "requirement"},
		List: []bzl.Expr{&bzl.StringExpr{Value: req.pypiName}},
	}
}
```

I've got the requirement macro call working, _but_ the symbol isn't `load`ed into the build file. I was hoping that `FixLoads` would magically somehow see the identifier and match it to the `rule.LoadInfo` I have for `requirement`, but it seems to only iterate on the rules in a file, which seems reasonable for all existing language plugins.

In essence, I'd like to have gazelle load the `requirement` symbol as it does for rules, but I'm not sure how to do that. Any suggestions? If need be, I would be happy to upstream changes necessary to `merger/fix.go`'s `FixLoads` function.

wolfd:
**What type of PR is this?**

Feature

all

This PR allows gazelle to merge, sort, and identify loads from "simple call expressions", i.e. starlark call expressions with a single string argument.

For example, rules_python uses a macro to expand a string to a full (but mangled) label, the mangling algorithm might change from rules_python version to version. I'm attempting to write a gazelle plugin for Python, and ran into a roadblock here. It felt like a much better idea to emit the macro than to have a moving target of emitting mangled names, mostly for git diff and developer readability reasons.

A simple example of a build file that a Python gazelle plugin might try to sync:

```
load("@rules_python//python:defs.bzl", "py_library")
load("@pip3//:requirements.bzl", "requirement")

py_library(
    name = "mylibrary",
    srcs = [
        "__init__.py",
        "fun.py",
    ],
    deps = [
        "//other/label",
        requirement("numpy"),
    ],
)
```

Currently gazelle allows for the rule syntax above to be emitted, but the macro won't be automatically loaded, and if being merged with an existing build file, the existing list gets clobbered, as gazelle can't handle lists of strings + other things. The change adds the ability for gazelle to reason about very simple call expressions, specifically ones with a single string argument, and allows those to be included in lists of strings.

Fixes #978

**Other notes for review**

This review ended up being fairly invasive as I wanted to support these [simpleValue](https://github.com/wolfd/bazel-gazelle/blob/a1020867fa059d6020ed70471c22b40f9b9931d6/rule/expr.go#L192-L198) expressions in a number of places. The `requirement("pypi_name")` pattern is common to a few different Python related rules, [the official rules](https://github.com/bazelbuild/rules_python/blob/bbe0712276ee2c9cd216323783fbdbfa15da7c99/examples/pip_install/BUILD#L34) and [a popular alternative way of using pip with the official rules](https://github.com/ali5h/rules_pip/blob/0559f4dfb1bfce200dc8db5d8f1c011d1eb1aaff/examples/src/BUILD#L9), but at a cursory glance, it seems somewhat unique to Python.

So, I'd be happy to discuss whether this method seems reasonable to you, and if need be, come up with alternatives.

Also, I think I'd want to add a test case to test the squash behavior, but wanted to get your opinion first on the rest of this.

### DIFF
--- a/merger/BUILD.bazel
+++ b/merger/BUILD.bazel
@@ -8,15 +8,18 @@ go_library(
     ],
     importpath = "github.com/bazelbuild/bazel-gazelle/merger",
     visibility = ["//visibility:public"],
-    deps = ["//rule"],
+    deps = [
+        "//rule",
+        "@com_github_bazelbuild_buildtools//build:go_default_library",
+    ],
 )
 
 go_test(
     name = "merger_test",
     size = "small",
     srcs = ["merger_test.go"],
-    embed = [":merger"],
     deps = [
+        ":merger",
         "//language",
         "//language/go",
         "//language/proto",
--- a/merger/fix.go
+++ b/merger/fix.go
@@ -58,7 +58,7 @@ func FixLoads(f *rule.File, knownLoads []rule.LoadInfo) {
 	}
 
 	// Make a map of all the symbols from known files used in this file.
-	// We walk the whole AST to find every call expression that LHS is an
+	// We walk the whole AST to find every call expression where LHS is an
 	// identifier.
 	usedKinds := make(map[string]map[string]bool)
 	bzl.WalkPointers(f.File, func(x *bzl.Expr, _ []bzl.Expr) {
--- a/merger/fix.go
+++ b/merger/fix.go
@@ -59,12 +59,13 @@ func FixLoads(f *rule.File, knownLoads []rule.LoadInfo) {
 	// Make a map of all the symbols from known files used in this file.
 	usedKinds := make(map[string]map[string]bool)
 	for _, r := range f.Rules {
-		kind := r.Kind()
-		if file, ok := knownKinds[kind]; ok && !otherLoadedKinds[kind] {
-			if usedKinds[file] == nil {
-				usedKinds[file] = make(map[string]bool)
+		for _, symbol := range r.CalledSymbols() {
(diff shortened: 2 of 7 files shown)


wolfd:
@googlebot I signed it!

wolfd:
Doing some more thinking, I think this isn't worth the change. My plan is to basically alias these `requirement` calls by writing out a file with rules that call `requirement`, and using these generated labels.

PR #986, opened 2021-01-25
Upgrade Google go-cmp to v0.5.4

achew22:
**What type of PR is this?**

Updating dep on go-cmp

### DIFF
--- a/deps.bzl
+++ b/deps.bzl
@@ -142,8 +142,8 @@ def gazelle_dependencies(
         go_repository,
         name = "com_github_google_go_cmp",
         importpath = "github.com/google/go-cmp",
-        sum = "h1:/exdXoGamhu5ONeUJH0deniYLWYvQwW66yvlfiiKTu0=",
-        version = "v0.4.1",
+        sum = "h1:L8R9j+yAqZuZjsqh/z+F1NCffTKKLShY6zXTItVIZ8M=",
+        version = "v0.5.4",
     )
 
     _maybe(
--- a/go.mod
+++ b/go.mod
@@ -9,7 +9,7 @@ require (
 	github.com/bmatcuk/doublestar v1.2.2
 	github.com/davecgh/go-spew v1.1.1 // indirect
 	github.com/fsnotify/fsnotify v1.4.7
-	github.com/google/go-cmp v0.4.1
+	github.com/google/go-cmp v0.5.4
 	github.com/kr/pretty v0.1.0 // indirect
 	github.com/pelletier/go-toml v1.2.0
 	github.com/pmezard/go-difflib v1.0.0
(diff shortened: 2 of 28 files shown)


jayconrod (commented):
Please also run `go mod vendor`. The gazelle binary that `go_repository` uses is built from vendored sources, and it's important that uses the same version of everything.

achew22:
Interesting, would it be better to have all of Gazelle us the vendored sources? Ran into this issue because this was effectively pinning my repo to an old version of go-cmp

jayconrod:
For the build failure, it looks like the build files need to be deleted from `vendor/github.com/bazelbuild/buildtools`. I forgot that step.

> Interesting, would it be better to have all of Gazelle us the vendored sources?

That would be ideal if Gazelle were fully self-contained. But because vendor directories are private to a repository, and because Gazelle can link in extension packages developed in other repos, we'd likely end up with multiple copies of the same package linked into the same binary. I'm not sure if Gazelle depends on anything where that would be really problematic, but I'd like to avoid it if possible.

I'd love to be able to get rid of the vendor directory, but I it would be hard to build `fetch_repo` and `gazelle` without it.

> Ran into this issue because this was effectively pinning my repo to an old version of go-cmp

If you declare the `go-cmp` `go_repository` before calling `gazelle_dependencies`, you should be able to use a higher version. `gazelle_dependencies` won't override anything that's already declared.

achew22:
> For the build failure, it looks like the build files need to be deleted from `vendor/github.com/bazelbuild/buildtools`. I forgot that step.

No worries. Easy fix.

> > Interesting, would it be better to have all of Gazelle us the vendored sources?
> 
> That would be ideal if Gazelle were fully self-contained. But because vendor directories are private to a repository, and because Gazelle can link in extension packages developed in other repos, we'd likely end up with multiple copies of the same package linked into the same binary. I'm not sure if Gazelle depends on anything where that would be really problematic, but I'd like to avoid it if possible.
>
> I'd love to be able to get rid of the vendor directory, but I it would be hard to build `fetch_repo` and `gazelle` without it.

Hrm... that's a good point. I'll think on it.
 
> > Ran into this issue because this was effectively pinning my repo to an old version of go-cmp
> 
> If you declare the `go-cmp` `go_repository` before calling `gazelle_dependencies`, you should be able to use a higher version. `gazelle_dependencies` won't override anything that's already declared.

I dont' think you can pre-declare any of these since I get `go_repository` from gazelle, which needs the `gazelle_dependencies`. Am I missing something?

jayconrod:
> I dont' think you can pre-declare any of these since I get go_repository from gazelle, which needs the gazelle_dependencies. Am I missing something?

You can declare a `go_repository` before calling `gazelle_dependencies`. Repository rules are evaluated lazily, so they can be declared in pretty much any order.

I think it's good to declare direct dependencies first, then indirect dependencies (inside a function like `gazelle_dependencies`) should be careful not to override them (using `_maybe`). Gazelle inserts new dependencies following that convention, but it's a fairly recent change.

jayconrod (approved):

PR #1015, opened 2021-03-19
language/go: don't let 'go mod download' edit go.sum

### LINKED ISSUE #990: language/go: update-repos adds sums to go.sum with Go 1.16

e4496b956eb46bdf8c9bf95b8d1d85e3a086c4be

46c56d280eeee5b6e71b710f9afb6ab49ffcdf51

4.0.0

yes

darwin amd64

### What did you do?

* Remove module content hashes from `go.sum` (those without the `/go.mod` suffix)
* Run `gazelle update-repos -from_file=go.mod`.

Gazelle does not modify `go.sum` or `go.mod`.

### What did you see instead?

Gazelle adds missing sums to `go.sum`.

This happens because Gazelle runs `go mod download` to find missing sums. In Go 1.16, `go mod download` adds module content sums to `go.sum`. Previously, `go mod download` only updated `go.sum` after loading the build list, so it would only add `/go.mod` hashes. That was probably a bug before we didn't notice.

jayconrod:
Starting in Go 1.16, 'go mod download' adds missing sums to go.sum.
We don't want that in update-repos though. 'go mod download' should be
run in a temporary directory with the same arguments.

Fixes #990

### DIFF
--- a/cmd/gazelle/integration_test.go
+++ b/cmd/gazelle/integration_test.go
@@ -4210,3 +4210,47 @@ go_library(
 `,
 	}})
 }
+
+// TestUpdateReposDoesNotModifyGoSum verifies that commands executed by
+// update-repos do not modify go.sum, particularly 'go mod download' when
+// a sum is missing. Verifies #990.
+//
+// This could also be tested in language/go/update_import_test.go, but that
+// test relies on stubs for speed, and it's important to run the real
+// go command here.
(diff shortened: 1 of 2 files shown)


achew22 (approved):
I love any change thats 2x tests to 1x code 😄

jayconrod:
> I love any change thats 2x tests to 1x code 😄

😁

sluongng:
It would be nice to get a minor tag release for this change

PR #1022, opened 2021-03-30
Respect .bazelignore

### LINKED ISSUE #318: Respect .bazelignore files
These were introduced in Bazel 0.17.1. They allow directories to be excluded.

Michaelhobo:
**What type of PR is this?**

Feature

cmd/gazelle

Gazelle now can recognize .bazelignore files

Fixes #318

**Other notes for review**

### DIFF
--- a/walk/config.go
+++ b/walk/config.go
@@ -16,8 +16,9 @@ limitations under the License.
 package walk
 
 import (
+	"bufio"
 	"flag"
-	"io/ioutil"
+	"fmt"
 	"log"
 	"os"
 	"path"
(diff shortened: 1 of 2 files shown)


walk/config.go
```diff
@@ -84,6 +87,28 @@ func (cr *Configurer) Configure(c *config.Config, rel string, f *rule.File) {
 	*wcCopy = *wc
 	wcCopy.ignore = false
 
+	// Extract patterns to ignore from .bazelrc, and add it to the list of excludes.
+
+	bazelIgnore, err := ioutil.ReadFile(path.Join(c.RepoRoot, ".bazelignore"))
```
achew22, line 92: A couple of proposed refactors that I think would make this clearer.

1. this process is all to set some additional fields on `wcCopy`. I think we can use that fact and extract this into its own function, which will be trivial to test in isolation if we deem that necessary 😄
2. I think loading the whole file into memory and splitting on newlines can probably be done more idiomatically with a `bufio.Scanner`[0].

What do you think about something like

```go
func (cr *Configurer) loadBazelIgnore(c *config.Config, wc *walkConfig, rel string) error {
    file, err := os.Open(path.Join(c.RepoRoot, ".bazelignore"))
    if err != nil && !os.IsNotExist(err) {
        return fmt.Errorf(".bazelignore exists but couldn't be read: %w", err)
    }
    defer file.Close()

    scanner := bufio.NewScanner(file)
    for scanner.Scan() {
       ignore := scanner.Text() // This is going to be each line of the file
       // Copy the major body of this change right here and it should work.
    }

    if err := scanner.Err(); err != nil {
        return err
    }
    return nil
}
```

Then you would invoke the new `loadBazelIgnore` where this code lives right now.

What do you think?

[0] https://golang.org/pkg/bufio/#Scanner

Michaelhobo: Thanks, scanner is definitely better, especially for larger files. I should get rid of my habit of loading the whole thing into memory

achew22 (approved):
Looks good to me, leaving it for 24h in case Jay has any more thoughts

achew22:
Thanks so much for contributing!

PR #1023, opened 2021-03-30
Support Label @repo -> @repo//:repo shorthand

wolfd:
Matches Bazel's label parsing behavior

**What type of PR is this?**

Bug fix

all

`label.Parse` doesn't handle shorthand for specifying `Repo` and `Name` together via `@name`

This behavior is present in Bazel itself:
- [test case](https://github.com/bazelbuild/bazel/blob/09c621e4cf5b968f4c6cdf905ab142d5961f9ddc/src/test/java/com/google/devtools/build/lib/cmdline/LabelTest.java#L56-L61)
- [relevant parsing code](https://github.com/bazelbuild/bazel/blob/09c621e4cf5b968f4c6cdf905ab142d5961f9ddc/src/main/java/com/google/devtools/build/lib/cmdline/Label.java#L137-L143)

Thought PR was small enough to not warrant an issue. Could open one if extra discussion is warranted.

**Other notes for review**


achew22 (approved):
Neat trick! LGTM but I'll give Jay 24h to review it in case he wants to weigh in.

jayconrod (approved):
LGTM too. Didn't know about that shorthand before now.

PR #1024, opened 2021-03-31
language/go: rewrite embedResolver to use a tree structure

### LINKED ISSUE #1013: language/go: `embedsrcs` not being generated for nested directories
0.23.0

0.26.0

4.0.0

Yes

macOS 11.2.3 amd64

### What did you do?

Given this small files as `main.go`

```go
package main

import (
	"embed"
	"fmt"
	"io"
)

//go:embed config/*
var config embed.FS

func main() {
	reader, err := config.Open("config/config.yaml")
	if err != nil {
		panic(err)
	}

	bytes, err := io.ReadAll(reader)
	if err != nil {
		panic(err)
	}

	fmt.Println(string(bytes))
}
```

and with this as `config/config.yaml`

```yaml
dev: true
```

running `go run main.go` properly prints out: `dev: true` as one would expect.

But when trying to use gazelle to generate a BUILD file `gazelle -repo_root . -go_prefix github.com/jacobfoard/demo` it prints out `gazelle: /Users/jacobfoard/go/src/github.com/jacobfoard/demo/main.go:9:2: pattern config/*: matched no files`

and generates this BUILD file

```bazel
load("@io_bazel_rules_go//go:def.bzl", "go_binary", "go_library")

go_library(
    name = "demo_lib",
    srcs = ["main.go"],
    importpath = "github.com/jacobfoard/demo",
)

go_binary(
    name = "demo",
    embed = [":demo_lib"],
)
```

I would expect the yaml file to be added into `embedsrcs`

```bazel
load("@io_bazel_rules_go//go:def.bzl", "go_binary", "go_library")

go_library(
    name = "demo_lib",
    srcs = ["main.go"],
    embedsrcs = ["config/config.yaml"],
    importpath = "github.com/jacobfoard/demo",
)

go_binary(
    name = "demo",
    embed = [":demo_lib"],
)
```

I was able to generate the file above by modifying the check for `matched` in the [embedResolver](https://github.com/bazelbuild/bazel-gazelle/blob/master/language/go/embed.go#L166-L169)
```go 
for i < len(er.files) {
	p := er.files[i].path

	if er.files[i].isDir {
		p += "/"
	}

	matched, _ := path.Match(embed.path, p)
	advance(matched)
}
```

jayconrod:
When embedResolver is created, we now gather all embeddable files and
directories into a tree instead of a list.

When embedResolver.resolve matches a pattern, it visits every node in
the tree. This ensures we can recursively embed all files in a
directory except hidden files.

Fixes #1013

### DIFF
--- a/language/go/embed.go
+++ b/language/go/embed.go
@@ -30,18 +30,26 @@ import (
 // embedResolver maps go:embed patterns in source files to lists of files that
 // should appear in embedsrcs attributes.
 type embedResolver struct {
-	// files is a list of embeddable files and directories, stored in
-	// depth-first pre-order (nearly the same order as filepath.Walk, except
-	// files in the top-level directory appear before subdirectories.
-	files []embeddableFile
+	// files is a list of embeddable files and directory trees, rooted in the
+	// package directory.
+	files []*embeddableFile
 }
 
 type embeddableFile struct {
-	path  string
-	isDir bool
+	path    string
+	entries []*embeddableFile // non-nil for directories
 }
 
-// newEmbedResolver builds a list of files that may be embedded. This is
+func (f *embeddableFile) isDir() bool {
+	return f.entries != nil
+}
+
+func (f *embeddableFile) isHidden() bool {
+	base := path.Base(f.path)
+	return strings.HasPrefix(base, ".") || strings.HasPrefix(base, "_")
+}
+
+// newEmbedResolver builds a set of files that may be embedded. This is
 // approximately all files in a Bazel package including explicitly declared
 // generated files and files in subdirectories without build files.
 // Files in other Bazel packages are not listed, since it might not be possible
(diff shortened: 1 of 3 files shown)


language/go/embed.go
```diff
@@ -30,18 +30,26 @@ import (
...
-	// files is a list of embeddable files and directories, stored in
-	// depth-first pre-order (nearly the same order as filepath.Walk, except
-	// files in the top-level directory appear before subdirectories.
-	files []embeddableFile
+	// files is a list of embeddable files and directory trees, rooted in the
+	// package directory.
+	files []*embeddableFile
 }
 
 type embeddableFile struct {
```
achew22, line 38: Is this still best thought of as an `embeddableFile` or is it now an `embeddableNode` since it's a tree?

achew22: Or maybe `embeddableFileNode`

achew22 (approved):

PR #1031, opened 2021-04-15
Change merge behavior to remove attrs instead of panicking

wolfd:
**What type of PR is this?**

Bug fix

all

I have an optional list argument that may or may not exist on a rule, in the current configuration, `mergeExprs` can return nil if it encounters an empty destination list, which panics when passed to `SetAttr`.

I think a simple fix for this is to `DelAttr` on values that return `nil` (with no errors) in `mergeExprs`.

Didn't make an issue for this.

**Other notes for review**

### DIFF
--- a/merger/merger_test.go
+++ b/merger/merger_test.go
@@ -20,7 +20,7 @@ import (
 	"testing"
 
 	"github.com/bazelbuild/bazel-gazelle/language"
-	"github.com/bazelbuild/bazel-gazelle/language/go"
+	golang "github.com/bazelbuild/bazel-gazelle/language/go"
 	"github.com/bazelbuild/bazel-gazelle/language/proto"
 	"github.com/bazelbuild/bazel-gazelle/merger"
 	"github.com/bazelbuild/bazel-gazelle/rule"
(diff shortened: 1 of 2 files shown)


achew22 (changes requested):
Could you add a test case that fails before this change was applied and demonstrates how your change fixes it? I'm having a hard time conceptualizing how this would affect things without more context.

wolfd:
What I realized is that due to how `mergeExprs` works, it's impossible to set an attr in `MergeableAttrs` to an empty list because [`mergeList` returns `nil` if the merged list is length 0](https://github.com/bazelbuild/bazel-gazelle/blob/351d2e65f53407c348dc95e73ec73865478c24c5/rule/merge.go#L180-L182). If `dst`'s attr is a list with items and `src`'s attr is a list without items, `mergeExprs` returns `nil, nil`. There's no error, so it just passes `nil` into `SetAttr`, which then panics.

In my plugin right now, I just check if we're trying to set a mergeable attr to an empty list, and call `DelAttr` directly to prevent this panic from happening.

The other use of `mergeExprs` in `MergeRules` [currently has the same condition](https://github.com/bazelbuild/bazel-gazelle/blob/351d2e65f53407c348dc95e73ec73865478c24c5/rule/merge.go#L58-L60), actually, and treats `nil` as an indicator to call `DelAttr` (although it also passes `nil` as the source attr value).

Regardless, adding a test case is a good idea :smiley:. I confirmed that it does panic without the added condition.

achew22 (approved):

achew22:
Thanks so much for adding the test! That will help us have stuff like this never happen again in the future.

PR #1033, opened 2021-04-18
Add per-file mode for proto_library generation

### LINKED ISSUE #1032: proto plugin should support a library-per-file mode
The docs for [`proto_library` state](https://docs.bazel.build/versions/master/be/protocol-buffer.html#proto_library) that the recommended organization is to make a `proto_library` rule for every file:

> Recommended code organization:
>
>  - One proto_library rule per .proto file.
>  - A file named foo.proto will be in a rule named foo_proto, which is located in the same package.
>  - A [language]_proto_library that wraps a proto_library named foo_proto should be called foo_[language]_proto, and be located in the same package.

Currently the only way to get more fine-grained `proto_library`s is to:
- move proto files around to different directories.
- use `package` mode and change package names for all proto files that you want to bundle differently.
- use `package` mode with `proto_group` option enabled, and add that option to your proto files.

I propose a new `file` mode for the proto plugin that follows the recommendations above. The rule name/grouping could still be overridden via an `option` like in `package` mode, if need be.

wolfd:
**What type of PR is this?**

> Feature

> language/proto

Allows `proto_library` rules to be instantiated per-file, which helps in situations where `package` is too coarse. Can also alleviate pain when files between two packages form import cycles when bundled at the package level instead of the file level, if you find yourself in that unfortunate situation.

Fixes #1032

**Other notes for review**

### DIFF
--- a/language/proto/testdata/file_mode/BUILD.want
+++ b/language/proto/testdata/file_mode/BUILD.want
@@ -1,0 +1,13 @@
+load("@rules_proto//proto:defs.bzl", "proto_library")
+
+proto_library(
+    name = "bar_proto",
+    srcs = ["bar.proto"],
+    visibility = ["//visibility:public"],
+)
+
+proto_library(
+    name = "foo_proto",
+    srcs = ["foo.proto"],
+    visibility = ["//visibility:public"],
+)
--- a/language/proto/testdata/file_mode/foo.proto
+++ b/language/proto/testdata/file_mode/foo.proto
@@ -1,0 +1,3 @@
+syntax = "proto3";
+
+package file_mode;
--- a/language/proto/config.go
+++ b/language/proto/config.go
@@ -97,6 +97,9 @@ const (
 	// PackageMode generates a proto_library for each set of .proto files with
 	// the same package name in each directory.
 	PackageMode
+
+	// FileMode generates a proto_library for each .proto file.
+	FileMode
 )
 
 func ModeFromString(s string) (Mode, error) {
@@ -111,6 +114,8 @@ func ModeFromString(s string) (Mode, error) {
 		return LegacyMode, nil
 	case "package":
 		return PackageMode, nil
+	case "file":
+		return FileMode, nil
 	default:
 		return 0, fmt.Errorf("unrecognized proto mode: %q", s)
 	}
@@ -128,6 +133,8 @@ func (m Mode) String() string {
 		return "legacy"
 	case PackageMode:
 		return "package"
+	case FileMode:
+		return "file"
 	default:
 		log.Panicf("unknown mode %d", m)
 		return ""
--- a/language/proto/generate.go
+++ b/language/proto/generate.go
@@ -126,11 +126,17 @@ func buildPackages(pc *ProtoConfig, dir, rel string, protoFiles, genFiles []stri
 					break
 				}
 			}
+		} else if pc.Mode == FileMode {
+			key = strings.TrimSuffix(name, ".proto")
 		}
+
 		if packageMap[key] == nil {
 			packageMap[key] = newPackage(info.PackageName)
 		}
 		packageMap[key].addFile(info)
+		if key != info.PackageName {
+			packageMap[key].RuleName = key
+		}
 	}
 
 	switch pc.Mode {
@@ -147,7 +153,7 @@ func buildPackages(pc *ProtoConfig, dir, rel string, protoFiles, genFiles []stri
 		}
 		return []*Package{pkg}
 
-	case PackageMode:
+	case PackageMode, FileMode:
 		pkgs := make([]*Package, 0, len(packageMap))
 		for _, pkg := range packageMap {
 			pkgs = append(pkgs, pkg)
@@ -212,7 +218,7 @@ func generateProto(pc *ProtoConfig, rel string, pkg *Package, shouldSetVisibilit
 	if pc.Mode == DefaultMode {
 		name = RuleName(goPackageName(pkg), pc.GoPrefix, rel)
 	} else {
-		name = RuleName(pkg.Options[pc.groupOption], pkg.Name, rel)
+		name = RuleName(pkg.RuleName, pkg.Name, rel)
 	}
 	r := rule.NewRule("proto_library", name)
 	srcs := make([]string, 0, len(pkg.Files))
--- a/language/proto/package.go
+++ b/language/proto/package.go
@@ -21,6 +21,7 @@ import "path/filepath"
 // same package name. This translates to a proto_library rule.
 type Package struct {
 	Name        string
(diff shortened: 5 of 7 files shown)


README.rst
```diff
@@ -721,6 +721,8 @@ The following directives are recognized:
 | * ``default``: ``proto_library``, ``go_proto_library``, and ``go_library``                 |
 |   rules are generated using ``@io_bazel_rules_go//proto:def.bzl``. Only one                |
 |   of each rule may be generated per directory. This is the default mode.                   |
+| * ``file``: a ``proto_library`` rule is generated for every .proto file.                   |
+|   .proto files can be grouped using ``proto_group`` option.                                |
```
jayconrod, line 725: This doesn't really make sense to me. If `file` mode puts one `.proto` file in each `proto_library` target, it doesn't seem meaningful to group them at all. Why would `proto_group` have any effect?

jayconrod (commented):
Thanks for working on this! Looks pretty good. One question below though.

README.rst
wolfd: It was mostly a side-effect of the way I implemented the per-file option, but I think it's pretty useful. The `proto_group` option allows you to override the default name of the `proto_library` rule. If for some reason someone wanted to logically group proto files, but still default to single proto file libraries, this allows that.

You could also use this option to rename a `proto_library`, if that was desired for whatever reason.

I think I could improve my wording here. How about:

> The `proto_library` a .proto file ends up in can be overridden by using the `proto_group` option.

jayconrod: > The proto_group option allows you to override the default name of the proto_library rule.

I wonder if a better way to do that in this mode would be to match on `srcs` when finding targets to merge? In language/proto/kinds.go, it looks like `proto_library` will already do that. That seems a much more intuitive workflow: users can just change the name of the target in the build file, and Gazelle keeps using the new name thereafter. `proto_group` seems much less intuitive to me for this use case.

> If for some reason someone wanted to logically group proto files, but still default to single proto file libraries, this allows that.

Gazelle's probably a poor fit for that level of customization. I'd recommend manually maintaining build files for directories like this.

In any case, if `proto_group` does this, it needs to be tested to make sure it doesn't break later on.

wolfd: Thanks for bringing this up. After doing some more digging I found that the `proto_group` method didn't even work in file mode :facepalm: (else if vs if). Untested code is broken code... Anyway, I dropped the docs about it and the (faulty) code that would have made it work, as suggested.

I was unaware of `MatchAttrs`, that should handle most use cases that I was thinking about.

wolfd: @jayconrod have you had a chance to look at this?

jayconrod: Regrettably, I no longer have bandwidth to work on Gazelle. I'll add @bazelbuild/go-maintainers in case anyone else can take a look.

wolfd: @achew22 are you able to take a quick look at this?

language/proto/testdata/file_mode/foo.proto
```diff
@@ -0,0 +1,3 @@
+syntax = "proto3";
+
+package file_mode;
```
achew22, line 3: Can you have one of these import from the other to confirm that the dep detection logic works

wolfd: I have added an import statement to this file, which could test that I haven't broken the parsing logic, but fwiw these tests are just for the generate behavior, not resolve. I haven't made any changes to the resolve behavior, which is tested separately.

Right now I don't believe that resolve is affected by mode changes at all. I could loop through all the available modes on the config here to test that the existing tests pass with all modes? https://github.com/bazelbuild/bazel-gazelle/blob/b1e37463c2ef0997c5cf0e90cdf03e696caaed27/language/proto/resolve_test.go#L357

What do you think?

pcj:
FWIW I've tested this branch out in a sample repo https://github.com/magma/magma and it seems to work as intended.  I'm +1 to get this merged.

danny-skydio:
Ah, just noticed that @achew22 responded to me about the patch in the Bazel Slack a while back. I need to just add a better test for the resolve behavior and then it should be good to go.

language/proto/testdata/file_mode/foo.proto
danny-skydio: Re: your comment in Slack
> I'm not concerned about it changing the result mode, I want to make sure that it actually does the resolve correctly when there is single file target generation

I've added a simple test that imports a single proto given an index of three different protos with separate `proto_library`s.

The previous test that I added for the comment above has been removed.

Should be good now? @achew22

language/proto/testdata/file_mode/BUILD.want
```diff
@@ -0,0 +1,13 @@
...
+
+proto_library(
+    name = "bar_proto",
+    srcs = ["bar.proto"],
+    visibility = ["//visibility:public"],
+)
+
+proto_library(
+    name = "foo_proto",
+    srcs = ["foo.proto"],
```
achew22, line 11: Shouldn't this have a `deps` entry for `":bar_proto"`?

wolfd: Hmm, I don't think so. The resolve tests are entirely separate from the generate tests, is that where the confusion is from?

I think I should remove the import there (just did so in https://github.com/bazelbuild/bazel-gazelle/pull/1033/commits/a5bd0e0069da3a0b1c6cf704d8bf12f848756989), as I think it's just confusing.
> I have added an import statement to this file, which could test that I haven't broken the parsing logic, but fwiw these tests are just for the generate behavior, not resolve. I haven't made any changes to the resolve behavior, which is tested separately.

To be clear, it was possible to have many different `proto_library` packages in a directory before, but gazelle wouldn't auto-generate them for you, you would need to use different `package`s (in the same dir) or an `option` in the proto file that defined a grouping behavior (with appropriate gazelle config changes). The resolve behavior for single proto file `proto_library`s should have worked before and after this change regardless of the change to the generate.go behavior.

Anyway, the new entry in resolve_test.go should be what you're asking for?

As it stands there isn't an end-to-end test for the proto language plugin. I have thought about adding something like it, but it would be a larger change and deserve it's own PR. Basic thought would be to have the testdata directory, but instead of BUILD.old and BUILD.want, you'd have BUILD.old, BUILD.generate, BUILD.resolve, (and maybe BUILD.merge?). I would need to dig into it a bit more, but it would be refactoring all of the existing tests in this plugin (and maybe also for the go plugin).

wolfd: As a counterpoint, adding a new stage for those tests might be redundant. The resolve_test.go file defines a similar structure that isn't backed by .proto source files, and it can create arbitrarily complex situations for the resolver that may not be output by generate.go.

Michaelhobo:
Any chance we can cut a new release with this change?

PR #1042, opened 2021-05-05
Galleze -> Gazelle

tanyabouman:
**What type of PR is this?**
Documentation

Fixes a typo


tanyabouman:
@googlebot I signed it!

jayconrod (approved):
Thanks for fixing this. Looks like that's been there a while without anyone noticing.

PR #1046, opened 2021-05-14
Updated documentation of go_prefix flag

tanyabouman:
**What type of PR is this?**
Documentation

It updates the description of the `go_prefix` flag, since the [go_prefix rule no longer exists](https://github.com/bazelbuild/rules_go/issues/721).  Hopefully my understanding is correct.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -381,9 +381,9 @@ The following flags are accepted:
 | :flag:`-go_prefix example.com/repo`                          |                                        |
 +--------------------------------------------------------------+----------------------------------------+
 | A prefix of import paths for libraries in the repository that corresponds to                          |
-| the repository root. Equivalent to the ``# gazelle:prefix`` directive and the                         |
-| ``prefix`` attribute of the gazelle rule. If neither of those are set, this                           |
-| option is mandatory.                                                                                  |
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -381,8 +381,9 @@ The following flags are accepted:
 | :flag:`-go_prefix example.com/repo`                          |                                        |
 +--------------------------------------------------------------+----------------------------------------+
 | A prefix of import paths for libraries in the repository that corresponds to                          |
-| the repository root. Gazelle infers this from the ``go_prefix`` rule in the                           |
-| root BUILD.bazel file, if it exists. If not, this option is mandatory.                                |
+| the repository root. Gazelle infers this from the ``# gazelle:prefix`` directive                      |
```
jayconrod, line 384: It would be more accurate to say that setting `-go_prefix` on the command line is equivalent to putting a `# gazelle:prefix` directive in the root build file. The `prefix` attribute on the `gazelle` rule actually causes `gazelle` to be invoked with this flag.

tanyabouman: Ah, I think I get it.  It's updated.  Does it makes sense now?

jayconrod: That works. Thanks!

PR #1047, opened 2021-05-17
Include reference to proto_import_prefix Gazelle directive

tanyabouman:
**What type of PR is this?**
Documentation

It includes a reference to the Gazelle directive documentation in `proto_import_prefix` flag description, because there is more explanation of `proto_import_prefix` in the directives section.


jayconrod (approved):
Looks good, thanks!

PR #1053, opened 2021-05-24
Allow ListRepositories to load macros not in Workspace

### LINKED ISSUE #1049: ListRepositories can't load macro directives from outside of workspace file

v0.23.0

v0.24.13

4.0.0

Yes

Mac OS 11.2.3 Intel

### What did you do?

When trying to load many repository_macro files. I added a `repository_macro` directive from outside of the workspace file. Bazel can properly resolve and use dependencies through gazelle rules that are loaded from other macro files, so `gazelle update-repos` should respect the placement of these directives and load them with `ListRepositories(...)`. 

Proper loading of rules from the `repository_macro` file

### What did you see instead?

The rules were not found by ListRepositories

tfrench-uber:
**What type of PR is this?**

Feature

repo

The PR allows a user to load other repository macros from within a top-level repository macro, so that it does not have to be added to the Workspace file. These macros with a second layer are identified by prepending a "+" to the beginning of the path in the repository_macro directive:

e.g. `# gazelle:repository_macro +repos.bzl%go_repositories` would load `repos.bzl%go_repositories` and any other files loaded from within the macro.

This is needed because the functionality already exists and is legal in bazel: A `go_repository( ... )` rule can be added to a macro, loaded by another macro, and loaded by the workspace. However, gazelle will not resolve this repository, and create a duplicate rule, thereby creating incorrect/duplicate Starlark content.

In very large repositories, this is important when dependency management needs to happen in many different file spaces.
This change uses a highly conservative opt-in approach, although since it is legal in Bazel, perhaps this "opt-in" requirement can be removed in the future. With this implementation, ListRepositories will correctly list all of the `go_repository( ` imports in that file, whether directly or loaded from another macro, as Bazel would similarly load.

For all existing uses (macro files existing as references only from Workspace file), no changes would be made. Only in cases where a user specifically decides to load another macro, and identify this desire in the gazelle directive.

Fixes #1049


linzhp:
Is it safe to assume all repository_macros only contains repository definitions or other repository_macros? In that case, we don't need to specify `# gazelle:repository_macro`  in a repository macro: if a statement is not a `Rule`, then we recursively load it as a repository  macro

tfrench-uber:
> Is it safe to assume all repository_macros only contains repository definitions or other repository_macros? In that case, we don't need to specify `# gazelle:repository_macro` in a repository macro: if a statement is not a `Rule`, then we recursively load it as a repository macro

In my personal opinion, I feel that it would be not safe to do this because it opens the potential for more unnecessary recursive calls. Firstly, this function finds all the `Rule` direct loads of repositories (which are Bazel lingo not gazelle). Next it checks for the gazelle directives only of the type `go_repository`. I think the best approach is the more conservative possible, the better. Since rules could be of other types, and directives could be unrelated. With this approach, the user has to specify, the same way that they are already required to do from the Workspace file, that this other file is a repository macro, only for loading go repositories (since directives are supposed to be primarily handwritten).

repo/repo.go
```diff
@@ -98,33 +91,95 @@ func ListRepositories(workspace *rule.File) (repos []*rule.Rule, repoFileMap map
...
+func loadRepositoriesFromMacro(la *loadArgs) error {
+	macroFile, err := rule.LoadMacroFile(la.f, "", la.defName)
+	if err != nil {
+		return err
+	}
+	for _, repo := range macroFile.Rules {
+		if name := repo.Name(); name != "" {
+				la.repos = append(la.repos, repo)
+				la.repoFileMap[name] = macroFile
+				la.repoIndexMap[name] = len(la.repos) - 1
```
tfrench-uber, line 139: Indentation

linzhp:
@jayconrod We are adding a new syntax 

```
# gazelle:repository_macro +some/file.bzl%go_repositories
```

The `+` sign tells Gazelle to look for other repository macros called by this repository macro. It's to avoid unnecessary recursion into non-repository macros. Do you think it a good idea? Would you prefer a simpler assumption that all repository macros should **only** contain either repository rules or other repository macros?

repo/repo.go
```diff
@@ -98,33 +91,95 @@ func ListRepositories(workspace *rule.File) (repos []*rule.Rule, repoFileMap map
...
+				la.checked[la.f+"%"+la.defName] = true
+				if err := loadRepositoriesFromMacro(la); err != nil {
+					return err
 				}
-				repoFileMap[repo.Name()] = macroFile
 			}
 		}
 	}
-	return repos, repoFileMap, nil
+	la.repos, err = loadExtraRepos(macroFile, la.repos, la.repoFileMap, la.repoIndexMap)
```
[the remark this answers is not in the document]
tfrench-uber: This is for backwards compatibility: Gazelle still operates normally with directives declared using a `# gazelle:repository` macro instead of a bazel rule, so I simply just wanted to still allow this. For example, see repositories declared in https://github.com/bazelbuild/bazel-gazelle/blob/master/WORKSPACE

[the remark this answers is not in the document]
linzhp: I see. gazelle:repository directive was already supported in macros before this change

internal/go_repository_config.bzl
```diff
@@ -40,13 +38,19 @@ def _go_repository_config_impl(ctx):
         )
         if result.return_code:
             fail("generate_repo_config: " + result.stderr)
+        if result.stdout:
+            for f in result.stdout.split("\n"):
```
linzhp, line 42:
```suggestion
            for f in result.stdout..splitlines():
```

cmd/generate_repo_config/generate_repo_config.go
```diff
@@ -60,22 +63,26 @@ func main() {
 	if flag.NArg() != 0 {
 		log.Fatal("generate_repo_config does not accept positional arguments")
 	}
-	if err := generateRepoConfig(*configDest, *configSource); err != nil {
+	macros, err := generateRepoConfig(*configDest, *configSource)
+	if err != nil {
 		log.Fatal(err)
 	}
+	for _, m := range macros {
+		fmt.Fprintf(os.Stdout, m)
```
linzhp, line 71:
```suggestion
		fmt.Fprintlnf(os.Stdout, m)
```

linzhp (commented):
This PR keeps failing on Windows. Please fix the test on Windows.

linzhp (approved):
@jayconrod @blico I don't see further issues. Do you want to take another look?

repo/repo.go
```diff
@@ -98,33 +91,94 @@ func ListRepositories(workspace *rule.File) (repos []*rule.Rule, repoFileMap map
...
+			// TODO: Also handle the case where one macro calls another macro in the same bzl file
+			if len(f) == 0 {
+				continue
 			}
-			for _, repo := range extraRepos {
-				if i, ok := repoIndexMap[repo.Name()]; ok {
-					repos[i] = repo
-				} else {
-					repos = append(repos, repo)
+			if !la.checked[la.f+"%"+la.defName] {
```
blico, line 155: I think we want this instead?
```
if !la.checked[f + "%" + la.defName]
```

repo/repo.go
```diff
@@ -98,33 +91,94 @@ func ListRepositories(workspace *rule.File) (repos []*rule.Rule, repoFileMap map
...
+
+type loadArgs struct {
+	leveled bool
+	workspace string
+	f string
+	defName string
+	repos []*rule.Rule
+	repoFileMap map[string]*rule.File
+	repoIndexMap map[string]int
+	checked map[string]bool
```
blico, line 127: NIT: `visited` is a little more accurate than `checked`

repo/repo.go
```diff
@@ -98,33 +91,94 @@ func ListRepositories(workspace *rule.File) (repos []*rule.Rule, repoFileMap map
...
+}
+
+func loadRepositoriesFromMacro(la *loadArgs) error {
```
blico, line 130: I think the following function signature would be a little less error-prone:
```
func loadRepositoriesFromMacro(la *loadArgs, leveled bool, f, defName string) error
```
`loadArgs` are things that are shared between the recursive calls, whereas the explicitly passed in parameters change between every recursive call

blico (commented):
Please also modify the `repository_macro` doc in `readme.rst` to describe the added functionality

cmd/generate_repo_config/generate_repo_config.go
```diff
@@ -124,8 +131,17 @@ func generateRepoConfig(configDest, configSource string) error {
 	buf.WriteString("\n")
 	buf.Write(destFile.Format())
 	if err := ioutil.WriteFile(configDest, buf.Bytes(), 0666); err != nil {
-		return err
+		return nil, err
+	}
+
+	var macros []string
+	for _, m := range sortedFiles {
+		f, err := filepath.Rel(filepath.Dir(configSource), m.Path)
```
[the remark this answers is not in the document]
tfrench-uber: This ran into problems testing because the output dir is not set to the repository (workspace) root, instead it is set to a temp file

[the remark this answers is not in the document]
tfrench-uber: This was actually found to be an issue based on how ListRepositories always prepends the config-source to the macro file path, since it is passed `workspace=config_path` as an argument. Thus, it is still safe to `TrimPrefix` of the config source
See https://github.com/bazelbuild/bazel-gazelle/issues/1068

repo/repo.go
```diff
@@ -98,33 +91,94 @@ func ListRepositories(workspace *rule.File) (repos []*rule.Rule, repoFileMap map
...
+	macroFile, err := rule.LoadMacroFile(la.f, "", la.defName)
+	if err != nil {
+		return err
+	}
+	for _, repo := range macroFile.Rules {
+		if name := repo.Name(); name != "" {
+			la.repos = append(la.repos, repo)
+			la.repoFileMap[name] = macroFile
+			la.repoIndexMap[name] = len(la.repos) - 1
+		} else if la.leveled {
```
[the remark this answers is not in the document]
tfrench-uber: Yes, Also we need to `continue` if the rule isn't a call (has `name != ""`), I do this by also adding a `continue` to previous `if`

cmd/generate_repo_config/generate_repo_config.go
```diff
@@ -17,7 +17,9 @@ limitations under the License.
 // WORKSPACE and generates a stripped version of the file. The generated
 // config file should contain only the information relevant to gazelle for
 // dependency resolution, so go_repository rules with importpath
-// and name defined, plus any directives.
+// and name defined, plus any other files that load repsitories to this
```
blico, line 20: NIT/wording: The macro files aren't included in the generated config file, they are printed to stdout by this command to be read by the `go_repository_config` rule.

cmd/gazelle/update-repos.go
```diff
@@ -481,7 +481,8 @@ func ensureMacroInWorkspace(uc *updateReposConfig, insertIndex int) (updated boo
 	// be called somewhere else.
 	macroValue := uc.macroFileName + "%" + uc.macroDefName
 	for _, d := range uc.workspace.Directives {
-		if d.Key == "repository_macro" && d.Value == macroValue {
+		m, defName, _, _ := repo.ParseRepositoryMacroDirective(d.Value)
```
blico, line 484: We should avoid calling `ParseRepositoryMacroDirective` unless `d.Key == "repository_macro"` (sadly leading to an extra nested `if`):
```
if d.Key == "repository_macro" {
   if m, defName, _, _ := repo. ParseRepositoryMacroDirective(d.Value); m == uc.macroFileName && defName == uc.macroDefName
}
```

internal/go_repository_test.go
```diff
@@ -113,15 +113,80 @@ func TestDirectives(t *testing.T) {
...
-	if err := bazel_testing.RunBazel("build", "@bazel_gazelle_go_repository_config//:all"); err != nil {
-		t.Fatal(err)
-	}
+func TestRepoConfig(t *testing.T, m *testing.M) {
 	outputBase, err := getBazelOutputBase()
 	if err != nil {
 		t.Fatal(err)
 	}
 	outDir := filepath.Join(outputBase, "external/bazel_gazelle_go_repository_config")
+	testArgs := bazel_testing.Args{
```
blico, line 122: Any reason why we are creating a new `bazel_testing.Args` here rather than using the one defined at the top of this file?

repo/repo.go
```diff
@@ -71,62 +71,126 @@ func FindExternalRepo(repoRoot, name string) (string, error) {
...
+				break
+			}
+		}
+		// TODO: Also handle the case where one macro calls another macro in the same bzl file
+		if f == "" {
+			continue
+		}
+		if !la.visited[f+"%"+defName] {
+			la.visited[f+"%"+defName] = true
+			if err := loadRepositoriesFromMacro(la, false, f, defName); err != nil {
```
blico, line 155: NIT: https://github.com/uber-go/guide/blob/master/style.md#avoid-naked-parameters
```
if err := loadRepositoriesFromMacro(la, false /* leveled */, f, defName); err != nil {
```

repo/repo.go
```diff
@@ -71,62 +71,126 @@ func FindExternalRepo(repoRoot, name string) (string, error) {
...
 			}
-			for _, repo := range extraRepos {
-				if i, ok := repoIndexMap[repo.Name()]; ok {
-					repos[i] = repo
-				} else {
-					repos = append(repos, repo)
-				}
-				repoFileMap[repo.Name()] = macroFile
-			}
+			repos = la.repos
```
blico, line 108: Hmmm, I think it might be a little clearer if we update `loadRepositoriesFromMacro` to return just the `repos` that it sees. Then we can remove `repos` from `loadArgs` and from here we can call:
```
repos = append(repos, loadedRepos...)

**Review by blico — COMMENTED** [review]
Looks good! Just a few more minor comments.

File: internal/go_repository_test.go (hunk 12 above)
Comment (tfrench-uber) [reply internal/go_repository_test.go]: This was an accident: was trying to update the test configuration to address the config issue

File: repo/repo.go (hunk 14 above)
Comment (tfrench-uber) [reply repo/repo.go:108]: We are unable to do this because `len(repos)` is required to update `repoIndexMap`, so therefore a complete version of repos must be passed every time we are gathering new repos for `repoFileMap`. This is so a repo can easily be switched without needing to search through the `repos`. Used here:
```
	for _, repo := range extraRepos {
		if i, ok := repoIndexMap[repo.Name()]; ok {
			(*repos)[i] = repo
		} else {
			*repos = append(*repos, repo)
		}
		repoFileMap[repo.Name()] = f
	}
```

blico (approved):

PR #1055, opened 2021-05-27
Add reference to bazel-skylib Gazelle plugin

achew22:
Also removes reference to abandoned `rules_sass` plugin.


jayconrod (approved):
Thanks, Skylib should definitely be in here.

achew22:
These feel kind of hidden away. What would you think about making the list of supported languages more prominent (in a separate CL)?

jayconrod:
Making these more prominent (in README?) sounds good to me.

(Feel free to merge when ready; you should be able to)

PR #1056, opened 2021-05-28
Include links to go_naming_convention directive

tanyabouman:
**What type of PR is this?**
Documentation

It includes a reference to the Gazelle directive documentation in the `build_naming_convention` attribute of the `go_repository` rule and the `go_naming_convention` flag description, because there is more explanation of `go_naming_convention` in the directives section.


achew22:
Thank you so much for contributing to the documentation!

PR #1061, opened 2021-06-07
gazelle: fix #902, cache modpath for super large repos

### LINKED ISSUE #902: go_repository is very slow for some packages
v0.24.1

v0.22.0

3.5.0

The versions used are latest

MacOS/amd64

### What did you do?
Generate repositories.bzl with gazelle from go.mod
Run `bazel build //...`

Go repositories to be fetched quickly

### What did you see instead?
Package github.com/docker/docker needs over 90s to be fetched

Screenshot 2020-09-08 at 12 10 59

lenage:
**What type of PR is this?**

> Performance Improvement

> cmd/gazelle

This PR fixes an issue which `bazel run //:gazelle` take lots of time when an external repo has many directories(for example 2000+), and fix as below:

1. only call NewRemoteCache once during walk directories.
2. cache `fetch_repo` when `repo.NewRemoteCache`: which is called `go get -d` to download source code
3. also cache modpath to avoid too many `go list` call

Fixes #902

### DIFF
--- a/cmd/gazelle/fix-update.go
+++ b/cmd/gazelle/fix-update.go
@@ -272,6 +272,16 @@ func runFixUpdate(wd string, cmd command, args []string) (err error) {
 	// Visit all directories in the repository.
 	var visits []visitRecord
 	uc := getUpdateConfig(c)
+
+	// Resolve dependencies.
+	rc, cleanupRc := repo.NewRemoteCache(uc.repos)
+
+	defer func() {
+		if cerr := cleanupRc(); err == nil && cerr != nil {
+			err = cerr
+		}
+	}()
+
 	walk.Walk(c, cexts, uc.dirs, uc.walkMode, func(dir, rel string, c *config.Config, update bool, f *rule.File, subdirs, regularFiles, genFiles []string) {
 		// If this file is ignored or if Gazelle was not asked to update this
 		// directory, just index the build file and move on.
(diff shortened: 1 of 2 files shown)


lenage:
@googlebot I fixed it.

repo/remote.go
```diff
@@ -389,20 +385,28 @@ func (r *RemoteCache) Mod(importPath string) (modPath, name string, err error) {
...
 			path: modPath,
 			name: label.ImportPathToBazelRepoName(modPath),
 		}, nil
-	})
+	}
+
+	v, err := r.mod.ensure(importPath, mvalue)
 	if err != nil {
 		return "", "", err
 	}
```
sluongng, line 402: Is this part needed? Seems to be a no-ops style change?

repo/remote.go
```diff
@@ -419,7 +423,6 @@ func defaultModInfo(rc *RemoteCache, importPath string) (modPath string, err err
 
 	goTool := findGoTool()
 	env := append(os.Environ(), "GO111MODULE=on")
-
```
sluongng, line 422: same here, seems to be a no-ops style improvement that could be a separate PR

cmd/gazelle/fix-update.go
```diff
@@ -272,6 +272,16 @@ func runFixUpdate(wd string, cmd command, args []string) (err error) {
 	// Visit all directories in the repository.
 	var visits []visitRecord
 	uc := getUpdateConfig(c)
+
+	// Resolve dependencies.
+	rc, cleanupRc := repo.NewRemoteCache(uc.repos)
+
+	defer func() {
```
sluongng, line 279: I would suggest remove this empty line to keep the code consistent with original style

lenage: Fixed

repo/remote.go
lenage: fixed

repo/remote.go
lenage: nice catch, fixed

blico:
Go Modules supports multiple modules for a single repository, meaning it is possible for one modpath to be a prefix of another modpath, e.g. `github.com/go-redis/redis` and `github.com/go-redis/redis/v7`. Meaning with this change if an import for a top-level module is seen before a nested module of the same repo, Gazelle will incorrectly resolve the nested module to the top-level one.

Let's discuss alternative solutions in #902

PR #1066, opened 2021-06-18
do not update files if content did not change

### LINKED ISSUE #1065: update-repos: skip updating macro and WORKSPACE files if no changes
**Feature request.**

`gazelle update-repos` should not modify macro and WORKSPACE files if the content of files does not change.

Right now gazelle always rewrites these files and this causes the full analysis of the dependency graph. In some projects it adds up to additional 8-10 seconds during build.

dududko:
fixes #1065

Skip saving files if file content did not change.
This will improve the speed of analysis phase after invoking gazelle.

### DIFF
--- a/cmd/gazelle/update-repos.go
+++ b/cmd/gazelle/update-repos.go
@@ -16,6 +16,7 @@ limitations under the License.
 package main
 
 import (
+	"bytes"
 	"errors"
 	"flag"
 	"fmt"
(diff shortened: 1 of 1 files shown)


blico:
#825 addressed a similar issue for BUILD files. Rather than modifying `file.Save` you could add a check to `update-repos.go`  before `file.Save` is called, e.g.
```
	newContent := f.Format()
	if !bytes.Equal(f.Content, newContent) {
		f.Save()
	}
```

dududko:
@blico Thank you for the tip!

sluongng:
Is bytes comparison the best way to do this? Couldnt we have a flag that signal content has been added / updated / removed instead?

excavador:
> Is bytes comparison the best way to do this? Couldnt we have a flag that signal content has been added / updated / removed instead?

It's only possible in case if you have "ibazel" or similar tool. Usually you do not have any active live process to watch file system notifications (in case of gazelle at least)

blico:
LGTM!

PR #1069, opened 2021-06-24
Honor host GOMODCACHE when GO_REPOSITORY_USE_HOST_CACHE

linzhp:
**What type of PR is this?**
Feature

go_repository

`GOMODCACHE` was added in Go 1.15. Let's read it when `GO_REPOSITORY_USE_HOST_CACHE==1`.

### DIFF
--- a/internal/go_repository_cache.bzl
+++ b/internal/go_repository_cache.bzl
@@ -38,27 +38,26 @@ def _go_repository_cache_impl(ctx):
     go_root = str(ctx.path(go_sdk_label).dirname)
     go_path = str(ctx.path("."))
     go_cache = str(ctx.path("gocache"))
+    go_mod_cache = ""
     if ctx.os.environ.get("GO_REPOSITORY_USE_HOST_CACHE", "") == "1":
         extension = executable_extension(ctx)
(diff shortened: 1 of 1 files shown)


internal/go_repository_cache.bzl
```diff
@@ -38,27 +38,26 @@ def _go_repository_cache_impl(ctx):
...
+        if not go_mod_cache and not go_path:
+            fail("GOPATH or GOMODCACHE must be set when GO_REPOSITORY_USE_HOST_CACHE is enabled.")
+        go_cache = read_go_env(ctx, go_tool, "GOCACHE")
+        if not go_cache:
             fail("GOCACHE must be set when GO_REPOSITORY_USE_HOST_CACHE is enabled.")
-        go_cache = res.stdout.strip()
 
     cache_env = {
         "GOROOT": go_root,
         "GOPATH": go_path,
```
blico, line 55: Now that we optionally allow either `GOPATH` or `GOMODCACHE` to be set, we should also optionally set `GOPATH`:
```
if go_path:
  cache_env["GOPATH"] = go_path
 ```

blico (approved):

PR #1083, opened 2021-07-13
autogazelle can build on windows

asuffield:
I don't know for sure why this was disabled, but the initial commit and
the change to golang to support AF_UNIX on windows were at roughly the
same time in 2018, so that might be the reason.

**What type of PR is this?**

Other

cmd/autogazelle

This makes 'bazel test //...' work on windows. Currently it dies to a build error, which appears to be unnecessary.


robfig:
Thanks!

PR #1086, opened 2021-07-17
Have label properly resolve directives starting with "@//"

### LINKED ISSUE #1025: label/label.go does not correctly resolve external references to the main repository

0e5dc95b1ac5ac199f90530831be3fa9022ebfd7

v0.24.13

 3.7.0

Yes

Mac OS Big Sur Intel

### What did you do?

Was passing a label via a gazelle:resolve directive for a go_repository rule including a repo starting with @// to reference to the main repository. This is legal via [bazel_specs](https://docs.bazel.build/versions/master/build-ref.html#labels) for referencing the main repository from an external reference.

`BUILD file: "@//path"`,
No error

### What did you see instead?

`BUILD file: "//path"`, the resolution is removing the "@" from the path for the external reference to the main repository. 

This is causing an error: `no such package ...`

tfrench-uber:
**What type of PR is this?**

Bug fix

label

Allows external references to the main repository (starting with "@//") to properly work in gazelle resolve statements:
`gazelle:resolve something @//some_path:to_something`. This scenario is legal in Bazel, and therefore is a bug that needs fixing/patches because it is not properly handled with Gazelle.

From [bazel specs](https://docs.bazel.build/versions/main/build-ref.html#labels):
> Labels starting with @// are references to the main repository, which will still work even from external repositories. Therefore @//a/b/c is different from //a/b/c when referenced from an external repository. The former refers back to the main repository, while the latter looks for //a/b/c in the external repository itself. This is especially relevant when writing rules in the main repository that refer to targets in the main repository, and will be used from external repositories.

Fixes #1025

### DIFF
--- a/label/label.go
+++ b/label/label.go
@@ -62,7 +62,7 @@ func New(repo, pkg, name string) Label {
 var NoLabel = Label{}
 
 var (
-	labelRepoRegexp = regexp.MustCompile(`^$|^[A-Za-z][A-Za-z0-9_]*$`)
+	labelRepoRegexp = regexp.MustCompile(`^@$|^[A-Za-z][A-Za-z0-9_]*$`)
 	labelPkgRegexp  = regexp.MustCompile(`^[A-Za-z0-9/._-]*$`)
 	labelNameRegexp = regexp.MustCompile(`^[A-Za-z0-9_/.+=,@~-]*$`)
 )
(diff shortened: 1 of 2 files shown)


label/label.go
```diff
@@ -77,9 +77,12 @@ func Parse(s string) (Label, error) {
 	if strings.HasPrefix(s, "@") {
 		relative = false
 		endRepo := strings.Index(s, "//")
-		if endRepo > 0 {
+		if endRepo > len("@") {
 			repo = s[len("@"):endRepo]
 			s = s[endRepo:]
+		} else if endRepo == len("@") {
```
linzhp, line 83: Can you add a comment explaining what this scenario is?

label/label.go
```diff
@@ -134,8 +137,10 @@ func (l Label) String() string {
 	}
 
 	var repo string
-	if l.Repo != "" {
+	if l.Repo != "" && l.Repo != "@" {
 		repo = fmt.Sprintf("@%s", l.Repo)
+	} else {
+		repo = l.Repo
```
linzhp, line 143: Add comment for this case too.

PR #1090, opened 2021-07-31
Read .netrc in go_repository

### LINKED ISSUE #847: Add `netrc` (like that of `http_archive`) to `go_repository`?

linzhp:
**What type of PR is this?**
Feature

go_repository

The URLs that `go_repository` downloads libraries from may need authentication. We need to support the authentication tokens saved in `.netrc` similar to what `http_archive` does.

** Which issues(s) does this PR fix?
Fixes #847

** Other notes for review
This is a re-implementation of #848 by addressing @jayconrod's comments there. Besides fixes in the code, these are the responses to Jay's comments:

> Could you confirm that Bazel 1.2.0 provides these functions, and if not, what is the oldest version of Bazel that does? I'd like to make sure we're not breaking compatibility here. If these were added recently, it would be better to copy them here.

Bazel 1.2.0 provides these functions (read_netrc, use_netrc): https://github.com/bazelbuild/bazel/blob/1.2.0/tools/build_defs/repo/utils.bzl. However, the third argument to use_netrc was added in Bazel 3.1. So this pull request will require Bazel 3.1, which shouldn't be a big issue as rules_go already requires Bazel 4.0

> It looks like this will only work for HTTP mode. It also needs to work for module mode though (the ctx.attr.version branch). I think that just means making sure NETRC is set to the correct file in env_keys, but please verify.

In module mode, go_repository calls Go toolchain, which already has logic to support netrc without requiring `NETRC` environment variable.

> Please add documentation to go_repository.bzl for these attributes.

Let's not add these attributes to keep it simple and use the netrc file from default location with default auth pattern for now until some people run into a scenario where customization is needed.

### DIFF
--- a/internal/go_repository.bzl
+++ b/internal/go_repository.bzl
@@ -19,21 +19,21 @@ load("@bazel_tools//tools/build_defs/repo:utils.bzl", "read_netrc", "use_netrc")
 # We can't disable timeouts on Bazel, but we can set them to large values.
 _GO_REPOSITORY_TIMEOUT = 86400
 
-
-# Copied from @bazel_tools//tools/build_defs/repo:http.bzl
+# Inspired by @bazel_tools//tools/build_defs/repo:http.bzl
 def _get_auth(ctx, urls):
     """Given the list of URLs obtain the correct auth dict."""
-    if "HOME" in ctx.os.environ and not ctx.os.name.startswith("windows"):
+    netrcfile = ""
+    if "NETRC" in ctx.os.environ:
+        netrcfile = ctx.os.environ["NETRC"]
+    elif ctx.os.name.startswith("windows"):
+        if "USERPROFILE" in ctx.os.environ:
+            netrcfile = "%s/_netrc" % (ctx.os.environ["USERPROFILE"])
+    elif "HOME" in ctx.os.environ:
         netrcfile = "%s/.netrc" % (ctx.os.environ["HOME"])
-        if ctx.execute(["test", "-f", netrcfile]).return_code == 0:
-            netrc = read_netrc(ctx, netrcfile)
-            return use_netrc(netrc, urls, {})
 
-    if "USERPROFILE" in ctx.os.environ and ctx.os.name.startswith("windows"):
-        netrcfile = "%s/.netrc" % (ctx.os.environ["USERPROFILE"])
-        if ctx.path(netrcfile).exists:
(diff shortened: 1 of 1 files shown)


blico:
This has been attempted once before in the same fashion with #848, could you check out Jay's comments there?

internal/go_repository.bzl
```diff
@@ -14,10 +14,29 @@
 
 load("@io_bazel_rules_go//go/private:common.bzl", "env_execute", "executable_extension")
 load("@bazel_gazelle//internal:go_repository_cache.bzl", "read_cache_env")
+load("@bazel_tools//tools/build_defs/repo:utils.bzl", "read_netrc", "use_netrc")
 
 # We can't disable timeouts on Bazel, but we can set them to large values.
 _GO_REPOSITORY_TIMEOUT = 86400
 
+# Inspired by @bazel_tools//tools/build_defs/repo:http.bzl
+def _get_auth(ctx, urls):
```
blico, line 23: > I expect folks will want to have just one netrc file shared by all rules, so let's add these attributes to `go_repository_config` as well and make sure they can be read by `go_repository` if they're set there.

internal/go_repository.bzl
```diff
@@ -33,6 +52,7 @@ def _go_repository_impl(ctx):
             sha256 = ctx.attr.sha256,
             stripPrefix = ctx.attr.strip_prefix,
             type = ctx.attr.type,
+            auth = _get_auth(ctx, ctx.attr.urls),
```
blico, line 55: Have you verified this works? Looks like we are setting the `auth` param but not using it when we download.

blico (commented):
Once we settle on the new parameters needed, we will need to update the documentation for them too.

internal/go_repository.bzl
linzhp: It works. I believe `ctx.download_and_extract` uses the auth param

internal/go_repository.bzl
blico: Sync'd offline, and realized this is not feasible because `get_auth` results cannot be passed in as parameters to `go_repository` rule.

blico (approved):

PR #1008, opened 2021-03-09
Correctly propagate runfiles from gazelle_bin to gazelle

HALtheWise:
**What type of PR is this?**

Bug fix

**What package or component does this PR mostly affect?**

wrapper script

**What does this PR do? Why is it needed?**

#911 previously attempted to fix #827, but did not actually do so.

In particular, the runfiles for the gazelle_bin were correctly collecting transitive runfiles for the language plugins, but the runfiles for the gazelle wrapper script were not correctly aggregating them

This small fix repairs that by merging the runfiles from the binary into the wrapper script.

**Which issues(s) does this PR fix?**

Fixes #827

**Other notes for review**

PR #1094, opened 2021-08-19
Correctly propagate runfiles from data dependencies to gazelle

alexeagle:
This is a followup to #1008 which fixed this problem for the
gazelle_binary but not for the data attribute of the gazelle rule.

**What type of PR is this?**

Bug fix

cmd/gazelle

We are working on a python extension, where the gazelle binary will have a py_binary in its data deps.
Without this fix, the runfiles of that py_binary are absent at runtime and gazelle fails like
FileNotFoundError: [Errno 2] No such file or directory: '/home/alexeagle/.cache/bazel/_bazel_alexeagle/cdec52671ea8b1684766cfc23dcfc4be/execroot/rules_python/bazel-out/k8-fastbuild/bin/gazelle/g.runfiles/bazel_tools/tools/python/py3wrapper.sh'

### DIFF
--- a/def.bzl
+++ b/def.bzl
@@ -74,9 +74,11 @@ def _gazelle_runner_impl(ctx):
     runfiles = ctx.runfiles(files = [
         ctx.executable.gazelle,
         go_tool,
-    ] + ctx.files.data).merge(
+    ]).merge(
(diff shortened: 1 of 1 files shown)


achew22 (approved):

achew22:
@alexeagle thanks for the contribution!

PR #1095, opened 2021-08-19
Merge the private attribute

alexeagle:
**What type of PR is this?**

Bug fix

cmd/gazelle

Currently the `private` attribute isn't copied.

Fixes #

**Other notes for review**

### DIFF
--- a/rule/BUILD.bazel
+++ b/rule/BUILD.bazel
@@ -26,6 +26,7 @@ go_test(
     name = "rule_test",
     srcs = [
         "directives_test.go",
+        "merge_test.go",
         "rule_test.go",
         "value_test.go",
     ],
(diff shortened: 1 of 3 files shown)


alexeagle:
/fyi @f0rmiga I'll need some help here filling in the template, I'm not sure how this bug manifests. The test failure it fixes isn't very clear what `private` is.

achew22:
I'm also a little confused about what this is fixing. Is there any way we could pull together a quick testcase of this so we don't regress in the future?

f0rmiga:
We are currently using a private attribute to assign a UUID to the `py_library` target. Any `py_test` and `py_binary` in the same Bazel package will then "import" this UUID, which gets resolved later in the `Resolve` method.

The private attribute gets passed around when it's first generated but on subsequent runs, after the target is already generated, the private attributes are purged. This patch allows the private attributes to stay in the targets during the resolution stage.

achew22:
Generally speaking I prefer to have some level of testing to prevent regressions. Could we add a test to prevent this backsliding?

f0rmiga:
@achew22 any chance we could have this so we don't need to patch Gazelle from rules_python?

PR #1101, opened 2021-09-10
use forward slashes during embed discovery

tgolsson:
**What type of PR is this?**

Bug fix.

`language/go`

On Windows, `//go:embed` directives with a folder (e.g. `foo/*.ext`) will not generate any `embedsrcs` due to using `\\` path separators during directory walks instead of `/` which is expected when matching paths.

This PR converts those double slashes during traversal, and forwards the fixed path to the later steps. This fixed our issues when compiling [this embed directive](https://github.com/cloudspannerecosystem/yo/blob/master/v2/module/builtin/module.go#L31).

No relevant issue (e.g., subdirectory embeds on Windows) found.

### DIFF
--- a/language/go/embed.go
+++ b/language/go/embed.go
@@ -170,11 +170,12 @@ func (er *embedResolver) resolve(embed fileEmbed) (list []string, err error) {
 	// would not include "a/.e".
 	var visit func(*embeddableNode, bool)
 	visit = func(f *embeddableNode, add bool) {
-		match, _ := path.Match(embed.path, f.path)
+		convertedPath := filepath.ToSlash(f.path)
+		match, _ := path.Match(embed.path, convertedPath)
 		add = match || (add && !f.isHidden())
 		if !f.isDir() {
 			if add {
-				list = append(list, f.path)
+				list = append(list, convertedPath)
 			}
 			return
 		}


achew22:
Thanks so much for helping out!

To prevent regressing, all changes to the repo must be accommodated by test cases that fail before and pass after the patch is applied. Do you think you could add a test to confirm the functionality you're looking for here?

tgolsson:
@achew22 Yes, it looks like I should be able to setup an extra target under `language/go/testdata` to validate it from `generate_test.go`. I'll look into it during the week.

tgolsson:
@achew22 There *is* a test already, `language/go/testdata/embedsrcs` -- I can't see any commits that should affect this in the last few months (we were pinned to master as of 2021-06-18, d038863ba2e096792c6bb6afca31f6514f1aeecd) and the test content has commit-date of 2021-03-08. So I'm slightly confused as to why this breaks in our Windows builds but bazel-gazelle tests pass.

tgolsson:
@achew22 Please see my latest commit. The tests don't run at all on Windows -- adding a panic to test_generate.go breaks on all other platforms. We don't need a new test -- we need to fix the test system (as this'd have been caught by the original test as far as I can tell). :-)

robfig:
It sounds like this specific test is skipped on windows due to testdata directories not being linked into runfiles? That seems like a pretty significant functionality gap.. @jayconrod any idea what the latest state here is?

```
func TestGenerateRules(t *testing.T) {
	if runtime.GOOS == "windows" {
		// TODO(jayconrod): set up testdata directory on windows before running test
		if _, err := os.Stat("testdata"); os.IsNotExist(err) {
			t.Skip("testdata missing on windows due to lack of symbolic links")
		} else if err != nil {
			t.Fatal(err)
		}
	}
```

robfig:
OK, well it sounds like there's nothing too actionable right now to improve the Windows build. If you remove the panic, I think this is fine to merge.

tgolsson:
Sure, will try to update the PR before end of week, had my hands full recently!

PR #1107, opened 2021-09-22
Escape vars in cgo flags with an extra $

### LINKED ISSUE #1106: Gazelle emitting single `$` in strings subject to Make variable substitution

v0.23.0

v0.24.0

4.2.1

Yes

Linux, x86_64

### What did you do?

When upgrading to `rules_go` v0.24.0, the make variable substitution added in https://github.com/bazelbuild/rules_go/pull/2865 caused `clinkopts` to be flagged as incorrect if they used a single `$` in them.

If I change it to `$$`, `gazelle` removes the second `$`.

`gazelle` should expect the behavior that `rules_go` now enforces and do the same.

### What did you see instead?
`gazelle` removes the second `$` required by `rules_go`

dierksen:
**What type of PR is this?**

Bug fix

language/go

`gazelle` and `rules_go` are currently in conflict with the use of `$`-escaped variables in cgo flags. This PR should resolve the discrepancy.

Fixes #1106 

**Other notes for review**

### DIFF
--- a/language/go/generate.go
+++ b/language/go/generate.go
@@ -825,5 +825,7 @@ func escapeOption(opt string) string {
 		"\t", "\\\t",
 		"\n", "\\\n",
 		"\r", "\\\r",
+		"$(", "$(",
+		"$", "$$",
 	).Replace(opt)
 }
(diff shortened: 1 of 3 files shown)


robfig:
Thank you!

PR #1111, opened 2021-10-07
Update everything before minor release

jayconrod:
* Updated to rules_go v0.29.0. This is now the minimum supported version, since Gazelle no longer generates `go_tool_library` targets in `org_golang_x_tools` which are needed when that repository is generated with `go_repository` in older versions of rules_go.
* Updated to Go 1.17.1.
* Updated all dependencies.
* Updated hardcoded Go and proto import resolution tables (only used when Gazelle is built by the go tool, not by Bazel).


BUILD.bazel
```diff
@@ -97,3 +97,9 @@ bzl_library(
         "@bazel_tools//tools/build_defs/repo:git.bzl",
     ],
 )
+
+bzl_library(
```
robfig, line 101: What's this for?

robfig (approved):
Nice, thanks! Just one question

BUILD.bazel
jayconrod: Good catch. Added accidentally.

PR #1113, opened 2021-10-11
language/go: set golang.org/x/tools/go/analysis/internal/facts to public visibility

robfig:
ce6d56176c9d2d2714aaaf57cf4ce755de6bfeee broke configurations which
bring their own x/tools go_repository and use nogo, because the
rules_go public visibility patch is lost and nogo's dependency on the
internal/facts package breaks the build.

Having Gazelle set that package's go_library to public seems to be the
best fix here.


achew22 (approved):

robfig:
Oh! nogo is referring to the alias, not the real library rule, and that one still has restricted visibility. Working on it..

Issue #763, opened 2020-04-17
Investigate migrating documentation to Stardoc


jayconrod:
[Stardoc](https://github.com/bazelbuild/stardoc) is a documentation generator for Bazel rules, providers, and other definitions.

Currently, Gazelle documentation is maintained separately from the implementation, so it's easy for it to drift.

We should investigate whether Stardoc is a viable replacement for the current documentation. If it is, we should migrate the current documentation to generated Stardoc.

Analogous rules_go issue bazelbuild/rules_go#762.

alexeagle:
@achew22 now that we've committed to having some markdown in the repository, for documentation that's output by stardoc, what do you think about having a mix of rst and markdown in the long-term? Or should I also run the remaining docs through something like pandoc to translate them to markdown as well so there's only one markup language? IMO we should do the latter.

achew22:
I think that's a great idea. Let's leave this issue open to group all the work that needs to be done, but I appreciate all the work. Thank you!

PR #1123, opened 2021-10-21
Stardoc generates repository.md

alexeagle:
This is a mirror of the content in repository.rst, allowing us to manually inspect the delta between them:
LHS https://github.com/bazelbuild/bazel-gazelle/blob/master/repository.rst
RHS https://github.com/aspect-dev/bazel-gazelle/blob/stardoc/repository.md

As soon as we are satisfied that the generated content is correct, we'll delete repository.rst and update any links to it.
(This could happen as part of this PR, or subsequent)

Technique: copied doc strings out of existing rst file into the starlark files. Transcribed rst syntax to markdown.

First step in addressing #763
Remaining steps will be to rinse-and-repeat for other .rst files containing rule APIs


alexeagle:
yup, I noticed the same thing while diffing them.

even with stardoc given
```
symbol_names = [
    "go_repository",
    "git_repository",
    "http_archive",
],
```
they still appear alphabetically in the generated file. So I don't think this is possible with stardoc's current public API

PR #1127, opened 2021-11-01
Added `rule.AddAlias` func to support loading aliased symbols

### LINKED ISSUE #1126: Add a function `rule.AddAlias` to support loading aliased symbols
latest (0.24.0)

0.29.0

4.2.1

Yes

### What did you do?
I'm developing a custom Gazelle extension.
However, when creating load() statements in a BUILD file, I need to add multiple symbols that all have the same name (e.g. `outs`).

A method like `rule.Add(sym string)`, that allows setting the alias.
Perhaps `rule.AddAlias(sym, to string)`?
I can create a PR for it if agreed upon.

Thanks!

dorfire:
Resolves #1126

### DIFF
--- a/rule/rule.go
+++ b/rule/rule.go
@@ -608,6 +608,19 @@ func (l *Load) Add(sym string) {
 	}
 }
 
+// AddAlias inserts a new aliased symbol into the load statement. This has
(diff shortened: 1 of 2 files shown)


robfig:
Thanks!

Issue #1135, opened 2021-11-21
Gazelle cannot be built when cgo is enforced by the Go SDK


uhthomas:
### What version of gazelle are you using?

v0.24.0

### What version of rules_go are you using?

v0.29.0

### What version of Bazel are you using?

4.2.1

### Does this issue reproduce with the latest releases of all the above?

Yes.

### What operating system and processor architecture are you using?

linux/amd64

### What did you do?

We're using the [dev.boringcrypto](https://go.googlesource.com/go/+/dev.boringcrypto) branch of Go. We introduced a change which hard-fails compilation if cgo is not enabled.

### What did you expect to see?

A successful build.

### What did you see instead?

The build fails as [Gazelle explicitly sets `CGO_ENABLED=0`](https://github.com/bazelbuild/bazel-gazelle/blob/3e83644bd8b1ed373d322b898cf7c05927b39201/internal/go_repository_tools.bzl#L57-L58) for its tools.

PR #1136, opened 2021-11-21
Allow tools to be built with cgo

uhthomas:
**What type of PR is this?**

Bug fix

cmd/gazelle

See the linked issue for more information. It allows Gazelle's tools to be built with cgo as builds fail when the Go SDK enforces it.

Fixes https://github.com/bazelbuild/bazel-gazelle/issues/1135

**Other notes for review**

### DIFF
--- a/internal/go_repository_tools.bzl
+++ b/internal/go_repository_tools.bzl
@@ -54,8 +54,6 @@ def _go_repository_tools_impl(ctx):
         "GO111MODULE": "off",
         # workaround: avoid the Go SDK paths from leaking into the binary
         "GOROOT_FINAL": "GOROOT",
-        # workaround: avoid cgo paths in /tmp leaking into binary
-        "CGO_ENABLED": "0",
     })
 
     if "PATH" in ctx.os.environ:


achew22:
In order to get something like this in, I would like to see some kind of regression test added so we can better understand what is broken right now, and what functionality this enables going forward.

uhthomas:
That makes sense, thank you. I have some ideas in mind for regression tests. Is there anything specific you had in mind?

I'd really like to understand the context behind forcefully disabling cgo in the first place. It was introduced as part of https://github.com/bazelbuild/bazel-gazelle/pull/382, but I'm not aware of cgo making builds non-deterministic? Please correct me if I'm wrong!

achew22:
Somehow you came to the conclusion that this was a necessary change. Canonicalizing that and putting it in the repo as a regression test is where I would start.

I can't speak to why something I didn't review happened. Sorry about that! We should also double check that the binaries are still reproducibly built after making this change. I have vague recollections of `/tmp/$RANDOMSTRING` being a part of my binaries in the long ago times and I wonder if the linked PR is what fixed that. It'd be unfortunate to go back on that.

uhthomas:
It looks like https://github.com/bazelbuild/rules_go/issues/3010 should resolve any issues with non-deterministic cgo builds.

Sorry for the lack of movement on this, will get round to writing a regression test soon.

PR #1139, opened 2021-11-23
Allow dots and dashes at repo names in labels

### LINKED ISSUE #1082: label.Parse fails on labels with hyphens in the repository name

v0.23.0

v0.27.0

4.1.0

Yes.

nixos x86_64

### What did you do?

```go
lbl, err := label.Parse("@stackage-exe//tasty-discover:tasty-discover")
if err != nil {
    log.Fatal(err)
}
```

I expected the label to be parsed since hyphens are accepted in repository names by bazel. 
https://github.com/bazelbuild/bazel/issues/11837
https://github.com/bazelbuild/bazel/commit/e574d558da17cfd0f818e7a937a07926aa270069

### What did you see instead?

```
gazelle: label parse error: repository has invalid characters: "@stackage-exe//tasty-discover:tasty-discover"
```

facundominguez:
Bug fix
cmd/gazelle

This PR has label.Parse accept repo names with dashes and dots. `gazelle` needs it to process bazel repositories that use these characters in repository names.

Fixes #1082

### DIFF
--- a/label/label.go
+++ b/label/label.go
@@ -62,7 +62,7 @@ func New(repo, pkg, name string) Label {
 var NoLabel = Label{}
 
 var (
-	labelRepoRegexp = regexp.MustCompile(`^@$|^[A-Za-z][A-Za-z0-9_]*$`)
+	labelRepoRegexp = regexp.MustCompile(`^@$|^[A-Za-z.-][A-Za-z0-9_.-]*$`)
 	labelPkgRegexp  = regexp.MustCompile(`^[A-Za-z0-9/._-]*$`)
 	labelNameRegexp = regexp.MustCompile(`^[A-Za-z0-9_/.+=,@~-]*$`)
 )
--- a/label/label_test.go
+++ b/label/label_test.go
@@ -75,6 +75,8 @@ func TestParse(t *testing.T) {
 		{str: "@a", want: Label{Repo: "a", Pkg: "", Name: "a"}},
 		{str: "@a//b", want: Label{Repo: "a", Pkg: "b", Name: "b"}},
 		{str: "@a//b:c", want: Label{Repo: "a", Pkg: "b", Name: "c"}},
+		{str: "@..//b:c", want: Label{Repo: "..", Pkg: "b", Name: "c"}},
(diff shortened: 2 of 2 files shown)


facundominguez:
Bazel source code does not seem to allow leading dots and dashes in repository names:
https://github.com/bazelbuild/bazel/blob/cbad324d2c8520195d2ddafcabeb2d2ca96da0a1/src/main/java/com/google/devtools/build/lib/packages/WorkspaceGlobals.java#L53

However, when declaring repositories with `http_archive` I was able to define and reference repositories  like `--` and `..`.

blico:
Thanks for the fix!

> However, when declaring repositories with http_archive I was able to define and reference repositories like -- and ... I tested bazelbuild/bazel@cbad324.

`http_archive` rules do not create a new WORKSPACE, explaining why you do not see a failure. If you name a `go_repository` rule `..` then you will see:
```
Error in repository_rule: invalid repository name '@..': workspace names are not allowed to be '@..'
```

Given that we cannot differentiate between repository rules that create a new WORKSPACE and those that don't, we should match the pattern in [Bazel's source code](https://github.com/bazelbuild/bazel/blob/cbad324d2c8520195d2ddafcabeb2d2ca96da0a1/src/main/java/com/google/devtools/build/lib/packages/WorkspaceGlobals.java#L53) and require repo name's start with a letter.

facundominguez:
> Given that we cannot differentiate between repository rules that create a new WORKSPACE and those that don't, we should match the pattern in Bazel's source code and require repo name's start with a letter.

I feel inclined to argue for the opposite. Given that we cannot differentiate between repository rules that create a new WORKSPACE and those that don't, `gazelle` must accept leading dots and dashes so it can digest the most repos that might be encountered in the wild.

Perhaps there is some reason for having `gazelle` being stricter than `bazel`, which I'm failing to see.

blico:
Good point, from the perspective of Gazelle's usage of `label.Parse` it makes sense for it to be less strict. It is possible there are other consumers of `label.Parse` that would prefer it to be more strict, but it is unlikely.

PR #1141, opened 2021-11-26
Update to install@latest command for builds via Go tooling (cmd/gazelle)

wamuir:
Documentation
cmd/gazelle

Revises doc language for installing cmd/gazelle via Go tooling as `go install pkg@version` has replaced now-deprecated `go get ...` when building and installing executables in module mode.


blico:
Thanks for the contribution!

PR #1152, opened 2022-01-03
don't include self-imports

### LINKED ISSUE #948: language/proto: don't include self-imports
Our monorepo has created a `CrossResolver` for resolving `proto` imports without the need for a rule index. In addition, all of our `proto` rules are generated using the `default` proto mode. 

Currently, without an index, `CrossResolver` has no way of detecting self imports (imports between the same package) meaning it adds an unwanted `dep` label for them. For example:
```
proto_library(
    name = "protopb_proto",
    srcs = [
        "foo.proto",
        "bar.proto",
    ],
    deps = [":protopb_proto"],
)

go_proto_library(
    name = "protopb_go_proto",
    compilers = [ "@io_bazel_rules_go//proto:gogoslick_proto],
    importpath = "example.com/foo"
    proto = ":protopb_proto",
    deps = [":go_default_library"],
)

go_library(
    name = "go_default_library",
    embed = [":protopb_go_proto"],
    importpath = "example.com/foo"
)
```

So far my only idea to avoid this issue is to detect self-imports when generating `proto` rules and then avoid adding them to `r.PrivateAttr(config.GazelleImportsKey)`. Meaning we then don't have to worry about resolving them for `proto_library` and `go_proto_library` rules.

tfrench-uber:
**What type of PR is this?**

Bug fix

language/proto

This PR makes a change where build file generation does not include self-imports in proto `BUILD.bazel` files.

If a package has a dependency on itself, it does not need to define this in the imports attribute.

Fixes #948 

**Other notes for review**


blico (commented):
Minor comments

PR #1155, opened 2022-01-06
Report error details from `go mod download`

illicitonion:
**What type of PR is this?**

Bug fix

language/go

Report error details from `go mod download`

Currently if there are errors from `go mod download`, e.g. if a URL
couldn't be resolved or connected to, the output is swallowed because we
ignore the JSON which describes the error.

This change attempts to best-effort parse the output JSON to find error
description strings, and attach them to the error we report to users.

### DIFF
--- a/language/go/modules.go
+++ b/language/go/modules.go
@@ -41,7 +41,7 @@ func importReposFromModules(args language.ImportReposArgs) language.ImportReposR
 	// List all modules except for the main module, including implicit indirect
 	// dependencies.
 	type module struct {
-		Path, Version, Sum string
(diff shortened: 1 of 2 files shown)


blico (commented):
Thanks for the contribution. If it doesn't turn out to be too much of a hassle, could you try to add a new test case covering this code here: https://github.com/bazelbuild/bazel-gazelle/blob/master/language/go/update_import_test.go#L30

illicitonion:
@blico Thanks for looking - I think I added something reasonable there, PTAL!

PR #1158, opened 2022-01-18
language/proto: don't skip same-package imports in per-file mode

### LINKED ISSUE #1157: Per-file proto mode doesn't find dependencies in the same package
I'm testing the new per-file mode added in #1033

HEAD as of this issue - 34c2040f691e1068f991fb981d5584478f34481c

v0.29.0

4.2.2

Yes

Linux, x86

### What did you do?

See example repo: https://github.com/jvolkman/gazelle-proto-per-file-dep-bug

Committed BUILD files are the result of `bazel run //:gazelle`

The `//proto/pkg1:test_proto` target should depend on `//proto/pkg1:foo_proto`. 

### What did you see instead?
The dependency is missing, and `bazel build //proto/pkg1:test_proto` fails.  However, the target in another package at `//proto/pkg2:test2_proto` does correctly depend on `//proto/pkg1:foo_proto` and builds correctly. 

Upon closer inspection, this probably broke following #1152.

jvolkman:
**What type of PR is this?**
Bug fix

language/proto

With the recently-added per-file mode for the proto generator, imports of `.proto` files within the same package don't generate dependencies.

Fixes #1157

**Other notes for review**

### DIFF
--- a/language/proto/generate.go
+++ b/language/proto/generate.go
@@ -235,8 +235,9 @@ func generateProto(pc *ProtoConfig, rel string, pkg *Package, shouldSetVisibilit
 	r.SetPrivateAttr(PackageKey, *pkg)
 	imports := make([]string, 0, len(pkg.Imports))
 	for i := range pkg.Imports {
-		// If the proto import is a self import (an import between the same package), skip it
-		if getPrefix(pc, path.Dir(i)) == getPrefix(pc, rel) {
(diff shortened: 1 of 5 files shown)


jvolkman:
Just realized that the test I added also passes without the change. Looks like I need to add a test somewhere in `generate_test.go`.

blico:
Thank you for the quick fix @jvolkman!

PR #1164, opened 2022-01-25
FixLoad loads helper functions, not just rules

illicitonion:
**What type of PR is this?**

Feature

`merger`

Some language implementations may try to add helper functions as
`CallExpr`s, e.g. in rules_python the `requirement` function, or in
rules_jvm_external the `artifact` function. This allows for parsing
these out and adding `load` statements for them.

See also: #981, #978, #1087 (but does not fully fix them).

**Other notes for review**

This approach was suggested in [this comment](https://github.com/bazelbuild/bazel-gazelle/issues/978#issuecomment-755431161).

### DIFF
--- a/merger/fix.go
+++ b/merger/fix.go
@@ -20,6 +20,7 @@ import (
 	"strings"
 
 	"github.com/bazelbuild/bazel-gazelle/rule"
+	bzl "github.com/bazelbuild/buildtools/build"
 )
 
 // FixLoads removes loads of unused go rules and adds loads of newly used rules.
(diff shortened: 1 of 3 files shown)


merger/fix.go
```diff
@@ -57,16 +58,30 @@ func FixLoads(f *rule.File, knownLoads []rule.LoadInfo) {
...
 
 	// Make a map of all the symbols from known files used in this file.
-	usedKinds := make(map[string]map[string]bool)
-	for _, r := range f.Rules {
-		kind := r.Kind()
-		if file, ok := knownKinds[kind]; ok && !otherLoadedKinds[kind] {
-			if usedKinds[file] == nil {
-				usedKinds[file] = make(map[string]bool)
+	usedSymbols := make(map[string]map[string]bool)
+	bzl.Walk(f.File, func(x bzl.Expr, stk []bzl.Expr) {
```
blico, line 62: I am pretty sure the following logic would lead to less iteration, unless I am missing the reason why you need to loop through `stk`:
```
bzl.Walk(f.File, func(x bzl.Expr, stk []bzl.Expr) {
  ce, ok := x.(*bzl.CallExpr)
  if !ok {
    return
  }
  ...
}

blico:
@illicitonion thanks for the great PR! Left one comment, but besides that the change LGTM.

merger/fix.go
illicitonion: Good call - updated!

PR #1165, opened 2022-01-27
Move list of supported languages to top of README

alexeagle:
Preview: https://github.com/alexeagle/bazel-gazelle/tree/patch-1#supported-languages


achew22:
Thanks for your contribution @alexeagle!

PR #1168, opened 2022-02-03
fix(platforms): remove iOS 386 and arm targets

### LINKED ISSUE #1167: iOS 386 and arm(32) are not supported targets
With these included our builds would fail with rules_go@v0.30.0. Digging a little further it seems like [Go only compiles to ios arm64 and ios x86_64 (amd64)](https://github.com/golang/go/blob/master/misc/ios/clangwrap.sh)

6bbfc47f1b0a27ee1efeddcc6671f3e4e03235dc

v0.30.0

4.2.2

Yes

Apple MacOS 12.2

### What did you do?

TBD

Successful build

### What did you see instead?

Failing on unsupported target.

JamyDev:
**What type of PR is this?**

Bug fix

language/go

With these included our builds would fail with rules_go@v0.30.0. Digging a little further it seems like [Go only compiles to ios arm64 and ios x86_64 (amd64)](https://github.com/golang/go/blob/master/misc/ios/clangwrap.sh)

I'll try to submit a sample package with a failing build on this issue, but the following patched worked for our project.

Fixes #1167

**Other notes for review**


nicolai86:
this works for me as well. thanks @JamyDev

PR #1182, opened 2022-02-25
Add experimental flags to save and load the rule index between runs

### LINKED ISSUE #1181: Support saving and loading the rule index between Gazelle runs
As discussed in various issues and in-person at the Gazelle maintainers sync, indexing can become a dominant part of the Gazelle runtime for large repositories, so even "incremental" gazelle runs that only update a couple directories still end up being slow. The current best solution is to turn off indexing, but that doesn't work for some uses, particularly non-Go languages without strong target naming conventions.

I propose that we add a pair of command line flags to Gazelle that enable it to save the state of the index at the end of a run, and load that saved state at the beginning of a future invocation. The second invocation would then be able to process just a limited set of files, but still resolve imports correctly using the cached index.

There's a few ways this could be used:
* Autogazelle, or a similar tool, could run the full Gazelle once, save the index, then watch for file modifications to trigger limited reruns
* An editor auto-save hook could run Gazelle in the current directory, loading, updating, and saving the index each time
* The index could be checked into the repo, with unit tests verifying that it is up to date.

For this to work...
* Something external needs to decide which directories to re-run on. Using the cached index won't necessarily yield the correct result if a package that needs updating isn't passed to the Gazelle run.
* The index needs to, on load, invalidate any entries that were originally created by packages which _will_ be re-run in this invocation, so that the obsolete index entries don't stick around.

I've put together a prototype (PR below) which seems to decrease runtime on our repo by ~80% when only re-processing a single directory, compared to having indexing enabled.

eric-skydio:
**What type of PR is this?**

Feature

`resolve/index.go`

This PR implements experimental flags to save and load the Gazelle rule index between runs. When Gazelle is given an index file to load, it skips traversing the entire repo, and instead uses the cached data in the provided index.

To make this work, a few things needed to change:
- The index is now structured with (closer to) one entry per indexed rule, and doesn't retain references to the File or Rule object that created them. For this to work, resolution of embeds is performed lazily at index query time. This also has the effect of removing the need to explicitly finalize the index.
- `buildUpdateRelMap` from walk.go has been refactored into a helper object, so that it can also be used to filter the index as it is loaded.
- index.go now has a unit test.

Fixes #1181

**Other notes for review**

- Removing the `.Finish()` method from the RuleIndex is technically a breaking change for plugin authors, although a pretty minor one that will probably only affect test code. We could leave it in and make it a no-op if this is a concern.
- `index.go` cannot depend on `walk.go` due to forming an import cycle, so the exclude list is passed as a function. This seems to work fine.


resolve/index_test.go
```diff
@@ -0,0 +1,152 @@
...
+	}{
+		{"parsed", ix},
+		{"loaded", loaded},
+	}
+
+	for _, index := range indexes {
+		t.Run(index.name, func(t *testing.T) {
+			for _, tC := range testCases {
+				t.Run(tC.desc, func(t *testing.T) {
+					got := ix.FindRulesByImport(tC.query, tC.lang)
```
sluongng, line 129:
```suggestion
					got := index.FindRulesByImport(tC.query, tC.lang)
```

This should be `index`?

resolve/index_test.go
```diff
@@ -0,0 +1,152 @@
...
+					for _, val := range got {
+						gotNames = append(gotNames, val.Label.Name)
+						var embeds []string
+						for _, embed := range val.Embeds {
+							embeds = append(embeds, embed.Name)
+						}
+						gotEmbeds = append(gotEmbeds, embeds)
+					}
+
+					if !reflect.DeepEqual(gotNames, tC.wantNames) {
```
jayconrod, line 141: Nit: prefer `cmp.Equal` (from `github.com/google/go-cmp/cmp` over `reflect.DeepEqual`. Can also use `cmd.Diff` in the error message.

resolve/index_test.go
```diff
@@ -0,0 +1,152 @@
...
+						t.Errorf("Querying for rule %v: expected %v got %v", tC.query, tC.wantNames, gotNames)
+					}
+
+					if tC.wantEmbeds != nil && !reflect.DeepEqual(gotEmbeds, tC.wantEmbeds) {
```
jayconrod, line 145: Same thing about `cmp.Equal` / `cmp.Diff`.

resolve/index_test.go
```diff
@@ -0,0 +1,152 @@
+package resolve_test
```
jayconrod, line 1: New .go files need a copyright notice to appease our legal overlords.

resolve/index_test.go
```diff
@@ -0,0 +1,152 @@
...
+
```
jayconrod, line 2: Thanks for writing a unit test. Please also write an integration test that makes use of the new flags. I found tests that run Gazelle and check the real output are much better at finding bugs and regressions than unit tests. They're also generally simpler to write since we have the machinery in place already.

Integration tests are in `cmd/gazelle/integration_test.go`. For this, I'd suggest something like:

1. Run with `-experimental_write_index_path=index.json` in a small workspace with .go files for two libraries, one importing the other from different directories.
2. Check that the build files have the expected content.
3. Change the build file for the imported library to have a different label.
4. Run again with `-experimental_read_index_path=index.json` only in the importing library's directory. Nothing should change.
5. Run again without the index. The `deps` for the importing library should change.

jayconrod (commented):
@linzhp nudged me a couple times to review this. Sorry for the long delay. I'm not an active maintainer (for now, maybe again soon) so take my feedback with a grain of salt.

I think we should land this PR (modulo comments). I'm generally cautious about code that relies on saved state that may be out of sync with reality. Gazelle treads a fine line here (build files of course may be out of sync with reality and Gazelle reads and writes them). That said, a feature like this is needed: I've heard several people say they've run into scaling problems with Gazelle because it takes a long time to index large repos.

I don't know exactly what the UX should look like. There's a good chance you'd want CI to generate the index files, triggered by CI, and Gazelle might need to fetch those. Or maybe it should query a service instead of dealing with JSON. `goimports` inside Google does something like that.

In any case, I think it's a good idea to start out with JSON files and flags clearly marked as experimental so we can get a feel for it.

resolve/index_test.go
```diff
@@ -0,0 +1,152 @@
...
+// testResolver implements the resolve.Resolve interface, with a configurable language name and
+// Imports/Embeds taken from the rule attributes.
+type testResolver struct{ name string }
+
+func (tr testResolver) Name() string { return tr.name }
+
+// imports is expected to be a list of the form ["Lang|Imp"]
+func (tr testResolver) Imports(c *config.Config, r *rule.Rule, f *rule.File) []resolve.ImportSpec {
+	if r.Attr("imports") == nil {
```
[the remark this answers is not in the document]
eric-skydio: This check is here to force Imports() to return nil in that case, rather than an empty slice. The semantics of Imports makes those two things different, and so we mimic that here by having the semantics of `imports=[]` not be the same as if `imports` is missing entirely.

resolve/index_test.go
eric-skydio: Thanks.

resolve/index.go
```diff
@@ -302,3 +326,78 @@ func passesLanguageFilter(langFilter []string, langName string) bool {
...
+
+	contents := indexFileData{
+		Version: indexFileVersion,
+	}
+	for _, v := range ix.labelMap {
+		contents.Data = append(contents.Data, v)
+	}
+
+	sort.Slice(contents.Data, func(i, j int) bool {
+		return contents.Data[i].Label.String() < contents.Data[j].Label.String()
```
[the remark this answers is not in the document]
eric-skydio: Good idea. I'm not going to right now, but will leave a comment.

cmd/gazelle/fix-update.go
```diff
@@ -272,6 +277,15 @@ func runFixUpdate(wd string, cmd command, args []string) (err error) {
 	// Visit all directories in the repository.
 	var visits []visitRecord
 	uc := getUpdateConfig(c)
+
+	ruleIndex := resolve.NewRuleIndex(mrslv.Resolver, exts...)
+	if uc.readIndexFile != "" {
+		err := ruleIndex.ReadFromFile(uc.readIndexFile, walk.NewUpdateFilter(c.RepoRoot, uc.dirs, uc.walkMode).ShouldIndex)
```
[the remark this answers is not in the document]
eric-skydio: Done

resolve/index.go
```diff
@@ -133,85 +139,50 @@ func NewRuleIndex(mrslv func(r *rule.Rule, pkgRel string) Resolver, exts ...inte
...
-	ix.labelMap[record.label] = record
-}
 
-// Finish constructs the import index and performs any other necessary indexing
-// actions after all rules have been added. This step is necessary because
-// a rule may be indexed differently based on what rules are added later.
-//
-// Finish must be called after all AddRule calls and before any
-// FindRulesByImport calls.
-func (ix *RuleIndex) Finish() {
```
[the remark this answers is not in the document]
eric-skydio: Done

eric-skydio:
@fmeum I finally got around to resurrecting this quite-old PR, which adds experimental support for saving and loading the rule index. This is intended to serve as a building block for other tooling (either company-specific or in an improved version of autogazelle) that understands what files have changed and executes Gazelle on only those directories.

Solving that second problem is tricky especially in a fully general and provably correct sense, so I think it makes sense to leave it out of Gazelle for now.

I left the new changes in a separate commit for easy reviewing, since @jayconrod looked at an earlier draft of this pretty thoroughly a year ago.

sluongng:
FWIW, there was an excellent talk on BUILD file generation at Google at Bazelcon 2023 https://youtu.be/IXimf4DCAoY?t=8020. Might be worth a watch.

fmeum:
@jayconrod Would you be available to give this a final look? I'm not too familiar with this part of the code, but can look into it if you are busy.

cmd/gazelle/fix-update.go
```diff
@@ -305,6 +310,16 @@ func runFixUpdate(wd string, cmd command, args []string) (err error) {
 	// Visit all directories in the repository.
 	var visits []visitRecord
 	uc := getUpdateConfig(c)
+
+	ruleIndex := resolve.NewRuleIndex(mrslv.Resolver, exts...)
+	if uc.readIndexFile != "" {
+		exclude := walk.NewUpdateFilter(c.RepoRoot, uc.dirs, uc.walkMode).ShouldIndex
```
jayconrod, line 316: Not sure I understand why we're excluding parts of the index when reading.

Previously, we decided whether to index directories based on a mix of `-r`, `-index`, and the command line arguments. If you wanted to update build files in a specific directory, you'd still usually need to walk the entire repo in order to build the index in memory (unless `-index=false` was explicitly set).

I think a major advantage of having an index file is that you can do fast updates in specific directories, for example, when saving a file. You can still have the index for the entire repo but you don't have to walk the full directory tree. This seems though like we're only loading parts of the index for directories being updated.

resolve/index.go
```diff
@@ -303,3 +347,80 @@ func passesLanguageFilter(langFilter []string, langName string) bool {
...
+type indexFileData struct {
+	Version int           `json:"version"`
+	Data    []*ruleRecord `json:"data"`
+}
+
+// WriteToFile saves all data in the index to a JSON file at the path provided.
+//
+// This file is not expected to have a stable format between Gazelle versions, but
+// a best-effort attempt is made to keep it sorted so it _should_ remain stable
+// between runs.
```
jayconrod, line 362: Let's comment for both `WriteToFile` and `ReadToFile` that these functions are experimental, and users outside Gazelle should not rely on them.

I'd also suggest not committing to JSON in the doc comment. This may not scale for very large repos: we might not want to read the file into memory. We might want to switch to SQLite or something similar. JSON is good for initial experimentation obviously though.

jayconrod (commented):
Apologies for the slow response, I'm finding new responsibilities lately that involve a lot more meetings :) Couple comments and thoughts below. Generally I'm in favor of this though, especially as it's marked experimental and can change.

linzhp:
> I don't know exactly what the UX should look like. There's a good chance you'd want CI to generate the index files, triggered by CI, and Gazelle might need to fetch those. 

We have a "incremental index" mode in Uber: when users run "gazelle <some dir>", Gazelle updates (with help of a global index file) and indexes all build targets in <some dir> and its subdirectories. All the entries in the global index file under <some dir> is also updated. So both index and build files can be incrementally updated. We never need to run Gazelle from root to populate the index. In fact, that's not possible in Git sparse checkout.

Our "index file" is actually a long list of `gazelle:resolve` at the root BUILD.bazel file, which is not ideal, but I hope this PR would preserve the "incremental index" feature.

jbedard:
@eric-skydio has any more thought been put into this? Are you using this in your repo or is this poc abandoned?

aiuto:
@linzhp  Would you be willing to share or upstream your incremental indexer?

linzhp:
The incremental index is based on a convention plugin that we have in Uber, in which we define strict conventions to map imports to Bazel targets. A more flexible implementation for that is now available in Gazelle called "lazy indexing" (#1891), which will significantly reduce the size of the index.

We probably still need the incremental indexing for the targets that doesn't conform to any conventions, and thus not handled by lazy indexing. However, if we ever come to a point of upstreaming it, it will be based on the lazy indexing, not in the current form. (cc @blico, @tyler-french)

I would encourage you trying out lazy indexing first. If the number of imports that cannot be resolved with lazy indexing is manageable, you don't need incremental indexing at all and instead ask developers to manually write "gazelle:resolve".

Issue #75, opened 2018-01-05
Test running gazelle rule in another workspace


jayconrod:
Test should:

* Create a new workspace with a WORKSPACE file, BUILD.bazel file, and some trivial source files, using the recommended boilerplate code.
* `bazel run //:gazelle`
* `bazel build` a newly generated `go_binary` rule.

This test should work similarly to rules_go integration tests (see [bazel_tests.bzl](https://github.com/bazelbuild/rules_go/blob/master/tests/bazel_tests.bzl)).

achew22:
I have done something like this in the iBazel project. I opted to just use sh_library in my examples but the same logic applies.

https://github.com/bazelbuild/bazel-watcher/blob/a2ceacd14ddf2cb7e1afe718f181441cc95162b3/e2e/simple/simple_test.go#L19-L31

It also has the added bonus of allowing you to verify against multiple versions of Bazel in unison by passing in a set of versions.

jayconrod:
@achew22 Thanks, that looks promising.

PR #1183, opened 2022-02-25
feat(testing): testing utility for running the gazelle binary.

aptenodytes-forsteri:
- Run gazelle on a test workspace and check the results.

**What type of PR is this?**

> Feature

> testtools

`rules_python` has a nice framework for running full integration tests against an example workspace. This PR pulls that behavior into inte the main gazelle repo. It also adds the ability to update snapshots, so tests can be fixed by re-running with an environment variable set.
https://github.com/bazelbuild/rules_python/blob/c6357a68ef2588f4248d8a76d44bc984445bf358/gazelle/python_test.go#L78

Fixes https://github.com/bazelbuild/bazel-gazelle/issues/75 (Possibly)

**Other notes for review**

### DIFF
--- a/extend.md
+++ b/extend.md
@@ -20,6 +20,46 @@ To extend Gazelle, you must do three things:
   `bazel run //:gazelle`, your binary will be built and executed instead of
   the default binary.
 
+Tests
+-----
+
+To write tests for your gazelle extension, you can use `gazelle_generation_test`,
+which will run a gazelle binary of your choosing on a set of test workspaces.
+
+```starlark
+    load("@bazel_gazelle//:def.bzl", "gazelle_generation_test")
+
+    gazelle_generation_test(
+        name = "my_generation_test",
+        # The name of the gazelle binary target in your repo.
+        gazelle_binary_name = "gazelle_local",
+        # Optional, the workspace relative path to the gazelle binary.
+        # Defaults to the root of the workspace.
+        gazelle_binary_dir = "",
+        test_data = glob([
+            "testdata_full_gen_test/**",
+        ]),
+        test_data_dir = "path/to/testdata_full_gen_test",
+    )
+```
+
+The generation test expects a file structure like the following.
+```
+|-- <testDataPath>
+    |-- some_test
+        |-- WORKSPACE
+        |-- README.md --> README describing what the test does.
+        |-- test.yaml --> YAML file for test configuration.
+        |-- app
+            |-- sourceFile.foo
+            |-- BUILD.in --> BUILD file prior to running gazelle.
+            |-- BUILD.out --> BUILD file expected after running gazelle.
+```
+
+To run update your desired BUILD.out files, you can run:
+
+`UPDATE_SNAPSHOTS=true bazel run //path/to:my_generation_test`.
+
 Supported languages
 -------------------
 
--- a/extend.md
+++ b/extend.md
@@ -23,44 +23,9 @@ To extend Gazelle, you must do three things:
 Tests
 -----
 
-To write tests for your gazelle extension, you can use `gazelle_generation_test`,
+To write tests for your gazelle extension, you can use [gazelle_generation_test](#gazelle_generation_test),
 which will run a gazelle binary of your choosing on a set of test workspaces.
 
-```starlark
-    load("@bazel_gazelle//:def.bzl", "gazelle_generation_test")
-
-    gazelle_generation_test(
-        name = "my_generation_test",
-        # The name of the gazelle binary target in your repo.
-        gazelle_binary_name = "gazelle_local",
-        # Optional, the workspace relative path to the gazelle binary.
-        # Defaults to the root of the workspace.
-        gazelle_binary_dir = "",
-        test_data = glob([
-            "testdata_full_gen_test/**",
-        ]),
-        test_data_dir = "path/to/testdata_full_gen_test",
-    )
-```
-
-The generation test expects a file structure like the following.
-```
-|-- <testDataPath>
-    |-- some_test
-        |-- WORKSPACE
-        |-- README.md --> README describing what the test does.
-        |-- expectedStdout.txt --> Expected stdout for this test.
-        |-- expectedStderr.txt --> Expected stderr for this test.
-        |-- expectedExitCode.txt --> Expected exit code for this test.
-        |-- app
-            |-- sourceFile.foo
-            |-- BUILD.in --> BUILD file prior to running gazelle.
-            |-- BUILD.out --> BUILD file expected after running gazelle.
-```
-
-To run update your desired BUILD.out files, you can run:
-
-`UPDATE_SNAPSHOTS=true bazel run //path/to:my_generation_test`.
 
 Supported languages
 -------------------
@@ -168,5 +133,46 @@ proto extension stores metadata in hidden attributes of generated
 | :------------- | :------------- | :------------- | :------------- | :------------- |
 | <a id="gazelle_binary-name"></a>name |  A unique name for this target.   | <a href="https://bazel.build/docs/build-ref.html#name">Name</a> | required |  |
 | <a id="gazelle_binary-languages"></a>languages |  A list of language extensions the Gazelle binary will use.<br><br>            Each extension must be a [go_library] or something compatible. Each extension             must export a function named <code>NewLanguage</code> with no parameters that returns             a value assignable to [Language].   | <a href="https://bazel.build/docs/build-ref.html#labels">List of labels</a> | required |  |
+
+
+<a id="#gazelle_generation_test"></a>
+
+## gazelle_generation_test
+
+<pre>
+gazelle_generation_test(<a href="#gazelle_generation_test-name">name</a>, <a href="#gazelle_generation_test-gazelle_binary">gazelle_binary</a>, <a href="#gazelle_generation_test-test_data">test_data</a>, <a href="#gazelle_generation_test-build_in_suffix">build_in_suffix</a>, <a href="#gazelle_generation_test-build_out_suffix">build_out_suffix</a>)
+</pre>
+
+    gazelle_generation_test is a macro for testing gazelle against workspaces.
+
+The generation test expects a file structure like the following:
+
+```
+|-- <testDataPath>
+    |-- some_test
+        |-- WORKSPACE
+        |-- README.md --> README describing what the test does.
+        |-- expectedStdout.txt --> Expected stdout for this test.
+        |-- expectedStderr.txt --> Expected stderr for this test.
+        |-- expectedExitCode.txt --> Expected exit code for this test.
+        |-- app
+            |-- sourceFile.foo
+            |-- BUILD.in --> BUILD file prior to running gazelle.
+            |-- BUILD.out --> BUILD file expected after running gazelle.
+```
+
+To update the expected files, run `UPDATE_SNAPSHOTS=true bazel run //path/to:the_test_target`.
+
+
+**PARAMETERS**
+
+
+| Name  | Description | Default Value |
+| :------------- | :------------- | :------------- |
+| <a id="gazelle_generation_test-name"></a>name |  The name of the test.   |  none |
+| <a id="gazelle_generation_test-gazelle_binary"></a>gazelle_binary |  The name of the gazelle binary target. For example, //path/to:my_gazelle.   |  none |
+| <a id="gazelle_generation_test-test_data"></a>test_data |  A target of the test data files you will pass to the test. This can be a https://bazel.build/reference/be/general#filegroup.   |  none |
+| <a id="gazelle_generation_test-build_in_suffix"></a>build_in_suffix |  The suffix for the input BUILD.bazel files. Defaults to .in. By default, will use files named BUILD.in as the BUILD files before running gazelle.   |  <code>".in"</code> |
+| <a id="gazelle_generation_test-build_out_suffix"></a>build_out_suffix |  The suffix for the expected BUILD.bazel files after running gazelle. Defaults to .out. By default, will use files named check the results of the gazelle run against files named BUILD.out.   |  <code>".out"</code> |
 
 
--- a/extend.md
+++ b/extend.md
@@ -49,7 +49,9 @@ The generation test expects a file structure like the following.
     |-- some_test
         |-- WORKSPACE
         |-- README.md --> README describing what the test does.
-        |-- test.yaml --> YAML file for test configuration.
+        |-- expectedStdout.txt --> Expected stdout for this test.
+        |-- expectedStderr.txt --> Expected stderr for this test.
+        |-- expectedExitCode.txt --> Expected exit code for this test.
         |-- app
             |-- sourceFile.foo
             |-- BUILD.in --> BUILD file prior to running gazelle.
--- a/internal/extend_docs.bzl
+++ b/internal/extend_docs.bzl
@@ -23,44 +23,9 @@ To extend Gazelle, you must do three things:
 Tests
 -----
 
-To write tests for your gazelle extension, you can use `gazelle_generation_test`,
+To write tests for your gazelle extension, you can use [gazelle_generation_test](#gazelle_generation_test),
 which will run a gazelle binary of your choosing on a set of test workspaces.
 
-```starlark
-    load("@bazel_gazelle//:def.bzl", "gazelle_generation_test")
-
-    gazelle_generation_test(
-        name = "my_generation_test",
-        # The name of the gazelle binary target in your repo.
-        gazelle_binary_name = "gazelle_local",
-        # Optional, the workspace relative path to the gazelle binary.
-        # Defaults to the root of the workspace.
-        gazelle_binary_dir = "",
-        test_data = glob([
-            "testdata_full_gen_test/**",
-        ]),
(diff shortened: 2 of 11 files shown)


achew22:
This is great! It looks like it is based off of the one I wrote for [`skylib`](https://github.com/bazelbuild/bazel-skylib/blob/main/gazelle/bzl/gazelle_test.go) so I'm reasonably familiar with it.

I think this is a great contribution to the core of Gazelle and can be used to trivially produce more test cases. One enhancement I've long wanted was a quick and easy build rule that you could give a `gazelle` binary and a `testdata` directory and it would do the rest. You're most of the way to that with this code, I think all one would need to do would be to make a `go_test` that reads in command line flags and fills in a `TestGazelleGenerationArgs` and passes it in to `TestGazelleGenerationOnPath`. Then we'd have a 5 line macro that everyone could benefit from.

```
gazelle_generation_test(
  name = "my_test",
  gazelle = "//path/to:gazelle",
  testdata = "//path/to:testdata",
)
```

WDYT?

aptenodytes-forsteri:
> WDYT?

I might be able to take a crack at that here, or might make sense as a follow-on PR?

achew22:
If you'd like to do it in a follow on, works for me, just tell me and I'll review this one properly

aptenodytes-forsteri:
> Then we'd have a 5 line macro that everyone could benefit from.

This is done, but looks like I have some checks to fix.

internal/go_repository_tools.bzl
```diff
@@ -77,7 +77,7 @@ def _go_repository_tools_impl(ctx):
                 "run",
                 ctx.path(ctx.attr._list_repository_tools_srcs),
                 "-dir=src/github.com/bazelbuild/bazel-gazelle",
-                "-check=internal/go_repository_tools_srcs.bzl",
+                "-generate=internal/go_repository_tools_srcs.bzl",
```
aptenodytes-forsteri, line 80: Oops, will revert.

extend.md
```diff
@@ -20,6 +20,46 @@ To extend Gazelle, you must do three things:
...
+        test_data = glob([
+            "testdata_full_gen_test/**",
+        ]),
+        test_data_dir = "path/to/testdata_full_gen_test",
+    )
+```
+
+The generation test expects a file structure like the following.
+```
+âââ <testDataPath>
```
aptenodytes-forsteri, line 48: Yikes this looks off.

internal/generationtest/generation_test.go
```diff
@@ -0,0 +1,76 @@
...
+// TestFullGeneration runs the gazelle binary on a few example
+// workspaces and confirms that the generated BUILD files match expectation.
+func TestFullGeneration(t *testing.T) {
+	tests := map[string][]bazel.RunfileEntry{}
+	runfiles, err := bazel.ListRunfiles()
+	if err != nil {
+		t.Fatalf("bazel.ListRunfiles() error: %v", err)
+	}
+	var manifest *manifestYAML
+	for _, f := range runfiles {
```
aptenodytes-forsteri, line 23: Kind of silly to iterate through the runfiles twice. I think there's a better way.

aptenodytes-forsteri: Eh, maybe not. Issue is we can't really pass the test anything because this version of go_test doesn't take env vars.

aptenodytes-forsteri:
@achew22 I think this is read for review now

internal/generationtest/generation_test.bzl
```diff
@@ -0,0 +1,41 @@
...
+        name: the name of the test.
+        gazelle_binary_name: the name of the gazelle binary target. For example, if you have a
+        gazelle_binary target where you pass the name "my_gazelle", this would be "my_gazelle".
+        test_data: a glob of the test data files you will pass to the test.
+        test_data_dir: the workspace relative path where the testdata file is.
+        gazelle_binary_dir: the workspace relative path where the gazelle_binary target is. For example,
+        if "my_gazelle" is defined in a BUILD.bazel file in <wkspace>/my/path, this arg should be
+        "my/path".
+    """
+    native.genrule(
```
achew22, line 21: This looks like a lot of work to figure out where the template is and all that stuff. What would you think about modifying `go_test` below to have some [`args`](https://docs.bazel.build/versions/main/be/common-definitions.html#common-attributes-tests)? We use stuff like this throughout the repo. For example [here](https://github.com/bazelbuild/rules_go/blob/bd0dd3fe48a53227f2c457e8c2097e1c097d1f38/tests/core/go_test/BUILD.bazel#L51-L61).

1. Change the `gazelle_generation_test` to be:

```
def gazelle_generation_test(name, gazelle_binary, test_data):
    """
    gazelle_generation_test is a macro for testing gazelle against workspaces.
    Args:
        name: the name of the test.
        gazelle_binary: A path to the binary under test. For example: "//path/to:my_gazelle".
        test_data: a target of the test data files you will pass to the test. This can be a
          https://docs.bazel.build/versions/main/be/general.html#filegroup
    """
   ```

2.  Remove the genrule and change the `go_test` to directly manage those inputs:

   ```
go_test(
        name = name,
        args = [
                "--gazelle_binary_path=$(location %s)" % gazelle_binary,
                "--test_data=$(locations %s)" % test_data,
        ],
        deps = [
            "//testtools",
            "@io_bazel_rules_go//go/tools/bazel:go_default_library",
        ],
        data = test_data + [
            gazelle_binary,
            ":generation_test_manifest.yaml",
        ],
```

internal/generationtest/generation_test.go
```diff
@@ -0,0 +1,77 @@
...
+	"io/ioutil"
+	"path"
+	"strings"
+	"testing"
+
+	"github.com/bazelbuild/bazel-gazelle/testtools"
+	"github.com/bazelbuild/rules_go/go/tools/bazel"
+	yaml "gopkg.in/yaml.v2"
+)
+
```
achew22, line 13: Then in this file you'd do something like (after adding an import for `flag`) to add in some flags:

```
var (
  gazelleBinaryPath = flag.String("gazelle_binary_path", "", "Path to the gazelle binary to test")
  testData = flag.String("test_data", "", "Path to the data you want to test")
)
```

Now you can use those strings to get the values. I believe that `testData` is going to be some kind of delimited string, like `"foo.txt,bar.txt"` or maybe space delimited, I'm not sure, but you should be able to use [`strings.Split`](https://pkg.go.dev/strings#Split) to turn it into a list of input files. I think you can iterate through that list of files looking for a `WORKSPACE` and use that as the root for execution of Gazelle.

[0] https://blog.jbowen.dev/2019/08/using-go-flags-in-tests/

testtools/files.go
```diff
@@ -129,3 +139,232 @@ func CheckFiles(t *testing.T, dir string, files []FileSpec) {
...
+						}
+					}
+					t.Logf("%q exists in %v", relativePath, testdataDir)
+					return nil
+				})
+				if !shouldUpdate {
+					t.Logf(`
+=====================================================================================
+
+Run UPDATE_SNAPSHOTS=true bazel run //path/to/this/test to update BUILD.out files.
```
achew22, line 278: I believe you can use `os.Getenv("TEST_TARGET")`[0][1] to get the actual target inside the test, so this would become

```
t.Logf(`
Run UPDATE_SNAPSHOTS=true bazel run %s to update BUILD.out files.
`, os.Getenv("TEST_TARGET"))
```

[0] https://pkg.go.dev/os#Getenv
[1] https://docs.bazel.build/versions/main/test-encyclopedia.html#initial-conditions

testtools/files.go
```diff
@@ -129,3 +139,232 @@ func CheckFiles(t *testing.T, dir string, files []FileSpec) {
...
+//└── <testDataPath>
+//    └── some_test
+//        ├── WORKSPACE
+//        ├── README.md --> README describing what the test does.
+//        ├── test.yaml --> YAML file for test configuration.
+//        └── app
+//            └── sourceFile.foo
+//            └── BUILD.in --> BUILD file prior to running gazelle.
+//            └── BUILD.out --> BUILD file expected after running gazelle.
+func TestGazelleGenerationOnPath(t *testing.T, args *TestGazelleGenerationArgs, files []bazel.RunfileEntry) {
```
achew22, line 181: Can we get away with out taking in the set of `files`? If the `TestGazelleGenerationArgs` has the `TestDataPath` I believe we can get all of these from listing the files on disk and make the interface more robust.

testtools/files.go
```diff
@@ -129,3 +139,232 @@ func CheckFiles(t *testing.T, dir string, files []FileSpec) {
...
+	}
+}
+
+// TestGazelleGenerationOnPath runs a full gazelle binary on a testdata directory.
+// With a test data directory of the form:
+//└── <testDataPath>
+//    └── some_test
+//        ├── WORKSPACE
+//        ├── README.md --> README describing what the test does.
+//        ├── test.yaml --> YAML file for test configuration.
```
achew22, line 176: The YAML file feels kind of unnecessary to me. WDYT about making a file called `stderr` and `stdout` describing the outputs and passing the `$(location)` of those in as arguments to the `gazelle_generation_test` rule? You can also pass in the `exit_code` that way and that completely eliminates the need to do YAML un/marshalling in this code.

testtools/files.go
```diff
@@ -129,3 +139,232 @@ func CheckFiles(t *testing.T, dir string, files []FileSpec) {
...
+		}
+
+		testdataDir, cleanup := CreateFiles(t, inputs)
+		defer cleanup()
+		defer func() {
+			if t.Failed() {
+				shouldUpdate := os.Getenv("UPDATE_SNAPSHOTS") != ""
+				buildWorkspaceDirectory := os.Getenv("BUILD_WORKSPACE_DIRECTORY")
+				if buildWorkspaceDirectory == "" {
+					// Default to ~/<workspace>.
```
achew22, line 245: If `BUILD_WORKSPACE_DIRECTORY` is empty, I think we should just `t.Fatal` out instead of trying to guess where things should be written. We are probably going to guess wrong and we should reinstruct the caller to use `bazel run`

testtools/files.go
```diff
@@ -129,3 +139,232 @@ func CheckFiles(t *testing.T, dir string, files []FileSpec) {
...
 }
+
+type TestGazelleGenerationArgs struct {
+	// name is the name of the test.
+	Name string
+	// testDataPath is the workspace relative path to the test data directory.
+	TestDataPath string
+	// gazelleBinaryDir is the workspace relative path to the location of the gazelle binary
+	// we want to test.
+	GazelleBinaryDir string
```
achew22, line 150: Can you collapse `GazelleBinaryDir` and `GazelleBinaryName` into a single attribute?

testtools/files.go
```diff
@@ -129,3 +139,232 @@ func CheckFiles(t *testing.T, dir string, files []FileSpec) {
...
+	BuildInSuffix string
+
+	// The suffix for all test output build files. Includes the ".".
+	// Default: ".out", so out BUILD files should be named BUILD.out.
+	BuildOutSuffix string
+}
+
+func NewTestGazelleGenerationArgs() *TestGazelleGenerationArgs {
+	return &TestGazelleGenerationArgs{
+		BuildInSuffix:  ".in",
```
achew22, line 165: Above it says that `BuildInSuffix` defaults to `.in`, if that's the case we shouldn't explicitly set the value here (same for `BuildOutSuffix`).

testtools/files.go
```diff
@@ -129,3 +139,232 @@ func CheckFiles(t *testing.T, dir string, files []FileSpec) {
...
+	// The suffix for all test input build files. Includes the ".".
+	// Default: ".in", so input BUILD files should be named BUILD.in.
+	BuildInSuffix string
+
+	// The suffix for all test output build files. Includes the ".".
+	// Default: ".out", so out BUILD files should be named BUILD.out.
+	BuildOutSuffix string
+}
+
+func NewTestGazelleGenerationArgs() *TestGazelleGenerationArgs {
```
achew22, line 163: Honestly, I'm not sure when you would want to call this function. Can you help me understand why this is better than the caller saying

```
&TestGazelleGenerationArgs{}
```
?

testtools/files.go
```diff
@@ -129,3 +139,232 @@ func CheckFiles(t *testing.T, dir string, files []FileSpec) {
...
+//    └── some_test
+//        ├── WORKSPACE
+//        ├── README.md --> README describing what the test does.
+//        ├── test.yaml --> YAML file for test configuration.
+//        └── app
+//            └── sourceFile.foo
+//            └── BUILD.in --> BUILD file prior to running gazelle.
+//            └── BUILD.out --> BUILD file expected after running gazelle.
+func TestGazelleGenerationOnPath(t *testing.T, args *TestGazelleGenerationArgs, files []bazel.RunfileEntry) {
+	t.Run(args.Name, func(t *testing.T) {
```
achew22, line 182: I would put in some logic right here, before the t.Run to do some setup.

```
t.Helper() // Make the stack trace a little bit more clear
if args.BuildInSuffix == "" {
  args.BuildInSuffix = ".in"
}
if args.BuildOutSuffix == "" {
  args.BuildOutSuffix = ".out"
}
```

internal/generationtest/generation_test.go
```diff
@@ -0,0 +1,77 @@
...
+			}
+
+			tests[parts[1]] = append(tests[parts[1]], f)
+		}
+	}
+	if len(tests) == 0 {
+		t.Fatal("no tests found")
+	}
+
+	testArgs := testtools.NewTestGazelleGenerationArgs()
```
achew22, line 60: I think this could be replaced with

```
testArgs := &testtools.TestGazelleGenerationArgs {
  GazelleBinaryPath: gazelleBinaryPath,
  TestDataPath: testDataDir,
}
```

internal/generationtest/generation_test.go
```diff
@@ -0,0 +1,77 @@
...
+	for testName, files := range tests {
+		testArgs.Name = testName
+		testArgs.TestDataPath = testDataDir
+		testArgs.GazelleBinaryDir = gazelleBinaryDir
+		testArgs.GazelleBinaryName = gazelleBinaryName
+		testtools.TestGazelleGenerationOnPath(t, testArgs, files)
+	}
+}
+
+type manifestYAML struct {
```
achew22, line 73: I think we can drop this as well if we pass everything in args.

achew22 (changes requested):

testtools/files.go
aptenodytes-forsteri: Oh nice!

testtools/files.go
aptenodytes-forsteri: Yep, totally agree.

internal/generationtest/generation_test.bzl
aptenodytes-forsteri: Yep, this seems much better.

> This can be a
>           https://docs.bazel.build/versions/main/be/general.html#filegroup

Looks like that doc has changed locations to https://bazel.build/reference/be/general#filegroup

testtools/files.go
aptenodytes-forsteri: I guess the issue is that we need a different one per test? Could we just look for files named that in each test directory when walking the tree?

internal/generationtest/generation_test.bzl
aptenodytes-forsteri: Hm another issue is that `$(location %s)` gives a workspace relative path, when I need an absolute path (I get `gazelle_local_/gazelle_local: no such file or directory`).

How do I use that value cleanly?

aptenodytes-forsteri: Ah, looks like I can use https://github.com/bazelbuild/rules_go/blob/v0.30.0/go/tools/bazel/runfiles.go#L43

internal/extend_docs.bzl
```diff
@@ -20,6 +20,48 @@ To extend Gazelle, you must do three things:
...
+To write tests for your gazelle extension, you can use `gazelle_generation_test`,
+which will run a gazelle binary of your choosing on a set of test workspaces.
+
+```starlark
+    load("@bazel_gazelle//:def.bzl", "gazelle_generation_test")
+
+    gazelle_generation_test(
+        name = "my_generation_test",
+        # The name of the gazelle binary target in your repo.
+        gazelle_binary_name = "gazelle_local",
```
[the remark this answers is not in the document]
aptenodytes-forsteri: Done.

testtools/files.go
```diff
@@ -129,3 +139,250 @@ func CheckFiles(t *testing.T, dir string, files []FileSpec) {
...
+			}
+
+			content, err := ioutil.ReadFile(path)
+			if err != nil {
+				t.Errorf("ioutil.ReadFile(%q) error: %v", path, err)
+			}
+
+			// Read in expected stdout, stderr, and exit code files.
+			if d.Name() == expectedStdoutFilename {
+				config.Stdout = string(content)
```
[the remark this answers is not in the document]
aptenodytes-forsteri: Done.

achew22 (approved):
@aptenodytes-forsteri this has to be one of my favorite PRs into Gazelle in a while. Thank you for putting the time and energy into making this so easy to reuse and to document it so well!

aptenodytes-forsteri:
> this has to be one of my favorite PRs into Gazelle in a while

Hooray! Thanks again for all the helpful feedback!

PR #1184, opened 2022-02-26
doublestar: upgrade to v4

### LINKED ISSUE #786: Possible optimization: doublestar.Match memory usage

robfig:
The latest release has been rewritten for speed.

Fixes #786


walk/config_test.go
```diff
@@ -16,7 +16,6 @@ func TestCheckPathMatchPattern(t *testing.T) {
 		{pattern: "**/*.pb.go", err: nil},
 		{pattern: "**/*.pb.go", err: nil},
 		{pattern: "[]a]", err: doublestar.ErrBadPattern},
-		{pattern: "[a-]", err: doublestar.ErrBadPattern},
```
achew22, line 19: Why remove this test case?

achew22 (approved):

walk/config_test.go
robfig: This is a behavior change in the upgrade. It didn't seem like a critical behavior to maintain, so it doesn't seem like a huge deal. It could be worth filing an issue to ensure that's working as expected upstream. At a glance, it seems fine if that's a char class that matches a or -, since I think that's what regexes do.

PR #1185, opened 2022-03-02
Limit Gazelle's walk to -universe

### LINKED ISSUE #1180: Defining universe for Gazelle to visit
Currently, when `-index true`, Gazelle would visit the whole repository to build the index: 

https://github.com/bazelbuild/bazel-gazelle/blob/4d2e96a20ad1a32b48bd2dc5c44d0627d34063bf/cmd/gazelle/fix-update.go#L123-L124

In works fine in a small repository, but it is inefficient in large repositories, especially one with more than one language. For example, in large repository with Go and Java, we know that Go will never import any Java code, so it is sufficient to only visit and index the part of the repo that contains all the Go files. It is a waste disk I/O and memory to visit a potentially large number of Java rules and store a large index of Java rules that it is not going to use.

We can add a `-universe` option to Gazelle to limit the scope it visits when `-index true`

linzhp:
**What type of PR is this?**
Feature
> Documentation
> Other

* config
* walk

Gazelle now takes a `-universe` parameter, which contains a list of directories. Instead of walking from the repo root, it walks from each directories in the `-universe`. This is needed to speed up traversal in large repositories, where we know some part of the repository is not relevant to Gazelle.

Fixes #1180 

**Other notes for review**

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -443,6 +443,12 @@ The following flags are accepted:
 | should use the index to resolve dependencies. If this is switched off, Gazelle would rely on               |
 | ``# gazelle:prefix`` directive or ``-go_prefix`` flag to resolve dependencies.                             |
 +-------------------------------------------------------------------+----------------------------------------+
+| :flag:`-universe dir1,dir2`                                       |                                        |
++-------------------------------------------------------------------+----------------------------------------+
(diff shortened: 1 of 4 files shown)


linzhp:
This approach doesn't work

PR #1187, opened 2022-03-07
feat: "include" directive

f0rmiga:
**What type of PR is this?**

Feature

all

Adds the `include` directive that takes precedence over `exclude`.

None

**Other notes for review**

### DIFF
--- a/walk/config.go
+++ b/walk/config.go
@@ -33,8 +33,9 @@ import (
 
 type walkConfig struct {
 	excludes []string
-	ignore   bool
 	follow   []string
+	ignore   bool
+	includes []string
 }
 
 const walkName = "_walk"
@@ -48,15 +49,23 @@ func (wc *walkConfig) isExcluded(rel, base string) bool {
 		return true
 	}
 	f := path.Join(rel, base)
-	for _, x := range wc.excludes {
-		matched, err := doublestar.Match(x, f)
-		if err != nil {
-			// doublestar.Match returns only one possible error, and only if the
-			// pattern is not valid. During the configuration of the walker (see
-			// Configure below), we discard any invalid pattern and thus an error
-			// here should not be possible.
-			log.Panicf("error during doublestar.Match. This should not happen, please file an issue https://github.com/bazelbuild/bazel-gazelle/issues/new: %s", err)
+	isIncluded := false
+	if len(wc.includes) == 0 {
+		isIncluded = true
+	} else {
+		for _, x := range wc.includes {
+			matched := must(doublestar.Match(x, f))
+			if matched {
+				isIncluded = true
+				break
+			}
 		}
+	}
+	if !isIncluded {
+		return true
+	}
+	for _, x := range wc.excludes {
+		matched := must(doublestar.Match(x, f))
 		if matched {
 			return true
 		}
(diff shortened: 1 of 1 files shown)


f0rmiga:
Is this desired? If so, I can adjust the tests.

aptenodytes-forsteri:
> I think it would be better to start with a problem statement rather than a solution.

Here is the problem we encountered:

Many of our developers have additional directories inside the workspace directory. For example, they might have additional copies of the repo to be able to toggle between the two. They might have additional data folders, e.g. when the app runs locally, it deposits some data in a folder in the workspace. They might have back ups of node_modules from a prior state of the repo.

In all of these cases, gazelle happily hangs forever as it tries to parse a node modules backup or a copy of the repo or lots of data files. In some cases (as in with the copy of the repo), gazelle even crashes due to duplicate rules encountered.

Without this patch, gazelle was nearly impossible to adopt for us.

aptenodytes-forsteri:
Filed https://github.com/bazelbuild/bazel-gazelle/issues/1188 to track.

walk/config.go
```diff
@@ -48,34 +49,54 @@ func (wc *walkConfig) isExcluded(rel, base string) bool {
...
+	if err != nil {
+		// doublestar.Match returns only one possible error, and only if the
+		// pattern is not valid. During the configuration of the walker (see
+		// Configure below), we discard any invalid pattern and thus an error
+		// here should not be possible. Same below.
+		log.Panicf("error during doublestar.Match. This should not happen, please file an issue https://github.com/bazelbuild/bazel-gazelle/issues/new: %s", err)
+	}
+	return matched
+}
+
```
sluongng, line 86: This design can lead to mis use of this helper func for non-doublestar cases.

Instead of doing `must(doublestar.Match))`, I would suggest to create something like
```
func dsMustMatch(x, f string) bool {
  err := doublestar.Match(x, f)
  // handle err
}
```
```

File: walk/config.go [hunk 2]
```diff
@@ -48,34 +49,54 @@ func (wc *walkConfig) isExcluded(rel, base string) bool {
...
-			// Configure below), we discard any invalid pattern and thus an error
-			// here should not be possible.
-			log.Panicf("error during doublestar.Match. This should not happen, please file an issue https://github.com/bazelbuild/bazel-gazelle/issues/new: %s", err)
+	isIncluded := false
+	if len(wc.includes) == 0 {
+		isIncluded = true
+	} else {
+		for _, x := range wc.includes {
+			matched := must(doublestar.Match(x, f))
+			if matched {
```
sluongng, line 58: should be able to inline the if clause

walk/config.go
```diff
@@ -48,34 +49,54 @@ func (wc *walkConfig) isExcluded(rel, base string) bool {
...
+				isIncluded = true
+				break
+			}
 		}
+	}
+	if !isIncluded {
+		return true
+	}
+	for _, x := range wc.excludes {
+		matched := must(doublestar.Match(x, f))
```
sluongng, line 68: Nit: overall these additional if-else clauses make the func harder to read.

I would recommend try refactoring this into `if {... <return early>}` instead.

sluongng (commented):
The direction LGTM, pending doc and test.

Personally my mental model for this PR is something like this:

Previously the gazelle set the default value of `include` to the entire repository, then apply `exclude` directives to each found path.

This PR provides a knob where we can narrow down `include` setting to a set of sub paths instead. `exclude` is still applied on top of the found paths.

walk/config.go
```diff
@@ -48,34 +49,54 @@ func (wc *walkConfig) isExcluded(rel, base string) bool {
...
-	for _, x := range wc.excludes {
-		matched, err := doublestar.Match(x, f)
-		if err != nil {
-			// doublestar.Match returns only one possible error, and only if the
-			// pattern is not valid. During the configuration of the walker (see
-			// Configure below), we discard any invalid pattern and thus an error
-			// here should not be possible.
-			log.Panicf("error during doublestar.Match. This should not happen, please file an issue https://github.com/bazelbuild/bazel-gazelle/issues/new: %s", err)
+	isIncluded := false
+	if len(wc.includes) == 0 {
```
aptenodytes-forsteri, line 53: This will fail on `"."` and never walk anything.

Locally, I have ` if len(wc.includes) == 0 || f == "." {`

Whoaa512:
I'd love to help get this in. @f0rmiga do you have bandwidth to address the comments? If not happy to contribute on your fork

f0rmiga:
@Whoaa512 - I can address it later today. Thanks for the ping. Somehow this one got lost on my list.

Issue #1173, opened 2022-02-12
language/proto: dependency resolution from prefixed roots


alexeagle:
I'm working in a large monorepo migrating to Bazel. The proto imports do not currently match the repo structure.
It seems there is no way to instruct the generation of proto_library targets to search subdirectories to resolve import statements.

At prior clients I've been forced to pre-factor all the code by adjusting proto import statements to match the repository structure, but it won't work this time, as the prior build system needs to keep working during the migration.

The affordances for generating [import_prefix](https://docs.bazel.build/versions/main/be/protocol-buffer.html#proto_library.import_prefix) and [strip_import_prefix](https://docs.bazel.build/versions/main/be/protocol-buffer.html#proto_library.strip_import_prefix) only affect how `protoc` is called, not what `deps` gazelle will generate in the `proto_library` targets.

## Repro

Since `protoc` allows multiple include directories, the existing protos are laid out this way:

```
├── BUILD
├── path
│   └── to
│       ├── app
│       │   └── api.proto
│       ├── root1
│       │   └── libA
│       │       └── a.proto
│       └── root2
│           └── libB
│               └── b.proto
└── WORKSPACE
```

And the dependency graph is `api.proto -> b.proto -> a.proto`. However the import statements look like so:

`api.proto`:
```
package app;

import "libB/b.proto";
```

`b.proto`:
```
package libB;

import "libA/a.proto";
```

(full tree: https://github.com/alexeagle/gazelle_proto_resolve_roots/commit/2ce22df2eb98ed1dede8a7cdc02f2e0c3fe45a6d )

Running gazelle in this repo produces non-functional `proto_library` targets, no surprise:

```
proto_library(
    name = "app_proto",
    srcs = ["api.proto"],
    visibility = ["//visibility:public"],
    deps = ["//libB:libB_proto"],  # <--- this is not the right package, needs path/to/root2
)
```

(full gazelle output: https://github.com/alexeagle/gazelle_proto_resolve_roots/commit/00edb8935c8b07db520d576dc126b74a10997f0a)

## Analysis

Of course Gazelle does give an affordance for overriding dependency resolution with the `#gazelle:resolve` directive. In my repro I would add

```
# gazelle:resolve proto libA/a.proto //path/to/root1/libA:a_proto
# gazelle:resolve proto libB/b.proto //path/to/root2/libB:b_proto
```
to the root BUILD file (or some common ancestor of the code to be generated). This fixes the problem.

[...]

PR #1195, opened 2022-03-13
Proto test

aptenodytes-forsteri:
**What type of PR is this?**

> Feature
> Test

> language/proto

Addresses the case where protobufs are scattered throughout a monorepo and meant to be included relative to a few different include paths.

Fixes [#1173](https://github.com/bazelbuild/bazel-gazelle/issues/1173
)

**Other notes for review**


aptenodytes-forsteri:
Aha looks like this fails on Windows.

kevinxw:
Instead of a `include_paths` directive, I would hope for `gazelle:proto_prefix_map` directive instead.
The example usage could be:
```
# maps "github.com/project1/protoA.proto" to @com_github_project1//proto:protoA_proto
gazelle:proto_prefix_map /github.com/project1 @com_github_project1//proto

# maps "root1/protoA.proto" to //root2:protoA_proto
gazelle:proto_prefix_map /root1 //root2 
```

In my case, the multiple external repositories may have conflicting file paths for different proto files. e.g.
`@external_repo1//proto/common.proto` and `@external_repo2//proto/common.proto` . If in local repo we want to import a `proto/common.proto`, a simple include directive won't resolve the conflict.

Even for the problem described in #1173 , `root1` and `root2` may have conflicting proto paths: `root1/common.proto` vs `root2/common.proto`

At last, I hope in the future gazelle can handle special import path automatically without manually specifying the mapping.

e.g. we can use import "@com_github_project1//protoA.proto" directly.

linzhp:
Posted a comment on the original issue: https://github.com/bazelbuild/bazel-gazelle/issues/1173#issuecomment-1094302769

Issue #1199, opened 2022-03-15
gazelle_generation_test fails on windows


aptenodytes-forsteri:
### What version of gazelle are you using?

[This PR](https://github.com/bazelbuild/bazel-gazelle/pull/1195)

### What version of rules_go are you using?

027d78bed3952d73c6fb7099b3247f903aa7318d

### What version of Bazel are you using?

5.0.0

### Does this issue reproduce with the latest releases of all the above?
Yes

### What operating system and processor architecture are you using?
Windows

### What did you do?

On the above PR, I added a usage of the `gazelle_generation_test` macro.

### What did you expect to see?
I expected the test to pass all checks on all operating systems.

### What did you see instead?
My test failed the Windows check, indicating that the macro won't work on Windows.

PR #1200, opened 2022-03-15
fix(tests): use filepath instead of path to handle Windows.

aptenodytes-forsteri:
**What type of PR is this?**

Bug fix

testtools

On https://github.com/bazelbuild/bazel-gazelle/pull/1195, I discovered that `generation_test` does not work on windows.

Fixes https://github.com/bazelbuild/bazel-gazelle/issues/1199


blico:
LGTM, thanks for the quick fix.

PR #1201, opened 2022-03-15
language/go: introduce static dependency resolution mode

### LINKED ISSUE #602: go_repository: statically resolve external imports

blico:
**What type of PR is this?**

Feature

language/go
repo
go_repository

Creates a new `dependencyMode` called `static`, which avoids going out to the network when resolving imports to external labels and instead solely relies on the dependencies Gazelle knows about within WORKSPACE. If no matching dependency is found, Gazelle will skip generating a label for it.

This can greatly improve the performance of `go_repository` rules by eliminating unnecessary network calls when an external dependency contains importpaths that are not used within the main repository (and therefore not needed). 

`static` is now the default `dependencyMode` for `go_repository` rules, given that all known dependencies can be provided via WORKSPACE, and any needed unknown dependencies should already result in build failure.

Fixes #602 

**Other notes for review**
A rough experiment shows in `static` mode it took a `go_repository` rule for `github.com/docker/docker` ~16.5s to fetch, compared to ~633 seconds in `external` mode:
```
$ vi WORKSPACE # invalidate rule
$ time bazel fetch @com_github_docker_docker_external//...
Loading: 849 packages loaded

real	10m33.033s
user	0m0.621s
sys	0m0.320s
```
```
$ vi WORKSPACE # invalidate rule
$ time bazel fetch @com_github_docker_docker_static//...
Loading: 849 packages loaded

real	0m16.587s
user	0m0.499s
sys	0m0.233s
```


repo/remote.go
```diff
@@ -212,6 +212,29 @@ var knownPrefixes = []struct {
 	{prefix: "github.com", missing: 2},
 }
 
+// RootStatic checks the cache to see if the provided importpath matches any known roots.
+// If no matches are found, rather than going out to the network to determine the root,
+// nothing is returned.
+func (r *RemoteCache) RootStatic(importPath string) (root, name string, err error) {
+	prefix := importPath
+	for {
+		v, ok, err := r.root.get(prefix)
```
linzhp, line 221: Is the value of `ok` still reliable when `err != nil`? Should we check `err != nil` first?

repo/remote.go
```diff
@@ -212,6 +212,29 @@ var knownPrefixes = []struct {
 	{prefix: "github.com", missing: 2},
 }
 
+// RootStatic checks the cache to see if the provided importpath matches any known roots.
+// If no matches are found, rather than going out to the network to determine the root,
+// nothing is returned.
+func (r *RemoteCache) RootStatic(importPath string) (root, name string, err error) {
+	prefix := importPath
+	for {
```
linzhp, line 220:
```suggestion
	for prefix := importPath; prefix != "." && prefix != "/"; prefix = path.Dir(prefix){
```

repo/remote.go
blico: Yeah, `err` is a cached value, and `ok` indicates whether or not there was a cache hit, so it is safe to check for `ok` first.

linzhp (approved):

PR #1206, opened 2022-03-18
go_repository: avoid unnecessary fetches in HTTP mode

### LINKED ISSUE #1175: go_repository rule restarts

0.24.0

0.30.0

5.0.0

Yes

Linux/x64

### What did you do?

Clean rebuild of a large project.

No messages about restarting fetches.  Or at least none that last long enough to be visible.

### What did you see instead?

Messages like
```
Fetching @com_github_aws_aws_sdk_go; Restarting.
```

From my own experience writing repository rules, there's a little bit of a gotcha which is that calls to `ctx.path()` on a label referencing files in a repository which hasn't been fetched yet will result in restarting execution of your rule, [as documented](https://docs.bazel.build/versions/5.0.0/skylark/repository_rules.html#when-is-the-implementation-function-executed) though not super clearly and not really emphasized.

In particular,  lines like
https://github.com/bazelbuild/bazel-gazelle/blob/4d2e96a20ad1a32b48bd2dc5c44d0627d34063bf/internal/go_repository.bzl#L150
or
https://github.com/bazelbuild/bazel-gazelle/blob/4d2e96a20ad1a32b48bd2dc5c44d0627d34063bf/internal/go_repository.bzl#L194
will result in a restart, because as the docs say
> To avoid unnecessary restarts (which are expensive, as network access might have to be repeated), label arguments are prefetched, provided all label arguments can be resolved to an existing file. Note that resolving a path from a string or a label that was constructed only during execution of the function might still cause a restart.

That's not great, because that means re-running the fetch from https://github.com/bazelbuild/bazel-gazelle/blob/4d2e96a20ad1a32b48bd2dc5c44d0627d34063bf/internal/go_repository.bzl#L102 (or from other code paths).

Possible workarounds for this are to pre-fetch the result of `ctx.path()` at the beginning of the rule implementation function, before anything more expensive has happened, and/or for static strings add them as private attributes to the rule so that bazel can know to fetch them prior to beginning execution of the rule.

blico:
Currently, if a `go_repository` rule is in HTTP mode and has `build_file_generation = on` it can result in unnecessary calls to `download_and_extract` while the `@bazel_gazelle_go_repository_tools` and `@bazel_gazelle_go_repository_cache` labels are still not available. Note that redundant fetches can be unavoidable when in HTTP mode and `build_file_generation = auto` because the rule needs to fetch and examine the repo to determine if it requires `@bazel_gazelle_go_repository_tools`.

For more info about repository rule restarting see:
https://docs.bazel.build/versions/5.0.0/skylark/repository_rules.html#when-is-the-implementation-function-executed 

Resolve #1175

### DIFF
--- a/internal/go_repository.bzl
+++ b/internal/go_repository.bzl
@@ -93,6 +93,15 @@ def _go_repository_impl(ctx):
     # TODO(#549): vcs repositories are not cached and still need to be fetched.
     # Download the repository or module.
     fetch_repo_args = None
+    gazelle_path = None
+
+    # Declare Label dependencies at the top of function to avoid unnecessary fetching:
+    # https://docs.bazel.build/versions/main/skylark/repository_rules.html#when-is-the-implementation-function-executed
+    go_env_cache = str(ctx.path(Label("@bazel_gazelle_go_repository_cache//:go.env")))
+    generate = ctx.attr.build_file_generation == "on"
+    _gazelle = "@bazel_gazelle_go_repository_tools//:bin/gazelle{}".format(executable_extension(ctx))
+    if generate:
+        gazelle_path = ctx.path(Label(_gazelle))
 
     if ctx.attr.urls:
         # HTTP mode
@@ -147,7 +156,7 @@ def _go_repository_impl(ctx):
     else:
         fail("one of urls, commit, tag, or importpath must be specified")
 
-    env = read_cache_env(ctx, str(ctx.path(Label("@bazel_gazelle_go_repository_cache//:go.env"))))
+    env = read_cache_env(ctx, go_env_cache)
     env_keys = [
         # Respect user proxy and sumdb settings for privacy.
         # TODO(jayconrod): gazelle in go_repository mode should probably
@@ -212,7 +221,7 @@ def _go_repository_impl(ctx):
             existing_build_file = name
             break
 
-    generate = (ctx.attr.build_file_generation == "on" or (not existing_build_file and ctx.attr.build_file_generation == "auto"))
+    generate = generate or (not existing_build_file and ctx.attr.build_file_generation == "auto")
 
     if generate:
         # Build file generation is needed. Populate Gazelle directive at root build file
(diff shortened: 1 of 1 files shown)


internal/go_repository.bzl
```diff
@@ -93,6 +93,15 @@ def _go_repository_impl(ctx):
...
     fetch_repo_args = None
+    gazelle_path = None
+
+    # Declare Label dependencies at the top of function to avoid unnecessary fetching:
+    # https://docs.bazel.build/versions/main/skylark/repository_rules.html#when-is-the-implementation-function-executed
+    go_env_cache = str(ctx.path(Label("@bazel_gazelle_go_repository_cache//:go.env")))
+    generate = ctx.attr.build_file_generation == "on"
+    _gazelle = "@bazel_gazelle_go_repository_tools//:bin/gazelle{}".format(executable_extension(ctx))
+    if generate:
+        gazelle_path = ctx.path(Label(_gazelle))
```
adam-azarchs, line 104: Does this really need to be conditional on `generate`?  I think (but am not 100% sure) that `bazel_gazelle_go_repository_tools` gets loaded at  some point along the line regardless.  In the fairly common case where `build_file_generation == "auto"`, we could potentially save a lot of trouble with
```suggestion
    if generate or ctx.attr.build_file_generation == "auto":
        gazelle_path = ctx.path(Label(_gazelle))
```
(and then removing the bit down at https://github.com/bazelbuild/bazel-gazelle/pull/1206/files#diff-ffef820c4796a4da27ec8c4f4c743464559ffcd596aec22bef4ff85e7ccef8c3R236-R237)

internal/go_repository.bzl
```diff
@@ -212,7 +221,7 @@ def _go_repository_impl(ctx):
             existing_build_file = name
             break
```
adam-azarchs, line 222: The loop that these are in (which I can't comment on entirely due to GitHub's UI) includes some use of `ctx.path`, so could potentially result in a restart.  Moving `build_file_paths = [ctx.path(name) for name in ctx.attr.build_file_name.split(",")]` to before the fetch step might help a lot.

adam-azarchs (commented):
Looks like this will help.  Thanks!

internal/go_repository.bzl
blico: Hmmm, so when we are in HTTP mode and `build_file_generation == "auto"` then it is possible for us to not need to load `go_repository_tools` (if the repo already has an existing BUILD file). So if we made this unconditional, it could result in unnecessary loads. I don't think we should do that, especially given how long it can take to fetch `go_repository_tools`.

This comment did cause me to realize we will always load `go_repository_tools` when `not ctx.attr.urls`, so I moved that logic to the top, which should also help.

PR #1207, opened 2022-03-18
resolve: improve matching of two resolvers (#1162)

### LINKED ISSUE #1162: [proto] Unable to properly find the list of embedded targets when the native rule is overridden

`v0.24.0`

`None`

`None`

Yes

`Linux` / `amd64`

### What did you do?

I created a macro to override the `go_library` rule and I run `gazelle` to update a `BUILD` file for a `.proto` file which import another `.proto` file.

To highlight the issue, we can use this minimal setup below:

- An empty `WORKSPACE` file
- `BUILD` file

```skylark
# gazelle:prefix github.com/lbcjbb/test
# gazelle:map_kind go_library my_go_library //:my.bzl
```

- `my.bzl`

```skylark

my_go_library = go_library
```

- `a/a.proto`

```proto
syntax = "proto3";

package test;
option go_package = "github.com/lbcjbb/test/a";
```

- `b/b.proto`

```proto
syntax = "proto3";

package test;
option go_package = "github.com/lbcjbb/test/b";

import "a/a.proto";
```

And execute this command below inside this directory:

```sh
$ gazelle update
```

- No warnings from `gazelle`
- A `b/BUILD.bazel` with a `go_proto_library` rule with a `deps` attribute containing the `//a` label.

### What did you see instead?

- This warning below:

```
gazelle: multiple rules (//a:a_go_proto and //a) may be imported with "a/a.proto" from //b:b_go_proto
```

- No `deps` attribute for the `go_proto_library` rule in `b/BUILD.bazel`. 

All the other rules in `a/BUILD.bazel` and `b/BUILD/bazel` files are correctly added.

lbcjbb:
A Resolver can be overriden, this is the case by example for all rules
from the map_kind operation.

This layer must be transparent while comparing two resolvers.

**What type of PR is this?**

Bug fix

resolve

These modifications provide an abstraction of the inverseMapKindResolver to compare two Resolver objects.

Fixes #1162

**Other notes for review**

Thank you @blico for the test <3.

### DIFF
--- a/cmd/gazelle/integration_test.go
+++ b/cmd/gazelle/integration_test.go
@@ -2935,6 +2935,107 @@ go_library(
 	})
 }
 
+// TestMapKindEmbeddedResolve tests the gazelle:map_kind properly resolves
+// dependencies for embedded rules (see #1162).
+func TestMapKindEmbeddedResolve(t *testing.T) {
+	files := []testtools.FileSpec{
+		{
+			Path: "WORKSPACE",
+		}, {
+			Path: "BUILD.bazel",
+			Content: `
+# gazelle:prefix example.com/mapkind
+# gazelle:map_kind go_library my_go_library //:my.bzl
+`,
+		}, {
+			Path: "a/a.proto",
+			Content: `
+syntax = "proto3";
+
+package test;
+option go_package = "example.com/mapkind/a";
+`,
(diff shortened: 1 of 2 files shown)


blico:
Thank you for the fix!

Issue #1133, opened 2021-11-15
embedsrcs ignores directories with BUILD files


uhthomas:
### What version of gazelle are you using?

v0.24.0

### What version of rules_go are you using?

v0.29.0

### What version of Bazel are you using?

4.2.1

### Does this issue reproduce with the latest releases of all the above?

Yes.

### What operating system and processor architecture are you using?

Tested on MacOS and Linux.

### What did you do?

Assuming the directory structure

```
cmd/something
├── BUILD.bazel
├── data
│   ├── BUILD.bazel
│   ├── something.txt
│   ├── some_other_thing
│   │   ├── some_other_thing.txt
└── main.go
```

and with `main.go`

```go
package main

import "embed"

//go:embed data
var content = embed.FS

func main() {}
```

Gazelle won't generate an `embedsrcs` for the "data" directory.

```sh
❯ bazel run //:gazelle
gazelle: cmd/something/main.go:5:2: pattern data: matched no files
```

### What did you expect to see?

`embedsrcs` should be generated.

### What did you see instead?

Gazelle doesn't generate `embedsrcs` for directories with BUILD files.

andyliuliming:
we need this too.

achew22:
I don't think this can actually be done automatically. By making the directory that you want to include have a `BUILD` file, bazel wouldn't be able to reach into the directory using something like `glob`. It would require synchronizing a `filegroup` or something in the inner directory which is not going to be something that's completely standardized between projects (some will want direct file enumeration vs `exports_files` vs `filegroup` vs custom type'd target, etc.). If you have a need for this, move the targets that you defined in `data` up a level into the `cmd/something/BUILD` file and then refer to the files in the inner directory as `data/something.txt`.

If you feel really strongly about this and would like to pick one of those implementations and contribute it as a patch, let's open another issue and discuss the way it would be done.

uhthomas:
The new `subpackages` function may be helpful?

https://github.com/bazelbuild/bazel/pull/14780

Design doc: https://docs.google.com/document/d/13UOT0GoQofxDW40ILzH2sWpUOmuYy6QZ7CUmhej9vgk/edit#heading=h.5mcn15i0e1ch

PR #1208, opened 2022-03-22
language/go: support embedding subdirs containing buildable subdirs

### LINKED ISSUE #1192: `go:embed` does not work for subdirs containing buildable subdirs

`v0.24.0` (also tried latest commit in master)

`v0.29.0` (would use `0.30.0` but it's broken 😄 without a devel version of Gazelle)

`5.0.0` (also tried `4.2.1`)

Yeah.

Linux x86_64

### What did you do?

Try to build https://github.com/pressly/goose.

Reproduction: https://github.com/uhthomas/bazel-gazelle-issue-1192

Successful build.

### What did you see instead?

```
compilepkg: /home/thomas/.cache/bazel/_bazel_thomas/b57cca483c8e5e6ea6286570e14f5568/sandbox/linux-sandbox/223/execroot/__main__/external/com_github_pressly_goose_v3/goose_test.go:158:12: could not embed examples/sql-migrations/*.sql: no matching files found
FAILED: Build did NOT complete successfully
```

blico:
**What type of PR is this?**

Bug fix

language/go

Fixes a bug where Gazelle was failing to embed files of a subdir, if it contained buildable subdirs.

For example, Gazelle would not find `bar.txt` for `//go:embed foo/bar.txt` if `foo` also contained a package `baz` which had go files.

`goPkgRels` is currently being used to determine if a directory is embeddable and if a `testdata` directory can be added as a `data` attribute. These two cases are the same unless a subdir contains buildable subdirs, in which the directory should still be embeddable but if it is `testdata` it should not be added as a `data` attribute.

The fix is to ensure `goPkgRels` correctly reflects whether a dir is a go package or not. To ensure `testdata` still conforms to the old logic, rather than checking for the value of `goPkgRels` we check if a value exists for the given dir.

Fixes #1192

### DIFF
--- a/language/go/generate.go
+++ b/language/go/generate.go
@@ -110,7 +110,8 @@ func (gl *goLang) GenerateRules(args language.GenerateArgs) language.GenerateRes
 	var hasTestdata bool
 	for _, sub := range args.Subdirs {
 		if sub == "testdata" {
-			hasTestdata = !gl.goPkgRels[path.Join(args.Rel, "testdata")]
+			_, ok := gl.goPkgRels[path.Join(args.Rel, "testdata")]
+			hasTestdata = !ok
 			break
 		}
 	}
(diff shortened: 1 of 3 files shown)


uhthomas:
@blico Does this also address https://github.com/bazelbuild/bazel-gazelle/issues/1133 or am I misunderstanding the nature of this fix?

blico:
@uhthomas no this only addresses #1192, as pointed out in the issue a fix for #1133 will be more involved.

PR #1213, opened 2022-03-24
Add -print0 to print names of rewritten files

### LINKED ISSUE #1028: Add -print0 to make gazelle scripting friendly

v0.22.2

v0.24.7

4.0

We would love to use gazelle in pre-commit-hook to update build files automatically. When doing so, we need to know which build file is rewritten and needs to be added to the staging area for committing.

We propose to add `-print0` flag for `-mode=fix`. When set, let gazelle print the names of rewritten files separated with '\0' (like other Unix tools,) then the output could be piped to `xargs -r0 git add`.

dr-dime:
This CL adds a config to fix/update routine to output BUILD files being
rewritten by gazelle when using the fix mode. This will make scripting
around gazelle much easier, eg. run gazelle in a pre-commit hook then
commit changed BUILD files:

    bazel run //:gazelle -- -print0 | xargs -0 git add

`-print0` is borrowed from `find(1)` to use null character as the delimiter
between file names, so other tools can correctly process file names with
whitespaces.

The alternative is to use the patch file mode, apply the patch then
extract the changed files from the patch. However, it depends on many
tools and requires some serious shell skills to get it right. Which I deem
not script friendly.

Fixes #1028


achew22:
Hey @dr-dime, thanks for sending this in. I think it's a good feature to have. Could I have you add a test, as I described [here](https://github.com/bazelbuild/bazel-gazelle/pull/1214#issuecomment-1083433450)?

dr-dime:
@achew22 PTAL

achew22 (approved):

PR #1216, opened 2022-03-25
Add an e2e test confirming no output on success

achew22:
**What type of PR is this?**

Other - adding a test.

`testing`

This test is an example of how to use the new e2e testing framework to write a basic test.


robfig (commented):
This test framework is great, thank you.

PR #1222, opened 2022-03-30
Update extend.md with a practical languages example

Anthony-Bible:
**What type of PR is this?**

Documentation

N/A

I want to add a practical example of using multiple languages in gazelle_binary. This was a pain point for me when first trying to use gazelle for both go and python so hopefully this is a good addition to provide a foundation for others to use.

N//A

**Other notes for review**

### DIFF
--- a/extend.md
+++ b/extend.md
@@ -35,8 +35,6 @@ Moved to [/README.rst](/README.rst#supported-languages)
 Example
 -------
 
-**TODO:** Add a self-contained, concise, realistic example.
-
 Gazelle itself is built using the model described above, so it may serve as
 an example.
 
@@ -54,7 +52,13 @@ load("@bazel_gazelle//:def.bzl", "DEFAULT_LANGUAGES", "gazelle_binary")
 
 gazelle_binary(
     name = "gazelle",
-    languages = DEFAULT_LANGUAGES,
+    languages = [
+        "@rules_python//gazelle",  # Use gazelle from rules_python.
+        "@bazel_gazelle//language/go",  # Built-in rule from gazelle for Golang.
+        "@bazel_gazelle//language/proto",  # Built-in rule from gazelle for Protos.
+         # Any languages that depend on Gazelle's proto plugin must come after it.
+        "@external_repository//language/gazelle",  # External languages can be added here.
+    ],
     visibility = ["//visibility:public"],
 )
 ```
--- a/internal/extend_docs.bzl
+++ b/internal/extend_docs.bzl
@@ -35,8 +35,6 @@ Moved to [/README.rst](/README.rst#supported-languages)
 Example
 -------
 
-**TODO:** Add a self-contained, concise, realistic example.
-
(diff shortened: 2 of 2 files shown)


extend.md
```diff
@@ -54,7 +54,14 @@ load("@bazel_gazelle//:def.bzl", "DEFAULT_LANGUAGES", "gazelle_binary")
 
 gazelle_binary(
     name = "gazelle",
-    languages = DEFAULT_LANGUAGES,
+    languages = [
+        "@rules_python//gazelle", #use gazelle from rules_python
```
achew22, line 58: Nit: in `.bzl` files, comments at the end of the line have two spaces then the `#` and then a space and then the comment. The comment should be a sentence and end with a period.

```bzl
        "@rules_python//gazelle",  # Use Gazelle from rules_python.
```

extend.md
```diff
@@ -54,7 +54,14 @@ load("@bazel_gazelle//:def.bzl", "DEFAULT_LANGUAGES", "gazelle_binary")
...
+        "@bazel_gazelle//language/go", #built in rule from gazelle for golang
+        "@bazel_gazelle//language/proto", #built in rule from gazelle for protos
+        # must be after the proto extension (order matters)
```
achew22, line 61: What do you think about saying

```
# Any languages that depend on Gazelle's proto plugin must come after it.
```

?

extend.md
```diff
@@ -54,7 +54,14 @@ load("@bazel_gazelle//:def.bzl", "DEFAULT_LANGUAGES", "gazelle_binary")
...
+        "@build_stack_rules_proto//language/protobuf", #use gazelle to create protobufs see https://github.com/stackb/rules_proto for more  info
```
achew22, line 62: I would rather not include an external plugin in the documentation for fear that it will look like an endorsement by the project. I'm sure that stack.build's rules_proto is great, but I've not used it and so I'd rather we make up another 3rd party plugin.

Could we go with something like

```
        "@external_repository//language/gazelle"  # External languages can also be added here. 
```

achew22 (changes requested):
The `extend.md` docs are generated by generating StarDoc from a number of input files. The section you're editing, I believe, comes from

https://github.com/bazelbuild/bazel-gazelle/blob/08c7bc0d836dc09c433db2d35b300ecd8433734e/internal/extend_docs.bzl#L52-L60

Then you can `bazel run //docs:update` and the `.md` file will be written out.

extend.md
Anthony-Bible: I completely understand, I went ahead and implemented your solution since it solves both a generic plugin while still showing an usable example.

extend.md
Anthony-Bible: I reformatted my comments to follow that convention. If it needs further intervention I'm happy to make more changes.

achew22:
I think you need to update the generated files, but otherwise looks good to me

achew22 (approved):

achew22:
@Anthony-Bible thanks for your contribution!

PR #1226, opened 2022-04-04
fix: Remove gazelle_binary import collision

illicitonion:
**What type of PR is this?**

Bug fix

internal/gazelle_binary

These changes make the import aliases for the generated Go code in gazelle_binary longer to avoid collisions.

Without theses changes, importing 2 gazelle extensions with the same base path would fail to build the gazelle binary:

```starlark
gazelle_binary(
    name = "gazelle_bin",
    languages = [
        # importpath = github.com/bazelbuild/bazel-skylib/gazelle
        "@bazel_skylib//gazelle/bzl",
        # importpath = github.com/bazelbuild/rules_python/gazelle
        "@rules_python//gazelle:gazelle_python_binary",
    ],
)
```

```console
$ bazel build //:gazelle_bin
[...  snip]
bazel-out/darwin-fastbuild/bin/gazelle_bin_/langs.go:10:2: gazelle_ redeclared in this block
	bazel-out/darwin-fastbuild/bin/gazelle_bin_/langs.go:9:2: other declaration of gazelle_
bazel-out/darwin-fastbuild/bin/gazelle_bin_/langs.go:10:2: imported and not used: "github.com/bazelbuild/rules_python/gazelle" as gazelle_
```

The **current** generated bazel-out/darwin-fastbuild/bin/gazelle_bin_/langs.go file:

```go
package main

import (
	"github.com/bazelbuild/bazel-gazelle/language"

	gazelle_ "github.com/bazelbuild/bazel-skylib/gazelle"
	gazelle_ "github.com/bazelbuild/rules_python/gazelle"
)

var languages = []language.Language{
	gazelle_.NewLanguage(),
	gazelle_.NewLanguage(),
}
```

With the proposed changes, this file would become:

```go
package main

import (
	"github.com/bazelbuild/bazel-gazelle/language"

	github_com_bazelbuild_bazel_skylib_gazelle_ "github.com/bazelbuild/bazel-skylib/gazelle"
	github_com_bazelbuild_rules_python_gazelle_ "github.com/bazelbuild/rules_python/gazelle"
)

var languages = []language.Language{
	github_com_bazelbuild_bazel_skylib_gazelle_.NewLanguage(),
	github_com_bazelbuild_rules_python_gazelle_.NewLanguage(),
}
```

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -88,3 +88,7 @@ http_archive(
         "https://github.com/bazelbuild/stardoc/releases/download/0.5.0/stardoc-0.5.0.tar.gz",
     ],
 )
+
+load("@bazel_skylib//lib:unittest.bzl", "register_unittest_toolchains")
+
+register_unittest_toolchains()
--- a/internal/BUILD.bazel
+++ b/internal/BUILD.bazel
@@ -1,5 +1,6 @@
 load("@io_bazel_rules_go//go/tools/bazel_testing:def.bzl", "go_bazel_test")
 load("@bazel_skylib//:bzl_library.bzl", "bzl_library")
+load(":gazelle_binary_test.bzl", "gazelle_binary_test_suite")
 
 # gazelle:exclude *_test.go
 go_bazel_test(
@@ -14,6 +15,8 @@ go_bazel_test(
     ],
     deps = ["//testtools"],
 )
+
+gazelle_binary_test_suite()
 
 # TODO(jayconrod): test fetch_repo error cases.
 
@@ -37,6 +40,7 @@ filegroup(
         "extend_docs.bzl",
         "gazelle.bash.in",
         "gazelle_binary.bzl",
+        "gazelle_binary_test.bzl",
         "go_repository.bzl",
         "go_repository_cache.bzl",
         "go_repository_config.bzl",
--- a/internal/gazelle_binary.bzl
+++ b/internal/gazelle_binary.bzl
@@ -119,10 +119,11 @@ def gazelle_binary_wrapper(**kwargs):
             fail("gazelle_binary attribute '%s' is no longer supported (https://github.com/bazelbuild/bazel-gazelle/issues/803)" % key)
     gazelle_binary(**kwargs)
 
+def _import_alias(importpath):
+    return importpath.replace("/", "_").replace(".", "_").replace("-", "_") + "_"
+
 def _format_import(importpath):
-    _, _, base = importpath.rpartition("/")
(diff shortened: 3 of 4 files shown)


achew22:
This seems like a good fix, but it's a smidge difficult to validate as just a code change. Is there any way you could add a [test case] for this?

You can create a new file `internal/gazelle_binary_test.bzl` with something like

```bzl
load("@bazel_skylib//lib:unittest.bzl", "asserts", "unittest")
load(":gazelle_binary.bzl", format_import="_format_import", format_call="_format_call")

def _gazelle_binary_test_impl(ctx):
  env = unittest.begin(ctx)
  asserts.equals(env, "formatted_call", format_call("asdf"))
  asserts.equals(env, "formatted_import", format_import("asdf"))
  return unittest.end(env)

_gazelle_binary_test = unittest.make(_gazelle_binary_test_impl)

def gazelle_binary_test_suite(name):
  # unittest.suite() takes care of instantiating the testing rules and creating
  # a test_suite.
  unittest.suite(
    name,
    _gazelle_binary_test,
    # more tests if you want
  )
```

In `internal/BUILD.bazel` you can do:

```bzl
load(":gazelle_binary_test.bzl", "gazelle_binary_test_suite")

gazelle_binary_test_suite(name = "gazelle_binary_tests")
```


[test case]: https://bazel.build/rules/testing#testing-starlark-utilities

achew22 (changes requested):
Oops, replied in the wrong place. See previous comment

illicitonion:
Done - do you know if there's a way to import private functions in tests?

I tried importing the private functions as you suggested in the example:

```starlark
load(":gazelle_binary.bzl", format_import="_format_import", format_call="_format_call")
```

But this does not work apparently:

```
ERROR: [snip]/internal/gazelle_binary_test.bzl:2:44: symbol '_format_call' is private and cannot be imported
ERROR: [snip]/internal/gazelle_binary_test.bzl:2:76: symbol '_format_import' is private and cannot be imported
```

If there is a way to import private functions, I would prefer it. If not I made the functions importable in these changes. I understand the files are in the internal directory already, so their "repo-local" nature is already encoded.

achew22 (approved):
I did a little sleuthing and I don't think that it's possible to do that. Fortunately, these functions live in `internal` so they are not importable to an external party, they would have to explicitly be imported and reexported by something outside that directory so I don't think the visibility of this is being significantly increased and I don't want to block `rules_python` being imported just because of this.

Thanks so much for your contribution!

PR #1227, opened 2022-04-05
Add label pattern matching logic

illicitonion:
**What type of PR is this?**

Feature

label

This implements target patterns as described at https://bazel.build/docs/build#specifying-build-targets

It is not immediately used in Gazelle, but as Gazelle has a solid label
parsing library, it felt like a useful place for this logic to live for
general use.


illicitonion:
I'm not really sure what's going on on CI - it looks like a bazel_test is expecting a version of `internal/go_repository_tools_srcs.bzl` without the new file listed, but I've not managed to work out why. Any pointers would be appreciated!

achew22:
Did you regenerate it using the command that is at the top of the file, or did you manually add your new file?

That file exists to make packaging the repo easier. In my experience, it might actually make things harder. I think it is necessary, unfortunately.

illicitonion:
> Did you regenerate it using the command that is at the top of the file, or did you manually add your new file?

I ran the command - if I run the command again, the file doesn't get changed... Very weird...

achew22:
Even stranger, the diff seems to work in Windows?! I'm quite lost on this one...

illicitonion:
@achew22 I think I fixed it up - turned out I was missing a data dependency on on new files, so they weren't getting copied in to the bazel workspace getting set up for the integration tests - please take a look when you have a moment!

achew22:
Sorry to be slow on this, every time I look at this I think "yep, this code is correct, but should this actually live in Gazelle". It feels like a thing that should be a part of [buildifier's definition of a label](https://pkg.go.dev/github.com/bazelbuild/buildtools@v0.0.0-20220413184717-25dab36ca732/labels) and we should just use that inside our codebase. At the same time, I fear the ship may have sailed on that. This is already exported as a package so I don't think we can take it away without causing everyone who's written a plugin to have to do a migration.

I guess we're SOL on this, but I wanted to at least ask, is there something we can do here that you can think of?

illicitonion:
Yeah, I had completely the same thoughts - it shouldn't obviously live here, but the label implementation here is more fully featured than the one in buildifier, but the `Matches` function needs to tie to _some_ `Label` implementation... I don't really mind where the code lives (and in fact I can just put it in my project if that works better), but figured I'd offer it up to avoid someone else having to write the same thing again.

Happy to do whatever makes sense with it :)

PR #1229, opened 2022-04-06
Broaden label name regex

illicitonion:
**What type of PR is this?**

Bug fix

label

Broadens the accepted names in labels.
These are documented at https://docs.bazel.build/versions/main/build-ref.html#name


achew22 (approved):

achew22:
Thanks for the fix!

PR #1231, opened 2022-04-06
gazelle_generation_test: redact workspace path from output

dr-dime:
gazelle_generation_test uses the golden output to check the gazelle run results.
However, gazelle may occasionally output the full workspace path in logs, which
makes the test unreproducible, thus impossible to generate a golden output.

This CL replaces the workspace path in the output with a constant token, to make
the tests reproducible.

(Actually, this won't work on windows, as the output path will be with backward
slashes and the golden files are with forward slashes. But I'll leave it as it
is for the sake of testing.)

### DIFF
--- a/testtools/files.go
+++ b/testtools/files.go
@@ -260,6 +260,7 @@ func TestGazelleGenerationOnPath(t *testing.T, args *TestGazelleGenerationArgs) 
 		})
 
 		testdataDir, cleanup := CreateFiles(t, inputs)
+		workspaceRoot := filepath.Join(testdataDir, args.Name)
 
 		var stdout, stderr bytes.Buffer
 		var actualExitCode int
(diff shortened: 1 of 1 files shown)


testtools/files.go
```diff
@@ -260,6 +260,7 @@ func TestGazelleGenerationOnPath(t *testing.T, args *TestGazelleGenerationArgs)
 		})
 
 		testdataDir, cleanup := CreateFiles(t, inputs)
+		workspaceRoot := filepath.Join(testdataDir, args.Name)
```
dr-dime, line 263: Alternatively, I can declare `redactWorkspacePath := func(s string) string` here, and have it captures the `workspaceRoot` to make call sites clearer. But this function is too damn long... I don't want to make it even more complex.

achew22: Yes, this function is already too complex for an inlined anonymous. A refactor would be in order but I think is for another day.

testtools/files.go
```diff
@@ -398,3 +398,9 @@ func updateExpectedConfig(t *testing.T, expected string, actual string, srcTestD
 		}
 	}
 }
+
+// redactWorkspacePath replaces workspace path with a constant to make the test
+// output reproducible.
+func redactWorkspacePath(s, wsPath string) string {
+	return strings.ReplaceAll(s, wsPath, "%WORKSPACEPATH%")
```
achew22, line 405: I think a bit more is needed to make this work in Windows

```
Executing tests from //tests:fix_mode_strict
-----------------------------------------------------------------------------
--- FAIL: TestFullGeneration (0.04s)
    --- FAIL: TestFullGeneration/fix_mode_strict (0.04s)
        files.go:187: expected gazelle stderr: gazelle: <WORKSPACE>/BUILD.bazel:3:13: syntax error near visibility
            gazelle: Exit as strict mode is on
            
            got: gazelle: <WORKSPACE>\BUILD.bazel:3:13: syntax error near visibility
            gazelle: Exit as strict mode is on
            
        files.go:306: 
            Run UPDATE_SNAPSHOTS=true bazel run //tests:fix_mode_strict to update BUILD.out and expected{Stdout,Stderr,ExitCode}.txt files.
FAIL
```

From [here](https://buildkite.com/bazel/bazel-gazelle/builds/1752#9a5fa00c-43f0-4e96-a5ce-68af6d78257e)

I think the difference is the direction of the `/` character. I wanted to quick flag flag this in case you've got a clever strategy for fixing this.

achew22: Actually, now that I read your comment it seems something you're aware of. Let's merge this and talk it over in the CL where the tests is actually failing.

PR #1234, opened 2022-04-12
Handle wrapped errors

illicitonion:
**What type of PR is this?**

Bug fix

language/go

Handle wrapped errors

`go mod download` returns a different error structure than `go list`, as
documented at https://go.dev/ref/mod#go-mod-download and https://go.dev/ref/mod#go-list-m

We were previously parsing these assuming they had the same structure,
instead, parse them separately.

Also, prevent infinite looping on invalid JSON.

### DIFF
--- a/language/go/modules.go
+++ b/language/go/modules.go
@@ -40,7 +40,8 @@ func importReposFromModules(args language.ImportReposArgs) language.ImportReposR
 
 	// List all modules except for the main module, including implicit indirect
 	// dependencies.
-	type module struct {
+	// Schema is documented at https://go.dev/ref/mod#go-list-m
+	type moduleFromList struct {
 		Path, Version, Sum, Error string
 		Main                      bool
 		Replace                   *struct {
@@ -48,14 +49,28 @@ func importReposFromModules(args language.ImportReposArgs) language.ImportReposR
 		}
 	}
 	// path@version can be used as a unique identifier for looking up sums
-	pathToModule := map[string]*module{}
+	pathToModule := map[string]*moduleFromList{}
 	data, err := goListModules(dir)
+	dec := json.NewDecoder(bytes.NewReader(data))
 	if err != nil {
+		// Best-effort try to adorn specific error details from the JSON output.
+		for dec.More() {
+			var dl moduleFromList
+			if decodeErr := dec.Decode(&dl); decodeErr != nil {
+				// If we couldn't parse a possible error description, just return the raw error.
+				err = fmt.Errorf("%w\nError parsing module for more error information: %v", err, decodeErr)
+				break
+			}
+			if dl.Error != "" {
+				err = fmt.Errorf("%w\nError listing %v: %v", err, dl.Path, dl.Error)
+			}
+		}
+		err = fmt.Errorf("error from go list: %w", err)
+
 		return language.ImportReposResult{Error: err}
 	}
-	dec := json.NewDecoder(bytes.NewReader(data))
 	for dec.More() {
-		mod := new(module)
+		mod := new(moduleFromList)
 		if err := dec.Decode(mod); err != nil {
 			return language.ImportReposResult{Error: err}
 		}
@@ -103,6 +118,20 @@ func importReposFromModules(args language.ImportReposArgs) language.ImportReposR
 		}
 	}
 
+	type downloadError struct {
+		Err string
+	}
+
+	// Schema is documented at https://go.dev/ref/mod#go-mod-download
+	type moduleFromDownload struct {
+		Path, Version, Sum string
+		Main               bool
+		Replace            *struct {
+			Path, Version string
+		}
+		Error *downloadError
+	}
+
 	if len(missingSumArgs) > 0 {
 		tmpDir, err := ioutil.TempDir("", "")
 		if err != nil {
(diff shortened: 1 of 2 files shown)


achew22:
Forgive my naivete on this one, but it seems that you're creating a new struct, which you don't create any instance of, and then call a method on it? Are you depending on struct equivalence to call your unwrap method?

Can we get away with doing a type switch on this to decide if it is an `error` or `string` and then put a struct in that has both of those as a fields and returns the one that is non-zero in the `Unwrap` method?

illicitonion:
We're using Go's built-in json deserialization [here](https://github.com/bazelbuild/bazel-gazelle/blob/ec1c643571ece7bb5cac359b96c17b3641638c5d/language/go/modules.go#L129-L131): so while we're not manually creating an instance of the `jsonError` struct, Go is doing so for us. [This StackOverflow post](https://stackoverflow.com/questions/48305425/json-key-can-either-be-a-string-or-an-object) suggested that this is the cleanest way to have the json deserializer automatically deserialize either a string or struct.

I don't _think_ here's a handy hook-point where we can insert our own logic here, as we're relying on the stdlib `json.Unmarshal`, but will gladly restructure if you can nudge me at how :)

sluongng:
I would suggest adding a test for this change, including an assertion for how `go mod download -json` outputs would look like if `InvalidPathError` was to return and how that is failing the decode. Then provide a fix and flip the test assertion.

That would help maintainers review and merge this change a lot easier (without having to go and read up all the link reference).

It's weird how you are experiencing decoding issue since we are essentially following go documentation https://golang.org/ref/mod#go-mod-download here:

```
> go mod help download
...
The -json flag causes download to print a sequence of JSON objects
to standard output, describing each downloaded module (or failure),
corresponding to this Go struct:

    type Module struct {
        Path     string // module path
        Version  string // module version
        Error    string // error loading module
        Info     string // absolute path to cached .info file
        GoMod    string // absolute path to cached .mod file
        Zip      string // absolute path to cached .zip file
        Dir      string // absolute path to cached source root directory
        Sum      string // checksum for path, version (as in go.sum)
        GoModSum string // checksum for go.mod (as in go.sum)
    }
...

> go version
go version go1.18.1 darwin/arm64
```

I am still not yet sure on the fix direction just yet, but I can review again once the test is provided.

illicitonion:
Sorry for the incomplete PR - I did some digging, and found that the root cause is actually that we use the same code to parse `go list` and `go mod download` output, but they actually use different formats. Added some tests for both cases - PTAL!

language/go/modules.go
```diff
@@ -40,22 +40,35 @@ func importReposFromModules(args language.ImportReposArgs) language.ImportReposR
...
 	if err != nil {
+		// Best-effort try to adorn specific error details from the JSON output.
+		for dec.More() {
+			var dl moduleFromList
+			if err := dec.Decode(&dl); err != nil {
+				// If we couldn't parse a possible error description, just return the raw error.
+				break
+			}
+			if dl.Error != "" {
+				err = fmt.Errorf("%w\nError downloading %v: %v", err, dl.Path, dl.Error)
```
sluongng, line 64: Should be `error listing go modules` instead of `error downloading` ?

language/go/modules.go
```diff
@@ -40,22 +40,35 @@ func importReposFromModules(args language.ImportReposArgs) language.ImportReposR
...
+	pathToModule := map[string]*moduleFromList{}
 	data, err := goListModules(dir)
+	dec := json.NewDecoder(bytes.NewReader(data))
 	if err != nil {
+		// Best-effort try to adorn specific error details from the JSON output.
+		for dec.More() {
+			var dl moduleFromList
+			if err := dec.Decode(&dl); err != nil {
+				// If we couldn't parse a possible error description, just return the raw error.
+				break
```
sluongng, line 61: Nit: could we wrap the error before breaking to let user know that this was caused specifically when we try to decode `go list` json output?

language/go/modules.go
```diff
@@ -114,20 +141,20 @@ func importReposFromModules(args language.ImportReposArgs) language.ImportReposR
 		if err != nil {
 			// Best-effort try to adorn specific error details from the JSON output.
 			for dec.More() {
-				var dl module
+				var dl moduleFromDownload
 				if err := dec.Decode(&dl); err != nil {
-					// If we couldn't parse a possible error description, just ignore this part of the output.
-					continue
+					// If we couldn't parse a possible error description, just return the raw error.
+					break
```
sluongng, line 147: ditto: please consider wrapping the decode error before breaking.

language/go/update_import_test.go
```diff
@@ -228,32 +229,120 @@ go_repository(
...
+				definitely.doesnotexist/ever v0.1.0
+			)
+			`,
 				}, {
 					Path: "go.sum",
 					Content: `
-					definitely.doesnotexist/ever v0.1.0/go.mod h1:HI93XBmqTisBFMUTm0b8Fm+jr3Dg1NNxqwp+5A1VGuJ=
-					`,
+								definitely.doesnotexist/ever v0.1.0/go.mod h1:HI93XBmqTisBFMUTm0b8Fm+jr3Dg1NNxqwp+5A1VGuJ=
+								`,
```
sluongng, line 247: ditto: not sure if this is needed

language/go/update_import_test.go
```diff
@@ -228,32 +229,120 @@ go_repository(
...
+			module github.com/bazelbuild/bazel-gazelle
 
-require (
-	definitely.doesnotexist/ever v0.1.0
-)
-`,
+			require (
+				definitely.doesnotexist/ever v0.1.0
+			)
+			`,
```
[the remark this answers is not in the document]
illicitonion: It's not! Not sure why my IDE decided to do so - fixed

illicitonion:
All addressed - PTAL

achew22:
This is much cleaner than that was happening before. Thank you for fixing it!

PR #1235, opened 2022-04-12
Add 'gazelle:proto_group' semantics for 'gazelle:proto file' mode.

pcj:
This makes it possible to include additional files in the srcs list of a 'file' mode generated rule.

**What type of PR is this?**

> Feature

> language/proto

This augments the 'file' mode of gazelle to be able to co-locate other srcs with the generated rule.
File mode is great for complex proto setups that otherwise would trigger dependency cycles, but 
sometimes you need to piggyback other srcs with the key proto file.

The primary use case is https://scalapb.github.io/docs/customizations/#package-scoped-options.  The 
scalapb compiler is picky in that it needs the package.proto to be in the srcs list of a protoc invocation, 
not in deps.

Happy to consider breaking this out into a different directive is people feel that overloading the meaning of 
proto_group is too much.

Fixes <TODO>


linzhp (changes requested):
Yea, I think this overload's `proto_group` too much. Correct me if I am wrong, I feel this `package.proto` is a special pattern used by ScalaPB. If that's the case, I don't think it's a good idea to support that in Gazelle. Instead, you can develop a Gazelle extension to append `package.proto` to every `proto_library`'s `srcs`.

pcj:
A separate extension was my initial attempt, but unfortunately this does not work.  In order for dependency resolution to proceed, the `pkg.Files` slice must be populated by the builtin `proto` extension.

Although prompted by `scalapb`, the feature is needed whenever `gazelle:proto file` mode is chosen and one requires additional "buddy" proto files to be present in each of the `srcs` list of generated `proto_library` rules.

I can break it out into a new directive like `gazelle:proto_file_group`.

linzhp:
Your custom extension can modify the rules generated by proto extension, by appending `package.proto` to its srcs attribute and append additional imports in `package.proto`.

Alternatively, you can have `# gazelle:file` so `package.proto` also gets a `proto_library` with all deps resolved. Then your customer extension can simply add `embed = [":package_proto"]` other `proto_library` targets.

pcj:
Revisiting this.  

> Your custom extension can modify the rules generated by proto extension, by appending package.proto to its srcs attribute and append additional imports in package.proto.

Unfortunately this is not the case.  While one can append additional `srcs` to the rule, adding additional strings under  `GazelleImportsKey` has no effect, because by the time the customer extension can interact with `args.OtherGen`, the imports list is already fixed (see https://github.com/bazelbuild/bazel-gazelle/blob/master/cmd/gazelle/fix-update.go#L314).

pcj:
@linzhp can we review this again?  Thanks!

linzhp:
How about this?

> Alternatively, you can have # gazelle:proto file so package.proto also gets a proto_library with all deps resolved. Then your customer extension can simply add embed = [":package_proto"] other proto_library targets.

pcj:
@linzhp I'm not sure I understand, `proto_library` does not have an `embed` attribute.  I thought you meant this:

```bazel
proto_library(
    name = "automated_trading_system_proto",
    srcs = [
        "automated_trading_system.proto",
        # "package.proto",
    ],
    embed = [":package_proto"],
    # deps = ["//thirdparty/protobuf/scalapb:scalapb_proto"],
)
```

```
no such attribute 'embed' in 'proto_library' rule
```

linzhp:
I see. Yeah, `proto_library` rule doesn't have `embed`. Can you create another scalapb extension and disable the default proto extension to handle this situation? I am still not convinced that this is common enough situation to be supported in the default proto extension.

linzhp:
By scalapb extension, I meant forking the proto extension to generate all `proto_library` targets, and add `package.proto` to all of them. You can call some functions from the proto extension though

PR #1237, opened 2022-04-13
fix(lang/proto): include imports from different targets

### LINKED ISSUE #1219: language/proto: Missing dependencies with `go_package` grouping

v0.25.0

v0.31.0

5.0, 5.1

Yes

Mac OS 12.3, x86_64-apple-darwin21.1.0

### What did you do?

Using Gazelle to generate `proto_library` and `go_proto_library` targets using `gazelle:proto package` and `gazelle:proto_group go_package` directives with Proto files that share a Proto package, but have separate Go packages. I've pushed a simple reproduction of the issue here: https://github.com/nickgooding/bazel-gazelle-1219-repro

- Generated `proto_library` and `go_proto_library` targets should be generated with appropriate `deps`
- Generated `proto_library` and `go_proto_library` targets should be buildable

### What did you see instead?

- Generated `proto_library` and `go_proto_library` targets are missing `deps` for `proto_library`/`go_proto_library` targets that belong to the same Proto package but have different Go packages
- Generated `proto_library` and `go_proto_library` targets fail to build due to the missing dependencies

---

Using `git bisect`, it appears that 6bbfc47 is the commit where these dependencies stopped being generated. A similar issue, bazelbuild/bazel-gazelle#1157 was raised regarding per-file Proto mode, but the fix for that does not appear to have resolved the issue when using `go_package` grouping.

nickgooding:
**What type of PR is this?**

Bug fix

language/proto

Fixes an issue where dependencies would be missed in the case where an
import was of a proto file in the same directory but from a different
target, such an when using `proto_group`. This would cause Gazelle to
generate `proto_library` and `go_proto_library` targets that could not
be built.

Fixes #1219

### DIFF
--- a/language/go/testdata/proto_package_mode/BUILD.want
+++ b/language/go/testdata/proto_package_mode/BUILD.want
@@ -17,7 +17,10 @@ proto_library(
         "foo1.proto",
         "foo2.proto",
     ],
-    _gazelle_imports = ["google/protobuf/any.proto"],
+    _gazelle_imports = [
+        "google/protobuf/any.proto",
+        "proto_package_mode/bar1.proto",
+    ],
     visibility = ["//visibility:public"],
 )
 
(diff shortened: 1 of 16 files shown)


achew22:
Sorry for the slow response on this one. Thank you for contributing a fix @nickgooding !

PR #1243, opened 2022-04-25
language/go: Add support for //go:build tags

### LINKED ISSUE #1099: go:build syntax is not supported

v0.23.0

v0.28.0

v4.2.0

Yes.

macOS 11.5.2 amd64.

### What did you do?

Ran `go run golang.org/x/tools/cmd/goimports@v0.1.5` and it added a `go:build tools` directive. Then I removed the old one `+build tools`.

```go
//go:build tools          <- was added
// +build tools            <- I removed this one

package build

// https://github.com/golang/go/wiki/Modules#how-can-i-track-tool-dependencies-for-a-module

import (
	_ "github.com/golang/mock/mockgen"
	_ "golang.org/x/tools/cmd/goimports"
)
```

Then I ran gazelle.

No changes to build files.

### What did you see instead?

Gazelle updated the build file for this package - added a `go_library` target.

This seem to be new behavior, added in Go 1.17. Please see `//go:build lines` under https://golang.org/doc/go1.17#go-command.

thempatel:
**What type of PR is this?**

Feature

language/go

Adds support for the new `//go:build` build tags syntax introduced in `go1.17`

Fixes #1099

**Other notes for review**


sluongng:
My advice is to split this PR into smaller ones for easier reviewing:
- 1 PR for refactoring test comparision and introduce `com_github_google_go_cmp`
- 1 PR to add support for `//go:build` tags

thempatel:
@sluongng #1244

thempatel:
@sluongng PTAL

cc: @achew22 in case you are interested and want to review

sluongng (changes requested):
@thempatel thanks so much for working on this.

I have several nits as commented. Generally I would recommend moving some of the added, mirror functionality from `go/build/constraint` to a new file. `fileinfo.go` is approaching 1000 lines status and the diff hunks were quite hard to consume.

Very glad that you were able to refactor the unit tests using the 2 new structs. Could you please also provide 1 simple test cases (which only uses `//go:build` and not the old `// +build`) in `testdata` as integration test for this? The testdata are used in `TestGenerateRules` in `generate_test.go` FYI.

language/go/fileinfo.go
```diff
@@ -31,6 +33,7 @@ import (
 	"strings"
 	"unicode"
 	"unicode/utf8"
+	_ "unsafe"
```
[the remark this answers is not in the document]
thempatel: unfortunately not, using the linker directive requires the use of `unsafe`

language/go/fileinfo.go
```diff
@@ -93,78 +96,172 @@ type fileEmbed struct {
...
+		} else if b != nil {
+			return b, nil
+		}
+	default:
+		return nil, fmt.Errorf("unknown constraint type: %T", x)
 	}
-	return true
+
+	return nil, nil
 }
```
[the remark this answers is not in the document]
thempatel: i added a comment

language/go/fileinfo.go
```diff
@@ -376,6 +473,28 @@ func goFileInfo(path, rel string) fileInfo {
 	return info
 }
 
+// matchAuto interprets text as either a +build or //go:build expression (whichever works).
+// Intended to match go/build.Context.matchAuto
+func matchAuto(tokens []string) (*buildTags, error) {
```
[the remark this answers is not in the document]
thempatel: updated wording

language/go/fileinfo.go
```diff
@@ -393,27 +512,29 @@ func saveCgo(info *fileInfo, rel string, cg *ast.CommentGroup) error {
...
+		}
 
 		// Parse options.
-		opts, err := splitQuoted(optstr)
+		opts, err := splitQuoted(argstr)
 		if err != nil {
 			return fmt.Errorf("%s: invalid #cgo line: %s", info.path, orig)
 		}
-		var ok bool
+
```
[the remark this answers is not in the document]
thempatel: it is already defined above

language/go/fileinfo.go
```diff
@@ -393,27 +512,29 @@ func saveCgo(info *fileInfo, rel string, cg *ast.CommentGroup) error {
 		}
 
 		// Split at colon.
-		line = strings.TrimSpace(line[4:])
-		i := strings.Index(line, ":")
-		if i < 0 {
+		line, argstr, ok := strings.Cut(strings.TrimSpace(line[4:]), ":")
```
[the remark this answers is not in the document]
thempatel: fair

language/go/fileinfo.go
```diff
@@ -564,80 +688,91 @@ func safeCgoName(s string, spaces bool) bool {
...
+				osSpecific = true
+			}
+			_, archOk := rule.KnownArchSet[tag]
+			if archOk {
+				archSpecific = true
 			}
 		}
 	}
+	checkTags(info.tags.tags())
+	checkTags(cgoTags.tags())
```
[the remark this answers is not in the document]
thempatel: yes

language/go/fileinfo.go
```diff
@@ -827,3 +974,14 @@ func parseGoEmbed(args string, pos token.Position) ([]fileEmbed, error) {
...
+// In order to correctly capture the subtleties of build tag placement
+// and to automatically stay up-to-date with the parsing semantics and
+// syntax, we link to the stdlib version of header parsing.
+//go:linkname parseFileHeader go/build.parseFileHeader
+func parseFileHeader(_ []byte) ([]byte, []byte, bool, error)
+
+// readComments is like io.ReadAll, except that it only reads the leading
+// block of comments in the file.
+//go:linkname readComments go/build.readComments
+func readComments(_ io.Reader) ([]byte, error)
```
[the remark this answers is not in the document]
thempatel: I am ok with your proposal

language/go/fileinfo_test.go
```diff
@@ -306,37 +320,66 @@ package main`,
...
+				t.Fatal(err)
+			}
+			path := f.Name()
+			defer os.Remove(path)
+			if err = f.Close(); err != nil {
+				t.Fatal(err)
+			}
+			if err = ioutil.WriteFile(path, []byte(tc.source), 0o600); err != nil {
+				t.Fatal(err)
+			}
```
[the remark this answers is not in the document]
thempatel: agree

language/go/generate_test.go
```diff
@@ -97,9 +98,8 @@ func TestGenerateRules(t *testing.T) {
 			}
 			want := string(wantBytes)
 			want = strings.ReplaceAll(want, "\r\n", "\n")
-
-			if got != want {
-				t.Errorf("GenerateRules %q: got:\n%s\nwant:\n%s", rel, got, want)
+			if diff := cmp.Diff(want, got); diff != "" {
+				t.Errorf("(-want, +got): %s", diff)
 			}
```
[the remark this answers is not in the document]
thempatel: i tried

thempatel:
@sluongng PTAL

language/go/fileinfo.go
```diff
@@ -393,27 +319,29 @@ func saveCgo(info *fileInfo, rel string, cg *ast.CommentGroup) error {
...
-		i := strings.Index(line, ":")
-		if i < 0 {
+		line, argstr, ok := strings.Cut(strings.TrimSpace(line[4:]), ":")
+		if !ok {
 			return fmt.Errorf("%s: invalid #cgo line: %s", info.path, orig)
 		}
-		line, optstr := strings.TrimSpace(line[:i]), strings.TrimSpace(line[i+1:])
 
-		// Parse tags and verb.
+		// Parse GOOS/GOARCH stuff.
```
sluongng, line 327: nit: original comment seems to be more correct?

language/go/fileinfo.go
```diff
@@ -665,22 +543,34 @@ func matchesOS(os, value string) bool {
...
+
+		} else if _, ok := rule.KnownArchSet[tag]; ok {
+			if arch == "" {
+				return false
+			}
+			return arch == tag
+
+		} else {
+			return goConf.genericTags[tag]
 		}
```
sluongng, line 570: nit: since we are returning early in each if block, we can probably rewrite this without using else

```
if foo {
return
}

if bar {
return
}

return
```

language/go/fileinfo_test.go
```diff
@@ -57,7 +59,7 @@ func TestOtherFileInfo(t *testing.T) {
 
 			// Only check that we can extract tags. Everything else is covered
 			// by other tests.
-			if diff := cmp.Diff(tc.wantTags, got.tags, fileInfoCmpOption); diff != "" {
+			if diff := cmp.Diff(tc.wantTag, got.tags, fileInfoCmpOption); diff != "" {
```
sluongng, line 62: small nit: renaming this from plural to singular now made the comparison singular vs plural, hence a bit harder to read.

sluongng (approved):
I only have small nits left.

I think this version is a lot cleaner and easier to digest. Breaking the diff into smaller files was definitely the right way to go.

`fileinfo.go` diff is still quite noisy to digest through. But the test passing is additional assurance.

There are small scope creeps here and there with the usage of cmp.Diff but IMO the end result is strictly better.

Pinging @achew22 @linzhp for additional reviews

thempatel:
@sluongng PTAL

thempatel:
@sluongng @achew22 @linzhp ping, hoping to get this in -- we have work blocked on this particular body of code making it's way to mainline

linzhp (commented):
When we have

```
-- collector_linux.go --
//go:build linux && amd64

package cpufreqcounters

import (
	"uber.com/inventory"
	"uber.com/file"
)

-- collector_other.go --
//go:build !(linux && amd64)

package cpufreqcounters

import (
	"uber.com/file"
)
```

This pull request would lead to huge list of `deps` in `go_library` like:

```
    deps = select({
        "@io_bazel_rules_go//go/platform:aix_ppc64": [
            "//uber.com/file",
        ],
        "@io_bazel_rules_go//go/platform:android_386": [
            "//uber.com/file",
        ],
        "@io_bazel_rules_go//go/platform:android_amd64": [
            "//uber.com/file",
            "//uber.com/inventory",
        ],
        "@io_bazel_rules_go//go/platform:android_arm": [
            "//uber.com/file",
        ],
        "@io_bazel_rules_go//go/platform:android_arm64": [
            "//uber.com/file",
        ],
        "@io_bazel_rules_go//go/platform:darwin_386": [
            "//uber.com/file",
        ],
        "@io_bazel_rules_go//go/platform:darwin_amd64": [
            "//uber.com/file",
        ],
        "@io_bazel_rules_go//go/platform:darwin_arm": [
            "//uber.com/file",
        ],
        "@io_bazel_rules_go//go/platform:darwin_arm64": [
            "//uber.com/file",
        ],
        "@io_bazel_rules_go//go/platform:dragonfly_amd64": [
            "//uber.com/file",
        ],
        "@io_bazel_rules_go//go/platform:freebsd_386": [
            "//uber.com/file",
        ],
        "@io_bazel_rules_go//go/platform:freebsd_amd64": [
            "//uber.com/file",
        ],
        "@io_bazel_rules_go//go/platform:freebsd_arm": [
            "//uber.com/file",
        ],
        "@io_bazel_rules_go//go/platform:freebsd_arm64": [
            "//uber.com/file",
        ],
        "@io_bazel_rules_go//go/platform:illumos_amd64": [
            "//uber.com/file",
        ],
        "@io_bazel_rules_go//go/platform:linux_amd64": [
            "//uber.com/file",
            "//uber.com/inventory",
        ],
        "//conditions:default": [],
```

Can you simplify it to something like:

```
    deps = [
            "//uber.com/file",
    ] + select({
        "@io_bazel_rules_go//go/platform:android": [
            "//uber.com/inventory",
        ],
        "@io_bazel_rules_go//go/platform:linux": [
            "//uber.com/inventory",
        ],
        "//conditions:default": [],
    }),

```

or 
```
    deps = select({
        "@io_bazel_rules_go//go/platform:android": [
            "//uber.com/file",
            "//uber.com/inventory",
        ],
        "@io_bazel_rules_go//go/platform:linux": [
            "//uber.com/file",
            "//uber.com/inventory",
        ],
        "//conditions:default":  [
            "//uber.com/file",
        ],
    }),

```

thempatel:
@linzhp I'm somewhat confused by that, does that mean that the existing build tags logic is broken? How does it work today? Is the long list a problem, these are generated files so I would lean toward not introducing scope creep to make the file prettier unless the generated code is causing bugs/issues

linzhp:
BUILD.bazel files are also supposed to be human readable and maintainable so people can customize their rules on top of what gazelle generates. So I wonder whether it's possible to avoid generating such overly verbose list of `select` containing many platforms that people don't care most of the times.

thempatel:
@linzhp I'm having a hard time reproducing, can you help?

Here's what I did

1. I setup a new repository

```
test_build/
  lib/
    collector_linux.go
    collector_other.go
```

collector_linux.go
```go
//go:build linux && amd64
// +build linux,amd64

package lib
```

collector_other.go
```go
//go:build !linux || !amd64
// +build !linux !amd64

package lib
```

WORKSPACE
```starlark
workspace(name = "test_build")

load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")

http_archive(
    name = "io_bazel_rules_go",
    sha256 = "f2dcd210c7095febe54b804bb1cd3a58fe8435a909db2ec04e31542631cf715c",
    urls = [
        "https://mirror.bazel.build/github.com/bazelbuild/rules_go/releases/download/v0.31.0/rules_go-v0.31.0.zip",
        "https://github.com/bazelbuild/rules_go/releases/download/v0.31.0/rules_go-v0.31.0.zip",
    ],
)

git_repository(
    name="bazel_gazelle",
    commit="c089265a8e6cacb02c5729179a98d3c47e8398c8",
    remote="https://github.com/thempatel/bazel-gazelle.git"
)

load("@io_bazel_rules_go//go:deps.bzl", "go_register_toolchains", "go_rules_dependencies")
load("//:deps.bzl", "go_dependencies")

# gazelle:repository_macro deps.bzl%go_dependencies
go_dependencies()

go_rules_dependencies()

go_register_toolchains(version = "1.18")

gazelle_dependencies()
```

Root BUILD
```starlark
load("@bazel_gazelle//:def.bzl", "gazelle")

# gazelle:prefix github.com/thempatel/test_build
gazelle(name = "gazelle")

gazelle(
    name = "gazelle-update-repos",
    args = [
        "-from_file=go.mod",
        "-prune",
    ],
    command = "update-repos",
)
```

Root go.mod
```go.mod
module github.com/thempatel/test_build

go 1.18
```

Here is the the generated BUILD in `test_build/lib`
```starlark

go_library(
    name = "lib",
    srcs = [
        "collector_linux.go",
        "collector_other.go",
    ],
    importpath = "github.com/thempatel/test_build/lib",
)
```

If I replace the above `git_repository` with the `v0.25.0` bazel-gazelle release, I get the same thing.
```starlark
http_archive(
    name = "bazel_gazelle",
    sha256 = "5982e5463f171da99e3bdaeff8c0f48283a7a5f396ec5282910b9e8a49c0dd7e",
    urls = [
        "https://mirror.bazel.build/github.com/bazelbuild/bazel-gazelle/releases/download/v0.25.0/bazel-gazelle-v0.25.0.tar.gz",
        "https://github.com/bazelbuild/bazel-gazelle/releases/download/v0.25.0/bazel-gazelle-v0.25.0.tar.gz",
    ],
)
```

```starlark

go_library(
    name = "lib",
    srcs = [
        "collector_linux.go",
        "collector_other.go",
    ],
    importpath = "github.com/thempatel/test_build/lib",
)
```

linzhp:
I created a workspace like this:

```
-- WORKSPACE --
workspace(name = "test_build")

load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")

http_archive(
    name = "io_bazel_rules_go",
    sha256 = "685052b498b6ddfe562ca7a97736741d87916fe536623afb7da2824c0211c369",
    urls = [
        "https://mirror.bazel.build/github.com/bazelbuild/rules_go/releases/download/v0.33.0/rules_go-v0.33.0.zip",
        "https://github.com/bazelbuild/rules_go/releases/download/v0.33.0/rules_go-v0.33.0.zip",
    ],
)

http_archive(
    name = "bazel_gazelle",
    sha256 = "d88a68381d5a91f07a5a03d9470ee8d7ef3ff0d841aad572ffb524b858428fa3",
    strip_prefix = "bazel-gazelle-milan-go-build-syntax",
    urls = [
        "https://github.com/thempatel/bazel-gazelle/archive/refs/heads/milan/go-build-syntax.zip",
    ],
)

load("@io_bazel_rules_go//go:deps.bzl", "go_register_toolchains", "go_rules_dependencies")
load("@bazel_gazelle//:deps.bzl", "gazelle_dependencies", "go_repository")
load("//:deps.bzl", "go_dependencies")

# gazelle:repository_macro deps.bzl%go_dependencies
go_dependencies()

go_rules_dependencies()

go_register_toolchains(version = "1.18")

gazelle_dependencies()
-- BUILD.bazel --
load("@bazel_gazelle//:def.bzl", "gazelle")

# gazelle:prefix github.com/thempatel/test_build
gazelle(name = "gazelle")

gazelle(
    name = "gazelle-update-repos",
    args = [
        "-from_file=go.mod",
        "-prune",
    ],
    command = "update-repos",
)
-- deps.bzl --
load("@bazel_gazelle//:deps.bzl", "go_repository")

def go_dependencies():
    go_repository(
        name = "org_golang_x_crypto",
        importpath = "golang.org/x/crypto",
        sum = "h1:ObdrDkeb4kJdCP557AjRjq69pTHfNouLtWZG7j9rPN8=",
        version = "v0.0.0-20191011191535-87dc89f01550",
    )
    go_repository(
        name = "org_golang_x_mod",
        importpath = "golang.org/x/mod",
        sum = "h1:OJxoQ/rynoF0dcCdI7cLPktw/hR2cueqYfjm43oqK38=",
        version = "v0.5.1",
    )
    go_repository(
        name = "org_golang_x_tools",
        importpath = "golang.org/x/tools",
        sum = "h1:aZzprAO9/8oim3qStq3wc1Xuxx4QmAGriC4VU4ojemQ=",
        version = "v0.0.0-20191119224855-298f0cb1881e",
    )

    go_repository(
        name = "org_golang_x_xerrors",
        build_extra_args = ["-go_naming_convention_external=go_default_library"],
        build_file_generation = "on",
        build_file_proto_mode = "disable",
        importpath = "golang.org/x/xerrors",
        sum = "h1:uF6paiQQebLeSXkrTqHqz0MXhXXS1KgF41eUdBNvxK0=",
        version = "v0.0.0-20220609144429-65e65417b02f",
    )
-- go.mod --
module github.com/thempatel/test_build

go 1.18

require (
	golang.org/x/mod v0.5.1
	golang.org/x/xerrors v0.0.0-20220609144429-65e65417b02f
)
-- go.sum --
golang.org/x/mod v0.5.1 h1:OJxoQ/rynoF0dcCdI7cLPktw/hR2cueqYfjm43oqK38=
golang.org/x/mod v0.5.1/go.mod h1:5OXOZSfqPIIbmVBIIKWRFfZjPR0E5r58TLhUjH0a2Ro=
golang.org/x/xerrors v0.0.0-20220609144429-65e65417b02f h1:uF6paiQQebLeSXkrTqHqz0MXhXXS1KgF41eUdBNvxK0=
golang.org/x/xerrors v0.0.0-20220609144429-65e65417b02f/go.mod h1:K8+ghG5WaK9qNqU5K3HdILfMLy1f3aNYFI/wnl100a8=
-- lib/collector_linux.go --
//go:build linux && amd64

package lib

import _ "golang.org/x/xerrors"
-- lib/collector_other.go --
//go:build !(linux && amd64)

package lib

import _ "golang.org/x/mod/modfile"
```

After running `bazel run //:gazelle`, I got large `lib/BUILD.bazel`

thempatel:
@linzhp I don't know what to tell you, I created https://github.com/thempatel/test_build just as you have outlined and am not seeing the large `lib/BUILD.bazel` you're describing

thempatel:
bazel info results:

```
bazel-bin: /private/var/tmp/_bazel_milan/30f51008ffea5c584261e337d59e37de/execroot/test_build/bazel-out/darwin-fastbuild/bin
bazel-genfiles: /private/var/tmp/_bazel_milan/30f51008ffea5c584261e337d59e37de/execroot/test_build/bazel-out/darwin-fastbuild/bin
bazel-testlogs: /private/var/tmp/_bazel_milan/30f51008ffea5c584261e337d59e37de/execroot/test_build/bazel-out/darwin-fastbuild/testlogs
character-encoding: file.encoding = ISO-8859-1, defaultCharset = ISO-8859-1
command_log: /private/var/tmp/_bazel_milan/30f51008ffea5c584261e337d59e37de/command.log
committed-heap-size: 230MB
execution_root: /private/var/tmp/_bazel_milan/30f51008ffea5c584261e337d59e37de/execroot/test_build
gc-count: 11
gc-time: 136ms
install_base: /var/tmp/_bazel_milan/install/d97d79f90938dfeec6f13f503ef09df1
java-home: /usr/local/Cellar/openjdk@11/11.0.15/libexec/openjdk.jdk/Contents/Home
java-runtime: OpenJDK Runtime Environment (build 11.0.15+0) by Homebrew
java-vm: OpenJDK 64-Bit Server VM (build 11.0.15+0, mixed mode) by Homebrew
max-heap-size: 8589MB
output_base: /private/var/tmp/_bazel_milan/30f51008ffea5c584261e337d59e37de
output_path: /private/var/tmp/_bazel_milan/30f51008ffea5c584261e337d59e37de/execroot/test_build/bazel-out
package_path: %workspace%
release: release 5.2.0-homebrew
repository_cache: /var/tmp/_bazel_milan/cache/repos/v1
server_log: /private/var/tmp/_bazel_milan/30f51008ffea5c584261e337d59e37de/java.log.milans-mac-mini.milan.log.java.20220612-133020.16473
server_pid: 16473
used-heap-size: 113MB
workspace: /Users/milan/src/test_build
```

thempatel:
@linzhp ok -- i fixed the two go files and now I do get the large output

thempatel:
@linzhp I get the same large result whether I use `go:build` or the `+build` syntax and even when I use the latest release:

https://github.com/thempatel/test_build/blob/3413cdcfa112cbeab3960a0b6685e529bd3d524d/WORKSPACE#L14-L21

Can you comment on whether you think this is a result of the proposed change here, or wether this is an existing problem with bazel-gazelle ?

linzhp:
For existing files with `go:build` and **without** `+build`, Gazelle doesn't process `go:build`, but it recognizes the `*_linux.go` and `*_other.go` suffix, so it would generate a `go_library` target like:

```starlark
go_library(
    name = "lib",
    srcs = [
        "collector_linux.go",
        "collector_other.go",
    ],
    importpath = "github.com/thempatel/test_build/lib",
    deps = [
        "@org_golang_x_mod//modfile",
    ] + select({
        "@io_bazel_rules_go//go/platform:android": [
            "@org_golang_x_xerrors//:xerrors",
        ],
        "@io_bazel_rules_go//go/platform:linux": [
            "@org_golang_x_xerrors//:xerrors",
        ],
        "//conditions:default": [],
    }),
)
```

For these packages, generating a huge `go_library` for them seems a degradation to me. 

One possible fix is to group the conditions of `select` by their values, and assign the group with largest number of conditions sharing the same value to `//conditions:default`.

thempatel:
> For these packages, generating a huge go_library for them seems a degradation to me.

Respectfully, what you view a degradation, I view as fixing a **major hole** in gazelle generated `BUILD` files. That build file is simply *not correct*. Moreover, if you actually added the correct `+build` tags back into that file, _you would get the same thing_. 

What about files that don't have an os/arch file suffix? Gazelle, wouldn't event produce the correct `BUILD` file in that scenario, despite the `go:build` tags specifying the OS and arch.

I think there are two paths forward,

1. You can file an issue with `bazel-gazelle` that BUILD files generated from files with build constraints are too big and work with the other maintainers to fix
2. You are more than welcome to push commits to this branch.

In the interesting of preventing this PR from becoming bloated, I suggest you take route 1 but I won't be offended with route 2.

thempatel:
👋🏽 @linzhp any luck implementing your suggestion?

linzhp:
A few targets in Uber failed to build after applying your PR. I am still trying to figure out what was going on

linzhp:
I am able to create a minimal reproduction of the failure. You can try the following command:
```
bazel build --platforms=@io_bazel_rules_go//go/toolchain:darwin_arm64 @com_github_rcrowley_go_metrics//:all
```
in the follow workpace:

```
-- WORKSPACE --
workspace(name = "test_build")

load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")

http_archive(
    name = "io_bazel_rules_go",
    sha256 = "685052b498b6ddfe562ca7a97736741d87916fe536623afb7da2824c0211c369",
    urls = [
        "https://mirror.bazel.build/github.com/bazelbuild/rules_go/releases/download/v0.33.0/rules_go-v0.33.0.zip",
        "https://github.com/bazelbuild/rules_go/releases/download/v0.33.0/rules_go-v0.33.0.zip",
    ],
)

http_archive(
    name = "bazel_gazelle",
    sha256 = "d88a68381d5a91f07a5a03d9470ee8d7ef3ff0d841aad572ffb524b858428fa3",
    strip_prefix = "bazel-gazelle-milan-go-build-syntax",
    urls = [
        "https://github.com/thempatel/bazel-gazelle/archive/refs/heads/milan/go-build-syntax.zip",
    ],
)

load("@io_bazel_rules_go//go:deps.bzl", "go_register_toolchains", "go_rules_dependencies")
load("@bazel_gazelle//:deps.bzl", "gazelle_dependencies", "go_repository")
load("//:deps.bzl", "go_dependencies")

# gazelle:repository_macro deps.bzl%go_dependencies
go_dependencies()

go_rules_dependencies()

go_register_toolchains(version = "1.18")

gazelle_dependencies()
-- BUILD.bazel --
load("@bazel_gazelle//:def.bzl", "gazelle")

# gazelle:prefix github.com/thempatel/test_build
gazelle(name = "gazelle")

gazelle(
    name = "gazelle-update-repos",
    args = [
        "-from_file=go.mod",
        "-prune",
    ],
    command = "update-repos",
)
-- deps.bzl --
load("@bazel_gazelle//:deps.bzl", "go_repository")

def go_dependencies():
    go_repository(
        name = "org_golang_x_xerrors",
        build_extra_args = ["-go_naming_convention_external=go_default_library"],
        build_file_generation = "on",
        build_file_proto_mode = "disable",
        importpath = "golang.org/x/xerrors",
        sum = "h1:uF6paiQQebLeSXkrTqHqz0MXhXXS1KgF41eUdBNvxK0=",
        version = "v0.0.0-20220609144429-65e65417b02f",
    )
    go_repository(
        name = "com_github_rcrowley_go_metrics",
        build_extra_args = ["-go_naming_convention_external=go_default_library"],
        build_file_generation = "on",
        build_file_proto_mode = "disable",
        importpath = "github.com/rcrowley/go-metrics",
        sum = "h1:MkV+77GLUNo5oJ0jf870itWm3D0Sjh7+Za9gazKc5LQ=",
        version = "v0.0.0-20200313005456-10cdbea86bc0",
    )
```

The cause is there is a `runtime_cgo.go` and a `runtime_no_cgo.go`, but the latter is not picked up  by the generated rule. So when cgo is disabled, the package fails to compile

thempatel:
@linzhp @sluongng 

This issue is pretty subtle and I had to build a workaround for this behavior to even get close to maintaining the same behavior. The issue comes down to the fact that ignored tags always evaluate to true, regardless of whether they are negated or not.

For example

```go
// +build !cgo appengine
```
and 
```go
// +build cgo appengine
```

both evaluate to true because the `cgo` tag is entirely ignored. The issue with `@com_github_rcrowley_go_metrics` boils down to this subtlety: the file that is being omitted from the library rule as a result of this change has a tag that looks like the above two, so it is retained. 

With this change, I tried to bring gazelle onto an evaluation mechanism that is more maintainable and more in inline with the way that the compiler would evaluate the build tags i.e. an actual boolean expression evaluation process, rather than iterating slices of string slices and performing ad-hoc evaluations. For example

```go
// +build !cgo appengine
```

turns into
```go
&constraints.OrExpr{
 X: &constraints.NotExpr{
  X: &constraints.TagExpr{
    Tag: "cgo",
  },
 },
 Y: &constraints.TagExpr{
  Tag: "appengine",
 },
}
```

In order to maintain as close of a behavior to the ignored tags behavior above, this change will simply drop the ignored tags from the expression, yielding

```go
&constraints.TagExpr{
  Tag: "appengine",
 }
```

The problem is that when you go to evaluate this to determine whether to keep the file, it evaluates to `false`, because it's not a defined tag. 

I'm not sure what the best path forward looks like. On the one hand the existing behavior doesn't really make any sense, ignored tags still participate in boolean expressions used to evaluate a file, but don't actually adhere to any sensible boolean evaluation criteria i.e. `!foo == foo == true`, which is impossible. On the other hand, like we've seen here, a lot of repo's depend on this behavior remaining the same, otherwise builds will break but the proposed method is largely incompatible with the existing behavior. I can likely come up with some hacks to get around this, but in the interest of saving time, I'd rather hear proposals from you guys.

sluongng:
Quick reaction: we should stick with w/e the behaviors in `go build . ` today and avoid introducing our own logic as much as possible. If there are issues in upstream go language then we better report and fix it there. 

I may not have time to review this in the next few days, but I will try to do some research meanwhile.

Please add a test to the PR to demonstrate the current behavior. It would help with the discussion if other maintainers/contributors want to join in the review. 🤔

linzhp:
Why cgo is an ignored tag? It seems to be a valid [build constraint](https://pkg.go.dev/cmd/go#hdr-Build_constraints). So it should be possible to produce a `srcs` attribute like:

```
srcs = select({
        "@io_bazel_rules_go//go/platform:darwin_amd64": [
            "runtime_no_cgo.go",
        ],
        "@io_bazel_rules_go//go/toolchain:darwin_amd64_cgo": [
            "runtime_cgo.go",
        ],
})
```

thempatel:
> Quick reaction: we should stick with w/e the behaviors in `go build . ` today and avoid introducing our own logic as much as possible. If there are issues in upstream go language then we better report and fix it there.
> 
> I may not have time to review this in the next few days, but I will try to do some research meanwhile.
> 
> Please add a test to the PR to demonstrate the current behavior. It would help with the discussion if other maintainers/contributors want to join in the review. 🤔

Added tests to demonstrate the existing behavior, the tests should fail with the new proposed changes. Sounds good, I will continue to investigate as I have time.

> Why cgo is an ignored tag? It seems to be a valid [build constraint](https://pkg.go.dev/cmd/go#hdr-Build_constraints). So it should be possible to produce a srcs attribute like:

Honestly, I'm not sure. This predates my involvement with `gazelle`. One hypothesis, is that `ignoredTags` are really intended to be `deferToGoCompiler` tags, i.e. there are a set of tags that are evaluated at compile time when all the sources for a library rule are compiled together, and so they are kept by gazelle.

linzhp:
It seems that Gazelle include all source files in `srcs` regardless of their OS/arch/cgo (essentially deferring to Go compiler), and only use the tags to select `deps`?

thempatel:
PTAL

I've added fixes that I think are reasonable that maintains gazelle's existing behavior. The updated build tags test cases hopefully show how gazelle's interpretation of build tags is fairly complex when translated to a boolean expression. 

@linzhp would you mind checking the latest changes in Uber's repo?

language/go/build_tags_test.go
```diff
@@ -0,0 +1,72 @@
...
+		},
+		{
+			desc:  "complex 1",
+			input: mustParseBuildTag(t, "!(cgo && (go1.8 || go1.9) || race || msan)"),
+			want:  mustParseBuildTag(t, "(cgo || (go1.8 && go1.9)) && race && msan"),
+		},
+		{
+			desc:  "complex 2",
+			input: mustParseBuildTag(t, "!(cgo && (go1.8 || go1.9 && (race && foobar)))"),
+			want:  mustParseBuildTag(t, "cgo || go1.8 && (go1.9 || (race || !foobar))"),
```
thempatel, line 54: ignored tags always evaluate to true (e.g. `!cgo == cgo == true`. This is the existing gazelle behavior. In order to replicate, we must first push negations down to the leaves of a boolean expression and then remove the negation for ignored tags. You can see here that because of this behavior, for test case `complex 2`, the user provided tag of `foobar` is never evaluated since the always true `cgo`, short circuits the boolean expression. I will leave it to y'all on how to feel about it.

language/go/testdata/platforms/tags_dual_no_cgo.go
```diff
@@ -0,0 +1,3 @@
+// +build !cgo appengine
```
thempatel, line 1: `!cgo OR appengine == true` since `!cgo == true`

language/go/testdata/platforms/tags_dual_cgo.go
```diff
@@ -0,0 +1,4 @@
+// +build cgo
```
thempatel, line 1: `cgo AND !appengine == true` since `cgo == true` and `appengine` is not defined

language/go/build_tags_test.go
thempatel: Unfortunately no. I took that approach first, but the new test cases I added (which caused the failures you pointed out) require that the ignored tags participate in the Boolean expression evaluation

linzhp: @sluongng What do you think?

language/go/build_tags.go
```diff
@@ -0,0 +1,381 @@
...
+
+func (b *buildTags) empty() bool {
+	if b == nil {
+		return true
+	}
+
+	return len(b.rawTags) == 0
+}
+
+// dropNegationForIgnoredTags wil drop negations for any concrete tags that should be ignored.
```
sluongng, line 147: nit:

```suggestion
// dropNegationForIgnoredTags drops negations for any concrete tags that should be ignored.
```

language/go/build_tags_test.go
sluongng: Gona put my thoughts in a separate review.

sluongng (commented):
I think the only concern left is a very uncommon edge case: users would have to be using `cgo` and is converting a code base from normal `go build` to `rules_go` for the first time to get affected.

Even when that happen, we provide a work around in gazelle by using `# keep` to fall back to manual BUILD file configuration.

Once users have converted their code base to Bazel / rules_go, subsequent CGO usages would/should mostly rely on Bazel's BUILD files instead of `go build` constraint. 

There might be some needs to keep `go build` constraint for IDE integration? But again, I think we are hitting above P95 of use cases then. We can always iterate on it in the future.

> Perfection is the enemy of progress 

The current state of this PR seems to be a net win, especially given how well tested it is and it's preserving all existing behaviors. Future imporvements toward this area should always be coupled with more tests to ensure backward compatibility.

---

The question here is mainly: how much interops between rules_go and `go build` do we want to support? at what cost?

Imo some of the API in `build_tags.go` should be provided and customizable from Go SDK and Gazelle should only call into them instead of currrent approach of `fork/vendor` the implementation. But it might take more time to convince Go team to implement that API.

---

Overall LGTM but there are some minor nits.

language/go/build_tags.go
thempatel: fixed

language/go/testdata/platforms/tags_dual_cgo.go
thempatel: done

language/go/build_tags.go
```diff
@@ -0,0 +1,381 @@
+/* Copyright 2022 The Bazel Authors. All rights reserved.
```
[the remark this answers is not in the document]
thempatel: I have updated some files to use constraints instead

thempatel:
> I think the only concern left is a very uncommon edge case: users would have to be using cgo and is converting a code base from normal go build to rules_go for the first time to get affected.

Could you help me understand this edge case (for my own edification), my understanding is if a new repo becomes managed by `gazelle` and they are using cgo, then the `BUILD` files should be generated correctly. 

 > Even when that happen, we provide a work around in gazelle by using # keep to fall back to manual BUILD file configuration.

this makes sense to me!

> The question here is mainly: how much interops between rules_go and go build do we want to support? at what cost?

I am not sure, to me it seems like it would be situational, maybe some existing examples would help bring clarity

> Imo some of the API in build_tags.go should be provided and customizable from Go SDK and Gazelle should only call into them instead of currrent approach of fork/vendor the implementation. But it might take more time to convince Go team to implement that API.

In an ideal world I agree, but I think it is unlikely we'll be able to convince the go team within any reasonable timeframe and we have many other things to do.

@linzhp PTAL

linzhp:
Sorry, I am getting swamped by the perf season. Would you mind rebasing on master so it's easier for me to test?

thempatel:
@linzhp rebased and pushed, PTAL

linzhp:
Some tests are failing

thempatel:
@linzhp pushed some old code, apologies. Things look green now

linzhp (approved):
All tests in Uber passed. Thanks for the contribution.

thempatel:
thanks for the collaboration everyone!

pcj:
The `strings.Cut` part causing upgrade pain.

linzhp:
Can you come up with a PR to fix it?

thempatel:
I'm sorry to hear that, I will point out that `bazel-gazelle`'s minimum supported version is [`go1.18`](https://github.com/bazelbuild/bazel-gazelle/blob/984e821189ddd636fc097bca59bccc5217ee24c5/go.mod#L3)

and given that [`strings.Cut`](https://pkg.go.dev/strings#Cut) was added in `go1.18`, it felt like a safe change to make. If possible, please consider upgrading your version of Go.

pcj:
Yeah, it happens.   Just one of those many cases where you think upgrading something will be a small task and it morphs into a bigger one.

PR #1245, opened 2022-04-27
Respect .bazelignore

### LINKED ISSUE #318: Respect .bazelignore files

Whoaa512:
**What type of PR is this?**

> Feature

> walk/config

The code for this re-purposes the original PR #1022 but makes a few
modifications for performance and correctness.

First we only load the .bazelignore once when `rel == ""` since that means
we're in the root; the previous implementation seemed to do the same load
for every directory visited. Next we only match files relative to repo
root to have parity with how Bazel itself parses the ignore file.

Fixes #318

**Other notes for review**


walk/config.go
```diff
@@ -104,6 +112,46 @@ func (cr *Configurer) Configure(c *config.Config, rel string, f *rule.File) {
...
+
+	scanner := bufio.NewScanner(file)
+	for scanner.Scan() {
+		ignore := strings.TrimSpace(scanner.Text())
+		if ignore == "" || string(ignore[0]) == "#" {
+			continue
+		}
+		// Bazel ignore paths are always relative to repo root.
+		// Glob patterns are not supported.
+		pat := path.Join(rel, ignore)
```
sluongng, line 141: I could be wrong here but I think .bazelignore pattern is counting from repoRoot and not from rel?

https://docs.bazel.build/versions/main/guide.html#bazelignore

Also if we have established `rel` to be effectively empty in begining of the func, do we still need this?

walk/config.go
```diff
@@ -104,6 +112,46 @@ func (cr *Configurer) Configure(c *config.Config, rel string, f *rule.File) {
...
+		if isGlobPattern(pat) {
+			log.Printf("the .bazelignore exclusion pattern must not be a glob %s", pat)
+			continue
+		}
```
sluongng, line 145: nit: we only need to check for glob pattern in `ignore` and skip it early?

walk/config.go
```diff
@@ -104,6 +112,46 @@ func (cr *Configurer) Configure(c *config.Config, rel string, f *rule.File) {
...
+		wc.excludes = append(wc.excludes, path.Join(rel, ignore))
+	}
+	return nil
+}
+
+func isGlobPattern(pattern string) bool {
+	return strings.ContainsAny(pattern, "*?[")
+}
```
sluongng, line 153: nit: should not need a dedicated func for this

walk/config.go
```diff
@@ -104,6 +112,46 @@ func (cr *Configurer) Configure(c *config.Config, rel string, f *rule.File) {
...
+	file, err := os.Open(ignorePath)
+	if os.IsNotExist(err) {
+		return nil
+	}
+	if err != nil {
+		return fmt.Errorf(".bazelignore exists but couldn't be read: %v", err)
+	}
+	defer file.Close()
+
+	wc.excludes = append(wc.excludes, path.Join(rel, ".bazelignore"))
```
sluongng, line 131: why do we need to exclude the `.bazelignore` file specifically? Could we make do without this? Seems like a very minor details for gazelle to specifically ignore 1 single file given that `rel` is effectively empty.

walk/config.go
```diff
@@ -104,6 +112,46 @@ func (cr *Configurer) Configure(c *config.Config, rel string, f *rule.File) {
 	c.Exts[walkName] = wcCopy
 }
 
+func (c *Configurer) loadBazelIgnore(repoRoot, rel string, wc *walkConfig) error {
+	// Only load .bazelignore at the root, and only once
+	if rel != "" {
+		return nil
+	}
```
sluongng, line 119: Ah so rel is effectively empty for the rest of the func.

Worth to note this detail as context for the rest

walk/walk_test.go
```diff
@@ -192,13 +192,24 @@ func TestExcludeFiles(t *testing.T) {
...
+		},
+		{
+			Path: ".bazelignore",
+			Content: `
+dir
+dir2/a/b
+dir3/
+
+# Random comment followed by a line
+a.file
```
sluongng, line 205: could you please provide a test case for glob ignoring logic as well?

sluongng (commented):
I reviewed this without knowing that this was a recreation of a previously merged+reverted PR so some questions might have been answered in the past unknowningly to me.

Overall this looks good to me but need some more work to take it over the finish line.

walk/config.go
Whoaa512: Agreed, this logic actually only works when `rel` is empty so we can probably just use the `ignore` value directly. I think I copied this from the original

walk/config.go
Whoaa512: Probably works without it; another bit copied from the original

walk/config.go
Whoaa512: agreed but going to keep the log as it's a better DX

Whoaa512:
@sluongng Updated to resolve comments and add the glob ignore test case

sluongng:
Thanks for the update. Latest version looks very clean to me. 👍 

My current concern of this PR is that I am not clear on whether `.bazelignore` would still be loaded if gazelle was invoked in a working directory that is different from root of the workspace/repository.

I.e. given a project like this
```
> tree project
.
├── BUILD.bazel
├── README.md
├── WORKSPACE
├── .bazelignore
├── foo
│   ├── BUILD.bazel
│   └── a.go
└── bar
    ├── BUILD.bazel
    └── b.go
```

Would `.bazelignore` still get loaded when somebody were to run `cd foo && gazelle`?

Imo try explore using sync.Once to load `.bazelignore` regardless of the current value of `rel` would be clearer and help avoid this situation regardless? WDYT?

sluongng:
Thanks for the update. Latest version looks very clean to me. 👍 

My current concern of this PR is that I am not clear on whether `.bazelignore` would still be loaded if gazelle was invoked in a working directory that is different from root of the workspace/repository.

I.e. given a project like this
```
> tree project
.
├── BUILD.bazel
├── README.md
├── WORKSPACE
├── .bazelignore
├── foo
│   ├── BUILD.bazel
│   └── a.go
└── bar
    ├── BUILD.bazel
    └── b.go
```

Would `.bazelignore` still get loaded when somebody were to run `cd foo && gazelle`?

Imo try explore using sync.Once to load `.bazelignore` regardless of the current value of `rel` would be clearer and help avoid this situation regardless? WDYT?

Whoaa512:
@sluongng Yea using `sync.Once` seems cleaner than relying on `rel`. Updated tho I don't know how to test this case that you mentioned as we usually run `gazelle` via bazel and always from the root

sluongng (approved):
LGTM

Pinging @blico @achew22 for additional reviews

Whoaa512:
friendly ping @blico @achew22

achew22:
LGTM

@linzhp and @blico, could I have you give this a once over in your more complex repo? Since you detected the rollback from the last time we tried this I'd like to be a bit cautious.

Thanks!

walk/config.go
```diff
@@ -104,6 +116,35 @@ func (cr *Configurer) Configure(c *config.Config, rel string, f *rule.File) {
 	c.Exts[walkName] = wcCopy
 }
 
+func (c *Configurer) loadBazelIgnore(repoRoot string, wc *walkConfig) error {
+	ignorePath := path.Join(repoRoot, ".bazelignore")
+	file, err := os.Open(ignorePath)
+	if os.IsNotExist(err) {
```
linzhp, line 122:
```suggestion
	if errors.Is(err, fs.ErrNotExist) {
```
New style check recommended by https://pkg.go.dev/os#IsNotExist

linzhp (approved):
It works in Uber

PR #1250, opened 2022-05-09
Implement very minimalistic support for go workspaces

### LINKED ISSUE #1232: Support for go 1.18 "workspace" mode?
Has anyone put any thought into gazelle support the new "workspace" mode added in go 1.18?

Some reference links:

Docs: https://go.dev/ref/mod#workspaces
Tutorial: https://go.dev/doc/tutorial/workspaces
Proposal: https://go.googlesource.com/proposal/+/master/design/45713-workspace.md
Discussion: https://github.com/golang/go/issues/45713

HakanSunay:
With this change, gazelle can invoke `update-repos` by processing `go.work` files.

Most of the common logic in `module.go` and `work.go` has been extracted into `utils.go`.
The main difference between the module and workspace mode lies in the action for fetching module sums.
For modules, only the missing sums are fetched through `go mod download`, but for workspaces the whole process is delegated to `go mod download`. This is due to the fact that `go list` does not show module sums yet. An upstream proposal was filed for the same, which can be seen in https://github.com/golang/go/issues/52792.

Given the fact the standard go tooling in 1.18 can work with `go.work` files, not many changes were actually necessary.
A smoke test was also added to `update_import_test.go` to verify that the latest version of govmomi and rest is always picked.

How to invoke:
```
gazelle update-repos -from_file=go.work -to_macro=deps.bzl%go_deps -prune=True
```

**What type of PR is this?**
Feature

language/go

Implements minimalistic support for GoLang workspaces. It would help folks with multiple `go.mod` files in the monorepo.

Fixes #1232

**Other notes for review**

Please, also do comment if we should be invoking `go work sync` as part of this workflow, which will sync all the child modules. Currently, this has not been implemented for the sake of not introducing any side effects to the execution of gazelle with regards to `go.mod` file changes.

**I am open to testing suggestions and other possible improvements, that I might have missed.**


sluongng:
This approachs makes sense to me, thought I wonder if we should mandate a `go work sync` call prior to generating things?

Could you please make sure this PR pass CI? It would help reviewing this a lot easier.

HakanSunay:
> This approach makes sense to me, though I wonder if we should mandate a `go work sync` call prior to generating things?
> 
> Could you please make sure this PR pass CI? It would help reviewing this a lot easier.

CI is green now.
I also thought about running `go work sync` and was wondering if gazelle executions should be causing any side effects to the submodules (their `go.mod` and `go.sum` files).

language/go/work.go
```diff
@@ -0,0 +1,142 @@
...
+			return language.ImportReposResult{Error: fmt.Errorf("finding module sums: %v", err)}
+		}
+		defer os.RemoveAll(tmpDir)
+		data, err := goModDownload(tmpDir, missingSumArgs)
+		dec = json.NewDecoder(bytes.NewReader(data))
+		if err != nil {
+			// Best-effort try to adorn specific error details from the JSON output.
+			for dec.More() {
+				var dl module
+				if err := dec.Decode(&dl); err != nil {
```
sluongng, line 98: let's rename this err var here to avoid confusion with the err returned from goModDownload()

language/go/work.go
```diff
@@ -0,0 +1,142 @@
...
+					// If we couldn't parse a possible error description, just ignore this part of the output.
+					continue
```
sluongng, line 100: I could be wrong but it seems like this should be a break with a possible action to handle the error before the break?

It seems like we are needlessly continueing the loop even though we have identified the error that we are interested in?

sluongng (commented):
I have some small nits, but overall this is looking real good.

There are obvious code duplication between work.go and modules.go that I wish we could apply some DRY... but perhaps YAGNI.

sluongng:
> > This approach makes sense to me, though I wonder if we should mandate a `go work sync` call prior to generating things?
> > Could you please make sure this PR pass CI? It would help reviewing this a lot easier.
> 
> CI is green now. I also thought about running `go work sync` and was wondering if gazelle executions should be causing any side effects to the submodules (their `go.mod` and `go.sum` files).

fair point.

May be add some documentation on this feature, then reminds people to run `go work sync` in the doc?

HakanSunay:
> I have some small nits, but overall this is looking real good.
> 
> There are obvious code duplication between work.go and modules.go that I wish we could apply some DRY... but perhaps YAGNI.

You are absolutely right, I think we should do it properly from the start (DRY). I have attempted to provide a more unified approach without that much code repetition for both modes, you can have a look.

language/go/work.go
```diff
@@ -0,0 +1,45 @@
...
+	}
+
+	pathToModule, err = fillMissingSums(pathToModule)
+	if err != nil {
+		return language.ImportReposResult{Error: fmt.Errorf("finding module sums: %v", err)}
+	}
+
+	// Translate to repository rules.
+	gen := toRepositoryRules(pathToModule)
+	return language.ImportReposResult{Gen: gen}
```
sluongng, line 44: nit: simplify

```suggestion
	return language.ImportReposResult{Gen: toRepositoryRules(pathToModule)}
```

sluongng (commented):
Small nits, but the rest LGTM.

Could you please update existing doc to add Go Workspace support

Once thats done, just remove draft status for other maintainers to review.

sluongng:
@linzhp @achew22 I was able to apply this patch in my repo and have things working as expected https://github.com/sluongng/nogo-analyzer/commit/4769fa6a9c18092bac055a02f4364505c82ea780

Could you guys take a look and consider merging this?

tomqwpl:
This PR explains some things I am confused about with gazelle. 
Bazel is supposed to be a tool for mono repos right? In a mono repo, I'm going to have lots of directories, each having a go.mod file. So far I've been confused by how I set this up, it looks like I have to set up targets in each of the modules to run the update-repos separately for each module, where running plain gazelle can be done over the whole thing from a BUILD.bazel file at the root. 

So just for clarification, I have:

````
WORKSPACE
BUILD.bazel
module1\
   go.mod
   BUILD.bazel
module2
  go.mod
  BUILD.bazel
````

The BUILD.bazel file at the root has 
````
# gazelle:prefix example.com/example
gazelle(name = "gazelle")
```

That appears to take care of generating the BUILD.bazel files for all the modules and their subdirectories. The bit I'm confused about is that I had naively tried adding 
````
gazelle(
    name = "gazelle-update-repos",
    args = [
        "-from_file=go.mod",
        "-prune",
    ],
    command = "update-repos",
)
````
to the same root BUILD.bazel file as well, but it appears that I have to put that individually into the BUILD.bazel file next to each go.mod file. I'm completely new to bazel, so maybe this shouldn't have been, but it felt like I should have to run that update-repos once across the whole monorepo.

Hopefully this PR helps that?

sluongng:
@tomqwpl I think you are conflating some definitions here. This PR provide support for `go.work` file, which is Go 1.18's new workspace feature.

If you have a hard time learning Bazel/Gazelle, I would suggest using Bazel's Community Slack or StackOverFlow to ask questions.

linzhp (approved):
Tested in Uber's repo, this doesn't break existing `gazelle update-repo -from_file=go.mod`.

PR #1271, opened 2022-06-09
Don't exclude spaces from the label name regex

illicitonion:
**What type of PR is this?**

Bug fix

label

Don't exclude spaces from the label name regex

In parsing the output of `bazel query`, we found that some files with
spaces in their names are included in the output.

It seems reasonable that all labels that come out of a `bazel query`
should be parsable by this `Label` type; let's just use the raw regex
from the Bazel documentation.

### DIFF
--- a/label/label.go
+++ b/label/label.go
@@ -65,9 +65,7 @@ var (
 	labelRepoRegexp = regexp.MustCompile(`^@$|^[A-Za-z.-][A-Za-z0-9_.-]*$`)
 	labelPkgRegexp  = regexp.MustCompile(`^[A-Za-z0-9/._@-]*$`)
 	// This was taken from https://docs.bazel.build/versions/main/build-ref.html#name
-	// Note: We've manually removed space from the regex, because though these technically parse
-	// with Bazel (and can appear in query results), they cannot actually be used in practice.
-	labelNameRegexp = regexp.MustCompile("^[A-Za-z0-9!%-@^_`\"#$&'()*-+,;<=>?\\[\\]{|}~/.]*$")
+	labelNameRegexp = regexp.MustCompile("^[A-Za-z0-9!%-@^_` \"#$&'()*-+,;<=>?\\[\\]{|}~/.]*$")
 )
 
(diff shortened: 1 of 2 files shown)


label/label.go
```diff
@@ -65,9 +65,7 @@ var (
 	labelRepoRegexp = regexp.MustCompile(`^@$|^[A-Za-z.-][A-Za-z0-9_.-]*$`)
 	labelPkgRegexp  = regexp.MustCompile(`^[A-Za-z0-9/._@-]*$`)
 	// This was taken from https://docs.bazel.build/versions/main/build-ref.html#name
-	// Note: We've manually removed space from the regex, because though these technically parse
-	// with Bazel (and can appear in query results), they cannot actually be used in practice.
```
achew22, line 69: Can you explain why this comment is no longer true?

achew22 (changes requested):
Thanks for the PR!

label/label.go
illicitonion: It was never really true - I made a limiting assumption which was incorrect. https://github.com/bazelbuild/bazel/issues/4327 is an issue tracking fixing gaps in this support elsewhere in Bazel, too, which is an open and accepted issue, not something closed as working as intended.

label/label.go
```diff
@@ -65,9 +65,7 @@ var (
...
-	labelNameRegexp = regexp.MustCompile("^[A-Za-z0-9!%-@^_`\"#$&'()*-+,;<=>?\\[\\]{|}~/.]*$")
+	labelNameRegexp = regexp.MustCompile("^[A-Za-z0-9!%-@^_` \"#$&'()*-+,;<=>?\\[\\]{|}~/.]*$")
```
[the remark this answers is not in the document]
illicitonion: Added a test :)

illicitonion:
@achew22 Could I nudge you at this again? Cheers!

achew22:
Thanks for adding that test, and for your continued support @illicitonion! Really appreciate your work

PR #1274, opened 2022-06-12
fix-update: defer construction of loads, kinds slices until after CheckFlags

### LINKED ISSUE #1275: fix-update: Custom extension flags cannot influence Loads() or Kinds()
The `.RegisterFlags()` / `.CheckFlags()` API exists to allow extensions to offer custom behavior via flags.  Currently, there is no way that these flags may influence the return value(s) of the `.Loads()` and/or `.Kinds()` API due to the fact that `.CheckFlags` is called after `.Loads()` and `.Kinds()`.  

master

0.27

4.0.0

Yes

pcj:
> Feature
> all

The `.RegisterFlags()` / `.CheckFlags()` API exists to allow extensions to offer custom behavior via flags.  Currently, there is no way that these flags may influence the return value(s) of the `.Loads()` and/or `.Kinds()` due to the fact that `.CheckFlags` is called after `.Loads()` and `.Kinds()`.  However, this is an arbitrary limitation based on the original code that decided to initialize the metaresolver, loads slice, and kinds slice early in `runFixUpdate`.  Fortunately, this is otherwise independent of initialization of the update configuration.

This PR simply moves `newFixUpdateConfiguration` before `.Loads` and `Kinds()` is called.

Fixes #1275

### DIFF
--- a/cmd/gazelle/fix-update.go
+++ b/cmd/gazelle/fix-update.go
@@ -241,6 +241,12 @@ func runFixUpdate(wd string, cmd command, args []string) (err error) {
 		&updateConfigurer{},
 		&walk.Configurer{},
 		&resolve.Configurer{})
+
+	c, err := newFixUpdateConfiguration(wd, cmd, args, cexts)
+	if err != nil {
+		return err
+	}
+
 	mrslv := newMetaResolver()
 	kinds := make(map[string]rule.KindInfo)
 	loads := genericLoads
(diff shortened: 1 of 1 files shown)


sluongng:
This code looks obviously correct.

But I would suggest provide an example in the form of a small test to guide future user on how to use this new features.
I was thinking that go's `update-repos` https://github.com/bazelbuild/bazel-gazelle/blob/aa57fa1a7ef4361fe5dee034b80430f9dd53bb56/language/go/config.go#L397-L423 could use a new flag that let us override the `Loads()` of `go_repository` https://github.com/bazelbuild/bazel-gazelle/blob/aa57fa1a7ef4361fe5dee034b80430f9dd53bb56/language/go/kinds.go#L181.

WDYT?

sluongng (approved):

pcj:
@sluongng reasonable, will plan to add test to https://github.com/bazelbuild/bazel-gazelle/blob/master/cmd/gazelle/integration_test.go before merging.

PR #1277, opened 2022-06-13
Sluongng/revert patch

sluongng:
This reverts commit 1dc3aaa5a964138b996c6912ec3d7069a6eccd14.

It seems like by using the native repository_ctx.path(file, strip)
in bazel_tools, we would be applying the patches to the repository
archive _before_ BUILD files generation.  This is not desirable as
patches on top of go_repository is often used to modify the BUILD files
themselves.

Another differences is that `patch` allow patch files without trailing
blank/empty lines while repository_ctx.patch() would fail to recognize
EOF in patch files without a trailing empty line:

  Expecting more chunk line at line xx

Let's revert this change until we find a more appropriate solution to
simplify go_repository structure.

Added docs to highlight the reason why we had to revert.
Also applied minor formatting changes by buildifier.


linzhp (approved):
Thanks for catching the bug and reverting

Issue #579, opened 2019-07-14
Nicer story for private modules, GOPROXY


steeve:
Currently the whole story around `GOPROXY`, `GONOPROXY`, `GOSUMDB` and `GONOSUMDB` involves passing those via the env when running bazel.

This is inconvenient in the sense we now need wrapper around bazel invocations to pass those.

This story becomes even mode complex for private modules.

With Go 1.13 around the corner, `go mod download` (and thus `fetch_repo`) always verifies downloads against the `GOSUMDB` service. This breaks downloads of private modules altogether if they are not added to `GONOSUMDB`.

Possible solutions, not all mutually exclusive:
- Pass `GOPROXY`, `GONOPROXY`, `GOSUMDB`, `GONOSUMDB`, `GOPRIVATE` to `fetch_repo` (easy)
- Allows to set them globally in `WORKSPACE`
- Always add the current fetch to the `GONOSUMDB`, since it's already verified with the `sum` in the `go_repository` attribute. It should, however be verified when that rule is created (via `update-repos`).
- Add a `is_private`, `no_proxy` and `no_sumdb` attributes to `go_repository` to leverage those environment variables automatically.

My preference would be to be able to get `GOPROXY` globally via the workspace, and manually add `is_private` to certain repos.

That said, them importing from `go.mod` there no way to tell gazelle that a certain repo is private. But it can live with a `make godeps` once in a while.

More info on `GOPRIVATE`: https://go-review.googlesource.com/c/go/+/181719/

jayconrod:
I need to tag a release shortly, but I'd like to do at least part of this before then. These points seem straightforward and clearly good ideas:

* Pass GOPROXY, GONOPROXY, GOSUMDB, GONOSUMDB, GOPRIVATE to fetch_repo (easy)
* Always add the current fetch to the GONOSUMDB, since it's already verified with the sum in the go_repository attribute. It should, however be verified when that rule is created (via update-repos).
 
Actually, I think we can just set `GOSUMDB=off` within `fetch_repo` since the `sum` attribute is mandatory.

Setting these variables globally in WORKSPACE will need more thought, but is probably the right thing to do.

qzmfranklin:
I just do not like the fact that gazelle goes to network for generating the BUILD files.

[...]

PR #1279, opened 2022-06-17
Allow GO env vars to be specified via update-repos

Whoaa512:
**What type of PR is this?**

> Feature

> * go_repository
> * update-repos

Short of being able to specify the `GOPROXY` and other vars via the workspace (which would be ideal), I needed the ability to specify an internal proxy as we're not allowed to hit public Go proxies at work.

This PR adds the `go_env` attr on `go_repository` and correctly feeds it into `fetch_repo`. Additionally it allows setting these env vars when calling `update-repos`.

Related to #579

**Other notes for review**
I'm mostly leaving this for others to use until a proper solution to configuring via WORKSPACE is put forth. Internally we're using this via a bazel patch.

Also I'm not a day-to-day Go programmer so very open to feedback on the implementation.

### DIFF
--- a/internal/go_repository.bzl
+++ b/internal/go_repository.bzl
@@ -194,7 +194,7 @@ def _go_repository_impl(ctx):
         fail("one of urls, commit, tag, or importpath must be specified")
 
     env = read_cache_env(ctx, go_env_cache)
-    env_keys = [
+    go_env_keys = [
         # Respect user proxy and sumdb settings for privacy.
         # TODO(jayconrod): gazelle in go_repository mode should probably
         # not go out to the network at all. This means *the build*
@@ -205,7 +205,8 @@ def _go_repository_impl(ctx):
         "GOPRIVATE",
         "GOSUMDB",
         "GONOSUMDB",
-
+    ]
+    env_keys = go_env_keys + [
         # PATH is needed to locate git and other vcs tools.
         "PATH",
 
@@ -229,6 +230,7 @@ def _go_repository_impl(ctx):
     env.update({k: ctx.os.environ[k] for k in env_keys if k in ctx.os.environ})
 
     if fetch_repo_args:
+        env.update({ k: ctx.attr.go_env[k] for k in go_env_keys if k in ctx.attr.go_env })
         # Disable sumdb in fetch_repo. In module mode, the sum is a mandatory
         # attribute of go_repository, so we don't need to look it up.
         fetch_repo_env = dict(env)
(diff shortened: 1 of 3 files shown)


internal/go_repository.bzl
```diff
@@ -415,6 +417,32 @@ go_repository = repository_rule(
...
+
+            Valid keys are:
+                GOPROXY
+                GONOPROXY
+                GOPRIVATE
+                GOSUMDB
+                GONOSUMDB
+            """,
+        ),
+        "go_env": attr.string_dict(
```
achew22, line 432: Is this supposed to be duplicate of the preceeding `go_env` entry?

language/go/config.go
```diff
@@ -125,6 +125,19 @@ type goConfig struct {
...
+	goEnv
+}
+
+// goEnv attributes for go_repository rules, set on the command line.
+type goEnv struct {
+	// GoProxy, GoNoProxy, GoPrivate, GoSumDb, and GoNoSumDb
+	// used to specify GO env vars for the go_repository rule.
+	GoProxy, GoNoProxy, GoPrivate, GoSumDb, GoNoSumDb goEnvKey
+}
+type goEnvKey struct {
```
achew22, line 139: nit: this is an entry or a pair not a key since it contains a value.

language/go/update.go
```diff
@@ -114,6 +115,23 @@ func setBuildAttrs(gc *goConfig, r *rule.Rule) {
 		buildTags := strings.Split(gc.buildTagsAttr, ",")
 		r.SetAttr("build_tags", buildTags)
 	}
+	env := make(map[string]string)
+
+	// Iterate of the goEnv struct using the `key` and `value` field with reflect
```
achew22, line 120: I think this might be a bit of an overreach to invoke reflection. Either switch to an iterable type (`map` seems  like a good candidate), or, since there are a fixed number of these and we can quite simply drop them inline. Empty environment variables being "" or being unset almost certainly behave the same way.

```go
env["GOPROXY"] = gc.goEnv.GoProxy
env["GONOPROXY"] = gc.goEnv.GoNoProxy
env["GOPRIVATE"] = gc.goEnv.GoPrivate
env["GOSUMDB"] = gc.goEnv.GoSumDb
env["GONOSUMDB"] = gc.goEnv.GoNoSumDb
```

language/go/config.go
```diff
@@ -137,6 +150,13 @@ func newGoConfig() *goConfig {
 		goProtoCompilers: defaultGoProtoCompilers,
 		goGrpcCompilers:  defaultGoGrpcCompilers,
 		goGenerateProto:  true,
+		goEnv: goEnv{
```
achew22, line 153: Would it be better to change the type to a `map[string]string`

```go
goEnv: map[string]string{
  "GOPROXY": "",
  "GONOPROXY": "",
  "GOPRIVATE": "",
  "GOSUMDB": "",
  "GONOSUMDB": "",
}
```

so that you have something you can iterate over when processing? You could go further and make a specific type so that nothing happens on accident.

```go
type goEnvValue string
```

achew22 (changes requested):
I think I can see the intent here, but it would be nice to have some kind of documentation modification to go with it so that people know it exists as well as a test case to confirm the functionality.

PR #1197, opened 2022-03-14
Code Quality Improvements

sluongng:
This PR is composed of multiple commits which aimed to clean up the code base according to the lint result of running `staticcheck` and `golangci-lint` and `gofumpt` over the Gazelle repository. It's best to review this PR per-commit instead of as a whole. There should be no functional changes as a result of this PR and exported interfaces were not touched.

```
> git log --oneline --reverse origin/master..HEAD
c8a1d88 Fix staticcheck ST1006
71a12aa fix ST1012
66f45db apply gofumpt
8c7397f fix U1000
2b03f00 fix S1011, S1017, SA9003, S1007, SA4009
035663f remove deadcode constants
11d9e29 (HEAD) fix errcheck
```

This was a results of me reading through the code base over the weekend while trying to make peace with my editor.
The code after should be a lot easier to read and maintained compared to before.

Worth to note that ST1005 was intentionally skipped as it does not make sense in most cases and generate a lot of noise.
```
> staticcheck ./...
cmd/fetch_repo/module.go:37:10: error strings should not end with punctuation or a newline (ST1005)
cmd/fetch_repo/vcs.go:34:15: error strings should not end with punctuation or a newline (ST1005)
cmd/gazelle/update-repos.go:63:10: error strings should not be capitalized (ST1005)
cmd/gazelle/update-repos.go:66:10: error strings should not be capitalized (ST1005)
cmd/gazelle/update-repos.go:90:11: error strings should not end with punctuation or a newline (ST1005)
cmd/gazelle/update-repos.go:98:11: error strings should not end with punctuation or a newline (ST1005)
cmd/gazelle/update-repos.go:324:15: error strings should not be capitalized (ST1005)
cmd/move_labels/move_labels.go:199:15: error strings should not be capitalized (ST1005)
cmd/move_labels/move_labels.go:214:15: error strings should not end with punctuation or a newline (ST1005)
cmd/move_labels/move_labels.go:222:15: error strings should not be capitalized (ST1005)
cmd/move_labels/move_labels.go:222:15: error strings should not end with punctuation or a newline (ST1005)
flag/flag.go:84:9: error strings should not be capitalized (ST1005)
language/proto/generate.go:187:14: error strings should not end with punctuation or a newline (ST1005)
merger/fix.go:186:9: error strings should not end with punctuation or a newline (ST1005)
repo/repo.go:263:15: error strings should not be capitalized (ST1005)
repo/repo.go:267:15: error strings should not be capitalized (ST1005)
```

[...]

PR #1280, opened 2022-06-18
Stop inferring import path for empty packages

linzhp:
**What type of PR is this?**

Bug fix

language/go

After #1197, Gazelle prints out the following warning every time it visits a directory without any `gazelle:prefix`, even the directory has no Go files:

```
gazelle: could not infer import path: /foo/bar: go prefix is not set, so importpath can't be determined for rules. Set a prefix with a '# gazelle:prefix' comment or with -go_prefix on the command line
```

The root cause is Gazelle tries to generate package and infer import path even when there is no Go or proto package. The error to infer import path was ignored before #1197. However, we should not infer import path for empty package because it will be inferred later when there is Go files: https://github.com/bazelbuild/bazel-gazelle/blob/32dab97bac88b1a35653d33a034b38e55d757983/language/go/generate.go#L165

### DIFF
--- a/language/go/generate.go
+++ b/language/go/generate.go
@@ -404,9 +404,8 @@ func emptyPackage(c *config.Config, dir, rel string, f *rule.File) *goPackage {
 		dir:  dir,
 		rel:  rel,
 	}
-	if err := pkg.inferImportPath(c); err != nil {
-		log.Printf("could not infer import path: %v", err)
-	}
+	// It's fine if it fails to infer import path
+	_ = pkg.inferImportPath(c)
 
 	return pkg
 }
--- a/language/go/generate.go
+++ b/language/go/generate.go
@@ -151,7 +151,7 @@ func (gl *goLang) GenerateRules(args language.GenerateArgs) language.GenerateRes
 					protoName = name
 					break
 				}
-			} else if len(protoPackages) > 1 {
+			} else {
 				pkg = emptyPackage(c, args.Dir, args.Rel, args.File)
 			}
 		} else {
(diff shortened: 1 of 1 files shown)


language/go/generate.go
```diff
@@ -404,9 +404,8 @@ func emptyPackage(c *config.Config, dir, rel string, f *rule.File) *goPackage {
 		dir:  dir,
 		rel:  rel,
 	}
-	if err := pkg.inferImportPath(c); err != nil {
-		log.Printf("could not infer import path: %v", err)
-	}
+	// It's fine if it fails to infer import path
+	_ = pkg.inferImportPath(c)
```
achew22, line 408: Is it the right thing to do to silently ignore this error? Why is it fine to ignore?

linzhp: See the PR summary. Let me know if it's still not clear or you have a better way to fix that.

achew22: Exactly. The information necessary to understand the silencing of the error isn't in the code. A later reader is very unlikely to be able to figure out why this error shouldn't be printed based upon the context that's available here and might bring it back. Since there isn't a test case, we have to depend on inline documentation to describe our intent.

IMO comments should explain *why* things are happening, not what is happening -- the code describes more accurately what is happening.

linzhp: How about removing this function call all together?

achew22: I think that's fine. `inferImportPath` does have a side effect

https://github.com/bazelbuild/bazel-gazelle/blob/1ab6bae363de3da4760501cb541fd80cb93031ec/language/go/package.go#L157-L167

Do we care about that though? I believe that it is just caching the import path inference, but want to double check that you think that's reasonable.

Assuming you think omission is safe, LGTM and let's merge it 👍

linzhp: The reason why this omission is safe is that later in the code, Gazelle checks whether `pkg.ImportPath` is set, and call `pkg.inferImportPath` when it's not: https://github.com/bazelbuild/bazel-gazelle/blob/32dab97bac88b1a35653d33a034b38e55d757983/language/go/generate.go#L164-L165

So the call I remove in this function is a duplicate call

PR #1281, opened 2022-06-18
Disable `go get -d -- importPath` for unknown modules in unused code

### LINKED ISSUE #902: go_repository is very slow for some packages

### LINKED ISSUE #1190: Some go_repository gazelle generations are slow, could be pre-computed or cached
A common pathological case is `@com_github_docker_docker` which can spend hundreds of seconds generating BUILD files.
Users are stuck waiting for this to run anytime the repository rule is invalidated.

@siggisim had the clever solution of checking in the resulting generated BUILD files rather than computing on-the-fly, using a script that converts to http_archive like https://github.com/buildbuddy-io/buildbuddy/blob/14ac4f6d9e3df578e31c129add910d176a63d075/deps.bzl#L876-L884

On Slack, @sluongng proposes the results are cachable:
https://bazelbuild.slack.com/archives/CDBP88Z0D/p1647012317774269
> naively speaking, a Gazelle cache is fairly easy to implement:
> Enable/disable by setting an environment variable pointing to a path
> Only cache if the go_repository SHA256 is set, use it as the cache key
> Under `$HOME/.cache/gazelle`, store the patch file under `./<sha256>.patch`
> If cache not found, generate then store in cache

bozaro:
**What type of PR is this?**

Bug fix

go_repository

Current behaviour:

 - my project use only small part of `github.com/docker/docker` package;
 - because `github.com/docker/docker` have not `go.mod` file, go mod commands don't see whole module dependency graph;
 - `go mod tidy` on my project adds to `go.mod` and `go.sum` transitive dependencies to little part of `github.com/docker/docker` dependencies;
 - gazelle generates WORKSPACE based by go build info (provided by `go mod download -json` command) and have incomplete `github.com/docker/docker` module dependency list (just enough to build my project). But gazelle generate BUILD files for whole `github.com/docker/docker` module;
 - for every unknown package gazelle call `go get -d` command and spend a lot of time (in my case summary it takes ~20 min).

As result gazelle spend a lot of time for resolving imports in unused third-party code.

Fixes #1190
Fixes #902

**Other notes for review**

We have been using this patch on our project since April 14 2022 without any issues. But perhaps there are some corner cases unknown to me.

### DIFF
--- a/repo/remote.go
+++ b/repo/remote.go
@@ -141,7 +141,7 @@ func NewRemoteCache(knownRepos []Repo) (r *RemoteCache, cleanup func() error) {
 		modVersion:            remoteCacheMap{cache: make(map[string]*remoteCacheEntry)},
 	}
 	r.ModInfo = func(importPath string) (string, error) {
-		return defaultModInfo(r, importPath)
+		return "", nil
 	}
 	r.ModVersionInfo = func(modPath, query string) (string, string, error) {
 		return defaultModVersionInfo(r, modPath, query)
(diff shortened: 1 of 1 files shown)


blico:
@bozaro what version of Gazelle is your repo on? As of [v.0.25.0](https://github.com/bazelbuild/bazel-gazelle/releases/tag/v0.25.0 )`go_repository` rules default to `static` resolution, meaning they do not call out to the network (similar to your change here).

bozaro:
We added this patch on v0.24.0 version. Currently we use v0.25.0 and I removed it and it looks like it's no longer needed.

PR #1293, opened 2022-06-30
Expose rule.Rule as Expr

illicitonion:
**What type of PR is this?**

Feature

`rule`

This allows for richer information gathering about existing rules in
BUILD files. In particular, it allows for calling `bzl.Walk` on an
entire `Rule`, whereas currently only individual args and attrs values
are exposed.

An example of where this is insufficient is that right now there's no
way to read the comment in this rule:

```starlark
filegroup(
    name = "fg",
    # This comment is important
    srcs = ["foo"],
)
```

because the existing API only surfaces the attr value as an Expr
(`["foo"]`), and not the `AssignExpr` (`srcs = ["foo"]`), to which this
comment  is actually attached.

### DIFF
--- a/rule/rule.go
+++ b/rule/rule.go
@@ -749,6 +749,10 @@ func (r *Rule) ShouldKeep() bool {
 	return ShouldKeep(r.expr)
 }
 
+func (r *Rule) Expr() bzl.Expr {
+	return r.expr
+}
+
 // Kind returns the kind of rule this is (for example, "go_library").
 func (r *Rule) Kind() string {
 	return r.kind
--- a/rule/rule_test.go
+++ b/rule/rule_test.go
@@ -362,6 +362,88 @@ func TestShouldKeepExpr(t *testing.T) {
 	}
 }
 
+func TestWalkForKeeps(t *testing.T) {
+	for _, tc := range []struct {
+		desc, src string
+		want      bool
+	}{
+		{
+			desc: "no keep",
+			src: `
+filegroup(
+    name = "fg",
+    srcs = ["foo"],
+)
+`,
+			want: false,
+		},
+		{
+			desc: "before",
+			src: `
+# keep
+filegroup(
+    name = "fg",
+    srcs = ["foo"],
+)
+`,
+			want: true,
+		},
+		{
+			desc: "before-attr",
+			src: `
+filegroup(
+    name = "fg",
+    # keep
+    srcs = ["foo"],
+)
+`,
+			want: true,
+		},
+		{
+			desc: "in-attr",
+			src: `
+filegroup(
+    name = "fg",
+    srcs = [
+        # keep
+        "foo",
+    ],
+)
+`,
+			want: true,
+		},
+		{
+			desc: "after-attr",
+			src: `
+filegroup(
+    name = "fg",
+    srcs = [
+        "foo", # keep
+    ],
+)
+`,
+			want: true,
+		},
+	} {
+		t.Run(tc.desc, func(t *testing.T) {
+			ast, err := bzl.Parse(tc.desc, []byte(tc.src))
+			if err != nil {
+				t.Fatal(err)
+			}
+			r := ruleFromExpr(0, ast.Stmt[0])
+			var hasKeep bool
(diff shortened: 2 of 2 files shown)


achew22:
Can you elaborate a bit on the usecase here? This feels like a pretty significant expansion of the parsing capabilities that can make plugins A LOT more complicated. Would, for example, a predicate be able to solve the situation you're describing here?

achew22:
Thanks for engaging in the conversation! I'm always happy to talk about this kind of stuff.

Just to confirm that I understand. You have a target defined:

```
java_test(
    name = "flaky-tests",
    srcs = ["ThirdTest.java"],
    flaky = True,
)
```

Is flaky a field that's detected from the, in this case, `.java` files or is this something that was manually added after the `BUILD` file was generated? I would expect a field like `flaky` to not be replaced in a later run if it was manually added. Are you seeing that behavior?

illicitonion:
It's manually added. Right now what we're doing in this case is manually writing:

```starlark
# gazelle:exclude ThirdTest.java

java_test(
    name = "flaky-tests",
    srcs = ["ThirdTest.java"],
    flaky = True,
)
```

which means our gazelle plugin doesn't try to add `ThirdTest.java` to the `java_test_suite` for the whole package, but _also_ means we don't get the `deps` of `flaky-tests` to be automatically maintained.

I'm looking for a way to say "Don't include `ThirdTest.java` in the standard generated `java_test_suite`, but still do _some_ automatic maintenance of it." Requiring a comment on the custom attributes seemed like the most self-documenting way of doing so.

achew22:
Just spitballing, so consider this a straw man to explore the design constraints. WDYT about doing something like

```
# gazelle:java_test_only ThirdFile.java
```

and using that to indicate that the third file should be generated in a different way? Apologies for not remembering more about Java testing, but is there some property of `ThirdFile.java`, maybe a particular import, that could be used to determine that it isn't intended to be a part of a `java_test_suite`?

illicitonion:
Thanks for giving this some real thought, it's definitely an interesting design space I'd be interested to explore more to get to a great solution!

> Just spitballing, so consider this a straw man to explore the design constraints. WDYT about doing something like
> 
> ```
> # gazelle:java_test_only ThirdFile.java
> ```

Yeah, this was one of the options I thought about in design here, and I think we could make it work reasonably, but the disadvantage I see with it is that it decouples the "why" from the "what".

Let's say that test stops being flaky, so someone removes the `flaky = True` attribute from the separate target, ideally that test would automatically get folded back into the test suite, rather than being keeping being a separate target - there's no longer anything special about it, so it can just get folded into the main one.

Similarly, being able to see at a glance "The reason this is a separate target is because it's marked `flaky = True` unlike the others" is valuable, whereas by using a separate directive, it relies on someone manually adding extra detail to the comment explaining why (and risks that comment becoming stale).

> is there some property of `ThirdFile.java`, maybe a particular import, that could be used to determine that it isn't intended to be a part of a `java_test_suite`?

No, for the most part this is an external, human-knowable-only constraint, e.g. "this test happens to perform network access" or "we've noticed this test is flaky" - the core issue here really is "a human wants to treat this specially, let's make that as easy as possible to automate". `java_test_suite` is basically a macro which says "for each java file, generate an identical `java_test` target, just for a different source file", and what we're looking to do here is add a way of expressing "we want to manually separate out one of these per-test-file targets, so we can override or supplement some attribute" - a human is going to need to express that constraint in the BUILD file _somehow_ (or I guess we could use a magic comment in the java file, but that's even less discoverable from looking at the BUILD file), and I think  that information living on the target is more clear than putting it separately...

achew22:
Okay, sorry -- one more question because I think I now understand something that I didn't before. Is `java_test_suite` a macro that produces a bunch of `java_test` targets effectively?

achew22:
It sounds to me like this is a property of the class that has the tests and not really a property of the build graph. The implications on the build graph are kind of an implementation detail.

What would you think about annotating the test class with something like `@Flaky`?

illicitonion:
Yeah, that sounds pretty reasonable... I'll put together a proof of concept and see if anything feels sketchy about it :)

Thanks for having a really useful conversation about it!

PR #1294, opened 2022-07-01
Use repo-relative labels everywhere

fmeum:
This makes it possible to load gazelle under any repo name, not
necessarily bazel_gazelle. This opens up the possibility to call the
gazelle Bazel module "gazelle".


achew22 (commented):
Do we not need the checked in BUILD files any more?

achew22:
> I dropped that commit out of an abundance of caution. I simply don't know, I couldn't find any reference outside of documentation. The tests passed locally with them deleted.

I think it's a good thing to do, but let's do it as a separate PR so that reversion is cleaner

fmeum:
> > I dropped that commit out of an abundance of caution. I simply don't know, I couldn't find any reference outside of documentation. The tests passed locally with them deleted.
> 
> I think it's a good thing to do, but let's do it as a separate PR so that reversion is cleaner

Sounds good! The current PR should be good enough for bzlmod anyway.

linzhp (approved):
Successfully ran Gazelle with this patch in Uber's Go monorepo

PR #1296, opened 2022-07-09
Fix RST URL errors for rules_jvm

qaisjp:
**What type of PR is this?**

Documentation

README

It uses Markdown, should use RST

n/a, see following PR summary template text:

> Small bug fixes and documentation improvements don't need issues.


achew22:
Thanks for fixing the link!

PR #1303, opened 2022-07-15
language: add BaseLang

sluongng:
In order to help folks extend Gazelle quicker and easier, provide a base
implementation of the language.Language interface where everything is
no-ops.

End users then can simply compose their downstream struct with this
to have a valid implementation to iterate upon.


dragonsinth (approved):

sluongng:
@linzhp @fmeum would appreciate if you guys take a look. 🙏

fmeum:
I can say that I would have found this useful back when I worked on my first Gazelle extension. @linzhp Do you have any concerns?

fmeum (approved):

linzhp (approved):
Shouldn't break anything

Issue #1238, opened 2022-04-13
go_repository should emit version information for auditing purposes


sluongng:
# Background

Software supply chain security has been a growing pain in recent years.

Typically, an organization would want to have the ability to scan through all of their external dependencies and check for malicious versions where possible. 

These information can then be fed through a security scanner to validate that there is no malicious dependency.
Furthermore, they can be used to build up a Software Bill of Materials (SBOM) for downstream consumption.

# Problem

Right now, `go_repository` does not emit any clue regarding when / where the repo came from once download. This make it hard for folks to consume the dependencies version programmatically and build automations and tests based on them.

Today, there are some alternative approaches that would give you close:

Bazel query
```
bazel query 'kind("go_repository rule", //external:*)' --output=xml
```

Bazel experimental_repository_resolved_file
```
bazel sync --experimental_repository_resolved_file=temp.bzl
```

Both of these approaches can be too verbose and hard to consumed programmatically as part of Bazel build graph.

# Propose solution

Provide option `create_version_file` in `go_repository`, default to `false`.
When this option is true, `go_repository` should create a `GO_REPOSITORY_VERSION.bzl` file at the root level of the repository.

This version file should follow this template:
```
NAME=<repository-name>

URL=<download-url>

COMMIT=<commit-sha>
TAG=<tag-name>

VERSION=<version>

SUM=<check-sum>
```

The file must have either one of `url` and `commit/tag` and `version` value provided according to how `go_repository` target was downloaded.

This file can then be consumed via various macros/aspect rules to test and generate SBOM for each of the go_binary releases.

achew22:
Presumably this could be made to work with the new in Go 1.18 [`debug/buildinfo`](https://tip.golang.org/doc/go1.18#debug/buildinfo) package. Is this a step in that direction? I think we will need to create some kind of compile time value to inject that meets this specification https://cs.opensource.google/go/go/+/refs/tags/go1.18.1:src/runtime/debug/mod_test.go;l=26, but haven't looked into it really at all.

sluongng:
@achew22 no, `buildinfo` is a compile time build stamping and can be used to identify which version of the current workspace I am building from. I am trying to identify which versions of all the dependencies that the go_binary is built with.

I think for this issue, I am looking for something that can be run/extract without needing to stamp the binary.

[...]

PR #1315, opened 2022-08-08
Record where dependencies come from

### LINKED ISSUE #1314: Information about where third party libraries are from should be present in the build graph
### Feature request

When using a `go_repository`, the repository rule knows the origin and precise version information of the third party repo being loaded. However, because this information is not encoded anywhere in the generated targets, it is impossible to retrieve this information in a build rule or aspect. This makes it impossible to walk the dependency graph of a target and find out which versions of which dependencies were used. 

In contrast, `rules_jvm_external` embeds this information in the `tags` of generated rules.

Having this information available makes it simpler to generate artifacts such as SBOMs (Software Bill of Materials) which can be consumed by other tooling.

0.26.0

Snapshot of cde7d7bc27a34547c014369790ddaa95b932d08d

5.2.0

Yes

N/A

shs96c:
**What type of PR is this?**

Feature

go_repository

This allows rules and aspects to determine where a particular
Go dependency has come from. This can be useful when (eg)
generating SBOMs directly from Bazel projects.

The approach of storing the information in `tags` is copied
from `rules_jvm_external`, which uses a similar technique to
store maven coordinates.

Build files are amended using Buildozer, since there is no
way to be sure whether or not a specific `go_repository` had
build files before being created. Where Buildozer is not
available, a warning is printed and no version information is
added to the tags of the generated libraries.

Fixes #1314 

**Other notes for review**


shs96c:
Looks like the failing CI builds are to do with `bzlmod`. I can make that work, but I'd appreciate any feedback on this PR in the meantime.

deps.bzl
```diff
@@ -47,6 +51,34 @@ def gazelle_dependencies(
...
+
+    _maybe(
+        http_file,
+        name = "buildozer_linux_x86",
+        urls = ["https://github.com/bazelbuild/buildtools/releases/download/5.0.1/buildozer-linux-amd64"],
+        sha256 = "78204dac0ac6a94db499c57c5334b9c0c409d91de9779032c73ad42f2362e901",
+        downloaded_file_path = "buildozer",
+        executable = True,
+    )
+
```
sluongng, line 81: I think a better approach here, instead of using buildozer, is to enhance gazelle with a flag to add a set of strings as tag attribute under all the generated build target. 🤔

internal/go_repository.bzl
```diff
@@ -310,6 +324,24 @@ def _go_repository_impl(ctx):
...
+    else:
+        buildozer = None
+        print("Unable to find buildozer binary, so omitting version information from tags")
+
+    if buildozer:
+        for tag in tags:
+            res = ctx.execute([buildozer, "add tags %s" % tag, "//...:%go_library"])
+            if res.return_code != 0:
+                fail("Unable to execute buildozer for tag %s" % tag)
+
```
sluongng, line 344: Aside from the earlier point I made about favoring usage of `gazelle` over `buildozer`, I think we should avoid modifying things after `patch(ctx)`. This way, user are always left with an option of overriding the default generated result with a custom patch. If we do these _after_ `patch(ctx)`, user will not have much options but to accept the default (or, patch Gazelle themselves).

sluongng (commented):
Also related to https://github.com/bazelbuild/bazel-gazelle/issues/1238

deps.bzl
shs96c: I would agree, but there are code-paths through `go_repository` that will make use of pre-existing build files in the repositories. That is, we don't always generate the build files, but we always want to tag them. By using buildozer, we can do this in a consistent way.

internal/go_repository.bzl
shs96c: But we're not sure if there are build files being modified by the `patch`. If there are, then the result of running `buildozer` may not allow patches to apply cleanly. The way things are approached here feels like the least risky, and we should only be adding metadata at this point.

achew22:
Hey @shs96c, thanks for sending in your first PR! This looks like a very fun one to get started on.

To echo the problem back at you so we can confirm I understand it before we go too much further. You want to:

1. Include a dependency on a 3rd party resource using `go_repository`
2. Have included with that dependency information about the origin of the materials
3. Make it possible to extract the license information from the 3rd party deps transitively using something like a `bazel query`

A couple of questions from me:

1. How tied are you to the idea of using `tags` for this?
  a. Are you limited in the version of Bazel that you can adopt to achieve your goal? Could you upgrade to > 5.2?
  b. Do you, for example, have preexisting code to parse these out of tags and are looking to reuse it?
1. Have you looked at any of the other efforts in the space? The one I'm the most aware of is [rules_license](https://github.com/bazelbuild/rules_license/blob/main/examples/vendor/constant_gen/BUILD#L14-L23). This project allows you to effectively attach a new attribute `license` on any target inside of a repo. You can attach arbitrary metadata to it there, but the list of already supported attributes contains: `package_name`, `package_url`, and `package_version`. If that information was available for any target inside the repository, would that satisfy your SBOM requirements? Here is an [example output](https://github.com/bazelbuild/rules_license/blob/main/examples/vendor/constant_gen/generator_licenses.golden) from their included rule.

shs96c:
Hi @achew22! Thanks for the response, and thanks for echoing back your understanding of the goals. My intent was slightly different, so if I may rephrase:

1. Include a dependency on a 3rd party resource using `go_repository`.
2. Have included with that dependency information about the origin of the materials, such as the version or git sha being used.
3. Make it possible to extract this version information from the 3rd party deps transitively using an `aspect` and custom build rule.

The licensing information is somewhat unrelated. The more important information is the source and version information of the dependency.

In answer to your questions:

> How tied are you to the idea of using tags for this?

I don't mind how this is achieved, but tags have worked well for this in `rules_jvm_external`, and they're a lightweight mechanism for adding this information. It's not entirely ideal that we're using well-known prefixes, but, again, this is a model that's followed successfully elsewhere.

> Are you limited in the version of Bazel that you can adopt to achieve your goal? Could you upgrade to > 5.2?

Being usable with earlier versions is a "nice to have", but not a hard requirement. Depending on Bazel 6, however, is untimely.

> b. Do you, for example, have preexisting code to parse these out of tags and are looking to reuse it?

I have a set of rules for generating [SBOMs](https://www.cisa.gov/sbom) in [CycloneDX format](https://cyclonedx.org/specification/overview/) from the build graph. The projects using this already apply this patch to `bazel-gazelle` in order to extract this information. 

> Have you looked at any of the other efforts in the space? The one I'm the most aware of is [rules_license](https://github.com/bazelbuild/rules_license/blob/main/examples/vendor/constant_gen/BUILD#L14-L23). This project allows you to effectively attach a new attribute license on any target inside of a repo. You can attach arbitrary metadata to it there, but the list of already supported attributes contains: package_name, package_url, and package_version. If that information was available for any target inside the repository, would that satisfy your SBOM requirements? Here is an [example output](https://github.com/bazelbuild/rules_license/blob/main/examples/vendor/constant_gen/generator_licenses.golden) from their included rule.

While the `license` rule doesn't currently support the same mechanisms for specifying versions that `go_repository` supports, I imagine that could be added. However, I have some concerns:

1. The `license` target would need to be mandatory attribute on every target generated by `go_repository`.
2. Version information is already known, but there's no guarantee that a valid license can be determined. There are three options: 
  - Omitting version information because we can't figure out the license would lose information unnecessarily.
  - Putting in a fake license could lead users to a false level of comfort.
  -  Omitting license information in a `license` rule feels like the rule isn't being used properly.
3. I think that the above demonstrate that "license" and "version" are two separate concepts, and mingling them with a `license` rule is confusing. 

To expand a little on point `1`, when generating an SBOM, we want to take a target rule (so a `go_binary` or a `go_image` and walk the dependency tree. This means that there's a requirement for us to be able to extract source and version information from the build graph, rather than inferring it in some other way. This means that the `license` must be properly set on all targets generated by `go_repository`.

shs96c:
I should also say that license conformance tends to fall under the purview of the legal team in many organisations, and gathering and acting on that information may be mandated in different ways by different companies. Because of this, it may not be possible for engineers to use `rules_license` in their projects.

achew22:
@shs96c, thanks for responding. I appreciate the backstory on how you got here and what tooling is already available. Great context!

> Being usable with earlier versions is a "nice to have", but not a hard requirement. Depending on Bazel 6, however, is untimely.

I poked around a little bit more and I think the lowest limit would be 5.0 if we were to really consider the `rules_license` approach, so no worries around hitting 6 compatibility.

> While the license rule doesn't currently support the same mechanisms for specifying versions that go_repository supports, I imagine that could be added. However, I have some concerns:

It can indeed be added and was [added](https://github.com/bazelbuild/rules_license/pull/41) last week. You have exceptional timing in bringing up this conversation 😉 . 

Apologies, one of the things I can't ever remember how to do in markdown is quote someone's numbered list and have the numbers stay the same. I promise I tried and these are in order, but they will probably render wrong. Sorry!

> 1. The license target would need to be mandatory attribute on every target generated by go_repository.

The license targe would need to be mandatory in the same way that a tag would need to be applied to every target inside the repo. It could be trivially added using something (very similar to what you're doing with tags) like:

```sh
buildozer 'add default_applicable_licenses //:license' //...:__pkg__
```

Then creating a [`license`](https://github.com/bazelbuild/rules_license/blob/main/rules/license.bzl#L87) target (which can omit any field would be) in the most basic case:

```bzl
load("@rules_license//rules:license.bzl", "license")

license(
  name = "license",
  license_kinds = [],
)
``` 

In cases where we know what the package url and version is, you could do:

```bzl
load("@rules_license//rules:license.bzl", "license")

license(
  name = "license",
  license_kinds = [],
  package_url = "https://github.com/bazelbuild/bazel-gazelle",
  package_version = "v0.31.0",
)
```

> 2. Version information is already known, but there's no guarantee that a valid license can be determined. There are three options:
> * Omitting version information because we can't figure out the license would lose information unnecessarily.

Since every field in the license field except `name` is optional, I don't think we're losing information.

> * Putting in a fake license could lead users to a false level of comfort.

To me this is a completely unacceptable outcome and I wouldn't assist in merging this.

> * Omitting license information in a license rule feels like the rule isn't being used properly.
> 3. I think that the above demonstrate that "license" and "version" are two separate concepts, and mingling them with a license rule is confusing.

I'm not sure I necessarily agree with these two points. The [design document](https://docs.google.com/document/d/1uwBuhAoBNrw8tmFs-NxlssI6VRolidGYdYqagLqHWt8/edit#) for `rules_license` explicitly states that they intend to use this for SBOM generation.

I think `rules_license` is a possibly overly specific name. All it really is is a way to staple a dictionary of data to another object which can be accessed via an aspect. One of the fields inside of there is the version and another the origin.

> I should also say that license conformance tends to fall under the purview of the legal team in many organisations, and gathering and acting on that information may be mandated in different ways by different companies. Because of this, it may not be possible for engineers to use rules_license in their projects.

I don't think having this information obligates anyone to anything. This is merely annotating the build graph with an extra edge pointing to a provider that contains an arbitrary key/value pair of data. We can fill it up with anything we like, but it doesn't obligate anyone in any legal department anywhere to do anything with the information.

Conceptually imagine that every target in your repo had an extra `data` dep on `//:metadata` which contained the url and version information. I believe these would be semantically equivalent.

In this sense, legal is no more obligated to do anything with this information in a `license` attribute than having the source URL as a tag attached to the object. In fact in a local repo I just set up for testing the output it I was able to add the `license` target to the root and run this one command and get very satisfying output.

```sh
buildozer 'add default_applicable_licenses //:license' //...:__pkg__; bazel build //:licenses_used && cat bazel-bin/licenses_used.json
```

[repo](https://github.com/achew22/rules_license_demo/), [sample output](https://github.com/achew22/rules_license_demo/blob/main/output.json)

sluongng:
Im just thinking out loud here: For rules_license approach to work with `go_repository` we would need to:

1. Create a gazelle extension that support generating `license()` targets on each `go_repository` archive.

2. Provide a mechanism to insert said extension into `go_repository_tools` / `go_repository_tools_srcs` version of gazelle.

3. Apply the same changes in this PR to `go_repository` to run Gazelle over the `go_repository` downloaded archive with some flags that provide version / commit / url / sha256 information to generate the `license` target with those attributes.

🤔 am I missing something?

shs96c:
>> While the license rule doesn't currently support the same mechanisms for specifying versions that go_repository supports, I imagine that could be added. However, I have some concerns:

> It can indeed be added and was https://github.com/bazelbuild/rules_license/pull/41 last week. You have exceptional timing in bringing up this conversation 😉 .

We're still missing `commit`, `sha256`, `importpath`, and `sum` from the list of versioning information that `go_repository` supports, so _no matter what_, we're going to have to modify the `tags` of the generated rules. It seems heavyweight to do this and also add a dependency on an additional ruleset that isn't strictly required.

For SBOMs to be used for toolchain analysis, the hash of the dependency is a key part of the puzzle.

I'm quite sure that `rules_license` has value, but it doesn't currently address the requirements that this PR does.

shs96c:
On the plus side, the CI builds are green now :)

shs96c:
A gentle ping on this PR. Is the formal request to rewrite this using `rules_license`, or was that just a query about why this approach was being taken?

shs96c:
Is it worth me rebasing this? The last review was in August, but I still need this functionality.

internal/go_repository.bzl
aiuto: FWIW... what we will be doing with rules_licenese is to rewrite the download after patch. We need to be able to rewrite license() rules based on scanning the code after download, and potentially updating a patched BUILD file.

aiuto:
I'ld love to see this use rules_license, especially the new, not yet checked in, stuff. But that should not block moving forward incrementally. I'm a little behind on my gazelle. Well, very behind, as I don't use it, so I can't comment on this PR this week. I need to get a 2 quarter task list for what we are doing in Bazel and rules license. That may inform what to do here.

shs96c:
I agree that it would be nice to be able to use `rules_license` for this, but it seems unlikely that will be ready in a reasonably short timeframe, so it would be good to merge this once it's rebased. Reading your comment, @aiuto, I think you're in agreement since this "moves forward incrementally"

This change is relatively non-invasive, and doesn't preclude switching to a more sophisticated method later (indeed, by using `buildifier`, I think it may lay some of the groundwork for that), but also exposes the information about the dependencies in the same way that `rules_jvm_external` does (via `tags`).

fmeum:
@shs96c Instead of using buildozer as a binary, could we instead have Gazelle depend on its Go package? It already comes with BUILD files and thus shouldn't complicate the bootstrapping situation for Gazelle.

shs96c:
@fmeum, this gets run as part of a repo rule, so I'm not sure we can reliably build binaries --- I think the "download pre-built" approach will be more robust. I'll extract the common bits now, as requested.

shs96c:
Wait a moment! It looks like we can use `go_repository_tools` for this. I shall investigate further.

shs96c:
On investigation, there are multiple codepaths which `go_repository` can use to load a dependency, not all of which go through tooling controlled by this ruleset (eg. the http path will use the bazel downloader). Because of that, it seems best to download the binary and execute it.

shs96c:
A gentle ping on this PR.

aiuto:
+1 to the gentle ping.  I don't maintain gazelle, so I don't want to force merge this.  @fmeum, you appear to be an owner if this. Can you approve and merge?

achew22:
@aiuto, as a maintainer of this project I would ask you not to merge this.

Ultimately buildozer is a CLI wrapper of the [`github.com/bazelbuild/buildtools` library](https://godoc.org/github.com/bazelbuild/buildtools), which we import and [extensively use](https://github.com/search?q=repo%3Abazelbuild%2Fbazel-gazelle%20%22github.com%2Fbazelbuild%2Fbuildtools%22&type=code). I do not see a circumstance where I would be comfortable taking on the additional packaging and distribution burden of shipping another binary with `gazelle`. `gazelle` it is often used to bootstrap people into the Bazel ecosystem. As a result is must be installable via the [standard `go` toolchain](https://github.com/bazelbuild/bazel-gazelle#running-gazelle-with-go). Merging this change would close onramp for the go community, which is something I'm deeply uncomfortable with.

I have already [commented extensively](https://github.com/bazelbuild/bazel-gazelle/pull/1315#issuecomment-1210144307) on the way I would like to see this PR advance. Further, the [Bazel 6.0.0 release](https://github.com/bazelbuild/bazel/releases/tag/6.0.0) seems to indicate this as the path forward:

> `package(default_package_metadata=[...])` is now the preferred alternative to `default_applicable_liceneses`. `default_applicable_liceneses` will be removed in a future release.

The primary objection of `licenses` being overly specific seems to have been mitigated. It seems to me rather daft to build an alternative system that we are going to have to support into the future (given our backward compatibility guarantees).

sluongng:
For whoever want to pick up the task:

It seems like https://github.com/bazelbuild/bazel-gazelle/pull/1343/ added a language extension that would include `package()` into BUILD files. Need additional instrumentation to manage `default_package_metadata` attribute before we could use it in `go_repository`

PR #1317, opened 2022-08-16
Look in call args for loadable symbols

illicitonion:
**What type of PR is this?**

Bug fix

`merger`

Since https://github.com/bazelbuild/bazel-gazelle/pull/1310 we can add
arguments to rules - this PR allows symbols to be loaded for these
arguments.


illicitonion:
@achew22 Could I nudge you at this small PR? :) Thanks!

tfrench-uber (approved):

Issue #1313, opened 2022-08-03
Language implementations cannot configure loads based on config


illicitonion:
## Background / Feature Request

Currently, gazelle has exactly one place where config is allowed to affect per-build-file generated loads, and that's using `map_kind`.

There are use-cases in practice for being able to do this more flexibly from arbitrary language plugins. For instance, say in Java you wanted to generate `java_test` targets most of the time, but if a test is marked as flaky (perhaps by annotating its class with an `@Flaky` annotation, as suggested in https://github.com/bazelbuild/bazel-gazelle/pull/1293#issuecomment-1171525421), you want to generate a `flaky_java_test` target, or if a test is marked as requiring network access (again, noted by annotation), you want to either generate `network_using_java_test` target.

In particular, this may not be a static global property of a plugin; for a shared common Java plugin, different teams may have different conventions requiring custom mappings (e.g. they may define their own target types they want to generate for). In a multi-team monorepo, these may even be per-directory conventions.

In the Java gazelle plugin, we wish to make a flexible configuration mechanism, where someone can in a BUILD file write something like:

```starlark
# gazelle:java_annotation_custom_rule com.example.annotations.Flaky @my_repo//rules:defs.bzl flaky_java_test
```

And to be able to customise what rules are generated as a result.

## The Problem

Unfortunately, a `language.Language` must statically declare up-front all of the kinds/loads it may generate. Languages can be configured globally using flags, but don't have any way of varying the kinds/loads they generate based on the contents of BUILD files, which means teams can't make their own local choices.

## Proposed Solution

Ideally, `language.Language.Kinds()` and `language.Language.Loads()` would take arguments: `c *config.Config, rel *string` to allow per-directory configuring to affect the output of `Kinds()` and `Loads()`.

But this is a breaking change! There are a couple of approaches I can reasonably imagine:
1. Just add the arguments. The functions are actually called in very few places (all of which in Gazelle have the required information to hand), and the migration of "adding two arguments you ignore" for each `language.Language` implementation isn't terribly complex. But it _is_ churn.
2. Add a second interface, and where we call `Loads()`/`Kinds()`, sniff whether the other interface is implemented, and if so, call that instead.

I'd prefer 1, but understand if it's too drastic. But in the long-term, I think we should converge on passing `Config` to these functions.

I'd also be open to any other ways of achieving this same aim, but this was the cleanest I could work out.

## Sample implementation

[...]

PR #1318, opened 2022-08-16
Register and parse flags before calling Kinds/Loads

### LINKED ISSUE #1275: fix-update: Custom extension flags cannot influence Loads() or Kinds()

illicitonion:
**What type of PR is this?**

Feature

all

This allows languages to configure their Kinds/Loads calls based on
command line flags.

Fixes #1275
Closes #1274


tests/loads_from_flag/arguments.txt
```diff
@@ -0,0 +1,2 @@
+-mode=diff
+-custom-load=@some//repo:rules.bzl,custom_rule
```
achew22, line 2: Is it possible to define custom directives that achieve the same effect?

illicitonion: It is not, though I would like it to be - see https://github.com/bazelbuild/bazel-gazelle/issues/1313

linzhp (approved):

Issue #696, opened 2020-01-13
Feature Request: Sorting repository rules


linzhp:
Current `gazelle update-repos` appends new repository rules to the end of the WORKSPACE file or macro. This approach can easily leads to merge conflict in a large repository with many developers, as they are trying to add different external dependencies to the end of the current list.

To reduce the chance of merge conflict, we should sort repository rules in some way, e.g., by name. New repo rules are likely to be inserted at different location of the current list.

There are a few design questions related to this:

## Should we sort both WORKSPACE file and macros?

I prefer sorting macros only, because WORSPACE file often includes rules (http_archive, http_file) that are manually maintained. People may prefer to keep some repo rules at the top because of their importance, e.g., loading rules_go, gazelle, setting up go toolchain. However, I am open to have a switch to turn on the sorting of WORSPACE file, but I definitely want to be able to turn it off for our repo.

## If the existing repo rules are not sorted, should we sort them during next `update-repos` run?

We probably should. As @jayconrod suggested, we can add a `rule.File.Sort` to sort and sync a file. This is related to #694 except this feature request will sort all rules, not just newly imported rules.

@jayconrod Any concerns?

jayconrod:
I agree we should not attempt to sort `WORKSPACE`. There's often some complicated interdependence between repositories declared there, so the order is important.

Sorting rules declared in macros seems less obvious, but Gazelle probably (already) can't update macros where the order is important, so I think it makes sense to always sort. If it turns out to be a problem, we can add a directive or something later to control it.

Making this work will be a little tricky. Inserted statements in the `rule` package have an index associated with them (in the embedded `stmt` struct). They aren't actually inserted until `rule.File.Sync` is called. I think we should have a `rule.File.Sort` method that calls `Sync` first (to handle pending insertions and deletions), then sorts `rule.File.Loads` and `rule.File.Rules`. Each `Swap` operation (in [`sort.Interface`](https://pkg.go.dev/sort?tab=doc#Interface)) should swap not only the high-level loads and rules but also the underlying `bzl.Expr` and indexes.

For the sort order, I think sorting loads by module label and sorting rules by name is good enough. I don't expect we'll need custom sort functions, and we could always add a `SortBy` method if it turns out we do.

[...]

PR #1321, opened 2022-08-25
SortMacro() should also sort the Loads

tfrench-uber:
**What type of PR is this?**

Feature

rule

This PR sorts the `.Loads` in a macro file when `SortMacro()` is called. This is needed because loads do not need to be provided in a specific order, and can get very messy as the number of Load statements gets large.

If the user is calling SortMacros, they already would prefer to have this macro in a clean/tidy state.

Adds additional feature improvement to #696

**Other notes for review**

### DIFF
--- a/rule/rule.go
+++ b/rule/rule.go
@@ -390,16 +390,18 @@ func (f *File) Format() []byte {
 	return bzl.Format(f.File)
 }
 
-// SortMacro sorts rules in the macro of this File. It doesn't sort the rules if
+// SortMacro sorts rules and loads in the macro of this File. It doesn't sort the rules if
 // this File does not have a macro, e.g., WORKSPACE.
 // This method calls Sync internally.
 func (f *File) SortMacro() {
 	f.Sync()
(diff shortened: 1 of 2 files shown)


rule/rule.go
```diff
@@ -390,16 +390,18 @@ func (f *File) Format() []byte {
...
-	if f.function != nil {
-		sort.Stable(byName{f.Rules, f.function.stmt.Body})
-	} else {
+
+	if f.function == nil {
 		panic(fmt.Sprintf("%s: not loaded as macro file", f.Path))
 	}
+	
+	sort.Stable(loadsByName{f.Loads, f.File.Stmt})
+	sort.Stable(byName{f.Rules, f.function.stmt.Body})
```
linzhp, line 404: Can you rename `byName` to `rulesByName` to be consistent with `loadsByName`?

tfrench-uber: Good point, updated. `byIndex` will stay because it can work with any statement type.

PR #1324, opened 2022-08-29
Allow configuring timeout of generation tests

illicitonion:
**What type of PR is this?**

Feature

`internal/generationtest`

Allow configuring timeout of generation tests

I've seen a slow generator, when running with high `--runs_per_test` to detect flaky tests, need more than 2 seconds.

### DIFF
--- a/internal/generationtest/generation_test.go
+++ b/internal/generationtest/generation_test.go
@@ -4,6 +4,7 @@ import (
 	"flag"
 	"path/filepath"
 	"testing"
+	"time"
 
 	"github.com/bazelbuild/bazel-gazelle/testtools"
 	"github.com/bazelbuild/rules_go/go/tools/bazel"
@@ -15,6 +16,7 @@ var (
 		" By default, will use files named BUILD.in as the BUILD files before running gazelle.")
 	buildOutSuffix = flag.String("build_out_suffix", ".out", "The suffix on the expected BUILD.bazel files after running gazelle. Defaults to .out. "+
 		" By default, will use files named BUILD.out as the expected results of the gazelle run.")
+	timeout = flag.Duration("timeout", 2*time.Second, "Time to allow the gazelle process to run before killing.")
 )
 
 // TestFullGeneration runs the gazelle binary on a few example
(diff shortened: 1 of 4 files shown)


internal/generationtest/generation_test.go
```diff
@@ -51,6 +53,7 @@ func TestFullGeneration(t *testing.T) {
 				GazelleBinaryPath:    absoluteGazelleBinary,
 				BuildInSuffix:        *buildInSuffix,
 				BuildOutSuffix:       *buildOutSuffix,
+				Timeout:              time.Duration(*timeoutSeconds) * time.Second,
```
tfrench-uber, line 56: It may make sense to pull this conversion outside of the loop

illicitonion: Moved - thanks!

tfrench-uber (approved):

tfrench-uber (approved):

internal/generationtest/generation_test.go
```diff
@@ -15,6 +16,7 @@ var (
 		" By default, will use files named BUILD.in as the BUILD files before running gazelle.")
 	buildOutSuffix = flag.String("build_out_suffix", ".out", "The suffix on the expected BUILD.bazel files after running gazelle. Defaults to .out. "+
 		" By default, will use files named BUILD.out as the expected results of the gazelle run.")
+	timeoutSeconds = flag.Int("timeout_seconds", 2, "Number of seconds to allow the gazelle process to run before killing.")
```
achew22, line 19: Could we use the built in [duration flag](https://pkg.go.dev/flag#Duration) instead of passing untyped integers? That will allow specifying `2s` or `5m` as valid values, which are self-unit-documenting.

achew22 (changes requested):

internal/generationtest/generation_test.go
illicitonion: Done, but left the bazel interface as still being int seconds - I can switch that to be a string if you'd prefer

illicitonion:
@achew22 I think these comments were addressed?

achew22:
You sure did! Thanks for following up and for your contribution

PR #1333, opened 2022-09-16
language/proto: gen_known_imports creates structs instead of function calls

eric-skydio:
The previous system, which generated calls to label.New(), was creating Go binaries that were over 80MB in size due to all the garbage collector metadata needed.

This PR instead generates struct literals, which reduces that to ~10MB total (in our configuration). There are probably still better ways to include the known imports maps, but this resolves the immediate issue for us.

**What type of PR is this?**

Bug fix

language/proto


eric-skydio:
This PR is an alternative to #1307 which I created before knowing that existed. I suspect (but have not tested) that this approach has _slightly_ lower runtime cost but creates a slightly larger binary, but ultimately am pretty ambivalent which one goes in.

pcj:
I'm OK with either one being merged, but given the semantic change in this one is less than #1307, I'd favor this one and close #1307.  @achew22 WDYT?  Let's get one of them in for sure.

linzhp (approved):
LGTM

linzhp:
@fmeum any idea about the test failure?

fmeum:
@linzhp Looks like the latest Bazel commit build didn't finish or didn't get uploaded. Could you retrigger the pipeline in case it's a temporary issue?

linzhp:
To make the CI repeatable, can we make it pinned to a specific Bazel commit instead of relying on the latest?

fmeum:
> To make the CI repeatable, can we make it pinned to a specific Bazel commit instead of relying on the latest?

We certainly can, but testing Bzlmod against non-HEAD also doesn't make that much sense given how frequent breaking changes to the new feature are - at least this way we learn of them. Alternatively, just ping me when you observe a breakage and let me fix it async in case we're at fault and otherwise ignore the status of that particular pipeline. I'm fine with both.

achew22:
@linzhp I think that's a good request, but given bzlmod's significance, status as a relative neophyte in the community, and that this is the first time I've seen it pop up as a problem, my vote is that we "deal with it later"™️. If we continue to have bzlmod issues then we should bring it up with the team responsible for that instead of an unrelated thread.

@eric-skydio + @pcj, thanks for sending both of your PRs in. I've gotten absolutely side-swiped by life. Apologies for not getting to the other PR earlier.

This PR feels like a pure win where we don't trade any of the compile time guarantees and push the pain into the code generation step instead of a parse step at startup. With that, I'm confident that this will continue to work like the old one did without any change in performance and my bias is to merge this one. So biased in fact that I'm going to hit the merge button.

Thank you to both of you for taking the time to contribute to the project. Legitimately, this thing wouldn't be very useful without contributions like that and I'm so appreciative. Cheers!

eric-skydio:
Massive shoutout to @pcj for that other implementation as well, in many ways it's a "less hacky" fix and may be worth revisiting, especially if these lists ever grow long enough to benefit from compression.

PR #1340, opened 2022-10-05
fix: pass `visibility` attribute for `gazelle` macro to resulting `sh_binary`

### LINKED ISSUE #1339: The `visibility` attribute value for `gazelle` is not passed to the `sh_binary`.

0.27.0

0.35.0

5.3.1

Yes.

MacOS, x86_64

### What did you do?

Added a `visibility` attribute the `gazelle` declaration.

```python
gazelle(
    name = "gazelle_update_repos",
    args = [
        "-build_file_proto_mode=disable_global",
        "-from_file=go.mod",
        "-prune",
    ],
    command = "update-repos",
    visibility = ["//:__subpackages__"],
)
```

I expected to be able to reference the target from a subpackage. However, I was not able to do so.

### What did you see instead?

```sh
ERROR: /Users/chuck/code/aspect-build/silo/update_go_deps/bazel/dependencies/BUILD.bazel:1:10: in sh_binary rule //bazel/dependencies:update_go_deps: target '//:gazelle_update_repos' is not visible from target '//bazel/dependencies:update_go_deps'. Check the visibility declaration of the former target if you think the dependency is legitimate
```

[The `visibility` attribute is not passed to the sh_binary.](https://github.com/bazelbuild/bazel-gazelle/blob/1cab9696f94a11d3468963a46a9058501f76d670/def.bzl#L148-L152)

cgrindel:
Closes #1339.

**What type of PR is this?**

Bug fix

gazelle macro

Passing a `visibility` value to the `gazelle` attribute should apply the `visibility` value to the resulting `sh_binary`.

Fixes #1339.

**Other notes for review**

None


bazel-gh1339_fix_gazelle_visibility_attr
```diff
@@ -0,0 +1 @@
+/private/var/tmp/_bazel_chuck/6e0f3973110f9a4ee9aec86165e58a3d/execroot/bazel_gazelle
```
achew22, line 1: I think this might be a leftover file?

tests/visibility_attr/child/print_gazelle.sh
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bash
+
+# --- begin runfiles.bash initialization v2 ---
```
achew22, line 3: Out of curiosity, why did you implement this in bash? Most (all?) of the testing in the rest of the repo is in go, and it seems to me like it would be a pretty simple test to add that way.

def.bzl
```diff
@@ -149,4 +151,5 @@ def gazelle(name, **kwargs):
         name = name,
         srcs = [runner_name],
         tags = tags,
+        visibility = visibility,
```
achew22, line 154: TBH, if you wanted to just make this single file change and not add any of the other testing I'd be happy to merge it.

achew22 (changes requested):
Hey @cgrindel. Thanks for the fix! Just a couple of comments. Removing the extra file is the only mandatory one.

bazel-gh1339_fix_gazelle_visibility_attr
cgrindel: Oops. Not sure how I did that. 😳

tests/visibility_attr/child/print_gazelle.sh
cgrindel: I can switch it to Go. My brain was thinking rule change.

cgrindel: It just occurred to me that I could just use `build_test` under `child`. Does that work for you?

achew22: Works great for me. Thanks!

achew22:
Thanks again @cgrindel. Great research and fix

PR #1346, opened 2022-10-14
bzlmod: Simplify go_grpc_library support

fmeum:
Also raise the module version to match the latest release.


fmeum:
@shs96c Could you check whether this still works in your example? Could be that protobuf is needed for something else.

Updating the modules in tandem without breaking CI is a mess unfortunately.

shs96c (approved):
Works in the modularised version of `contrib_rules_jvm` I'm working on.

PR #1351, opened 2022-10-18
Add size argument to `gazelle_generation_test`

charlesoconor:
**What type of PR is this?**

Other

`gazelle_generation_test`

This adds the `size` argument to the test. I currently have a very small test case for my custom plugins and want to set the size to small. I could imagine that people have a larger test they want to scale up so they can schedule appropriately.


achew22 (approved):

achew22:
@charlesoconor thank you so much for contributing the the Gazelle community!

PR #1359, opened 2022-10-26
bzlmod: Update version to 0.28.0

fmeum:


linzhp (approved):

linzhp:
Do you think we need to create another release 0.28.1 and bump the module version to 0.28.1 so they are in sync?

fmeum:
This is fine, the `MODULE.bazel` in the repo/source archive is actually ignored by Bazel in favor of the one checked into the registry.

PR #1371, opened 2022-11-14
fix updateStmt makeslice panic

### LINKED ISSUE #1370: panic: insert and delete Rule from empty File
This test case currently triggers a panic in the latest gazelle:

```go
func TestInsertDeleteSync(t *testing.T) {
	old := []byte("")
	f, err := LoadData(filepath.Join("old", "BUILD.bazel"), "", old)
	if err != nil {
		t.Fatal(err)
	}

	foo := NewRule("filegroup", "test")
	foo.Insert(f)
	foo.Delete()
	f.Sync()
	got := strings.TrimSpace(string(f.Format()))
	want := ""
	if got != want {
		t.Errorf("got:\n%s\nwant:%s", got, want)
	}
}
```

```
--- FAIL: TestInsertDeleteSync (0.00s)
panic: runtime error: makeslice: cap out of range [recovered]
	panic: runtime error: makeslice: cap out of range

goroutine 6 [running]:
testing.tRunner.func1.2({0x1221d40, 0x12a7950})
	/usr/local/Cellar/go/1.19.3/libexec/src/testing/testing.go:1396 +0x24e
testing.tRunner.func1()
	/usr/local/Cellar/go/1.19.3/libexec/src/testing/testing.go:1399 +0x39f
panic({0x1221d40, 0x12a7950})
	/usr/local/Cellar/go/1.19.3/libexec/src/runtime/panic.go:884 +0x212
github.com/bazelbuild/bazel-gazelle/rule.updateStmt(0xc00011e2a0, {0x0?, 0x0, 0x0}, {0xc000012090, 0x1, 0x4?}, {0x0, 0x0, 0x0})
...
FAIL	github.com/bazelbuild/bazel-gazelle/rule	0.248s
```

The problem is that updateStmt constructs a negative-value capacity argument in this scenario:

```go
func updateStmt(oldStmt *[]bzl.Expr, inserts, deletes, stmts []*stmt) {
        ...
	newStmt := make([]bzl.Expr, 0, len(*oldStmt)-len(deletes)+len(inserts))
	...
```

pcj:
**What type of PR is this?**

> Bug fix

> all

Fixes a panic in the file.Sync logic.

Fixes #1370

### DIFF
--- a/rule/rule.go
+++ b/rule/rule.go
@@ -357,7 +357,11 @@ func updateStmt(oldStmt *[]bzl.Expr, inserts, deletes, stmts []*stmt) {
 	sort.Stable(byIndex(deletes))
 	sort.Stable(byIndex(inserts))
 	sort.Stable(byIndex(stmts))
-	newStmt := make([]bzl.Expr, 0, len(*oldStmt)-len(deletes)+len(inserts))
+	cap := len(*oldStmt) - len(deletes) + len(inserts)
+	if cap < 0 {
+		cap = 0
+	}
+	newStmt := make([]bzl.Expr, 0, cap)
 	var ii, di, si int
 	for i, stmt := range *oldStmt {
 		for ii < len(inserts) && inserts[ii].index == i {
--- a/rule/rule_test.go
+++ b/rule/rule_test.go
@@ -224,6 +224,24 @@ x_library(name = "bar")
 	}
 }
 
+func TestInsertDeleteSync(t *testing.T) {
+	old := []byte("")
+	f, err := LoadData(filepath.Join("old", "BUILD.bazel"), "", old)
+	if err != nil {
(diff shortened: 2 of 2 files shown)


achew22 (approved):
Thanks for the fix on this one @pcj. Good find!

dnathe4th:
For what it's worth I think a better approach would have been to special case `s.deleted && s.inserted` to continue (i.e skip the following logic) [further up](https://github.com/bazelbuild/bazel-gazelle/pull/1371/files#diff-febcc69e222c8b8ebecd575798af76dee464594c827fb5ae105310adbb1e3660R317-R318). This would correct the bookkeeping so that bounds checking on capacity is not needed.

pcj:
@dnathe4th No opposition here to change it if that's the preferred approach

PR #1373, opened 2022-11-17
Update everything for release prep, add prep-release.sh script

dnathe4th:
**What type of PR is this?**

probably a bad one

all

Update dependencies in preparation for version release, and adds a script to help with release prep


tools/releaser/main.go
```diff
@@ -0,0 +1,289 @@
...
+
+	// update WORKSPACE
+	if verbose {
+		fmt.Println("Writing new WORKSPACE")
+	}
+	directivelessWorkspace := workspaceWithoutDirectives.Bytes()
+	newWorkspace := make([]byte, 0, len(directivelessWorkspace))
+	newWorkspace = append(newWorkspace, directivelessWorkspace[:directiveStart]...)
+	newWorkspace = append(newWorkspace, workspaceDirectivesBuff.Bytes()...)
+	newWorkspace = append(newWorkspace, directivelessWorkspace[directiveStart:]...)
```
linzhp, line 164: You can write to the file one part at a time without concatenating. It would have smaller memory footprint

tools/releaser/main.go
```diff
@@ -0,0 +1,289 @@
...
+		return err
+	}
+
+	workspaceBuffer := bytes.NewBuffer(workspace)
+	workspaceWithoutDirectives := new(bytes.Buffer)
+
+	if verbose {
+		fmt.Println("Preparing temporary WORKSPACE without gazelle directives.")
+	}
+	directiveStart, err := getWorkspaceWithouthDirectives(workspaceBuffer, workspaceWithoutDirectives)
```
linzhp, line 92: Can we always write the `gazelle:repository` directive at the bottom of `WORKSPACE` file? So we don't need to remember `directiveStart`

tools/releaser/main.go
```diff
@@ -0,0 +1,289 @@
...
+	if err = getBuffsFromTmp(depsMaybeBuff, workspaceDirectivesBuff); err != nil {
+		return err
+	}
+
+	// update deps
+	depsFile, err := os.ReadFile("deps.bzl")
+	if err != nil {
+		return err
+	}
+	existingDepsScanner := bufio.NewScanner(bytes.NewBuffer(depsFile))
```
linzhp, line 126:
```suggestion
	depsFile, err := os.Open("deps.bzl")
	if err != nil {
		return err
	}
	existingDepsScanner := bufio.NewScanner(depsFile)
```

If you are going to scan the file line by line, you can avoid reading the whole file into the memory

tools/releaser/main.go
```diff
@@ -0,0 +1,289 @@
...
+	tmpbzl, err := os.ReadFile(_tmpBzl)
+	if err != nil {
+		return err
+	}
+
+	attributeRegex := regexp.MustCompile(`^\s+(name|importpath) = "(.+)",$`)
+
+	var foundDef bool
+	var name string
+	tmpscanner := bufio.NewScanner(bytes.NewBuffer(tmpbzl))
```
linzhp, line 237:
```suggestion
	tmpbzl, err := os.Open(_tmpBzl)
	if err != nil {
		return err
	}

	attributeRegex := regexp.MustCompile(`^\s+(name|importpath) = "(.+)",$`)

	var foundDef bool
	var name string
	tmpscanner := bufio.NewScanner(tmpbzl)
```
Ditto

tools/releaser/main.go
```diff
@@ -0,0 +1,289 @@
...
+	for tmpscanner.Scan() {
+		if bytes.HasPrefix(tmpscanner.Bytes(), []byte("def ")) {
```
linzhp, line 239:
```suggestion
		if bytes.HasPrefix(tmpscanner.Bytes(), []byte("def gazelle_dependencies")) {
```
To be more strict. You can also extract `gazelle_dependencies` to a string and use it here and the `-to_macro` flag above.

You can also consider creating a variable `currentLine = tmpscanner.Text()`, and use `strings.HasPrefix`, `strings.TrimSpace`, etc., so you don't need to keep creating byte slices from strings below.

tools/releaser/main.go
```diff
@@ -0,0 +1,289 @@
...
+	return directiveStart, nil
+}
+
+func getBuffsFromTmp(depsMaybeBuff, workspaceDirectivesBuff *bytes.Buffer) error {
+	tmpbzl, err := os.ReadFile(_tmpBzl)
+	if err != nil {
+		return err
+	}
+
+	attributeRegex := regexp.MustCompile(`^\s+(name|importpath) = "(.+)",$`)
```
linzhp, line 233: You can call Gazelle's [LoadMacroFile](https://pkg.go.dev/github.com/bazelbuild/bazel-gazelle/rule#LoadMacroFile) function to parse the whole `tmpbzl` and iterate through the rules and build the rule data structure. At the end, you can call [Format](https://pkg.go.dev/github.com/bazelbuild/bazel-gazelle/rule#File.Format) to convert it to bytes.

dnathe4th: I originally went down this road but could not figure out how to modify directives using the returned rule.File since the documentation for `Directives` specifies it should not be modified after reading the file. Let me know if I missed an obvious alternative to manipulating this set.

tools/releaser/main.go
```diff
@@ -0,0 +1,315 @@
...
+			return err
+		}
+	}
+
+	workspace, err := os.ReadFile(workspacePath)
+	if err != nil {
+		return err
+	}
+
+	workspaceBuffer := bytes.NewBuffer(workspace)
```
linzhp, line 96:
```suggestion
	workspace, err := os.Open(workspacePath)
	if err != nil {
		return err
	}
        defer workspace.Close()
	workspaceBuffer := bufio.NewScanner(workspace)
```

tools/releaser/main.go
linzhp: This function has 3 parts:

1. Parsing the tmpbzl, for which you can use Gazelle's [LoadMacroFile](https://pkg.go.dev/github.com/bazelbuild/bazel-gazelle/rule#LoadMacroFile) function
2. Building depsMaybe. You can first create an [EmptyFile](https://pkg.go.dev/github.com/bazelbuild/bazel-gazelle/rule#EmptyFile), build `_maybe` rules, [insert](https://pkg.go.dev/github.com/bazelbuild/bazel-gazelle/rule#Rule.Insert) them into the File and [Save](https://pkg.go.dev/github.com/bazelbuild/bazel-gazelle/rule#File.Save) them to disk or [Format](https://pkg.go.dev/github.com/bazelbuild/bazel-gazelle/rule#File.Format) them into `[]byte`.
3. Building `# gazelle:repository` directives. This part is simple enough that I think it's fine with the current approach.

tools/releaser/main.go
```diff
@@ -0,0 +1,270 @@
...
+`)
+		flag.PrintDefaults()
+	}
+
+	flag.Parse()
+
+	workspacePath := path.Join(os.Getenv("BUILD_WORKSPACE_DIRECTORY"), "WORKSPACE")
+	depsPath := path.Join(os.Getenv("BUILD_WORKSPACE_DIRECTORY"), "deps.bzl")
+	_tmpBzl := "tmp.bzl"
+	tmpBzlPath := path.Join(os.Getenv("BUILD_WORKSPACE_DIRECTORY"), _tmpBzl)
```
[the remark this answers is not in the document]
dnathe4th: Ah I assumed for some reason this file needed to stay within the local workspace but looks like not.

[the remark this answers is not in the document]
dnathe4th: Actually I do not believe this is possible because the destination for to_macro must be local - https://github.com/bazelbuild/bazel-gazelle/blob/master/cmd/gazelle/update-repos.go#L229 and cannot back up directories https://github.com/bazelbuild/bazel-gazelle/blob/master/cmd/gazelle/update-repos.go#L66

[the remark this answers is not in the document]
linzhp: Oh, yes, you are right. So just address the `defer os.Remove(tmpBzl.Name())` part then. Although you can call `os.CreateTemp(os.Getenv("BUILD_WORKSPACE_DIRECTORY"), "tmp.*.bzl")`, it makes less sense if you are not using the system temp dir.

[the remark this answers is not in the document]
dnathe4th: I also don't think I can rely on `defer os.Remove` because I need that file to be gone before we do the final gazelle run. Let me see what I can do to be a little safer on errors though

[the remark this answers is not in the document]
linzhp: I see. If there is too much trouble, don't worry about that.

tools/releaser/main.go
```diff
@@ -0,0 +1,304 @@
...
+	return nil
+}
+
+func copyClosure(srcPath, destPath string) func() error {
+	return func() error {
+		src, err := os.Open(srcPath)
+		if err != nil {
+			return err
+		}
+		dest, err := os.OpenFile(destPath, os.O_WRONLY, 0644)
```
linzhp, line 222:
```suggestion
		dest, err := os.Create(destPath)
```
a bit simpler this way, when opening for write only

tools/releaser/main.go
```diff
@@ -0,0 +1,304 @@
...
+		if err != nil {
+			return err
+		}
+		_, err = io.Copy(src, dest)
```
linzhp, line 226:
```suggestion
		_, err = io.Copy(dest, src)
```
https://pkg.go.dev/io#Copy

tools/releaser/main.go
```diff
@@ -0,0 +1,304 @@
...
+	if verbose {
+		fmt.Println("Running final gazelle run, and copying some language specific build files.")
+	}
+	finalizationCommands := []struct {
+		cmd  string
+		args []string
+		then func() error
+	}{
+		{cmd: "bazel", args: []string{"run", "//:gazelle"}},
+		{cmd: "bazel", args: []string{"build", "//language/proto:known_imports"}, then: copyClosure(
```
linzhp, line 184: Optional: you can run `bazel build //language/proto:known_imports //language/proto:known_proto_imports //language/proto:known_go_imports` in one command and copy the 3 files later in their own loop, to avoid starting `bazel build` process 3 times

linzhp (approved):

PR #1390, opened 2022-12-09
Add rule.AttrAssignExpr function

### LINKED ISSUE #1389: Rule API: can't access the assigment statements
The rule.Rule API stores a map of `*bzl.AssignExpr` but you can't get the actual expr.  This means you can't read or write the comments there, which is a common place for users to put comments.

pcj:
**What type of PR is this?**

> Feature

> all

Allows programmatic access to the assignment expressions of a rule.

Fixes #1389

### DIFF
--- a/rule/rule.go
+++ b/rule/rule.go
@@ -827,6 +827,16 @@ func (r *Rule) Attr(key string) bzl.Expr {
 	return attr.RHS
 }
 
+// AttrAssignExpr returns the assignment value of the named attribute. nil is
+// returned when the attribute is not set.
+func (r *Rule) AttrAssignExpr(key string) *bzl.AssignExpr {
+	attr, ok := r.attrs[key]
+	if !ok {
+		return nil
+	}
+	return attr
+}
+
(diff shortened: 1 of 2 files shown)


achew22:
Seems like a reasonable path forward. Could you explain why defining directives as a part of your plugin doesn't achieve what you want? Maybe there is a control surface that we should be adding to Gazelle instead of pushing the complexity on plugin implementors

pcj:
@achew22 directives are fine way of reading configuration data from users in the build file.  

However, if plugin authors want to add comments to generated rules, they have limited options: Use the `rule.AddComment()` API, which makes it really hard to avoid duplicate comments when the rule already exists, or put it on the LHS of the assigment, which is where nobody ever actually writes comments.

pcj:
I could entertain a different function `rule.AttrComments(name string) *bzl.Comments` if that sounds better.

pcj:
Closing in favor of https://github.com/bazelbuild/bazel-gazelle/pull/1394.

PR #1394, opened 2022-12-15
Add rule.AttrComments func

### LINKED ISSUE #1389: Rule API: can't access the assigment statements

pcj:
**What type of PR is this?**

> Feature

> all

Allows programmatic access to the assignment expression comments.

Fixes #1389

### DIFF
--- a/rule/rule.go
+++ b/rule/rule.go
@@ -827,6 +827,17 @@ func (r *Rule) Attr(key string) bzl.Expr {
 	return attr.RHS
 }
 
+// AttrComments returns the Comments struct of the underlying binary assigment
+// expression for the given attribute name. nil is returned when the attribute
+// is not set.
+func (r *Rule) AttrComments(key string) *bzl.Comments {
+	attr, ok := r.attrs[key]
+	if !ok {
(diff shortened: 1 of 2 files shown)


achew22:
@pcj I think this is a lot more in line with what I would have expected. That said, it still has us exporting a non-primitive type from another package into our public API. Not asking you to do any code writing on this front, but curious about a couple of things.

* Is this a situation where we'd benefit from a `rule.AddComment` method? To me that seems even more discoverable as an API surface than `AttrComments`?
* Do you have more precise manipulation you're looking for here? Do you have maybe an example of the before/after that you're looking to get?

kusaeva:
Hi @pcj!
Could you please rebase the PR to make a patch applicable for the latest version of bazel-gazelle?

fmeum:
This was done in #1700

pcj:
Glad to see this was finally merged, thanks.

PR #1396, opened 2022-12-16
repo: opportunistically populate RemoteCache from go.mod

### LINKED ISSUE #1385: language/go: in go_repository, use go.mod as a hint for static dependency resolution

v0.28.0

v0.36.0

5.3.2

yes

darwin / arm64

### What did you do?

I've been trying to get a "hello world" Java workspace working with [`@rules_jvm_contrib//java/gazelle`](https://github.com/bazel-contrib/rules_jvm/tree/main/java/gazelle). That README has some instructions for adding `WORKSPACE` code, and I'm mostly trying to follow that as if I were a beginner.

The rules_jvm extension depends on gRPC. I think it starts a separate Java process in order to parse Java source files, and it communicates with that process over gRPC. So that pulls in a ton of dependencies.

Ideally, I'd like to just paste that code into my `WORKSPACE` file and have everything work without additional configuration. Let's assume this is a pure Java project and I know nothing about Go or Gazelle.

### What did you see instead?

I get a number of missing strict dependency errors like:

```
ERROR: /private/var/tmp/_bazel_jay/d38a4fa22ba3f230f840ba5933fdda32/external/com_github_golang_protobuf/internal/gengogrpc/BUILD.bazel:3:11: GoCompilePkg external/com_github_golang_protobuf/internal/gengogrpc/gengogrpc.a failed: (Exit 1): builder failed: error executing command bazel-out/darwin_arm64-opt-exec-2B5CBBC6-ST-e0d274b3679e/bin/external/go_sdk/builder compilepkg -sdk external/go_sdk -installsuffix darwin_arm64 -src ... (remaining 15 arguments skipped)

Use --sandbox_debug to see verbose messages from the sandbox and retain the sandbox build root for debugging
compilepkg: missing strict dependencies:
        /private/var/tmp/_bazel_jay/d38a4fa22ba3f230f840ba5933fdda32/sandbox/darwin-sandbox/37/execroot/java_tutorial/external/com_github_golang_protobuf/internal/gengogrpc/grpc.go: import of "google.golang.org/protobuf/compiler/protogen"
        /private/var/tmp/_bazel_jay/d38a4fa22ba3f230f840ba5933fdda32/sandbox/darwin-sandbox/37/execroot/java_tutorial/external/com_github_golang_protobuf/internal/gengogrpc/grpc.go: import of "google.golang.org/protobuf/types/descriptorpb"
No dependencies were provided.
Check that imports in Go sources match importpath attributes in deps.
Target @contrib_rules_jvm//java/gazelle:gazelle failed to build
```

If I look in the generated `BUILD` file for `@com_github_golang_protobuf//internal/gengogrpc`, I see no dependency on `@org_golang_google/protobuf//compiler/protogen`.

This `go_repository` was declared in `contrib_rules_jvm_gazelle_setup`. It has no explicitly set `build_external` attribute, so it defaults to `static`. Since I have no `go_repository` rules in my top-level `WORKSPACE` and no `# gazelle:repository` directives, Gazelle basically can't resolve any external dependencies.

### Proposal

When `-external=static` is used, Gazelle needs to map imports outside the current module to Bazel labels *without using the network*. It's difficult to do this without knowing which part of an import path corresponds to the module and which part is a directory inside the module.

I propose Gazelle gets that information `go.mod` in the module's top-level directory. That should contain requirements for all modules providing directly imported packages (i.e., the imports we want to resolve). We can statically transform those module paths into Bazel workspace names (so `google.golang.org/protobuf` gets transformed to `org_golang_google_protobuf`).

If a repository name is already known from the `go_repository` configuration file, that should take precedence, allowing for custom repository names. I only mean for this new mechanism to be a fallback, increasing the success rate of `static` dependency resolution.

jayconrod:
**What type of PR is this?**

> Bug fix

> language/go
> repo

Opportunistically populates `RemoteCache` from `go.mod` if the Go extension is being used and a `go.mod` file is in the repository root. This doesn't override roots added from `WORKSPACE` (or other configuration file). It should allow static external dependency resolution to succeed more often in `go_repository` when other `go_repository` declarations are outside the main `WORKSPACE` file, for example, in another dependency's macro.

Fixes #1385

**Other notes for review**

This unfortunately introduces more Go-specific code to `//cmd/gazelle`. I don't think there's a good way to do it without refactoring `//repo` into `//language/go` first, and that's a non-trivial amount of work.

### DIFF
--- a/cmd/gazelle/fix-update.go
+++ b/cmd/gazelle/fix-update.go
@@ -376,6 +376,9 @@ func runFixUpdate(wd string, cmd command, args []string) (err error) {
 			err = cerr
 		}
 	}()
+	if err := maybePopulateRemoteCacheFromGoMod(c, rc); err != nil {
+		log.Print(err)
+	}
 	for _, v := range visits {
 		for i, r := range v.rules {
 			from := label.New(c.RepoName, v.pkgRel, r.Name())
@@ -562,6 +565,41 @@ func findOutputPath(c *config.Config, f *rule.File) string {
 		return defaultOutputPath
 	}
 	return outputPath
+}
+
+// maybePopulateRemoteCacheFromGoMod reads go.mod and adds a root to rc for each
+// module requirement. This lets the Go extension avoid a network lookup for
+// unknown imports with -external=external, and it lets dependency resolution
+// succeed with -external=static when it might not otherwise.
+//
+// This function does not override roots added from WORKSPACE (or some other
+// configuration file), but it's useful when there is no such file. In most
+// cases, a user of Gazelle with indirect Go dependencies does not need to add
+// '# gazelle:repository' or '# gazelle:repository_macro' directives to their
+// WORKSPACE file. This need was frustrating for developers in non-Go
+// repositories with go_repository dependencies declared in macros. It wasn't
+// obvious that Gazelle was missing these.
+//
+// This function is regrettably Go specific and does not belong here, but it
+// can't be moved to //language/go until //repo is broken up and moved there.
+func maybePopulateRemoteCacheFromGoMod(c *config.Config, rc *repo.RemoteCache) error {
+	haveGo := false
+	for name := range c.Exts {
+		if name == "go" {
+			haveGo = true
+			break
+		}
+	}
+	if !haveGo {
+		return nil
+	}
+
+	goModPath := filepath.Join(c.RepoRoot, "go.mod")
+	if _, err := os.Stat(goModPath); err != nil {
+		return nil
+	}
+
+	return rc.PopulateFromGoMod(goModPath)
 }
 
 func unionKindInfoMaps(a, b map[string]rule.KindInfo) map[string]rule.KindInfo {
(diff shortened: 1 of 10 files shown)


linzhp:
Have you thought about adding a directive or command line flag to `fix-update` to provide the path to `go.mod`? That way we don't need to guess the location.

jayconrod:
> Have you thought about adding a directive or command line flag to fix-update to provide the path to go.mod? That way we don't need to guess the location.

I think for situations where this feature is useful, the `go.mod` file should always be in the workspace root directory.

Namely, I want this to work in `go_repository` when the author of the main `WORKSPACE` file might not be aware they have Go dependencies at all (because they are hidden inside a dependency macro). At this point, `go_repository` should nearly always be working in module mode; no point in supporting GOPATH mode any longer. In module mode, a `go.mod` can only appear in the module's root directory. A `go.mod` file in a subdirectory marks the root of another module, so that whole subdirectory is excluded from the module zip file.

For the main workspace, I'd encourage folks to add `# gazelle:repository_macro` or `# gazelle:repository` directives to the main `WORKSPACE` file instead of customizing the `go.mod` location, since those already work well and are more general.

linzhp (approved):

PR #1401, opened 2022-12-28
bzlmod: Add `dev_dependency` attribute to tags

fmeum:
The `dev_dependency` attribute on the `module` and `from_file` tags mimics the same attribute on `bazel_dep`, making it so that dev dependencies of non-root modules do not influence dependency resolution. In this way, Bazel modules can mark Go dependencies that are only used for e.g. tests, dev tooling or packaging tools and should not introduce overhead for dependent modules.

Also raises the minimum Bazel version for Bzlmod to 6.0.0, which is the first release in which it is no longer experimental.


fmeum:
Requires a change in Bazel to be useful.

fmeum:
@linzhp @achew22 I found a way to make this work without changes to Bazel.

fmeum:
I missed that there may be a Bazel-provided solution for this. Sorry for the confusion.

Issue #1407, opened 2023-01-06
testtools for ginkgo (and benchmark)


holyhope:
As a developer, I would like to test my custom bazelle lang that I wrote thanks to [`language.Language` package](https://pkg.go.dev/github.com/bazelbuild/bazel-gazelle/language#Language).

I found the [`/testtools`](https://github.com/bazelbuild/bazel-gazelle/tree/master/testtools) package that may help me.
The `testtool` package is only compatible with `testing.T` whereas I would like to use [`ginkgo`](https://onsi.github.io/ginkgo/).

I suggest to open the compatibility of `testtools` to ginkgo, but also benchmark tests (`testing.B`).

### What version of gazelle are you using?

v0.27.0

### What version of rules_go are you using?

v0.37.0

### What version of Bazel are you using?

v5.2.0

### Does this issue reproduce with the latest releases of all the above?

Yes

holyhope:
The more I try to use that, the more I would like a tests framework for gazelle language.
My first suggestion does not seems to be the right solution.

wdyt?

PR #1408, opened 2023-01-06
feat(testtool) Add support to ginkgo library

holyhope:
Hello,

Related to https://github.com/bazelbuild/bazel-gazelle/issues/1407
Here is an implementation of testtools supporting benchmarks and ginkgo library.

wdyt?

### DIFF
--- a/testtools/files.go
+++ b/testtools/files.go
@@ -57,11 +57,32 @@ type FileSpec struct {
 	NotExist bool
 }
 
+// Use an interface to allow tests to use *testing.T or *testing.B or external libraries like ginkgo.
+type TB interface {
+	Helper()
+	TempDir() string
+
+	FailNow()
+	Failed() bool
+
+	Log(args ...interface{})
+	Logf(format string, args ...interface{})
+	Error(args ...interface{})
+	Errorf(format string, args ...interface{})
+	Fatal(args ...interface{})
+	Fatalf(format string, args ...interface{})
+}
+
+var (
+	_ TB = (*testing.T)(nil)
+	_ TB = (*testing.B)(nil)
+)
+
 // CreateFiles creates a directory of test files. This is a more compact
 // alternative to testdata directories. CreateFiles returns a canonical path
 // to the directory and a function to call to clean up the directory
 // after the test.
-func CreateFiles(t *testing.T, files []FileSpec) (dir string, cleanup func()) {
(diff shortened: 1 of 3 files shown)


testtools/files.go
```diff
@@ -188,191 +209,189 @@ var (
...
-func TestGazelleGenerationOnPath(t *testing.T, args *TestGazelleGenerationArgs) {
-	t.Run(args.Name, func(t *testing.T) {
-		t.Helper() // Make the stack trace a little bit more clear.
-		if args.BuildInSuffix == "" {
-			args.BuildInSuffix = ".in"
-		}
-		if args.BuildOutSuffix == "" {
-			args.BuildOutSuffix = ".out"
+func TestGazelleGenerationOnPath(t TB, args *TestGazelleGenerationArgs) {
+	t.Helper() // Make the stack trace a little bit more clear.
```
achew22, line 213: It looks like quite a bit more stuff in this PR than changing the test to take a `TB` interface. If you'd like to change the inner workings of this file I would like to do it in a different PR than the one where we expand the accepted interface here.

achew22 (changes requested):
Defining a `TB` interface and taking that as an argument seems fine to me, but I'm not sure I understand the `ginkgo` connection. Could you help elaborate on what ginkgo has to do with that?

holyhope:
@achew22 
> Defining a `TB` interface and taking that as an argument seems fine to me, but I'm not sure I understand the `ginkgo` connection. Could you help elaborate on what ginkgo has to do with that?

Ginkgo does not provide access to the original `testing.T` of the current test, only a proxy to it.
From <https://onsi.github.io/ginkgo/#third-party-integrations>:
>  Unfortunately there is no formal interface for *testing.T however Ginkgo provides a function, GinkgoT() that returns a struct that implements all the methods that *testing.T implements. Most libraries accept the *testing.T object via an interface and you can usually simply pass in GinkgoT() and expect the library to work.

testtools/files.go
holyhope: I reverted the changes, but then I am not able to have a common interface for `testing.T` (`Run(name string, f func(t *T)) bool`) and `testing.B` (`Run(name string, f func(b *B)) bool`).

So I kept the `TestGazelleGenerationOnPath` function as it was (`func(t *testing.T, args *TestGazelleGenerationArgs)`)

I could imagine something using generics but it brings some complexity that we maybe do not want?

PR #1413, opened 2023-01-12
fix: add missing Starlark dependency and introduce `bzl_test`

cgrindel:
**What type of PR is this?**

Bug fix

Starlark dependencies

When trying to generate stardoc for [cgrindel/swift_bazel](https://github.com/cgrindel/swift_bazel), I received a missing deps error for `//internal/generationtest:generationtest.bzl`. This PR adds the missing dependency and introduces `bzl_test`. The `bzl_test` ensures that `bzl_library` dependencies are fully specified on one or more targets. I introduced [a similar version of this test in rules_swift](https://github.com/bazelbuild/rules_swift/pull/795/files).

NA

**Other notes for review**


tests/bzl_test.bzl
```diff
@@ -0,0 +1,41 @@
+"""Macro for Ensuring Starlark Dependencies are Specified Properly"""
+
+load("@bazel_skylib//:bzl_library.bzl", "bzl_library")
+load("@io_bazel_stardoc//stardoc:stardoc.bzl", "stardoc")
+
+def bzl_test(name, src, deps):
```
achew22, line 6: Is this to be used by consumers of gazelle or by gazelle authors only? If it is only for consumers of gazelle, this feels like a useful macro that might be better suited for the Stardoc repo. If this is only for internal use (it does seem useful btw), could we do something to clarify that like marking the visibility on it or annotating the docs? WDYT?

cgrindel: Frankly, I think this capability should be built into `bzl_library`. The real fix is to change `bzl_library` to perform these checks. I haven't tried to codify this into a published macro because it requires stardoc. 

The implementation in this PR is only relevant for gazelle. I am open to whatever you think will make it's purpose clear. I explicitly specified the visibility on the declaration. I also added another paragraph trying to provide some additional context.

achew22: Is it relevant to Gazelle consumers or the maintainers of Gazelle? Sorry, I'm still unclear about this.

cgrindel: Both. As a Gazelle consumer, stardoc choked because the gazelle target `//:def` was missing a dependency. It is relevant for the Gazelle maintainers because the build definition was incorrect. The `bzl_test` will help the Gazelle maintainers know that the Starlark build tree is correct. A Gazelle consumer will not use the `bzl_test`. 

I apologize if I am not clear. If it helps, I would be glad to jump on a quick video call to discuss.

achew22: I think fixing the dependency issue is definitely a good idea. I'm just unsure about adding a testing library that uses `bzl_library` and `stardoc` to accomplish this to the `gazelle` repo. I'm open to discussing why you think this should live here, but I feel that this test is a good one to put in the `stardoc` repo

cgrindel: I have removed `bzl_test` from this PR.

PR #1425, opened 2023-01-24
Apply map_kind to existing rules as well as new ones

### LINKED ISSUE #1391: map_kind should apply equally to all rules regardless of pre-existing state
The current behaviour around `map_kind` and existing rules is quite subtle and brittle. Much more so when a `map_kind` is used to just load a rule from another file, and not to rename it (e.g. `# gazelle:map_kind go_binary go_binary //tools/go:def.bzl`), and particularly around targets that gazelle _doesn't_ maintain.

I tried to fix a specific bug, and found that a lot of the behaviour is surprising, so wasn't sure which direction to make the bugfix.

I've pushed pushed a bunch of test-cases showing the current behaviours of `map_kind`. For existing hand-written targets unrelated to gazelle, gazelle will sometimes apply a `map_kind` that wasn't applied before, will sometimes _unapply_ a `map_kind` that _was_ manually applied before, or will sometimes preserve whatever existed before. See the top commit of this branch: https://github.com/illicitonion/bazel-gazelle/tree/map-kind-edge-case-tests

I believe that the correct behaviour should either be:
1. Always apply `map_kind` transformations in all BUILD files gazelle touches (even if that changes existing rules' kinds).
2. For existing rules that gazelle _isn't_ generating, always preserve the target kind that existed before gazelle got involved (which is fiddly, because it may require re-loading a rule kind with a unique new name), and for any rule that gazelle _is_ generating, apply any relevant `map_kind`.

I pretty strongly favour 1. I can't think of cases where you'd want to `map_kind` for gazelle but _not_ for non-gazelle, but if they exist, I think the BUILD file author should rename their manual import to achieve this.

In particular, I believe this test case is incorrect behaviour: https://github.com/bazelbuild/bazel-gazelle/blob/5e06b9479bc41f57d5730962feabf25dda938461/cmd/gazelle/integration_test.go#L2753-L2766 - if gazelle generates a rule, and there's a `map_kind` in scope for it, the `map_kind` _should_ be applied, regardless of what load was present before.

Before I put together a PR fixing my particular bug, I wanted to get people's opinions on what the true behaviour should be here.

illicitonion:
**What type of PR is this?**

Bug fix

cmd/gazelle

Applies map_kind directives to existing rules as well as new ones.

See https://github.com/bazelbuild/bazel-gazelle/issues/1391 for discussion.

Fixes #1391

**Other notes for review**

PR has two commits - the first is adding tests showing the existing behaviour, and the second is changing the behaviour and updating the tests.

### DIFF
--- a/cmd/gazelle/fix-update.go
+++ b/cmd/gazelle/fix-update.go
@@ -378,14 +378,16 @@ func runFixUpdate(wd string, cmd command, args []string) (err error) {
 		}
 	}
 
-	if len(errorsFromWalk) > 0 {
-		if len(errorsFromWalk) == 1 {
-			return errorsFromWalk[0]
-		}
+	if len(errorsFromWalk) == 1 {
+		return errorsFromWalk[0]
+	}
+
+	if len(errorsFromWalk) > 1 {
 		var additionalErrors []string
 		for _, error := range errorsFromWalk[1:] {
 			additionalErrors = append(additionalErrors, error.Error())
 		}
+
 		return fmt.Errorf("encountered multiple errors: %w, %v", errorsFromWalk[0], strings.Join(additionalErrors, ", "))
 	}
 
--- a/cmd/gazelle/fix-update.go
+++ b/cmd/gazelle/fix-update.go
@@ -332,7 +332,7 @@ func runFixUpdate(wd string, cmd command, args []string) (err error) {
 		for _, r := range allRules {
 			repl, err := lookupMapKindReplacement(c.KindMap, r.Kind())
 			if err != nil {
-				errorsFromWalk = append(errorsFromWalk, fmt.Errorf("couldn't handle loop in map_kinds: %w", err))
+				errorsFromWalk = append(errorsFromWalk, fmt.Errorf("looking up mapped kind: %w", err))
 				continue
 			}
 			if repl != nil {
--- a/cmd/gazelle/fix-update.go
+++ b/cmd/gazelle/fix-update.go
@@ -322,7 +322,13 @@ func runFixUpdate(wd string, cmd command, args []string) (err error) {
 			mappedKinds    []config.MappedKind
 			mappedKindInfo = make(map[string]rule.KindInfo)
 		)
-		for _, r := range gen {
+		// We apply map_kind to all rules, including pre-existing ones.
+		var allRules []*rule.Rule
+		allRules = append(allRules, gen...)
+		if f != nil {
+			allRules = append(allRules, f.Rules...)
+		}
+		for _, r := range allRules {
 			if repl, ok := c.KindMap[r.Kind()]; ok {
 				mappedKindInfo[repl.KindName] = kinds[r.Kind()]
 				mappedKinds = append(mappedKinds, repl)
--- a/cmd/gazelle/fix-update.go
+++ b/cmd/gazelle/fix-update.go
@@ -330,12 +330,12 @@ func runFixUpdate(wd string, cmd command, args []string) (err error) {
 			allRules = append(allRules, f.Rules...)
 		}
 		for _, r := range allRules {
-			repl, ok, err := lookupMapKindReplacement(c.KindMap, r.Kind())
+			repl, err := lookupMapKindReplacement(c.KindMap, r.Kind())
 			if err != nil {
 				errorsFromWalk = append(errorsFromWalk, fmt.Errorf("couldn't handle loop in map_kinds: %w", err))
 				continue
 			}
-			if ok {
+			if repl != nil {
 				mappedKindInfo[repl.KindName] = kinds[r.Kind()]
 				mappedKinds = append(mappedKinds, *repl)
 				mrslv.MappedKind(rel, *repl)
@@ -434,7 +434,11 @@ func runFixUpdate(wd string, cmd command, args []string) (err error) {
 	return exit
 }
 
-func lookupMapKindReplacement(kindMap map[string]config.MappedKind, kind string) (*config.MappedKind, bool, error) {
+// lookupMapKindReplacement finds a mapped replacement for rule kind `kind`, resolving transitively.
+// i.e. if go_library is mapped to custom_go_library, and custom_go_library is mapped to other_go_library,
+// looking up go_library will return other_go_library.
+// It returns an error on a loop, and may return nil if no remapping should be performed.
+func lookupMapKindReplacement(kindMap map[string]config.MappedKind, kind string) (*config.MappedKind, error) {
 	var mapped *config.MappedKind
 	seenKinds := make(map[string]struct{})
 	seenKindPath := []string{kind}
@@ -442,7 +446,7 @@ func lookupMapKindReplacement(kindMap map[string]config.MappedKind, kind string)
 		if replacement, ok := kindMap[kind]; ok {
 			seenKindPath = append(seenKindPath, replacement.KindName)
 			if _, alreadySeen := seenKinds[replacement.KindName]; alreadySeen {
-				return nil, false, fmt.Errorf("found loop of map_kind replacements: %s", strings.Join(seenKindPath, " -> "))
+				return nil, fmt.Errorf("found loop of map_kind replacements: %s", strings.Join(seenKindPath, " -> "))
 			}
 			seenKinds[replacement.KindName] = struct{}{}
 			mapped = &replacement
@@ -454,10 +458,7 @@ func lookupMapKindReplacement(kindMap map[string]config.MappedKind, kind string)
 			break
 		}
 	}
-	if mapped != nil {
-		return mapped, true, nil
-	}
-	return nil, false, nil
+	return mapped, nil
 }
 
 func newFixUpdateConfiguration(wd string, cmd command, args []string, cexts []config.Configurer) (*config.Config, error) {
--- a/cmd/gazelle/fix-update.go
+++ b/cmd/gazelle/fix-update.go
@@ -272,6 +272,7 @@ func runFixUpdate(wd string, cmd command, args []string) (err error) {
 	// Visit all directories in the repository.
 	var visits []visitRecord
 	uc := getUpdateConfig(c)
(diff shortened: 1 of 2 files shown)


cmd/gazelle/fix-update.go
```diff
@@ -322,7 +322,13 @@ func runFixUpdate(wd string, cmd command, args []string) (err error) {
 			mappedKinds    []config.MappedKind
 			mappedKindInfo = make(map[string]rule.KindInfo)
 		)
-		for _, r := range gen {
+		// We apply map_kind to all rules, including pre-existing ones.
+		var allRules []*rule.Rule
+		allRules = append(allRules, gen...)
+		if f != nil {
+			allRules = append(allRules, f.Rules...)
+		}
```
sluongng, line 330: Question: does the append ordering matter here?

I.e. if in `foo/BUILD` I were to map `a -> b` and in `foo/bar/BUILD` I were to map `b -> c`, what would be generated in `foo/bar/tee/BUILD`?

sluongng (approved):
1 question is pending but the overall change LGTM

To maintainers: Do we need to document this change to end users? I could imagine in some setups, this could cause a massive amount of unexpected updates + rebuild after a version bump.

cmd/gazelle/fix-update.go
illicitonion: That's an interesting question! I think ordering of rules doesn't matter here, but there is an interesting edge-case nonetheless:

For each rule, we look up in a pre-generated config for the `map_kind` directive, which was generated in the ordinary way for config. So the config object for `foo/bar/tee/BUILD` will contain: `a -> b` and `b -> c` but _doesn't_ resolve a transitive `a -> c`. Given this is a static pre-generated config, and the mappings aren't affected by the order the rules are consulted, I don't think there's an ordering problem here.

This means for things we _generate_, if we generate an `a` we'll end up with a `b`, and if we generate a `b` we'll end up with a `c`.

For things which _already existed_, if there was an `a` we'll end up with a `b`, and if there was a `b` we'll end up with a `c`. Which means that if you have an `a`, and run gazelle twice, you'll end up with different results after each generation.

I think if we wanted to fix this, we'd probably add a loop around the `if repl, ok := c.KindMap[r.Kind()]; ok {` to keep mapping until we reach an unmapped rule, or until we detect a loop.

I've pushed https://github.com/bazelbuild/bazel-gazelle/pull/1425/commits/0c06eb2321d1691a14a1d47915ee3255dec0df73 which does this transitive mapping, and which errors if a loop was encountered. I can imagine situations where people may have loops in their `map_kind` directives today, so this _may_ be a breaking change to some.

cmd/gazelle/integration_test.go
```diff
@@ -2910,10 +2910,10 @@ my_library(
 		{
 			Path: "enabled/existing_rules/unmapped/BUILD.bazel",
 			Content: `
-load("@io_bazel_rules_go//go:def.bzl", "go_library")
+load("//tools/go:def.bzl", "my_library")
 
-# An existing rule with an unmapped type is preserved
-go_library(
+# An existing rule with an unmapped type is updated
+my_library(
```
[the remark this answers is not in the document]
illicitonion: Yes, exactly :)

cmd/gazelle/fix-update.go
```diff
@@ -417,6 +434,32 @@ func runFixUpdate(wd string, cmd command, args []string) (err error) {
...
+			}
+			kind = replacement.KindName
+		} else {
+			break
+		}
+	}
+	if mapped != nil {
+		return mapped, true, nil
+	}
+	return nil, false, nil
```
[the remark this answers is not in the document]
illicitonion: Good call - done

illicitonion:
@sluongng I think everything's addressed here?

cmd/gazelle/fix-update.go
```diff
@@ -366,6 +378,17 @@ func runFixUpdate(wd string, cmd command, args []string) (err error) {
...
+		if len(errorsFromWalk) == 1 {
+			return errorsFromWalk[0]
+		}
+		var additionalErrors []string
+		for _, error := range errorsFromWalk[1:] {
+			additionalErrors = append(additionalErrors, error.Error())
+		}
+		return fmt.Errorf("encountered multiple errors: %w, %v", errorsFromWalk[0], strings.Join(additionalErrors, ", "))
+	}
+
```
sluongng, line 391: Nit:

```suggestion
	if len(errorsFromWalk) == 1 {
		return errorsFromWalk[0]
	}

	if len(errorsFromWalk) > 1 {
		var additionalErrors []string
		for _, error := range errorsFromWalk[1:] {
			additionalErrors = append(additionalErrors, error.Error())
		}

		return fmt.Errorf("encountered multiple errors: %w, %v", errorsFromWalk[0], strings.Join(additionalErrors, ", "))
	}

```

cmd/gazelle/fix-update.go
```diff
@@ -411,6 +434,33 @@ func runFixUpdate(wd string, cmd command, args []string) (err error) {
...
+			mapped = &replacement
+			if kind == replacement.KindName {
+				break
+			}
+			kind = replacement.KindName
+		} else {
+			break
+		}
+	}
+	return mapped, nil
```
sluongng, line 461: Nit: simplify by returning early on `else` condition

```suggestion
		replacement, ok := kindMap[kind]
		if !ok {
			break
		}

		seenKindPath = append(seenKindPath, replacement.KindName)
		if _, alreadySeen := seenKinds[replacement.KindName]; alreadySeen {
			return nil, fmt.Errorf("found loop of map_kind replacements: %s", strings.Join(seenKindPath, " -> "))
		}

		seenKinds[replacement.KindName] = struct{}{}
		mapped = &replacement
		if kind == replacement.KindName {
			break
		}

		kind = replacement.KindName
	}

	return mapped, nil
```

sluongng (approved):
Overall things look good to me. I have some minor style suggestions but they are completely optional.

pinging @achew22 @linzhp for a final maintainer review

cmd/gazelle/fix-update.go
```diff
@@ -322,11 +323,22 @@ func runFixUpdate(wd string, cmd command, args []string) (err error) {
 			mappedKinds    []config.MappedKind
 			mappedKindInfo = make(map[string]rule.KindInfo)
 		)
-		for _, r := range gen {
-			if repl, ok := c.KindMap[r.Kind()]; ok {
+		// We apply map_kind to all rules, including pre-existing ones.
+		var allRules []*rule.Rule
+		allRules = append(allRules, gen...)
```
linzhp, line 328:
```suggestion
		allRules := gen
```

cmd/gazelle/fix-update.go
```diff
@@ -322,11 +323,22 @@ func runFixUpdate(wd string, cmd command, args []string) (err error) {
...
+		if f != nil {
+			allRules = append(allRules, f.Rules...)
+		}
+		for _, r := range allRules {
+			repl, err := lookupMapKindReplacement(c.KindMap, r.Kind())
+			if err != nil {
+				errorsFromWalk = append(errorsFromWalk, fmt.Errorf("couldn't handle loop in map_kinds: %w", err))
```
linzhp, line 335: nit: the final error would be something like "couldn't handle loop in map_kinds: found loop of map_kind replacements: xxx", which is quite redundant. You can either append the error without any wrapping, or wrap with the context here like "looking up mapped kind: found loop of map_kind replacements: xxx".

cmd/gazelle/fix-update.go
```diff
@@ -411,6 +436,37 @@ func runFixUpdate(wd string, cmd command, args []string) (err error) {
 	return exit
 }
 
+// lookupMapKindReplacement finds a mapped replacement for rule kind `kind`, resolving transitively.
+// i.e. if go_library is mapped to custom_go_library, and custom_go_library is mapped to other_go_library,
+// looking up go_library will return other_go_library.
+// It returns an error on a loop, and may return nil if no remapping should be performed.
+func lookupMapKindReplacement(kindMap map[string]config.MappedKind, kind string) (*config.MappedKind, error) {
```
linzhp, line 443:
```suggestion
func lookupMapKindReplacement(kindMap map[string]config.MappedKind, kind string) (string, error) {
```
To simplify the interface, you can return the replacement kind here, because the caller of this function can get the result of info from `config.MappedKind`

linzhp (commented):
overall approach looks good to me. Just some stylish comments

cmd/gazelle/fix-update.go
illicitonion: This actually is semantically different code, and can potentially lead to overwriting values if both `allRules` and `gen` end up getting mutated because of the shared buffer. e.g. in this program:

```go
package main

import "fmt"

func main() {
	slice := make([]string, 0, 2)
	slice = append(slice, "original")

	other := slice

	slice = append(slice, "written-to-slice")
	fmt.Printf("slice: %v, other: %v\n", slice, other)
	other = append(other, "written-to-other")
	fmt.Printf("slice: %v, other: %v\n", slice, other)
}
```

the append to `other` ends up overwriting a value in `slice` as well, which we don't want to do here:
```
slice: [original written-to-slice], other: [original]
slice: [original written-to-other], other: [original written-to-other]
```

(While we don't append to `gen` right now, I'd like to guard against this kind of bug in the future, particularly as you only actually hit issues at capacity thresholds, so it's easy to miss in testing)

cmd/gazelle/fix-update.go
illicitonion: I'm not sure on this one - having a fully typed `*config.MappedKind` as a return value feels better documenting as to what the function is doing, and the caller is going to look up the output in `c.KindMap` again anyway, so it feels like returning a `*config.MappedKind` avoids potential errors such as a different KindMap being passed to the function and being used by the caller?

illicitonion:
@linzhp Could I nudge you at this one again? Thanks!

linzhp (approved):

PR #1438, opened 2023-02-03
feature: gazelle wrapper allows to pass environment variables

### LINKED ISSUE #946: Is there a way to configure gazelle to use ssh endpoints instead of http?

v0.20.0

v0.23.3

3.2.0-homebrew

I didn't see any mention of the improvements in the release notes for v0.20.1 and v0.20.2.

Mac v10.15.5.   x86-64

### What did you do?

I don't know if this is a bug or a feature request.  Is there a way to get an ssh prefix automatically on the import path?  

Currently developers have to run `git config --global url."git@github.com:".insteadOf "https://github.com/"` to fix this.

gazelle currently generates this:

```
go_repository(
    name = "com_github_promotedai_schema",
    importpath = "github.com/promotedai/schema",
    sum = "h1:GdqB7w0ZRiRf5HSBXrI/dGwrOy12/KnLs2FkfmNmNBM=",
    version = "v0.0.0-20201024035926-be6153f03f6e",
)
```

I'd like gazelle to generate this:

```
go_repository(
    name = "com_github_promotedai_schema",
    importpath = "ssh://git@github.com/promotedai/schema",
    sum = "h1:GdqB7w0ZRiRf5HSBXrI/dGwrOy12/KnLs2FkfmNmNBM=",
    version = "v0.0.0-20201024035926-be6153f03f6e",
)
```

I expected a flag or argument for gazelle to do this.

### What did you see instead?

No flags.

manuelnaranjo:
In certain cases it may be useful to have predefined values like GONOSUMDB or others passed into the gazelle helper, with this change the caller to helper can pass those values, and make them part of the code tree instead of a mess of environment variables in the command line.

For example this means someone can now add this to their BUILD.bazel and get a properly working gazelle setup against their ssh private repositories:

```
gazelle(
    name = "gazelle",
    env = {
        "GONOSUMDB": "git.somecompany.com",
        "GIT_CONFIG_COUNT": "1",
        "GIT_CONFIG_KEY_0": "url.git@git.somecompany.com:.insteadof",
        "GIT_CONFIG_VALUE_0": "https://git.somecompany.com",
    },
)
```

**What type of PR is this?**

Feature

cmd/gazelle

Since moving to bzlmod this feature got lost, now it can be reenabled.

Fixes #946

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -15,6 +15,9 @@ gazelle(
     name = "gazelle",
     command = "fix",
     gazelle = ":gazelle_local",
+    env = {
+        "FOO": "bar",
+    },
 )
 
 # gazelle_ci is called from CI to verify the repo is up-to-date, see: .bazelci/presubmit.yml
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -15,9 +15,6 @@ gazelle(
     name = "gazelle",
     command = "fix",
     gazelle = ":gazelle_local",
-    env = {
-        "FOO": "bar",
-    },
 )
 
 # gazelle_ci is called from CI to verify the repo is up-to-date, see: .bazelci/presubmit.yml
--- a/def.bzl
+++ b/def.bzl
@@ -45,6 +45,20 @@ DEFAULT_LANGUAGES = [
     Label("//language/go:go_default_library"),
 ]
 
+def _valid_env_variable_name(name):
+    """ Returns if a string is in the regex [a-zA-Z_][a-zA-Z0-9_]*
+
+    Given that bazel lacks support of regex, we need to implement
+    a poor man validation
+    """
+    if not name:
+        return False
+    for i, c in enumerate(name.elems()):
+        if c.isalpha() or c == "_" or (i > 0 and c.isdigit()):
+            continue
+        return False
+    return True
+
 def _gazelle_runner_impl(ctx):
     args = [ctx.attr.command]
     if ctx.attr.mode:
@@ -57,7 +71,11 @@ def _gazelle_runner_impl(ctx):
         args.extend(["-build_tags", ",".join(ctx.attr.build_tags)])
     args.extend([ctx.expand_location(arg, ctx.attr.data) for arg in ctx.attr.extra_args])
 
-    env = "\n".join([ "export %s='%s'" % (x, y) for (x, y) in ctx.attr.env.items() ])
+    for key in ctx.attr.env:
+        if not _valid_env_variable_name(key):
+            fail("Invalid environmental variable name: '%s'" % key)
+
+    env = "\n".join(["export %s=%s" % (x, shell.quote(y)) for (x, y) in ctx.attr.env.items()])
 
     out_file = ctx.actions.declare_file(ctx.label.name + ".bash")
     go_tool = ctx.toolchains["@io_bazel_rules_go//go:toolchain"].sdk.go
--- a/def.bzl
+++ b/def.bzl
@@ -57,6 +57,8 @@ def _gazelle_runner_impl(ctx):
         args.extend(["-build_tags", ",".join(ctx.attr.build_tags)])
     args.extend([ctx.expand_location(arg, ctx.attr.data) for arg in ctx.attr.extra_args])
 
+    env = "\n".join([ "export %s='%s'" % (x, y) for (x, y) in ctx.attr.env.items() ])
+
     out_file = ctx.actions.declare_file(ctx.label.name + ".bash")
     go_tool = ctx.toolchains["@io_bazel_rules_go//go:toolchain"].sdk.go
     substitutions = {
@@ -69,6 +71,7 @@ def _gazelle_runner_impl(ctx):
 """.format(label = str(ctx.label)),
         "@@RUNNER_LABEL@@": shell.quote(str(ctx.label)),
         "@@GOTOOL@@": shell.quote(go_tool.short_path),
+        "@@ENV@@": env,
     }
     ctx.actions.expand_template(
         template = ctx.file._template,
@@ -118,6 +121,7 @@ _gazelle_runner = rule(
         "prefix": attr.string(),
         "extra_args": attr.string_list(),
         "data": attr.label_list(allow_files = True),
+        "env": attr.string_dict(),
         "_template": attr.label(
             default = "//internal:gazelle.bash.in",
             allow_single_file = True,
--- a/internal/gazelle.bash.in
+++ b/internal/gazelle.bash.in
@@ -24,6 +24,8 @@ GAZELLE_LABEL=@@GAZELLE_LABEL@@
 ARGS=@@ARGS@@
 GOTOOL=@@GOTOOL@@
 
+@@ENV@@
+
 # find_runfile prints the location of a runfile in the source workspace,
 # either by reading the symbolic link or reading the runfiles manifest.
 function find_runfile {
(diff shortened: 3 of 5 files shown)


def.bzl
```diff
@@ -57,6 +57,8 @@ def _gazelle_runner_impl(ctx):
         args.extend(["-build_tags", ",".join(ctx.attr.build_tags)])
     args.extend([ctx.expand_location(arg, ctx.attr.data) for arg in ctx.attr.extra_args])
 
+    env = "\n".join(["export %s='%s'" % (x, y) for (x, y) in ctx.attr.env.items()])
```
sluongng, line 60: using `export VAR=VALUE` means that these values will remain in the user's shell after a Bazel's run?

Should we simply do `VAR=VALUE` instead?

sluongng (commented):
1 small nit, but the remaining LGTM. @fmeum @achew22 wdyt?

def.bzl
manuelnaranjo: In *nix you can't modify environment variables from the caller, you can only modify on the current process and it's childs, even on osx you can't not sure about Windows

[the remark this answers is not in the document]
manuelnaranjo: @fmeum so the problem with `isalnum` is that something like `GIT_CONFIG` will not pass and that's a valid name. I'm checking options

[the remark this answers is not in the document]
fmeum: True. If you look up the set of allowed characters, you can also use `x.elems()` and check the characters one by one.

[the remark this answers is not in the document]
manuelnaranjo: Yeah, just updated the PR, can you check it again? I did the stupid for char in string, not the most efficient ever, but also we shouldn't be getting 1M environmental variables for performance to be a concern

manuelnaranjo: @sluongng are you satisfied with my explanation? Can we resolve this thread?

sluongng: Yup, sounds good to me. 👍

def.bzl
```diff
@@ -57,6 +74,12 @@ def _gazelle_runner_impl(ctx):
         args.extend(["-build_tags", ",".join(ctx.attr.build_tags)])
     args.extend([ctx.expand_location(arg, ctx.attr.data) for arg in ctx.attr.extra_args])
 
+    for key in ctx.attr.env:
+        if not _valid_env_variable_name:
```
fmeum, line 78: You aren't calling the function here, you are taking it's truth value (which is always `True`). If you were to call it, I think it would fail since in Starlark strings aren't iterable. You need to use `name.elems()` to get the chars.

def.bzl
```diff
@@ -45,6 +45,23 @@ DEFAULT_LANGUAGES = [
...
+    a poor man validation
+    """
+    for character in name:
+        if (character >= "a" and character <= "z") or \
+           (character >= "A" and character <= "Z") or \
+           (character >= "0" and character <= "9") or \
+           character == "_":
+            continue
+        return False
+    character = name[0]
```
fmeum, line 61: We should fail descriptively if `name` is empty, just check for `if not name` at the very top of this function.

def.bzl
```diff
@@ -45,6 +45,23 @@ DEFAULT_LANGUAGES = [
     Label("//language/go:go_default_library"),
 ]
 
+def _valid_env_variable_name(name):
+    """ Returns if a string is in the regex [a-zA-Z][a-zA-Z0-9_]*
+
+    Given that bazel lacks support of regex, we need to implement
+    a poor man validation
+    """
+    for character in name:
```
fmeum, line 54: Instead of looping, you can use `all(... for c in name.elems())`.

def.bzl
manuelnaranjo: yeah should have gotten more coffee :P

def.bzl
```diff
@@ -45,6 +45,23 @@ DEFAULT_LANGUAGES = [
...
+        if (character >= "a" and character <= "z") or \
```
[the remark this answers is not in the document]
manuelnaranjo: ohh good call!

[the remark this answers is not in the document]
manuelnaranjo: ok I fixed it all

def.bzl
manuelnaranjo: this thing was going against buildifier linter, is the current solution good enough?

def.bzl
```diff
@@ -45,6 +45,21 @@ DEFAULT_LANGUAGES = [
     Label("//language/go:go_default_library"),
 ]
 
+def _valid_env_variable_name(name):
+    """ Returns if a string is in the regex [a-zA-Z][a-zA-Z0-9_]*
+
+    Given that bazel lacks support of regex, we need to implement
+    a poor man validation
+    """
+    if not len(name):
```
fmeum, line 54:
```suggestion
    if not name:
```

def.bzl
```diff
@@ -45,6 +45,21 @@ DEFAULT_LANGUAGES = [
...
+        return False
+    for (index, character) in enumerate(name.elems()):
```
fmeum, line 56:
```suggestion
    for index, character in enumerate(name.elems()):
```

def.bzl
```diff
@@ -45,6 +45,21 @@ DEFAULT_LANGUAGES = [
...
+        if (index == 0 and character.isalpha()) or \
```
fmeum, line 57: I don't think this correctly includes the case where the name starts with `_`. How about `character.isalpha() or (index != 0 and character == "_")`?

Also consider renaming character and index to c and I to shorten the line - for such a simple loop, long names don't add much.

BUILD.bazel
```diff
@@ -14,6 +14,9 @@ load("//:def.bzl", "gazelle", "gazelle_binary")
 gazelle(
     name = "gazelle",
     command = "fix",
+    env = {
```
fmeum, line 17: Is this accidental? Not exactly sure where a test for this would go, but probably not here.

fmeum (commented):
Thanks, just a few comments.

def.bzl
manuelnaranjo: ok good catch then the regex is something like `[a-zA-Z_][a-zA-Z0-9_]*` let me rearrange that

BUILD.bazel
manuelnaranjo: it was on purpose to show how it works, but I can take it off, and figure out how to write a test for it, as it's a new feature

fmeum: That would be great, let me know if you get stuck. To be honest, I'm not very familiar with Gazelle's testing infrastructure myself, but I can always ask the other maintainers.

manuelnaranjo: I think I found how to make a test that would pass, not so sure about one that would fail, we should test env variables starting with numbers, having spaces, or being empty as well as a few valid ones

def.bzl
manuelnaranjo: alright got it fixed and with a test!

BUILD.bazel
manuelnaranjo: It was slightly harder than what I expected, but I finally managed to get it working, testing of rules is such a pain, but we have to do it anyway

manuelnaranjo:
@fmeum I think now we have it all covered

fmeum (approved):
Thanks for going the extra mile and adding comprehensive test coverage!

manuelnaranjo:
Cool! Thanks for merging, it wasn't trivial to test, but will help with future contributions

PR #1442, opened 2023-02-10
Fix `go mod download` output expectation for errors

illicitonion:
**What type of PR is this?**

Bug fix

language/go

Fixes the expected format of a struct emitted by `go mod download`.

See https://go.dev/ref/mod#go-mod-download for the specification.


illicitonion:
@fmeum it looks like you fixed a similar issue in #1301, maybe you could take a look? :)

PR #1443, opened 2023-02-11
Adding more hooks to FinishableLanguage

linzhp:
**What type of PR is this?**
Feature

cmd/gazelle

`DoneGeneratingRules()` is called before resolving dependencies. However, some Gazelle extensions (e.g., [Python](https://github.com/bazelbuild/rules_python/tree/main/gazelle)) needs to start server processes to parse Python files and query standard modules before generating rules, and shutdown those processes after dependency resolution. Adding `BeforeGeneratingRules` and `DoneResolvingDeps` hooks to manage these server processes.

**Other notes for review**
This is a breaking change. Existing implementations of `FinishableLanguage` need to implement those two additional methods.

### DIFF
--- a/cmd/gazelle/fix-update.go
+++ b/cmd/gazelle/fix-update.go
@@ -269,6 +269,12 @@ func runFixUpdate(wd string, cmd command, args []string) (err error) {
 		return err
 	}
 
+	for _, lang := range languages {
+		if finishable, ok := lang.(language.FinishableLanguage); ok {
+			finishable.BeforeGeneratingRules()
+		}
+	}
+
 	// Visit all directories in the repository.
 	var visits []visitRecord
 	uc := getUpdateConfig(c)
@@ -372,6 +378,12 @@ func runFixUpdate(wd string, cmd command, args []string) (err error) {
 		}
 	})
 
+	for _, lang := range languages {
+		if finishable, ok := lang.(language.FinishableLanguage); ok {
+			finishable.DoneGeneratingRules()
+		}
+	}
+
 	if len(errorsFromWalk) == 1 {
 		return errorsFromWalk[0]
 	}
@@ -408,10 +420,9 @@ func runFixUpdate(wd string, cmd command, args []string) (err error) {
 		merger.MergeFile(v.file, v.empty, v.rules, merger.PostResolve,
 			unionKindInfoMaps(kinds, v.mappedKindInfo))
 	}
-
 	for _, lang := range languages {
 		if finishable, ok := lang.(language.FinishableLanguage); ok {
-			finishable.DoneGeneratingRules()
+			finishable.DoneResolvingDeps()
 		}
 	}
 
--- a/cmd/gazelle/fix-update.go
+++ b/cmd/gazelle/fix-update.go
@@ -17,14 +17,17 @@ package main
 
 import (
 	"bytes"
+	"context"
 	"flag"
 	"fmt"
 	"io/ioutil"
 	"log"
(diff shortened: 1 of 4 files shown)


f0rmiga:
How about extending the hooks and having:

- pre-generate
- post-generate
- post-resolve

There are other hook points that we may come up with, but those are the most obvious ones.

linzhp:
@achew22 Do you prefer a breaking change or simply move the call of `DoneGeneratingRules()` down?

cmd/gazelle/fix-update.go
```diff
@@ -372,12 +372,6 @@ func runFixUpdate(wd string, cmd command, args []string) (err error) {
...
 	})
 
-	for _, lang := range languages {
-		if finishable, ok := lang.(language.FinishableLanguage); ok {
-			finishable.DoneGeneratingRules()
-		}
-	}
-
 	if len(errorsFromWalk) == 1 {
 		return errorsFromWalk[0]
```
illicitonion, line 376: This now changes the semantics from "always called" to "only called if the walk didn't return any errors" - I don't have a problem with the call moving later, but it _is_ important to me that it's unconditionally called.

Otherwise, I have no preference for how many hooks or what they look like :) Thanks!

linzhp:
@illicitonion Can you take another look?

achew22:
I have a strong preference for no breaking changes

linzhp:
@achew22 I see only two options to make this not break:
1. like the original version of this PR, move  the calls to `DoneGeneratingRules()` down, but as @illicitonion pointed out, it would change the behavior 
2. add another interface to host the two new methods. The naming would become confusing though. Now we have a `FinishableLanguage` interface with `DoneGeneratingRules`, and have another interface with `BeforeGeneratingRules` and `DoneResolvingDeps`. If we are going this route, we are sacrificing readability for backward-compatibility

Any other options?

illicitonion:
LGTM, thanks! I'm happy to adapt my (one) extension to whatever ends up happening, so I'm not concerned with compatibility (but understand there are compatibility concerns).

language/lang.go
```diff
@@ -99,13 +99,21 @@ type Language interface {
 // FinishableLanguage allows a Language to be notified when Generate is finished
 // being called.
 type FinishableLanguage interface {
+	// Init is called before any calls to GenerateRules
+	// This allows for hooks to be called, for instance to starting a background
+	// server process.
+	Init()
```
achew22, line 105: Let's pass a context here so that requests can be killed when Gazelle shuts down

achew22 (approved):

internal/language/test_filegroup/lang.go
```diff
@@ -59,8 +65,15 @@ var kinds = map[string]rule.KindInfo{
 	},
 }
 
+func (l *testFilegroupLang) Init(ctx context.Context) {
```
achew22, line 68: I thought we were going to rename this to `Before`?

language/lang.go
```diff
@@ -99,13 +100,21 @@ type Language interface {
 // FinishableLanguage allows a Language to be notified when Generate is finished
 // being called.
 type FinishableLanguage interface {
+	// Init is called before any calls to GenerateRules
+	// This allows for hooks to be called, for instance to starting a background
+	// server process.
+	Init(ctx context.Context)
```
achew22, line 106: Same comment here. Am I reviewing the correct PR?

linzhp:
Superseded by #1475

PR #1447, opened 2023-02-15
Allow keep to be explained

stevebarrau:
**What type of PR is this?**

Feature

rule

Allow the # keep comment to take more text after them for the keep author to add a reference to a ticket or to explain the need for the keep.

### DIFF
--- a/rule/rule.go
+++ b/rule/rule.go
@@ -996,7 +996,7 @@ func (r *Rule) sync() {
 func ShouldKeep(e bzl.Expr) bool {
 	for _, c := range append(e.Comment().Before, e.Comment().Suffix...) {
 		text := strings.TrimSpace(strings.TrimPrefix(c.Token, "#"))
-		if strings.HasPrefix(text, "keep") {
+		if text == "keep" || strings.HasPrefix(text, "keep: ") {
 			return true
 		}
 	}
--- a/rule/rule.go
+++ b/rule/rule.go
@@ -996,7 +996,7 @@ func (r *Rule) sync() {
 func ShouldKeep(e bzl.Expr) bool {
 	for _, c := range append(e.Comment().Before, e.Comment().Suffix...) {
 		text := strings.TrimSpace(strings.TrimPrefix(c.Token, "#"))
-		if text == "keep" {
+		if strings.HasPrefix(text, "keep") {
 			return true
 		}
 	}
--- a/rule/rule_test.go
+++ b/rule/rule_test.go
@@ -305,6 +305,13 @@ x_library(name = "x")
 `,
 			want: true,
 		}, {
+			desc: "prefix",
+			src: `
+# keep because of ticket #42
+x_library(name = "x")
+`,
+			want: true,
+		}, {
 			desc: "compact_suffix",
 			src: `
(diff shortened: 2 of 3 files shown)


achew22 (approved):
If no one else has any comments on this by Monday, I'll merge it. I think it's a good idea

stevebarrau:
@achew22 did you receive any comments on this?

rule/rule.go
```diff
@@ -996,7 +996,7 @@ func (r *Rule) sync() {
 func ShouldKeep(e bzl.Expr) bool {
 	for _, c := range append(e.Comment().Before, e.Comment().Suffix...) {
 		text := strings.TrimSpace(strings.TrimPrefix(c.Token, "#"))
-		if text == "keep" {
+		if strings.HasPrefix(text, "keep") {
```
fmeum, line 999: This would also match `# keeping to mess this up` or `# keep this around for the time being` which may be unintentional. 

How about this?

```suggestion
		if text == "keep" || strings.HasPrefix(text, "keep: ") {
```

fmeum (changes requested):

rule/rule_test.go
```diff
@@ -302,6 +302,13 @@ func TestKeepRule(t *testing.T) {
 			src: `
 # keep
 x_library(name = "x")
+`,
+			want: true,
+		}, {
+			desc: "prefix",
+			src: `
+# keep because of ticket #42
```
fmeum, line 310: I think that this needs updating.

stevebarrau: Correct, updated in c0c70e2.

rule/rule_test.go
```diff
@@ -358,6 +365,13 @@ func TestShouldKeepExpr(t *testing.T) {
 			src: `
 # keep
 "s"
+`,
+			want: true,
+		}, {
+			desc: "before with description",
+			src: `
+# keep: we need it for the ninja feature
```
fmeum, line 373: Could you also add negative tests for the undesirable cases I mentioned in my comments?

rule/rule.go
```diff
@@ -996,7 +996,7 @@ func (r *Rule) sync() {
 func ShouldKeep(e bzl.Expr) bool {
 	for _, c := range append(e.Comment().Before, e.Comment().Suffix...) {
 		text := strings.TrimSpace(strings.TrimPrefix(c.Token, "#"))
-		if text == "keep" {
+		if text == "keep" || strings.HasPrefix(text, "keep: ") {
```
fmeum, line 999: Please also document this new capability under https://github.com/bazelbuild/bazel-gazelle#keep-comments.

fmeum (commented):
Thanks, looks good to me from the functional side. I left two comments as I noticed we hadn't updated the docs and had no negative tests. Will  right after.

rule/rule_test.go
stevebarrau: Added in dc7aa9c.

rule/rule.go
stevebarrau: Updated in 708dd8b

fmeum (approved):
Thanks!

PR #1450, opened 2023-02-15
feat(bzlmod): support go.mod replace directives

### LINKED ISSUE #1449: feat(bzlmod): support replace directives in go.mod

0.28.0

0.38.0

Bazel 6

Yes, this feature has not been implemented.

Linux AMD64

### What did you do?

Run `go_deps.from_file(go_mod = "//:go.mod")` in `MODULE.bazel` with a `go.mod` containing replace directives

The dependency to be properly replaced.

### What did you see instead?

Error:
```
INFO: Build option --test_env has changed, discarding analysis cache.
        File "/home/user/.cache/bazel/_bazel_tfrench/b97476d719d716accead0f2d5b93104f/external/gazelle~0.28.0/internal/bzlmod/go_deps.bzl", line 53, column 55, in _go_deps_impl
                additional_module_tags += deps_from_go_mod(module_ctx, from_file_tag.go_mod)
        File "/home/user/.cache/bazel/_bazel_tfrench/b97476d719d716accead0f2d5b93104f/external/gazelle~0.28.0/internal/bzlmod/go_mod.bzl", line 32, column 17, in deps_from_go_mod
                fail("No sum for {}@{} found in {}".format(require.path, require.version, go_sum_path))
```

tyler-french:
**What type of PR is this?**

Feature

internal/bzlmod/go_mod

This PR adds support for`replace` directives in `go.mod` files. There are multiple acceptible formats for `replace` directives:
```
    golang.org/x/net v1.2.3 => example.com/fork/net v1.4.5
    golang.org/x/net => example.com/fork/net v1.4.5 (SUPPORT ADDED)
    golang.org/x/net v1.2.3 => ./fork/net
    golang.org/x/net => ./fork/net
```
For now, we will ONLY allow the second one in the list: replacing a path with a new path and version. Local directories will not be allowed.

Fixes #1449

### DIFF
--- a/internal/bzlmod/go_deps.bzl
+++ b/internal/bzlmod/go_deps.bzl
@@ -9,6 +9,7 @@ load(
     "format_module_file_fixup",
     "format_rule_call",
     "get_directive_value",
+    "with_replaced_or_new_fields",
 )
 
 visibility("//")
@@ -141,6 +142,7 @@ def _go_deps_impl(module_ctx):
     root_versions = {}
     root_fixups = []
     sums = {}
+    replace_map = {}
 
     outdated_direct_dep_printer = print
     for module in module_ctx.modules:
@@ -177,14 +179,16 @@ def _go_deps_impl(module_ctx):
             )
         additional_module_tags = []
         for from_file_tag in module.tags.from_file:
-            module_tags_from_go_mod = deps_from_go_mod(module_ctx, from_file_tag.go_mod)
+            module_tags_from_go_mod, go_mod_replace_map = deps_from_go_mod(module_ctx, from_file_tag.go_mod)
             additional_module_tags += module_tags_from_go_mod
+
+            if module.is_root:
+                replace_map.update(go_mod_replace_map)
 
             # Load all sums from transitively resolved `go.sum` files that have modules.
             if len(module_tags_from_go_mod) > 0:
                 for entry, new_sum in sums_from_go_mod(module_ctx, from_file_tag.go_mod).items():
                     _safe_insert_sum(sums, entry, new_sum)
-
 
         # Load sums from manually specified modules separately.
         for module_tag in module.tags.module:
@@ -251,15 +255,23 @@ def _go_deps_impl(module_ctx):
                 ),
             )
 
-    for path, module in module_resolutions.items():
-        entry = (path, module.raw_version)
-        if entry not in sums:
-            fail("No sum for {}@{} found".format(path, module.raw_version))
+    # All `replace` directives are applied after version resolution.
+    # We can simply do this by checking the replace paths' existence
+    # in the module resolutions and swapping out the entry.
+    for path, replace in replace_map.items():
+        if path in module_resolutions:
+            module_resolutions[path] = with_replaced_or_new_fields(module_resolutions[path],
+                replace = replace.to_path,
+                version = semver.to_comparable(replace.version),
+                raw_version = replace.version,
+            )
 
+    for path, module in module_resolutions.items():
         go_repository(
             name = module.repo_name,
             importpath = path,
-            sum = sums[entry],
+            sum = _get_sum_from_module(path, module, sums),
+            replace = getattr(module, "replace", None),
             version = "v" + module.raw_version,
             build_directives = _get_directives(path, gazelle_overrides),
         )
@@ -301,6 +313,16 @@ one of these attributes and add "gazelle:go_naming_convention <value>" (for \
 build_naming_convention) or "gazelle:proto <value>" (for build_proto_file_mode) \
 to its 'directives' attribute.
 """ + format_module_file_fixup(root_fixups))
+
+def _get_sum_from_module(path, module, sums):
+    entry = (path, module.raw_version)
+    if hasattr(module, "replace"):
+        entry = (module.replace, module.raw_version)
+
+    if entry not in sums:
+        fail("No sum for {}@{} found".format(path, module.raw_version))
+
+    return sums[entry]
 
 def _safe_insert_sum(sums, entry, new_sum):
     if entry in sums and new_sum != sums[entry]:
--- a/internal/bzlmod/go_mod.bzl
+++ b/internal/bzlmod/go_mod.bzl
@@ -33,7 +33,7 @@ def deps_from_go_mod(module_ctx, go_mod_label):
             direct = require.direct,
         ))
 
-    return deps
+    return deps, go_mod.replace_map
 
 def parse_go_mod(content, path):
     # See https://go.dev/ref/mod#go-mod-file.
@@ -47,6 +47,7 @@ def parse_go_mod(content, path):
         "module": None,
         "go": None,
         "require": [],
+        "replace": {},
     }
 
     current_directive = None
@@ -103,6 +104,7 @@ def parse_go_mod(content, path):
         module = module,
         go = (int(major), int(minor)),
         require = tuple(state["require"]),
+        replace_map = state["replace"],
     )
 
 def _parse_directive(state, directive, tokens, comment, path, line_no):
@@ -120,8 +122,21 @@ def _parse_directive(state, directive, tokens, comment, path, line_no):
             version = tokens[1],
             direct = comment != "indirect",
         ))
+    elif directive == "replace":
+        # A replace directive might use a local file path beginning with ./ or ../
+        # These are not supported with gazelle~go_deps.
+        if len(tokens) == 3 and tokens[2][0] == ".":
+            fail("{}:{}: local file path not supported in replace directive: '{}'".format(path, line_no, tokens[2]))
 
-    # TODO: Handle exclude and replace.
+        if len(tokens) != 4 or tokens[1] != "=>":
+            fail("{}:{}: replace directive must follow pattern: 'replace from_path => to_path version' ".format(path, line_no))
+        from_path = tokens[0]
+        state["replace"][from_path] = struct(
+            to_path = tokens[2],
+            version = _canonicalize_raw_version(tokens[3]),
+        )
+
+    # TODO: Handle exclude.
 
 def _tokenize_line(line, path, line_no):
     tokens = []
--- a/tests/bcr/go.mod
+++ b/tests/bcr/go.mod
@@ -1,5 +1,9 @@
-// Only here to stop go mod from descending into this directory.
-
+// This will stop go mod from descending into this directory.
 module github.com/bazelbuild/bazel-gazelle/tests/bcr
 
 go 1.19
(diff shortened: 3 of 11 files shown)


internal/bzlmod/go_deps.bzl
```diff
@@ -109,11 +121,37 @@ def _go_deps_impl(module_ctx):
...
+    for path, replace in replace_map.items():
+        if path in module_resolutions:
+            module = module_resolutions[path]
+            module_resolutions[path] = struct(
+                module = module.module,
+                repo_name = module.repo_name,
+                replace = replace.to_path,
+                version = semver.to_comparable(replace.version),
+                raw_version = replace.version,
+                build_naming_convention = module.build_naming_convention,
```
fmeum, line 136: This pattern is error-prone as we need to enumerate all fields. 

How about introducing a `def with_replaced_fields(_struct, **replacements)` helper function that uses `dir` and `struct` to produce a shallow copy of a struct with the field given as keyword arguments replaced with the given values? That could come in handy in other places.

internal/bzlmod/go_deps.bzl
```diff
@@ -77,15 +92,13 @@ def _go_deps_impl(module_ctx):
...
-            if raw_version.startswith("v"):
-                raw_version = raw_version[1:]
+            raw_version = _canonicalize_raw_version(module_tag.version)
 
             # For modules imported from a go.sum, we know which ones are direct
             # dependencies and can thus only report implicit version upgrades
             # for direct dependencies. For manually specified go_deps.module
             # tags, we always report version upgrades.
-            if getattr(module, "is_root", False) and getattr(module_tag, "direct", True):
+            if not module.is_root and getattr(module_tag, "direct", True):
```
fmeum, line 101:
```suggestion
            if module.is_root and getattr(module_tag, "direct", True):
```

internal/bzlmod/go_deps.bzl
```diff
@@ -109,11 +121,37 @@ def _go_deps_impl(module_ctx):
...
+                replace = replace.to_path,
+                version = semver.to_comparable(replace.version),
+                raw_version = replace.version,
+                build_naming_convention = module.build_naming_convention,
+                build_file_proto_mode = module.build_file_proto_mode,
+            )
+
+
+    # Validate all the modules have valid sums after the version resolution.
+    # Modules that are discarded in resolution do not need to have sums defined
```
fmeum, line 142: That's true, but it makes it sound as if the converse is as well. Isn't it quite possible that a module that isn't discarded in resolution has a sum coming from a non-root module? For example, most Bazel modules using Go won't provide sums for all of Gazelle's Go dependencies.

internal/bzlmod/go_deps.bzl
```diff
@@ -109,11 +121,37 @@ def _go_deps_impl(module_ctx):
...
+    # in the root module.
+    for path, module in module_resolutions.items():
+        entry = (module.replace, module.raw_version) if hasattr(module, "replace") else (path, module.raw_version)
```
fmeum, line 145: Could use a helper function that gets the sum (or None) taking replacement into account. That function could then be used below.

internal/bzlmod/go_mod.bzl
```diff
@@ -126,8 +121,15 @@ def _parse_directive(state, directive, tokens, comment, path, line_no):
             version = tokens[1],
             direct = comment != "indirect",
         ))
+    elif directive == "replace":
+        from_path = tokens[0]
+        state["replace"][from_path] = struct(
+            # tokens[1] will be "=>" token
```
fmeum, line 127: Should we verify that with a `fail`?

internal/bzlmod/go_mod.bzl
```diff
@@ -120,8 +122,16 @@ def _parse_directive(state, directive, tokens, comment, path, line_no):
             version = tokens[1],
             direct = comment != "indirect",
         ))
-
-    # TODO: Handle exclude and replace.
+    elif directive == "replace":
+        from_path = tokens[0]
+        if tokens[1] != "=>":
```
fmeum, line 127: Please also check the expected length of `tokens` here so that users see the error message, not a generic out of bounds error.

internal/bzlmod/go_deps.bzl
```diff
@@ -300,6 +313,44 @@ build_naming_convention) or "gazelle:proto <value>" (for build_proto_file_mode)
...
+
+    Returns:
+        The resulting updated structure.
+    """
+    new_struct_assignments = {}
+
+    # structs have certain built-in methods or attributes that should not
+    # be overwritten. Load these by calling `dir` on an empty struct.
+    # e.g. ["to_proto", "to_json"]
+    protected_attributes = dir(struct())
```
fmeum, line 331: You can make this an unexported global variable.

internal/bzlmod/go_deps.bzl
```diff
@@ -300,6 +313,44 @@ build_naming_convention) or "gazelle:proto <value>" (for build_proto_file_mode)
...
+    for key in dir(_struct):
```
fmeum, line 332: This for loop could just be a dict comprehension.

fmeum (approved):
Thanks for making the changes, looks good to me with only a few comments.

internal/bzlmod/go_mod.bzl
tyler-french: Updated. Good idea!

internal/bzlmod/go_deps.bzl
tyler-french: Yes. I also decided to add a Unit Test for this function since it's functionality is important.

fmeum: While you are at it, could you move it to the new utils.bzl file?

internal/bzlmod/go_deps.bzl
```diff
@@ -11,7 +11,9 @@ load(
     "get_directive_value",
 )
 
-visibility("//")
+visibility([
+    "//tests/bzlmod/...",
```
fmeum, line 15:
```suggestion
    "//",
    "//tests/bzlmod/...",
```

fmeum (approved):

tests/bcr/go.mod
```diff
@@ -1 +1,7 @@
-// Only here to stop go mod from descending into this directory.
\ No newline at end of file
+// This will stop go mod from descending into this directory.
+module github.com/bazelbuild/bazel-gazelle/tests/bcr
+
+go 1.19
+
+// Validate go.mod replace directives can be properly used:
+replace github.com/bmatcuk/doublestar/v4 => github.com/bmatcuk/doublestar/v4 v4.5.0
```
fmeum, line 7: Looks good, could you add an actual dependency on this to the test (e.g., add the repo to `use_repo` and use some functionality in the `go_test`)?

internal/bzlmod/go_mod.bzl
```diff
@@ -201,6 +216,13 @@ def sums_from_go_mod(module_ctx, go_mod_label):
         go_mod_label.package,
         "go.sum",
     ))
+    go_sum_path = module_ctx.path(go_sum_label)
```
fmeum, line 219: I remember seeing `module_ctx.path(<some label>)` fail if the file doesn't exist, before there is even a chance to test for its existence. Did you check whether that's the case?

tests/bcr/pkg/mvs_test.go
```diff
@@ -17,3 +18,11 @@ deps = ["gazelle", "rules_go"]`)
 	require.NoError(t, err)
 	require.Equal(t, []string{"gazelle", "rules_go"}, tree.GetArray("modules.deps"))
 }
+
+func TestReplace(t *testing.T) {
+	// doublestar.StandardOS does NOT exist in doublestar/v4
```
fmeum, line 23: Looks great, thanks for adding this test!

internal/bzlmod/go_mod.bzl
tyler-french: If this is the case, I think it's ok to just keep it as is. It might add too much complexity to try to check if this exists and control how the failure happens.

yanndegat:
hi @tyler-french 

i'm having an issue with this new feature. 

the bazel structure of my repo is
root/mymod/go.mod
```go
replace github.com/lyraproj/hiera => github.com/yanndegat/hiera v0.4.7

require (
	github.com/lyraproj/hiera v0.0.0-00010101000000-000000000000
)
```
root/mymod/MODULE.bazel
```python
module(
    name = "mymod",
    version = "0.0.1",
)
...
...
```
root/MODULE.bazel
```python
...
bazel_dep(name = "mymod", version = "0.0.1")
local_path_override(
    module_name = "mymod",
    path = "./mymod/",
)
...
```

and when i step inside mymod directory and run 
bazel build //... it runs well

but if i build it from the root module such as
bazel build @mymod
then the replace directive in the go.mod seems to be lost as i hit this issue
```
        File "/.../.cache/bazel/_bazel_.../ff35e22318f59da43532c764b4d9e046/external/gazelle~0.33.0/internal/bzlmod/go_deps.bzl", line 390, column 44, in _go_deps_impl
                "sum": _get_sum_from_module(path, module, sums),
        File "/.../.cache/bazel/_bazel_...t/ff35e22318f59da43532c764b4d9e046/external/gazelle~0.33.0/internal/bzlmod/go_deps.bzl", line 439, column 13, in _get_sum_from_module
                fail("No sum for {}@{} found".format(path, module.raw_version))
Error in fail: No sum for github.com/lyraproj/hiera@0.0.0-00010101000000-000000000000 found
```

is this an issue with the feature and i shall log it in github ? or is it because i'm doing it wrong ?

thanks a lot

yanndegat:
it seems that this condition is prenventing the replace to happen if the module is not root

https://github.com/tyler-french/bazel-gazelle/blob/f1b344fc6c18cc1a59c036031565379ea4008efb/internal/bzlmod/go_deps.bzl#L185

fmeum:
This restriction is necessary as there is no obvious way to reconcile `replace` directives across transitive Bazel module dependencies. Having a transitive dependency switch out a dependency is also pretty unexpected and potentially problematic. This behavior is in line with that of regular Go tooling: Transitive `replace` directives [are ignored](https://go.dev/ref/mod#minimal-version-selection).

yanndegat:
hi @fmeum 
thanks a lot for your response
ok i think i get it. maybe my issue then is more with the way i structured my project

i reference the submodule in the root module so i can more easily build / test or run them

maybe there's a simpler way to to it. but i dont know any other way to build/run my submodule
without having to change directory

is there a simpler way to build and run all submodules of a mono repo without having to jump in each one of them ?

PR #1458, opened 2023-02-20
Update File.Sync to propertly position package() rule immediately after load()s

dnathe4th:
**What type of PR is this?**

Bug fix

cmd/gazelle

The documentation explicitly requires a package() rule to occur immediate after the load() calls in a BUILD.bazel file, but before any other rules. There was no enforcement of this within gazelle's file formatting functionality. Rather than requiring future extension authors manipulating package() rules to search for the proper insertion location, this PR updates the central Sync function (which does load() and rule() rearranging already) to position the package() invocation correctly.

**Other notes for review**

### DIFF
--- a/rule/rule.go
+++ b/rule/rule.go
@@ -315,6 +315,17 @@ func (f *File) Sync() {
 	for r, w = 0, 0; r < len(f.Rules); r++ {
 		s := f.Rules[r]
 		s.sync()
+
+		// package() rules are special in that they are REQUIRED to occur
+		// after all load()s and before any other rules. Therefore we
+		// manually update the index of any Rule{name:"package"} we
+		// find to the index of the last load(). This is safe because
+		// "statements at the same index will occur in order" and rules
+		// always occur after load()s in the call to updateStmt below.
+		if s.Kind() == "package" {
+			s.index = len(f.Loads)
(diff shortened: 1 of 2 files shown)


rule/rule.go
```diff
@@ -315,6 +315,17 @@ func (f *File) Sync() {
...
 		s.sync()
+
+		// the package() rule is special in that it is REQUIRED to occur
+		// after all load()s and before any other rules. Therefore we
+		// manually update the index of any Rule{kind:"package"} we
+		// find to the index of the last load(). This is safe because
+		// "statements at the same index will occur in order" and rules
+		// always occur after load()s in the call to updateStmt below.
+		if s.Kind() == "package" {
+			s.index = len(f.Loads)
```
linzhp, line 326: The index of last `load` is `len(f.Loads) - 1`, right? So this will give `package` the same index of the first rule statement.

Also, `index` of some statements are rewritten in `updateStmt`. Will the index of `package` also be rewritten?

dnathe4th:
Though tests pass this PR still has some issues I am working through

keith:
we can close this one, fixed by https://github.com/bazel-contrib/bazel-gazelle/pull/2249

PR #1459, opened 2023-02-21
Extend `labelRepoRegexp` regex pattern

alexander-scott:
**What type of PR is this?**

Bug fix

`label/label.go`

This change allows repo names to begin with a number. It is needed because valid labels which begin with a number are currently returned as invalid by `label.parse()`.

Fixes https://github.com/bazelbuild/bazel-gazelle/issues/1460

**Other notes for review**


fmeum:
Has been resolved at head by a different PR

PR #1462, opened 2023-02-24
feat: allow passing GIT_CONFIG_COUNT and it's related configs

### LINKED ISSUE #946: Is there a way to configure gazelle to use ssh endpoints instead of http?

manuelnaranjo:
**What type of PR is this?**

Feature

go_repository

Sometimes, specially when dealing with internal reposositories, it may be desirable to be able to pass variables into git through go mod tooling, this change allows someone to add this lines to their bazelrc and use ssh instead of https for their internal repositories:

```
build --repo_env=GONOSUMDB=gitserver.somecompany.com
build --repo_env=GIT_CONFIG_COUNT=1
build --repo_env=GIT_CONFIG_KEY_0=url.git@gitserver.somecompany.com:.insteadof
build --repo_env=GIT_CONFIG_VALUE_0=https://gitserver.somecompany.com
```

Fixes #946 

**Other notes for review**

### DIFF
--- a/internal/go_repository.bzl
+++ b/internal/go_repository.bzl
@@ -207,22 +207,23 @@ def _go_repository_impl(ctx):
         "GIT_SSL_CAINFO",
         "GIT_SSH",
         "GIT_SSH_COMMAND",
-        "GIT_CONFIG_COUNT"
+        "GIT_CONFIG_COUNT",
     ]
 
     # Git allows passing configuration through environmental variables, this will be picked
     # by go get properly: https://www.git-scm.com/docs/git-config/#Documentation/git-config.txt-GITCONFIGCOUNT
-    if 'GIT_CONFIG_COUNT' in ctx.os.environ:
-        count = ctx.os.environ['GIT_CONFIG_COUNT']
+    if "GIT_CONFIG_COUNT" in ctx.os.environ:
+        count = ctx.os.environ["GIT_CONFIG_COUNT"]
         if count:
             if not count.isdigit or int(count) < 1:
-                fail("GIT_CONFIG_COUNT has to be a positive over 0 integer")
-            for i in range(int(count)):
-                key = 'GIT_CONFIG_KEY_%d' % i
-                value = 'GIT_CONFIG_VALUE_%d' % i
+                fail("GIT_CONFIG_COUNT has to be a positive integer")
+            count = int(count)
+            for i in range(count):
+                key = "GIT_CONFIG_KEY_%d" % i
+                value = "GIT_CONFIG_VALUE_%d" % i
                 for j in [key, value]:
                     if j not in ctx.os.environ:
-                        fail("%s is not defined as an environmetal variable, but you asked for GIT_COUNT_COUNT=%i" % (j, count))
+                        fail("%s is not defined as an environment variable, but you asked for GIT_COUNT_COUNT=%d" % (j, count))
                 env_keys = env_keys + [key, value]
 
     env.update({k: ctx.os.environ[k] for k in env_keys if k in ctx.os.environ})
--- a/internal/go_repository.bzl
+++ b/internal/go_repository.bzl
@@ -209,11 +209,22 @@ def _go_repository_impl(ctx):
         "GIT_SSH_COMMAND",
         "GIT_CONFIG_COUNT"
     ]
+
+    # Git allows passing configuration through environmental variables, this will be picked
+    # by go get properly: https://www.git-scm.com/docs/git-config/#Documentation/git-config.txt-GITCONFIGCOUNT
     if 'GIT_CONFIG_COUNT' in ctx.os.environ:
         count = ctx.os.environ['GIT_CONFIG_COUNT']
-        if count and count.isdigit() and int(count) > 0:
+        if count:
+            if not count.isdigit or int(count) < 1:
+                fail("GIT_CONFIG_COUNT has to be a positive over 0 integer")
             for i in range(int(count)):
-                env_keys = env_keys + ['GIT_CONFIG_KEY_%d' % i, 'GIT_CONFIG_VALUE_%d' % i]
+                key = 'GIT_CONFIG_KEY_%d' % i
+                value = 'GIT_CONFIG_VALUE_%d' % i
+                for j in [key, value]:
+                    if j not in ctx.os.environ:
+                        fail("%s is not defined as an environmetal variable, but you asked for GIT_COUNT_COUNT=%i" % (j, count))
+                env_keys = env_keys + [key, value]
+
     env.update({k: ctx.os.environ[k] for k in env_keys if k in ctx.os.environ})
 
     if fetch_repo_args:
--- a/internal/go_repository.bzl
+++ b/internal/go_repository.bzl
@@ -207,7 +207,13 @@ def _go_repository_impl(ctx):
         "GIT_SSL_CAINFO",
         "GIT_SSH",
         "GIT_SSH_COMMAND",
+        "GIT_CONFIG_COUNT"
     ]
+    if 'GIT_CONFIG_COUNT' in ctx.os.environ:
+        count = ctx.os.environ['GIT_CONFIG_COUNT']
+        if count and count.isdigit() and int(count) > 0:
+            for i in range(int(count)):
+                env_keys = env_keys + ['GIT_CONFIG_KEY_%d' % i, 'GIT_CONFIG_VALUE_%d' % i]
     env.update({k: ctx.os.environ[k] for k in env_keys if k in ctx.os.environ})
 
     if fetch_repo_args:


fmeum:
I generally agree that passing in git config settings via env variables is useful. 

Having to set three vars for a single config seems a bit much though. Maintaining the correct count also looks error-prone.

If Starlark allows it, maybe `GIT_CONFIG_author.email=foo@bar.com` could work better?

We should also think about whether we want to/need to rerun the repo rule when these variables change.

manuelnaranjo:
> Having to set three vars for a single config seems a bit much though. Maintaining the correct count also looks error-prone.

fair enough, but that's what go it self expects

> If Starlark allows it, maybe `GIT_CONFIG_author.email=foo@bar.com` could work better?

I see your point, so the use case to override the repository url would be something like `GIT_CONFIG_url.git@gitserver.somecompany.com:.insteadof=https://gitserver.somecompany.com` I think it would be doable, is this what you're suggesting?

> We should also think about whether we want to/need to rerun the repo rule when these variables change.
Ok this is a very fair point, IMO we should, as this may change where you look for a given repo. Isn't it enough with changing the environmental variable in use by the repo rule enough for it to be reevaluated?

fmeum:
Sorry, I didn't know that Go itself uses this approach - if that's the case, then there is probably a good reason for it. I also didn't think of the many special characters that my approach would involve and that, as far as I know, aren't legal in names of environment variables.

Let's go with your original approach then, sorry for the misleading feedback.

Unfortunately the only way to have a repository rule rerun when an environment variable changes is to declare it in the `environ` attribute of `repository_rule`. Since the scheme is dynamic we can't do this perfectly, but we can at least go up to a certain number of variables (say 20).

manuelnaranjo:
> Unfortunately the only way to have a repository rule rerun when an environment variable changes is to declare it in the `environ` attribute of `repository_rule`. Since the scheme is dynamic we can't do this perfectly, but we can at least go up to a certain number of variables (say 20).

I don't think this is actually necessary, the use case I have for this variables is you have some company only available go repository, that repository can't be fetched over https only over ssh with git, then you set this variables. Until you haven't set them bazel can't use those dependencies so it will fail, if a new version needs to get downloaded then the go_repository will change it's argument and the repository rule is going to be invalidated anyway.

What do you think?

fmeum:
Sounds convincing!

internal/go_repository.bzl
```diff
@@ -208,7 +208,13 @@ def _go_repository_impl(ctx):
         "GIT_SSL_CAINFO",
         "GIT_SSH",
         "GIT_SSH_COMMAND",
+        "GIT_CONFIG_COUNT"
     ]
+    if 'GIT_CONFIG_COUNT' in ctx.os.environ:
+        count = ctx.os.environ['GIT_CONFIG_COUNT']
+        if count and count.isdigit() and int(count) > 0:
```
fmeum, line 215: Let's add an else branch with a `fail`.

internal/go_repository.bzl
```diff
@@ -208,7 +208,13 @@ def _go_repository_impl(ctx):
         "GIT_SSL_CAINFO",
         "GIT_SSH",
         "GIT_SSH_COMMAND",
+        "GIT_CONFIG_COUNT"
     ]
+    if 'GIT_CONFIG_COUNT' in ctx.os.environ:
```
fmeum, line 213: Could you add a link to the git docs?

manuelnaranjo:
@fmeum I applied both suggestions, do you think we're missing anything now?

internal/go_repository.bzl
```diff
@@ -208,7 +208,24 @@ def _go_repository_impl(ctx):
         "GIT_SSL_CAINFO",
         "GIT_SSH",
         "GIT_SSH_COMMAND",
+        "GIT_CONFIG_COUNT"
     ]
+
+    # Git allows passing configuration through environmental variables, this will be picked
+    # by go get properly: https://www.git-scm.com/docs/git-config/#Documentation/git-config.txt-GITCONFIGCOUNT
+    if 'GIT_CONFIG_COUNT' in ctx.os.environ:
```
fmeum, line 216: Please use double quotes around strings for consistency with the rest of the file.

internal/go_repository.bzl
```diff
@@ -208,7 +208,24 @@ def _go_repository_impl(ctx):
...
+        count = ctx.os.environ['GIT_CONFIG_COUNT']
+        if count:
+            if not count.isdigit or int(count) < 1:
+                fail("GIT_CONFIG_COUNT has to be a positive over 0 integer")
```
fmeum, line 220:
```suggestion
                fail("GIT_CONFIG_COUNT has to be a positive integer")
```

internal/go_repository.bzl
```diff
@@ -208,7 +208,24 @@ def _go_repository_impl(ctx):
...
+            for i in range(int(count)):
+                key = 'GIT_CONFIG_KEY_%d' % i
+                value = 'GIT_CONFIG_VALUE_%d' % i
+                for j in [key, value]:
+                    if j not in ctx.os.environ:
+                        fail("%s is not defined as an environmetal variable, but you asked for GIT_COUNT_COUNT=%i" % (j, count))
```
fmeum, line 226:
```suggestion
                        fail("%s is not defined as an environment variable, but you asked for GIT_COUNT_COUNT=%i" % (j, count))
```

fmeum (commented):
Looks good, just a few stylistic comments and then we can merge.

manuelnaranjo:
@fmeum hey Fabian, all suggestions had been applied, mind to review again?

internal/go_repository.bzl
```diff
@@ -207,7 +207,24 @@ def _go_repository_impl(ctx):
...
+        count = ctx.os.environ["GIT_CONFIG_COUNT"]
+        if count:
+            if not count.isdigit or int(count) < 1:
+                fail("GIT_CONFIG_COUNT has to be a positive integer")
+            for i in range(int(count)):
+                key = "GIT_CONFIG_KEY_%d" % i
+                value = "GIT_CONFIG_VALUE_%d" % i
+                for j in [key, value]:
+                    if j not in ctx.os.environ:
+                        fail("%s is not defined as an environmet variable, but you asked for GIT_COUNT_COUNT=%i" % (j, count))
```
fmeum, line 225:
```suggestion
                        fail("%s is not defined as an environment variable, but you asked for GIT_CONFIG_COUNT=%d" % (j, count))
```

internal/go_repository.bzl
manuelnaranjo: sounds good, I'll do that, maybe I cuold add a unit test for it to make sure it works as expected

fmeum: Unit tests for failures are unfortunately pretty tricky, so I won't make you do that. But manual test runs in the success and failure cases would be great.

manuelnaranjo: yeah I know, I've added a few tests to gazelle already, I'll do it as soon as I get sometime, don't expect this change to make it for this gazelle release anyway

manuelnaranjo: it's not unit tests for failures that are hard, it's unit test for repo rules are hard, didn't realize this is a repo rule, so no way I can write a unit test for it, I'll verify and keep you updated

manuelnaranjo:
```
bazel-gazelle/tests/bcr$ GIT_CONFIG_COUNT=1 bazel build gazelle
INFO: Repository gazelle~override~go_deps~com_github_bazelbuild_buildtools instantiated at:
  callstack not available
Repository rule go_repository defined at:
  /home/mnaranjo/.cache/bazel/_bazel_mnaranjo/4fd2935ef7ec68b38be352a7828e3f9b/external/gazelle~override/internal/go_repository.bzl:313:32: in <toplevel>
INFO: Repository gazelle~override~go_deps~org_golang_x_mod instantiated at:
  callstack not available
Repository rule go_repository defined at:
  /home/mnaranjo/.cache/bazel/_bazel_mnaranjo/4fd2935ef7ec68b38be352a7828e3f9b/external/gazelle~override/internal/go_repository.bzl:313:32: in <toplevel>
ERROR: An error occurred during the fetch of repository 'gazelle~override~go_deps~com_github_bazelbuild_buildtools':
   Traceback (most recent call last):
        File "/home/mnaranjo/.cache/bazel/_bazel_mnaranjo/4fd2935ef7ec68b38be352a7828e3f9b/external/gazelle~override/internal/go_repository.bzl", line 226, column 29, in _go_repository_impl
                fail("%s is not defined as an environment variable, but you asked for GIT_COUNT_COUNT=%d" % (j, count))
Error in fail: GIT_CONFIG_KEY_0 is not defined as an environment variable, but you asked for GIT_COUNT_COUNT=1
ERROR: <builtin>: fetching go_repository rule @gazelle~override//:gazelle~override~go_deps~com_github_bazelbuild_buildtools: Traceback (most recent call last):
        File "/home/mnaranjo/.cache/bazel/_bazel_mnaranjo/4fd2935ef7ec68b38be352a7828e3f9b/external/gazelle~override/internal/go_repository.bzl", line 226, column 29, in _go_repository_impl
                fail("%s is not defined as an environment variable, but you asked for GIT_COUNT_COUNT=%d" % (j, count))
Error in fail: GIT_CONFIG_KEY_0 is not defined as an environment variable, but you asked for GIT_COUNT_COUNT=1
INFO: Repository gazelle~override~go_deps~org_golang_x_sync instantiated at:
  callstack not available
Repository rule go_repository defined at:
  /home/mnaranjo/.cache/bazel/_bazel_mnaranjo/4fd2935ef7ec68b38be352a7828e3f9b/external/gazelle~override/internal/go_repository.bzl:313:32: in <toplevel>
INFO: Repository gazelle~override~go_deps~com_github_pelletier_go_toml instantiated at:
  callstack not available
Repository rule go_repository defined at:
  /home/mnaranjo/.cache/bazel/_bazel_mnaranjo/4fd2935ef7ec68b38be352a7828e3f9b/external/gazelle~override/internal/go_repository.bzl:313:32: in <toplevel>
ERROR: /home/mnaranjo/.cache/bazel/_bazel_mnaranjo/4fd2935ef7ec68b38be352a7828e3f9b/external/gazelle~override/language/go/BUILD.bazel:15:11: @gazelle~override//language/go:go depends on @gazelle~override~go_deps~com_github_bazelbuild_buildtools//build:go_default_library in repository @gazelle~override~go_deps~com_github_bazelbuild_buildtools which failed to fetch. no such package '@gazelle~override~go_deps~com_github_bazelbuild_buildtools//build': GIT_CONFIG_KEY_0 is not defined as an environment variable, but you asked for GIT_COUNT_COUNT=1
ERROR: Analysis of target '//:gazelle' failed; build aborted: 
INFO: Elapsed time: 0.616s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (0 packages loaded, 0 targets configured)
    Fetching repository @gazelle~override~go_deps~com_github_pmezard_go_difflib; starting
```

manuelnaranjo: I fixed a bug in while testing, turns out count was a string and not an int and %d was making it fail, now I believe we're fine to get a final review and merge unless there are further reviews. @fmeum

fmeum (approved):

PR #1463, opened 2023-02-24
Respect repo_name in MODULE.bazel for load generation

fmeum:
**What type of PR is this?**

Feature

cmd/gazelle

In MODULE.bazel, user's can specify custom apparent names for all
repositories and even if they don't, the name of a Bazel module can
differ from its WORKSPACE repository name. For example, the
"rules_go" module used to be imported as a repository called
"io_bazel_rules_go".

This commit adds a language.ModuleAwareLanguage interface that allows
any Gazelle language to be informed about the renamings applied in
MODULE.bazel when asked to generate its Load statements. The go, proto,
and visibility language are updated to implement this new interface.

The module-aware load generation is tested by having the BCR test
module's BUILD files generated by Gazelle.


fmeum:
@achew22 @linzhp After this and https://github.com/bazelbuild/bazel-gazelle/pull/1456 have been merged, I am finally in a position to write a coherent tutorial for rules_go and gazelle with Bzlmod.

cmd/gazelle/update-repos.go
```diff
@@ -136,10 +137,19 @@ func updateRepos(wd string, args []string) (err error) {
...
 
+	moduleToApparentName, err := module.ExtractModuleToApparentNameMapping(c.RepoRoot)
+	if err != nil {
+		return err
+	}
+
 	kinds := make(map[string]rule.KindInfo)
 	loads := []rule.LoadInfo{}
 	for _, lang := range languages {
-		loads = append(loads, lang.Loads()...)
```
linzhp, line 142: Can we simply replace all repos in the `loads` with their apparent names here automatically? Is there any case when an extension don't want to replace repo names with apparent names?

fmeum: I would have liked to do so, but I don't think it's possible without extra information: We somehow need to know that the WORKSPACE repo `io_bazel_rules_go` is equivalent to the `rules_go` module before we can figure out the apparent repo name for that.

Instead of `ApparentLoads`, we could introduce a `ModuleName` function that returns the module name for a given legacy WORKSPACE name. The resulting API would be simpler, but I don't like that the "interface" would still be WORKSPACE names, which are slated to go away.

internal/module/module.go
```diff
@@ -0,0 +1,83 @@
...
+		moduleToApparentName = collectApparentNames(moduleFile)
+	} else {
+		// If there is no MODULE.bazel file, return a function that always returns the empty string.
+		// Languages will know to fall back to the WORKSPACE names of repos.
+		moduleToApparentName = make(map[string]string)
+	}
+
+	return func(moduleName string) string {
+		return moduleToApparentName[moduleName]
+	}, nil
```
linzhp, line 44: Can we return the `moduleToApparentName` map instead of a function that simply looks up the map?

fmeum: We definitely can, it just means that languages can then loop over all modules declared in the MODULE.bazel file, which I'm not sure we want to encourage: Languages should only care about the modules from which they load and which they can explicitly list. 

I went with the function to soften Hyrum's law here and also because Go maps are mutable and we would need to defensively copy them. But happy to change that if you think the more convulsed signature is worse.

linzhp: Good reasons.

linzhp (approved):

PR #1471, opened 2023-03-09
Remove rules_go version detection logic

fmeum:
**What type of PR is this?**

Other

language/go

The logic is conditional on whether rules_go is loaded in version 0.23.0 or later, which is almost three years old at this point.

In addition to removing quite a bit of code, getting rid of the version detection also means that we don't have to port it to Bzlmod.


fmeum:
@jayconrod Could you review this? You are in the best position to determine whether this is safe to delete.

jayconrod (commented):
I don't think it's entirely safe to remove, but maybe the maintenance benefit is worth it.

This version check is important when you have a new version of Gazelle and an old version of rules_go: Gazelle can avoid generating build files that require features in a newer version of rules_go.

rules_go I hope is basically feature complete at this point, so the main risk is what happens when Go itself introduces support for a new GOOS or GOARCH. rules_go has a hard-coded list of platforms in `go/private/platforms.bzl` that needs to be updated. (You can get the list of platforms Go supports by running `go tool dist list`, but there's not a good way to run that in the load phase.)

Those values start appearing in build tags, and Gazelle needs to generate `select` expressions with the corresponding platform constraints. If the user has an old version of rules_go, they may get an inscrutable error message about those constraints being missing.

New platforms don't come out frequently of course, and they're not adopted immediately. Maybe it's good enough to document the error message in an FAQ that tells users to upgrade rules_go if they see it.

fmeum:
@linzhp @achew22 What do you think about the tradeoff here? I could test out the error message and add it to the docs or I could keep this code and make it work with Bzlmod.

linzhp:
Given the limited bandwidth we have to maintain rules_go and Gazelle, I prefer to reduce maintenance load and not to worry to much about rare scenario such as obscure platforms. However, Gazelle does require some minimal rules_go version to work properly. As for Gazelle 0.28, it requires rules_go 0.35 due to backward-incompatible changes in golang/x/tools. So we can safely remove the checks for rules_go 0.23.

With bzlmod, I think we can just specify rules_go version in MODULE.bazel and don't need to check it in the code, right? If so, we only need to maintain the version check for WORKSPACE mode.

PR #1477, opened 2023-03-14
fix(bzlmod): apply go.sum sums after version resolution

### LINKED ISSUE #1476: bug(bzlmod): only the root go.sum should be used

0.28.0

0.38.0

Bazel 6

Yes, this feature has not been implemented.

Linux AMD64

### What did you do?

When dependencies are resolved with `go mod`, `replace` directives are applied before the `go.sum` file is checked. Because of this, sums will be missing for modules that are replaced.

According to https://go.dev/ref/mod#go-sum-files

> The go command compares each hash with the corresponding line in the main module’s go.sum[ file](https://go.dev/ref/go-sum-files). If the hash is different from the hash in go.sum, the go command reports a security error and deletes the downloaded file without adding it into the module cache.

Currently, this is not the case since we check the `go.sum` file for each `go.mod` file we encounter.

Yes.

Linux AMD 64

### What did you do?

Load a repository with a `go.mod` file but with a missing `go.sum` file.

Only the root `go.sum` file should be honored.

### What did you see instead?

Each `go.mod` file's associated `go.sum` file is read and used.

tyler-french:
**What type of PR is this?**

> Bug fix

internal/bzlmod/go_deps.bzl

Currently, the `go_deps.from_file` implementation will validate each `go.mod` file against each `go.sum` file. This is not technically correct, since the root `go.sum` file may contain certain resolved entries _after_ version resolution.

The best example is `replace` directives. If a certain transitive requirement is replaced with a different entry, that transitive dependency will change from the entry in the transitive repo, and thus, a new entry will be placed in the root `go.sum` file for the accurate sum entry for this module.

This PR adjusts the implementation for gazelle's `go_deps` to more closely match this logic of`go mod`: Evaluate the repository `sums` after the version resolution.

If there are any conflicting entries of `go.sum` we will fail immediately. This makes sure duplicate entries from external repos can't lie about a fake sum of a module.

Fixes #1476


internal/bzlmod/go_deps.bzl
```diff
@@ -39,13 +39,14 @@ def _noop(s):
...
     module_resolutions = {}
     root_versions = {}
+    sums = {}
 
     outdated_direct_dep_printer = print
     for module in module_ctx.modules:
+
         # Parse the go_dep.config tag of the root module only.
         for mod_config in module.tags.config:
-            # bazel_module.is_root is only available as of Bazel 5.3.0.
```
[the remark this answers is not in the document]
fmeum: Yep, please just get rid of this. Bazel 5.4.0's Bzlmod implementation has too many issues to be a realistic target, we should assume everyone is on >= 6.0.0.

internal/bzlmod/go_deps.bzl
```diff
@@ -57,8 +58,25 @@ def _go_deps_impl(module_ctx):
...
+                    fail("Multiple mismatching sums for {}@{} found.".format(module_tag.path, sum_version))
+                sums[item] = new_sum
+
             additional_module_tags += deps_from_go_mod(module_ctx, from_file_tag.go_mod)
 
+        # Load sums from all manually specified modules separately.
+        # Fail on duplicates with mismatched sums.
+        for module_tag in module.tags.module:
+            sum_version = _canonicalize_raw_version(module_tag.version)
+            entry = (module_tag.path, sum_version)
```
fmeum, line 75: Nit: If `item` and `entry` refer to the same kind of tuple that I think they do, use the same name in both places.

internal/bzlmod/go_deps.bzl
```diff
@@ -57,8 +58,25 @@ def _go_deps_impl(module_ctx):
 
         additional_module_tags = []
         for from_file_tag in module.tags.from_file:
+
+            # We will load all sums from transitively resolved `go.sum` files. We will
+            # only use these after version resolution is completed.
+            for item, new_sum in sums_from_go_mod(module_ctx, from_file_tag.go_mod).items():
+                if item in sums and new_sum != sums[item]:
+                    fail("Multiple mismatching sums for {}@{} found.".format(module_tag.path, sum_version))
+                sums[item] = new_sum
```
fmeum, line 67: Could extract these three lines into a function that's reused below to ensure we get the same error message in each case.

internal/bzlmod/go_mod.bzl
```diff
@@ -183,10 +162,34 @@ def _tokenize_line(line, path, line_no):
 
     return tokens, None
 
+# sums_from_go_mod will return a dict of tuple to sum for
+# the modules where: dict[(path, version)] = sum
```
fmeum, line 166: Make this a `"""..."""` docstring, ideally with `Args:` and `Returns:` sections.

internal/bzlmod/go_mod.bzl
```diff
@@ -183,10 +162,34 @@ def _tokenize_line(line, path, line_no):
...
+def sums_from_go_mod(module_ctx, go_mod_label):
+
+    # We go through a Label so that the module extension is restarted if go.sum
+    # changes. We have to use a canonical label as we may not have visibility
+    # into the module that provides the go.sum.
+    go_sum_label = Label("@@{}//{}:{}".format(
+        go_mod_label.workspace_name,
+        go_mod_label.package,
+        "go.sum",
```
fmeum, line 175: What I missed earlier: Could you add a fail if `go_mod_label.name != "go.mod"`?

internal/bzlmod/go_mod.bzl
```diff
@@ -183,10 +162,34 @@ def _tokenize_line(line, path, line_no):
...
+    ))
+
+    go_sum_path = module_ctx.path(go_sum_label)
+    go_sum_content = module_ctx.read(go_sum_path)
```
fmeum, line 179:
```suggestion
    go_sum_content = module_ctx.read(go_sum_label)
```

tests/bcr/go.mod
```diff
@@ -1 +1,5 @@
-// Only here to stop go mod from descending into this directory.
\ No newline at end of file
+// Only here to stop go mod from descending into this directory.
+// We will run `go mod tidy` to make sure all checksums are properly evaluated.
```
fmeum, line 2: How are we doing that? ;-)

tests/bcr/MODULE.bazel
```diff
@@ -12,6 +12,9 @@ bazel_dep(name = "rules_go", version = "0.35.0")
 
 go_deps = use_extension("@gazelle//:extensions.bzl", "go_deps")
 
+# Load the go.mod file to load the go.sum entries.
+go_deps.from_file(go_mod = "//:go.mod")
```
fmeum, line 16: I'm slightly surprised to see this work: Aren't we essentially declaring every dep twice, once with `go_deps.module`, once in `go.mod`? We could add a separate module in a subdirectory that we load with a `local_path_override` to test this without making changes here.

fmeum (commented):
The design looks good to me, just left a few comments on execution.

tests/bcr/go.mod
tyler-french: Removed this. We can adjust and make a better test for the `go.mod` logic in a separate PR. 

This BCR test will still use the `go.sum` file in the root of the repo to validate, since we load gazelle as a `bazel_dep`.

tests/bcr/MODULE.bazel
tyler-french: Similar to other comment: Removed this. We can adjust and make a better test for the go.mod logic in a separate PR.

This BCR test will still use the go.sum file in the root of the repo to validate since we load gazelle as a bazel_dep.

internal/bzlmod/go_mod.bzl
```diff
@@ -183,10 +176,44 @@ def _tokenize_line(line, path, line_no):
...
+            passed from the MODULE.bazel call.
+        go_mod_label: a Label for a `go.mod` file. This label is used
+            to find the associated `go.sum` file.
+
+    Returns:
+        A Dict[(string, string) -> (string)] is retruned where each entry
+        is defined by a Go Module's sum:
+            (path, version) -> (sum)
+    """
+    if go_mod_label.name != "go.mod":
```
fmeum, line 193: Could also be extracted in a `_check_go_mod_name` function.

fmeum (approved):

PR #1492, opened 2023-03-29
bzlmod: Fix interaction between replace and implicit upgrade check

fmeum:
**What type of PR is this?**

Bug fix

bzlmod

Ensures that implicit version upgrade warnings are not emitted for the root module if it uses replace directives to update versions.


fmeum:
@tyler-french

PR #1495, opened 2023-04-04
Add a `go mod tidy` wrapper

fmeum:
**What type of PR is this?**

Feature

bzlmod

The wrapper is aware of generated sources: It runs a Bazel query to determine the external Go dependencies that are direct dependencies of Go targets in the repo and emits a temporary file with fake imports for them.


fmeum:
@tyler-french I could see this living in `rules_go` and being implicitly invoked with `bazel run @rules_go//go mod tidy`, with an option to disable this. What do you think?

grindlemire:
What is the status of this PR? I am running into an issue with `tidy` not liking generated code.

iamricard:
> What is the status of this PR? I am running into an issue with `tidy` not liking generated code.

👋🏼 same here - happy to lend a hand if that would be useful in getting this over the finish line.

fmeum:
@grindlemire @iamricard I am still trying to make sure that this really does cover the problems people experience with `go mod tidy`. Could you maybe share reproducers for the exact type of issue you are running into? That would help me test this PR.

iamricard:
hey! our issues have come from code in the source tree depending on code not available in the source tree, i think mostly proto and gomock packages (we ended up checking in generated code 😢). i can try to create a repro case when/if i get some time. what would be the more useful way to send it? PR to your PR? separate repo? something else?

fmeum:
@iamricard Just a small standalone sample git repo would be ideal. This can easily be turned into an integration test when the PR can address it.

PR #1497, opened 2023-04-05
language/go: Infer prefix from go.mod

fmeum:
**What type of PR is this?**

Feature

language/go

If no prefix has been set but a go.mod file is present, use the path in its `module` directive.

This allows simple Go projects to start using Gazelle without specifying any directives.

### DIFF
--- a/language/go/config.go
+++ b/language/go/config.go
@@ -36,6 +36,7 @@ import (
 	"github.com/bazelbuild/bazel-gazelle/repo"
 	"github.com/bazelbuild/bazel-gazelle/rule"
 	bzl "github.com/bazelbuild/buildtools/build"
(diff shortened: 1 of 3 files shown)


fmeum:
@tyler-french

language/go/config.go
```diff
@@ -36,6 +36,8 @@ import (
 	"github.com/bazelbuild/bazel-gazelle/repo"
 	"github.com/bazelbuild/bazel-gazelle/rule"
 	bzl "github.com/bazelbuild/buildtools/build"
+
+	"golang.org/x/mod/modfile"
```
tyler-french, line 40: I think it's best to only have two groups of imports: standard lib + others

PR #1509, opened 2023-04-18
Support Bzlmod for gazelle itself

fmeum:
All tests pass except for the Stardoc diff tests, which fail due to https://github.com/bazelbuild/bazel/issues/14140.


fmeum:
@tyler-french

PR #1512, opened 2023-04-19
language/go: Handle `unix` build tag.

### LINKED ISSUE #1465: Gazelle does not support `unix` build tag (`//go:build unix`)

v0.29.0

v0.38.1

v6.0.0

Yes.

Debian 11 amd64

### What did you do?

I initially tried to depend on github.com/jackc/pgx/stdlib in a bazel project that uses gazelle.

However, that's annoying to reproduce minimally and cleanly, so the error can be minimally reproduced as follows:
On a unix machine, use `git apply` to apply this patch to an empty workspace (empty directory on which `git init` has been run):

```diff
diff --git a/BUILD b/BUILD
new file mode 100644
index 0000000..a27a8fc
--- /dev/null
+++ b/BUILD
@@ -0,0 +1,4 @@
+load("@bazel_gazelle//:def.bzl", "gazelle")
+
+# gazelle:prefix github.com/example/project
+gazelle(name = "gazelle")
diff --git a/WORKSPACE b/WORKSPACE
new file mode 100644
index 0000000..42ee6b8
--- /dev/null
+++ b/WORKSPACE
@@ -0,0 +1,35 @@
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+
+http_archive(
+    name = "io_bazel_rules_go",
+    sha256 = "dd926a88a564a9246713a9c00b35315f54cbd46b31a26d5d8fb264c07045f05d",
+    urls = [
+        "https://mirror.bazel.build/github.com/bazelbuild/rules_go/releases/download/v0.38.1/rules_go-v0.38.1.zip",
+        "https://github.com/bazelbuild/rules_go/releases/download/v0.38.1/rules_go-v0.38.1.zip",
+    ],
+)
+
+http_archive(
+    name = "bazel_gazelle",
+    sha256 = "ecba0f04f96b4960a5b250c8e8eeec42281035970aa8852dda73098274d14a1d",
+    urls = [
+        "https://mirror.bazel.build/github.com/bazelbuild/bazel-gazelle/releases/download/v0.29.0/bazel-gazelle-v0.29.0.tar.gz",
+        "https://github.com/bazelbuild/bazel-gazelle/releases/download/v0.29.0/bazel-gazelle-v0.29.0.tar.gz",
+    ],
+)
+
+
+load("@io_bazel_rules_go//go:deps.bzl", "go_register_toolchains", "go_rules_dependencies")
+load("@bazel_gazelle//:deps.bzl", "gazelle_dependencies", "go_repository")
+
+############################################################
+# Define your own dependencies here using go_repository.
+# Else, dependencies declared by rules_go/gazelle will be used.
+# The first declaration of an external repository "wins".
+############################################################
+
+go_rules_dependencies()
+
+go_register_toolchains(version = "1.20.1")
+
+gazelle_dependencies()
diff --git a/foo/foo.go b/foo/foo.go
new file mode 100644
index 0000000..9d41b09
--- /dev/null
+++ b/foo/foo.go
@@ -0,0 +1,3 @@
+package foo
+
+var _ = &Foo{}
diff --git a/foo/no.go b/foo/no.go
new file mode 100644
index 0000000..486262e
--- /dev/null
+++ b/foo/no.go
@@ -0,0 +1,5 @@
+//go:build !unix
+
+package foo
+
+type Foo struct{}
diff --git a/foo/yes.go b/foo/yes.go
new file mode 100644
index 0000000..a76d81e
--- /dev/null
+++ b/foo/yes.go
@@ -0,0 +1,5 @@
+//go:build unix
+
+package foo
+
+type Foo struct{}
```

Then, in that directory, run `bazel run //:gazelle` followed by `bazel build //foo:foo`.

I expected to see, in the `foo/BUILD.bazel` file:

```python

go_library(
    name = "foo",
    srcs = [
        "foo.go",
        "yes.go",
    ],
    importpath = "github.com/example/project/foo",
)
```

with the relevant part being the `yes.go` file (which has the `//go:build unix` line) included in the `srcs` list, as, per https://pkg.go.dev/cmd/go@go1.20.1#hdr-Build_constraints, the `unix` build constraint should be satisfied "if GOOS is a Unix or Unix-like system", where the list of those systems can be found here: 

https://cs.opensource.google/go/go/+/refs/tags/go1.20.1:src/go/build/syslist.go;l=34-51 (Lines reproduced below for viewing convenience)
```go
// unixOS is the set of GOOS values matched by the "unix" build tag.
// This is not used for filename matching.
// This list also appears in cmd/dist/build.go and
// cmd/go/internal/imports/build.go.
var unixOS = map[string]bool{
	"aix":       true,
	"android":   true,
	"darwin":    true,
	"dragonfly": true,
	"freebsd":   true,
	"hurd":      true,
	"illumos":   true,
	"ios":       true,
	"linux":     true,
	"netbsd":    true,
	"openbsd":   true,
	"solaris":   true,
}
```

When I run `go env` on my machine, I get `GOOS="linux"`, so I should satisfy that build constraint.

I also expect that the build will then complete successfully.

### What did you see instead?

Instead, my `foo/BUILD.bazel` file is the following:

```python

go_library(
    name = "foo",
    srcs = [
        "foo.go",
        "no.go",
    ],
    importpath = "github.com/example/project/foo",
)
```

with the relevant part being the `no.go` file (which has the `//go:build !unix` line) included in the `srcs` list instead of the `yes.go` file. Further, the build fails with the following error:

```
foo/foo.go:3:10: undefined: Foo
compilepkg: error running subcommand external/go_sdk/pkg/tool/linux_amd64/compile: exit status 2
Target //foo:foo failed to build
```

because when `go` actually builds those files, the `!unix` build directive is not satisfied, so the `no.go` file is not built, and the `Foo` type is never defined. Note that if I change the `"no.go",` line to `"yes.go",`, the build succeeds.

Edit: After testing with the tag `linux` instead of `unix`, I see that both files are included in `srcs`. That would also be an acceptable outcome; I do not need `no.go` excluded, I just need `yes.go` included.

connyay:
**What type of PR is this?**

Bug fix

language/go and rules

Adds support for the "unix" build tag added via https://github.com/golang/go/issues/20322

jackc/pgx is a popular pg lib that currently fails to build with bazel. Requires a local patch to change pgx code. 
https://github.com/jackc/pgx/issues/1520

Fixes #1465

**Other notes for review**

### DIFF
--- a/rule/platform.go
+++ b/rule/platform.go
@@ -102,6 +102,23 @@ var OSAliases = map[string][]string{
 	"ios":     {"darwin"},
 }
 
+// UnixOS is the set of GOOS values matched by the "unix" build tag.
+// This list is from go/src/cmd/dist/build.go.
+var UnixOS = map[string]bool{
+	"aix":       true,
+	"android":   true,
+	"darwin":    true,
+	"dragonfly": true,
+	"freebsd":   true,
+	"hurd":      true,
+	"illumos":   true,
+	"ios":       true,
+	"linux":     true,
+	"netbsd":    true,
+	"openbsd":   true,
+	"solaris":   true,
+}
+
 var (
 	// KnownOSs is the sorted list of operating systems that Go supports.
 	KnownOSs []string
@@ -133,6 +150,8 @@ func init() {
 		KnownOSArchs[p.OS] = append(KnownOSArchs[p.OS], p.Arch)
 		KnownArchOSs[p.Arch] = append(KnownArchOSs[p.Arch], p.OS)
 	}
+	// unix is a pseudo os
+	KnownOSSet["unix"] = true
 	KnownOSs = make([]string, 0, len(KnownOSSet))
 	KnownArchs = make([]string, 0, len(KnownArchSet))
 	for os := range KnownOSSet {
--- a/rule/platform.go
+++ b/rule/platform.go
@@ -150,8 +150,6 @@ func init() {
 		KnownOSArchs[p.OS] = append(KnownOSArchs[p.OS], p.Arch)
 		KnownArchOSs[p.Arch] = append(KnownArchOSs[p.Arch], p.OS)
 	}
-	// unix is a pseudo os
-	KnownOSSet["unix"] = true
 	KnownOSs = make([]string, 0, len(KnownOSSet))
 	KnownArchs = make([]string, 0, len(KnownArchSet))
 	for os := range KnownOSSet {
--- a/language/go/fileinfo.go
+++ b/language/go/fileinfo.go
@@ -559,8 +559,7 @@ func checkConstraints(c *config.Config, os, arch, osSuffix, archSuffix string, t
 		if isIgnoredTag(tag) {
 			return true
 		}
-
-		if _, ok := rule.KnownOSSet[tag]; ok {
+		if _, ok := rule.KnownOSSet[tag]; ok || tag == "unix" {
 			if os == "" {
 				return false
 			}
--- a/language/go/fileinfo.go
+++ b/language/go/fileinfo.go
@@ -525,6 +525,9 @@ func matchesOS(os, value string) bool {
 	if os == value {
 		return true
 	}
+	if value == "unix" {
+		return rule.UnixOS[os]
+	}
 	for _, alias := range rule.OSAliases[os] {
 		if alias == value {
 			return true


rule/platform.go
```diff
@@ -133,6 +150,8 @@ func init() {
 		KnownOSArchs[p.OS] = append(KnownOSArchs[p.OS], p.Arch)
 		KnownArchOSs[p.Arch] = append(KnownArchOSs[p.Arch], p.OS)
 	}
+	// unix is a pseudo os
+	KnownOSSet["unix"] = true
```
connyay, line 154: This doesn't feel great, but neither does adding unix+arches to KnownPlatforms... unix is an alias to other oses, but the various oses have different arch support. Not sure the correct/expected behavior here.

connyay: This caused other implications. Changed in 276dbe9

connyay:
Hmm. I'm slightly nervous about the unintended consequences of this change. A simpler change would be to just add `unix` as an ignored tag and let the compiler handle it. 
https://github.com/bazelbuild/bazel-gazelle/blob/3d8aac016157e62b35b5774032ebcd0ed2ba55e6/language/go/build_constraints.go#L310-L314 

I confirmed that ignoring `unix` produced the expected results to build jackc/pgx successfully.


```diff
diff --git a/language/go/build_constraints.go b/language/go/build_constraints.go
index 1ce071a..3127077 100644
--- a/language/go/build_constraints.go
+++ b/language/go/build_constraints.go
@@ -313,7 +313,7 @@ func matchAuto(tokens []string) (*buildTags, error) {
 // build constraints for a file and will instead defer to the compiler at compile
 // time.
 func isIgnoredTag(tag string) bool {
-	if tag == "cgo" || tag == "race" || tag == "msan" {
+	if tag == "cgo" || tag == "race" || tag == "msan" || tag == "unix" {
 		return true
 	}
 	if len(tag) < 5 || !strings.HasPrefix(tag, "go") {
--
```

dlski:
Hi, any chance to get this merged soon?

fmeum (approved):

PR #1515, opened 2023-04-25
remove circular override hack

tyler-french:
This fixes a hack to `go_deps` which allows overrides from `rules_go` to still work (even when `rules_go` is not the main module.

This provides a dangerous experience by causing potential confusion regarding where an override could be coming from.

Because of that, this hack should be removed, since it is no longer needed with the latest `0.39.1` of rules_go.


seh (approved):
Thank you for remembering to remove that hack.

fmeum (approved):

PR #1508, opened 2023-04-14
Relax parsing of labels

purkhusid:
**What type of PR is this?**

 Bug fix

**What package or component does this PR mostly affect?**

Gazelle extension using the `label` package

**What does this PR do? Why is it needed?**

The parsing in the `labels` package is too strict. This PR updates the regex to allow more characters that Bazel is OK with.

My exact use case was that I was generating TypeScript targets for a NextJS application which has has the following folder structure:
```
pages/
    entity/
       [entityId]/
           index.ts
```

This caused issues with target-determinator(https://github.com/bazel-contrib/target-determinator) which uses the labels package to parse affected labels.

PR #1518, opened 2023-04-28
Support '~' in label repo

### LINKED ISSUE #1517: Gazelle's label validator for repo name is too strict

v0.25.1

n/a

6.1.2

Yes

Linux x86_64 and Mac x86_64 and Mac arm64

### What did you do?

Using target-determinator with Bazel 6 and rules_python configured via bzlmod results in a parse error of the rules_python repo names. target-determinator depends on Gazelle's label.Parse function to read labels. (Cross-linking issue filed there: https://github.com/bazel-contrib/target-determinator/issues/59).

No error, Bazel and rules_python are perfectly happy with a tilde in the repo name.

### What did you see instead?

```
label parse error: repository has invalid characters: "@rules_python~0.21.0~pip~pip//:boto3_pkg
```

See also: https://github.com/bazelbuild/bazel-gazelle/pull/1508 -- though at time of posting this doesn't touch the repo regex.

mark-thm:
**What type of PR is this?**

Bug fix

Gazelle extension using the label package

Relax label validation to support `~` characters in repo names.

Fixes #1517

**Other notes for review**

Related change #1508

### DIFF
--- a/label/label.go
+++ b/label/label.go
@@ -62,7 +62,7 @@ func New(repo, pkg, name string) Label {
 var NoLabel = Label{}
 
 var (
-	labelRepoRegexp = regexp.MustCompile(`^@$|^[A-Za-z.-][A-Za-z0-9_.-~]*$`)
+	labelRepoRegexp = regexp.MustCompile(`^@$|^[A-Za-z.-][A-Za-z0-9_.~-]*$`)
 	labelPkgRegexp  = regexp.MustCompile(`^[A-Za-z0-9/._@-]*$`)
 	// This was taken from https://docs.bazel.build/versions/main/build-ref.html#name
 	labelNameRegexp = regexp.MustCompile("^[A-Za-z0-9!%-@^_` \"#$&'()*-+,;<=>?\\[\\]{|}~/.]*$")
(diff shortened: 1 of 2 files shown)


label/label.go
```diff
@@ -62,7 +62,7 @@ func New(repo, pkg, name string) Label {
 var NoLabel = Label{}
 
 var (
-	labelRepoRegexp = regexp.MustCompile(`^@$|^[A-Za-z.-][A-Za-z0-9_.-]*$`)
+	labelRepoRegexp = regexp.MustCompile(`^@$|^[A-Za-z.-][A-Za-z0-9_.~-]*$`)
```
mark-thm, line 65: I opted for the smallest version of a fix here, looking at master in Bazel: 
https://github.com/bazelbuild/bazel/blob/71fb1e4188b01e582a308cfe4bcbf1c730eded1b/src/main/java/com/google/devtools/build/lib/cmdline/LabelValidator.java#L272-L278

...it seems as if there are no invalid characters, and that a repo name is any string between `@` and `//`. That calls for a bit more rework to this method, and I was reluctant to propose a bigger change as a first time contributor.

fmeum (approved):

Issue #1520, opened 2023-05-03
[bzlmod] Add build_file_generation directive


mortenmj:
### What version of gazelle are you using?

HEAD (23226de7388198e859f3ce8dfb54d3ebcde6dc93)

### What version of rules_go are you using?

0.39.1

### What version of Bazel are you using?

6.2.0rc1

### Does this issue reproduce with the latest releases of all the above?

Yes

### What operating system and processor architecture are you using?

Debian GNU/Linux 11 (bullseye)

### What did you do?

Trying to replicate the following (pre-bzlmod) setup:

```
    go_repository(
        name = "com_github_google_safetext",
        build_file_generation = "on",
        build_file_name = "BUILD.bazel",
        build_file_proto_mode = "disable_global",
        importpath = "github.com/google/safetext",
        sum = "h1:SJ+NtwL6QaZ21U+IrK7d0gGgpjGGvd2kz+FzTHVzdqI=",
        version = "v0.0.0-20220905092116-b49f7bc46da2",
    )
```

I am not able to express `build_file_generation = "on"` when using bzlmod.

mortenmj:
@fmeum Tagging you as requested

fmeum:
Cc @seh

PR #1522, opened 2023-05-03
bzlmod: Add `build_file_generation` to `gazelle_override`

fmeum:
This attribute is required as `build_file_generation` on `go_repository` doesn't have an equivalent directive.

Fixes https://github.com/bazelbuild/bazel-gazelle/issues/1520

### DIFF
--- a/internal/bzlmod/BUILD.bazel
+++ b/internal/bzlmod/BUILD.bazel
@@ -16,17 +16,11 @@ filegroup(
 )
 
 bzl_library(
-    name = "directives",
-    srcs = ["default_gazelle_overrides.bzl"],
-    visibility = ["//:__subpackages__"],
-)
-
-bzl_library(
     name = "go_deps",
     srcs = ["go_deps.bzl"],
     visibility = ["//:__subpackages__"],
     deps = [
-        ":directives",
+        ":default_gazelle_overrides",
         ":semver",
         "//internal:go_repository",
     ],
--- a/internal/bzlmod/BUILD.bazel
+++ b/internal/bzlmod/BUILD.bazel
@@ -5,7 +5,7 @@ filegroup(
     testonly = True,
     srcs = [
         "BUILD.bazel",
-        "directives.bzl",
+        "default_gazelle_overrides.bzl",
         "go_deps.bzl",
         "go_mod.bzl",
         "non_module_deps.bzl",
@@ -17,7 +17,7 @@ filegroup(
 
 bzl_library(
     name = "directives",
-    srcs = ["directives.bzl"],
+    srcs = ["default_gazelle_overrides.bzl"],
     visibility = ["//:__subpackages__"],
 )
 
@@ -58,5 +58,11 @@ bzl_library(
 bzl_library(
     name = "utils",
     srcs = ["utils.bzl"],
+    visibility = ["//:__subpackages__"],
+)
+
+bzl_library(
+    name = "default_gazelle_overrides",
+    srcs = ["default_gazelle_overrides.bzl"],
     visibility = ["//:__subpackages__"],
 )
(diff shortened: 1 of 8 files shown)


fmeum:
@seh Could you review?

FYI @mortenmj

fmeum:
I haven't added a default override for `github.com/google/safetext` as the most recent version doesn't seem to require any overrides anymore.

internal/bzlmod/BUILD.bazel
```diff
@@ -17,7 +17,7 @@ filegroup(
 
 bzl_library(
     name = "directives",
-    srcs = ["directives.bzl"],
```
seh, line 20: Do we still need to retain this `bzl_library` target, given the new one down at line 64?

tests/bcr/go.mod
```diff
@@ -6,4 +6,8 @@ go 1.19
 // Validate go.mod replace directives can be properly used:
 replace github.com/bmatcuk/doublestar/v4 => github.com/bmatcuk/doublestar v1.3.4
 
-require github.com/bmatcuk/doublestar/v4 v4.0.0
+
```
seh, line 9: Is this an extra blank line?
```suggestion
```

internal/bzlmod/go_deps.bzl
```diff
@@ -97,12 +101,23 @@ def _synthesize_gazelle_override(module, gazelle_overrides, fixups):
 def _safe_append_directives(module, gazelle_overrides, directives):
```
seh, line 101: Perhaps we should rename this function now, since it's no longer concerned solely with directives. Here is one suggested alternate name:
```suggestion
def _safe_update_overrides(module, gazelle_overrides, directives):
```

internal/bzlmod/go_deps.bzl
```diff
@@ -97,12 +101,23 @@ def _synthesize_gazelle_override(module, gazelle_overrides, fixups):
...
         directives = existing + directives,
+        build_file_generation = build_file_generation,
     )
 
+def _get_build_file_generation(path, gazelle_overrides):
+    override = gazelle_overrides.get(path)
+    if override:
+        return override.build_file_generation
+
+    return DEFAULT_BUILD_FILE_GENERATION_BY_PATH.get(path, "auto")
```
seh, line 119: Here's a possible rephrasing, which I'm not sure is better:
```suggestion
    return override.build_file_generation if override else DEFAULT_BUILD_FILE_GENERATION_BY_PATH.get(path, "auto")
```

internal/bzlmod/go_deps.bzl
```diff
@@ -433,6 +450,19 @@ _gazelle_override_tag = tag_class(
             extension within this Bazel module.""",
             mandatory = True,
         ),
+        "build_file_generation": attr.string(
+            default = "auto",
+            doc = """One of `"auto"` (default), `"on"`, `"off"`.
+
+            Whether Gazelle should generate build files for the Go module. In
```
seh, line 457: I was going to suggest capitalizing "build" as "BUILD" here and on the next line, but I see that [the existing documentation for `go_repository`'s "build_file_generation" attribute](https://github.com/bazelbuild/bazel-gazelle/blob/master/repository.md#user-content-go_repository-build_file_generation) writes it in lowercase too.

internal/bzlmod/go_deps.bzl
```diff
@@ -433,6 +450,19 @@ _gazelle_override_tag = tag_class(
...
+            `"auto"` mode, Gazelle will run if there is no build file in the Go
+            module's root directory.""",
+            values = [
+                "on",
+                "auto",
+                "off",
```
seh, line 463: Should we sort these values lexicographically?

internal/bzlmod/go_deps.bzl
fmeum: I'm also not completely sure which option I like better, but lean towards keeping the current one.

seh (approved):
Thank you for addressing my questions and suggestions.

PR #1526, opened 2023-05-04
Let `bazel_dep`s replace Go deps

### LINKED ISSUE #1525: [bzlmod] Add ability to express that a go dependency is fulfilled by a bazel_dep

HEAD (https://github.com/bazelbuild/bazel-gazelle/commit/23226de7388198e859f3ce8dfb54d3ebcde6dc93)

0.39.1

6.2.0rc1

Yes

Debian GNU/Linux 11 (bullseye), amd64

### What did you do?

Tried to use [github.com/google/go-jsonnet](https://github.com/google/go-jsonnet) as a go dependency. This package has dependencies that require that it's imported as a bzlmod module, but there is no way today to express that a go dependency is provided by a `bazel_dep`.

### LINKED ISSUE #1543: Allow "go_deps.module_override" tag in non-root Bazel modules
Gazelle Commit 23226de.
rules_go version 0.39.1 from bzlmod
Bazel 6.2 from nixpkgs
x86_64 Gentoo/Clang

- https://github.com/eomii/rules_ll/issues/149

Non-minimal reproducer:

```bash
git clone git@github.com/eomii/rules_ll
cd rules_ll/examples
nix develop
bazel build @rules_ll//devtools:cluster
```

### What did you do?

After https://github.com/bazelbuild/bazel-gazelle/issues/1521 we have something like this in our `MODULE.bazel`:

```python
bazel_dep(name = "rules_go", version = "0.39.1")
bazel_dep(name = "gazelle", version = "23226de")
go_deps.module_override(
    path = "github.com/cloudflare/circl",
    patches = [
        "@rules_ll//patches:go_circl_header_fix.diff",
    ],
    patch_strip = 1,
)
use_repo(
    go_deps,
    "com_github_pulumi_pulumi_kubernetes_sdk_v3",
    "com_github_pulumi_pulumi_sdk_v3",
    "io_k8s_sigs_kind",
)
```

Unfortunately this causes errors for importers of the module:

```bash
        File "external/gazelle~23226de/internal/bzlmod/go_deps.bzl", line 189, column 36, in _go_deps_i
mpl
                _fail_on_non_root_overrides(module, "module_override")
        File "external/gazelle~23226de/internal/bzlmod/go_deps.bzl", line 68, column 17, in _fail_on_no
n_root_overrides
                fail(_report_forbidden_override(module, tag_class))
Error in fail: Using the "go_deps.module_override" tag in a non-root Bazel module is forbidden, but module "rules_ll" requests it.

If you need this override for a Bazel module that will be available in a public registry (such as the Bazel Central Registry), please file an issue at https://github.com/baze
lbuild/bazel-gazelle/issues/new or submit a PR adding the required directives to the "directives.bzl" file at https://github.com/bazelbuild/bazel-gazelle/tree/master/internal
/bzlmod/directives.bzl.
```

This looks like the behavior of something like `local_path_override` applies to `go_deps.module_override` which seems wrong. The patch not being allowed to propagate to importers means that importers can't build targets the same way that upstream builds them.

I think it should be fine (and probably default behavior) to allow patches to propagate to importers or somehow bind them to the bazel module that declares them.

I've had similar issues with LLVM where we need to patch the repo but also have downstream users that need the ability to "repatch" it. We currently append the importer patches to the existing ones, which has worked fine so far (https://github.com/eomii/bazel-eomii-registry/blob/11e289ec5b9c2ed9c4ebb0455133ff8455831233/modules/llvm-project-overlay/17-init-bcr.3/patches/00_add_bzlmod_support.patch#L141).

That said, it also seems like a viable solution to just remove the `_fail_on_non_root_overrides` here:

https://github.com/bazelbuild/bazel-gazelle/blob/31064705f5a481149d060a94adb48cd2c03347b7/internal/bzlmod/go_deps.bzl#L230-L237

cc @seh @fmeum @tyler-french

fmeum:
When a Go dependency is brought in both as a `bazel_dep` and as a
regular Go dependency via e.g. go.mod, the `bazel_dep` repo should be
used instead of creating a separate `go_repository`, which would risk
duplicate packages.

This is realized by registering every Bazel module with a
`go_deps.from_file` tag as providing the Go module of the same version
with the module path given in `go.mod`.

Gazelle is aware of these tags and generates either a canonical label or
 a label using the apparent module repo name (the latter only for the
root module and only if it directly depends on the `bazel_dep`).

Fixes #1525 
Fixes #1543


fmeum:
@Wyverald @tyler-french Could you review?

I implemented two different approaches (first commit vs. both commits), with one making the association explicit and one deriving it from `go_deps.from_file`. I prefer the latter, but in case you see issues with this association being made implicitly, we can switch back to the first approach.

fmeum:
We probably need to introduce some kind of version resolution to ensure that outdated `bazel_dep`s do not override up-to-date `go.mod` deps.

fmeum:
I added version resolution.

@tyler-french @seh Could you take a look if time permits?

seh:
> Could you take a look if time permits?

I read it all, but I need to think about both the problem and the solution more before I have anything constructive to add.

fmeum:
@seh Thank, that's exactly what I'm looking for. Take your time!

tyler-french:
Tested at Uber

fmeum:
I rebased onto master and pushed a new commit that adds support for the generalized semver format used by Bazel module registries.

I will wait for https://github.com/bazelbuild/bazel-central-registry/pull/674 and add a test based on it before merging.

fmeum:
@seh I added a more realistic example of how this would be used in the form of [circl](https://github.com/cloudflare/circl).

seh:
>  I added a more realistic example of how this would be used in the form of [circl](https://github.com/cloudflare/circl).

I hadn't seen a module like _circl_ before, where there's [an upstream Go module](https://github.com/cloudflare/circl/tree/main) with no "awareness" of Bazel, but then we publish a Bazel module that [adds in all the necessary Bazel control files](https://github.com/bazelbuild/bazel-central-registry/tree/main/modules/circl/1.3.3/patches) to allow building and testing the library.

Now, coming back to _this_ patch, I understand it better now, in that we're reading all the uses of the `go_dep` extension's nomination of Go modules from non-root/child Bazel modules and taking them into consideration when choosing the maximum version for each Go module.

The description mentions a "provides" tag on the `go_dep` extension. Is that implemented?

fmeum:
@seh Sorry, the PR description was outdated, updated it to the description of the first commit.

seh:
> I hadn't seen a module like _circl_ before, where there's [an upstream Go module](https://github.com/cloudflare/circl/tree/main) with no "awareness" of Bazel, but then we publish a Bazel module that [adds in all the necessary Bazel control files](https://github.com/bazelbuild/bazel-central-registry/tree/main/modules/circl/1.3.3/patches) to allow building and testing the library.

What's not clear to me yet is why one would express such a Bazel module dependency, as opposed to just letting the _go.mod_ file express the dependency and rely on `go_deps.from_file` to take care of the rest. It seems like expressing the same idea twice, when our goal with the project is to eliminate as much of the redundant Bazel configuration as possible (e.g. getting rid of the need for `use_repo` with `go_deps`).

fmeum:
Having to use a Bazel module as a Go module should be rare, but there are important exceptions:

1. When a Go module requires patches beyond Gazelle's capabilities. A `go_deps.module_override` can fix that, but makes it so that the Go module can no longer be used by non-root Bazel modules. We *could* have a separate registry for patches similar to what we do to centralize `go_deps.gazelle_override`s, but that seems pretty complex to set up while the BCR provides this out of the box. 

The particular case of `circl` is interesting here as it requires adding `cc_library` targets, but those will eventually have to be loaded from `rules_cc`, which in turn requires visibility into this Bazel module - `go_deps`-generated repos don't have this, only `bazel_dep`s can declare additional dependencies in this way.

Another use case is for something like `bazelbuild/buildtools`, which provides both Starlark rules, Go tools and Go libraries. Most Bazel users will want the first two and thus probably a `bazel_dep`, but anyone who consumes the Go library wants to do so via `go_deps`. We should probably ensure that there aren't two different repos floating around that provide the same Go targets but potentially in different versions.

seh:
> there are important exceptions:

Thank you for the detailed explanation. The dependency landscape is more complex than I had considered.

aaronmondal:
I've just tested this for rules_ll which uses that exact circl dependency as a transitive dependency. We were previously running into an issue where users could not only not import our go tooling, but couldn't import rules_ll at all because the go toolchain misconfiguration would block the entire build.

Keeping patches like the ones required for circl maintained isn't nice if it requires custom deps and overrides. This patch fixes these issues in an IMO very elegant manner.

So this will also fix #1543 in a very satisfactory way :heart:

PR #1527, opened 2023-05-06
Option to include diff header line

siddharthab:
`git apply` works for existing files only when there is a "git diff"
header line for each file that begins with `git --diff`. Without this
header line, we get `error: new file ... depends on old contents`.

It may be reasonable to give Gazelle an option to output such lines to
play nice with git.

In general, `diff -u` also outputs a similar header line, although
`patch` does not seem to care as long as the patch conforms to the
unified diff format.

### DIFF
--- a/cmd/gazelle/fix-update.go
+++ b/cmd/gazelle/fix-update.go
@@ -51,6 +51,7 @@ type updateConfig struct {
 	walkMode       walk.Mode
 	patchPath      string
 	patchBuffer    bytes.Buffer
+	diffHeader     string
 	print0         bool
 }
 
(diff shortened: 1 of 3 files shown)


cmd/gazelle/fix-update.go
```diff
@@ -51,6 +51,7 @@ type updateConfig struct {
 	walkMode       walk.Mode
 	patchPath      string
 	patchBuffer    bytes.Buffer
+	diffHeader     string
```
siddharthab, line 54: Please let me know if you are OK with this variable name, or if you would like me to rename it to be `diffHeaderStyle` to be more clear.

siddharthab:
I just checked the source code for `git apply`. Looks like they do support the unidiff format, but they interpret the timestamp to mean a file creation/deletion event if it is the epoch timestamp.

https://github.com/git/git/blob/69c786637d7a7fe3b2b8f7d989af095f5f49c3a8/apply.c#L861

Instead of this PR, I can send in something to output a timestamp different from the epoch timestamp that is always being output now.

siddharthab:
Closing in favor of #1528.

PR #1530, opened 2023-05-09
Disable proto target generation for external repositories by default

fmeum:
**What type of PR is this?**

Other

go_repository

External non-Bazel-aware dependencies need to contain checked in .pb.go files for protos or they wouldn't support regular Go builds.

Fixes https://github.com/bazelbuild/rules_go/issues/3308

**Other notes for review**


jayconrod:
This would be an incompatible change, so the usual caution and caveats apply about that.

Having said that, I think it was a mistake for `go_repository` to generate proto targets by default, and I'd be happy to see a safe way to unwind that.

There are a couple reasons it was needed historically.

- If you have `proto_library` targets in your own workspace, and you want to import protos in other workspaces, you need `proto_library` targets to be generated there. For people migrating to Bazel from GOPATH, this was pretty common. Your GOPATH tree contained .proto files organized by Go import path, and it was easy to write a `//go:generate` line or Make rule to generate .pb.go files. I expect that's much less common with Go modules because it's hard to import stuff from the module cache.
- It was difficult to mix checked-in proto code with build-time-generated proto code. I forget what the difficulties were; pure `go_library` targets lacked a lot of metadata needed by the proto rules. I think it would compile but you'd get weird link-time conflicts. I'm not sure if that's still true today. It was especially problematic with GOPATH because many people checked in their vendor trees at different versions, and you'd end up with duplicated packages with incompatible types.

I generally think ease-of-use of building protos is an opportunity for Bazel to be a lot better than other tools. It's still hard to write or generate all the targets and dependencies, but Gazelle can help a lot with that with a good design. (Ideally one decoupled from Go.)

fmeum:
Thanks for sharing these insights. 

Do you think this would be safer if we only disabled `go_proto_library` generation, but kept `proto_library`? 

I also have a more targeted fix for the linked issue: Resolve import path conflicts in external repos by preferring the dep that doesn't embed protos, if one does and one doesn't. I will look into submitting this as an alternative approach.

jayconrod:
> Do you think this would be safer if we only disabled go_proto_library generation, but kept proto_library?

Better to disable both I think if you're disabling one. I don't think proto dependency resolution actually works across repos, right? So I imagine using those `proto_library` targets is uncommon. It would be better IMO to make `go_repository` pure Go.

linzhp:
In Uber, we disable proto target generation for external repositories by default, by doing something like `gazelle update-repos -build_file_proto_mode=disable`. Can we advise people to do that too?

fmeum:
I will close this for now and submitted a more targeted fix: https://github.com/bazelbuild/bazel-gazelle/pull/1537.

PR #1537, opened 2023-05-11
Prefer non-proto over proto deps in external repos

fmeum:
**What type of PR is this?**

Feature

go_repository

This resolves ambiguous resolver matches in favor of pre-generated `.pb.go` files, even if they live in a different directory than the `.proto` files.

Fixes https://github.com/bazelbuild/rules_go/issues/3308

**Other notes for review**


linzhp (commented):
I think we should do the opposite: prefer the proto over non-proto deps, because whether an external repo has pre-generated .pb.go is not controlled by user, but whether to generate go_proto_library can be controlled by users. In case the proto and the generated .pb.go files are out of sync, if users want the pre-generated .pb.go file, they can set `gazelle:go_generate_proto false`; if they want to use the go files generated from latest proto, they can set `gazelle:go_generate_proto true`.

To be compatible with `go build`, I think it makes sense to set the default of `gazelle:go_generate_proto` to false in external repos.

fmeum:
@linzhp Convinced. I would leave the flip of the default to a separate PR though.

linzhp (approved):

linzhp:
For simplicity in implementation of the flipping PR, we can flip the `-build_file_proto_mode` of `update-repos` to "disable", or `build_file_proto_mode` of `go_repository` to "disable", which stop generating `proto_library` and `go_proto_library` all together, like Jay suggested. In case, proto and .pb.go is out of sync and people want to use the .pb.go, it's quite dangerous to use the `proto_library`, so there is little value of generating them.

PR #1539, opened 2023-05-18
Fix `go:embed` failing to add `embedsrcs` for subdirectories named build

### LINKED ISSUE #1538: `go:embed` fails to add `embedsrcs` for subdirectories named `build`

0.30.0

0.39.0

6.2.0

Yes

This affects MacOS x86_64 (also tested in MacOS on an M2 chip).

### What did you do?

Use go:embed on a **sub**directory named  `build`  (just happens to be a [possible](https://github.com/bazelbuild/bazel-gazelle/blob/ee5f680/config/config.go#L137) build filename)

```
//go:embed ui/build/*
var staticContents embed.FS
```

An example of such usage can be seen in this Go library: https://github.com/hibiken/asynqmon/blob/28b1d46/handler.go#L100-L101. I encountered this issue while auto-generating the BUILD file for this library in our project.

Note: I've observed this happens when `build` is a **subdirectory** (i.e. nested)  , as seen in the example above `ui/build/*`

Expecting to see the `embedsrcs` to enumerate the content  of [ui/build](https://github.com/hibiken/asynqmon/tree/28b1d463d000e5da9843b9358fd3a2621aa7f0a4/ui/build)

```
go_library(
    name = "foo",
    srcs = [
        // ...
    ],
    embedsrcs = [
        "ui/build/android-chrome-192x192.png",
        "ui/build/android-chrome-512x512.png",
        "ui/build/apple-touch-icon.png",
        "ui/build/asset-manifest.json",
        "ui/build/favicon-16x16.png",
        "ui/build/favicon-32x32.png",
        "ui/build/favicon.ico",
        "ui/build/index.html",
        "ui/build/manifest.json",
        "ui/build/robots.txt",
        "ui/build/static/js/2.83624df2.chunk.js",
        "ui/build/static/js/2.83624df2.chunk.js.LICENSE.txt",
        "ui/build/static/js/2.83624df2.chunk.js.map",
        "ui/build/static/js/main.5adda2da.chunk.js",
        "ui/build/static/js/main.5adda2da.chunk.js.map",
        "ui/build/static/js/runtime-main.9fea6c1a.js",
        "ui/build/static/js/runtime-main.9fea6c1a.js.map",
        "ui/build/static/media/logo-color.c2b0c1f3.svg",
        "ui/build/static/media/logo-white.3fa2ac55.svg",
    ],
    //...
)
```
Note: There are no BUILD files or Go files within this directory.

### What did you see instead?

The `embedsrcs` was not populated at all and we get an error indicating there are no matched files.

```
❯ gazelle /path/asynqmon
gazelle: /path/asynqmon/handler.go:100:2: pattern ui/build/*: matched no files
```

This only happens when the subdirectory which contains the srcs to be embedded is named `build` (and presumably `build.bazel`). From digging around bazel-gazelle, I can surmise that this is because the check [here](https://github.com/bazelbuild/bazel-gazelle/blob/af02161/language/go/embed.go#L133) doesn't check if the possible build path is actually a file or a directory. This results in false positive for subdirectories named `build`.

Note: I've confirmed that this doesn't happen in Linux as linux is case-sensitive, and it doesn't recognize that the directory `ui/build` is a potential BUILD file named `ui/BUILD` because of the different case. Therefore, you will likely not encounter this issue in linux with the example above.

sgavinio:
**What type of PR is this?**

> Bug fix

> language/go

A bug  presently exists in `language/go/embed.go` where subdirectories with the name  `build` , e.g. `some/path/build` are being recognized as BUILD files, thereby resulting in those directories being skipped by the embed resolver. Please see more details in the linked issue.

To fix the issue, we only consider a path a `BUILD` or `BUILD.BAZEL` file if they are in fact not a directory.

Fixes #1538

### DIFF
--- a/language/go/testdata/embedsrcs/BUILD.want
+++ b/language/go/testdata/embedsrcs/BUILD.want
@@ -11,6 +11,7 @@ go_library(
         "m_static.txt",
         "n_/static.txt",
         "o_dir/_hidden.txt",
+        "p_dir/build.old/static.txt",
     ],
     importpath = "example.com/repo/embedsrcs",
     visibility = ["//visibility:public"],
--- a/language/go/embed.go
+++ b/language/go/embed.go
@@ -130,7 +130,7 @@ func newEmbedResolver(dir, rel string, validBuildFileNames []string, pkgRels map
 				return filepath.SkipDir
 			}
 			for _, name := range validBuildFileNames {
-				if _, err := os.Stat(filepath.Join(p, name)); err == nil {
+				if bFileInfo, err := os.Stat(filepath.Join(p, name)); err == nil && !bFileInfo.IsDir() {
 					// Directory already contains a build file.
 					return filepath.SkipDir
 				}
--- a/language/go/testdata/embedsrcs/embedsrcs.go
+++ b/language/go/testdata/embedsrcs/embedsrcs.go
@@ -2,5 +2,5 @@ package embedsrcs
 
 import "embed"
 
-//go:embed *m_* n_/* all:o*
+//go:embed *m_* n_/* p_dir/* all:o*
 var fs embed.FS
--- a/language/go/testdata/embedsrcs/p_dir/build.old/static.txt
+++ b/language/go/testdata/embedsrcs/p_dir/build.old/static.txt
@@ -1,0 +1,2 @@
+# p_dir/build.old/static.txt should be embedded because while the subdirectory is named `build.old`,
+# it is a directory and not an actual BUILD file.


language/go/testdata/embedsrcs/BUILD.want
```diff
@@ -11,6 +11,7 @@ go_library(
         "m_static.txt",
         "n_/static.txt",
         "o_dir/_hidden.txt",
+        "p_dir/build.old/static.txt",
```
sgavinio, line 14: Using `build.old` here as the BUILD file name in the tests is to `build.old` here: https://github.com/bazelbuild/bazel-gazelle/blob/f82a89d73ddb8a9ba32980159fbfe26a3a916148/language/go/generate_test.go#L69

fmeum (approved):

PR #1542, opened 2023-05-23
resolve: Add a resolve_regexp directive

### LINKED ISSUE #817: Support wildcards in resolve directives
This came up in discussion in #771.

Currently, `resolve` directives have two forms:

```
# gazelle:resolve lang import-string dep-string
# gazelle:resolve src-lang dep-lang import-string dep-string
```

We currently don't support any form of wildcard, so if a repository needs to use `resolve` directives for a large number of rules (for example, when using a custom naming convention or a large number of rules in a directory), it needs a large number of `resolve` directives, which can be very repetitively.

We should support some sort of wildcard in the `import-string` which could also appear in the `dep-string`. The simplest and most familiar way to do this would probably be to introduce a `resolve_regexp` directive, where `import-string` is a regular expression, and `dep-string` is something that could be passed to [`regexp.ReplaceAll`](https://pkg.go.dev/regexp?tab=doc#Regexp.ReplaceAll) (with `$1` and so on for matched groups).

DolceTriade:
Similar to the resolve directive, resolve_regexp will match the import string against a regex instead of a direct string match. This is useful when matching protos which are part of single rule.

Syntax is the same as the resolve directive except that the import string can be a regex.

ie: 
```
# gazelle:resolve_regexp py foo\..* //foo:py_lib                            
# gazelle:resolve_regexp proto go foo/.*\.proto //foo:foo_go_proto   
```

Partially addresses #817

**What type of PR is this?**

Feature

resolve/

**Other notes for review**

I didn't add tests yet since I wasn't sure if this implementation agreeable. If we think that this is something that is upstreamable, I will add tests.

Thanks!

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -921,6 +921,30 @@ The following directives are recognized:
 |   # gazelle:resolve proto go foo/foo.proto //foo:foo_go_proto                              |
 |                                                                                            |
 +---------------------------------------------------+----------------------------------------+
+| :direc:`# gazelle:resolve_regexp ...`                    | n/a                                    |
++---------------------------------------------------+----------------------------------------+
+| Specifies an explicit mapping from an import regex to a label for                         |
+| `Dependency resolution`_. The format for a resolve directive is:                           |
+|                                                                                            |
+| ``# gazelle:resolve source-lang import-lang import-string-regex label``                          |
+|                                                                                            |
+| * ``source-lang`` is the language of the source code being imported.                       |
+| * ``import-lang`` is the language importing the library. This is usually                   |
+|   the same as ``source-lang`` but may differ with generated code. For                      |
+|   example, when resolving dependencies for a ``go_proto_library``,                         |
+|   ``source-lang`` would be ``"proto"`` and ``import-lang`` would be ``"go"``.              |
+|   ``import-lang`` may be omitted if it is the same as ``source-lang``.                     |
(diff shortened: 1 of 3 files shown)


README.rst
```diff
@@ -921,6 +921,30 @@ The following directives are recognized:
...
+|                                                                                            |
+| ``# gazelle:resolve source-lang import-lang import-string-regex label``                          |
+|                                                                                            |
+| * ``source-lang`` is the language of the source code being imported.                       |
+| * ``import-lang`` is the language importing the library. This is usually                   |
+|   the same as ``source-lang`` but may differ with generated code. For                      |
+|   example, when resolving dependencies for a ``go_proto_library``,                         |
+|   ``source-lang`` would be ``"proto"`` and ``import-lang`` would be ``"go"``.              |
+|   ``import-lang`` may be omitted if it is the same as ``source-lang``.                     |
+| * ``import-string-regex`` is the regex used in source code to import a library.                 |
```
fmeum, line 937: This needs to be rewritten to explain what the regex is explained to.

fmeum (approved):

fmeum:
It does make sense to start with this limited capability. `$1` is extremely unlikely to show up in the dep-string, so we can always enable replacements as a follow-up change.

README.rst
DolceTriade: Done. PTAL!

PR #1552, opened 2023-05-30
cmd/gazelle: do not use the epoch as timestamp in diff output

### LINKED ISSUE #1528: Epoch timestamp in diff mode output confuses `git apply`

v0.30.0

v0.39.1

6.1.2

Yes.

Debian Linux. x86_64

### What did you do?

Stored the output of `gazelle update -mode=diff` in a file where a change is being made to an existing BUILD file, and then provided the file to `git apply`.

Patch correctly parsed and applied to the git worktree.

### What did you see instead?

git complains that a new file depends on old contents ([source code](https://github.com/git/git/blob/69c786637d7a7fe3b2b8f7d989af095f5f49c3a8/apply.c#L1778)).

This happens because gazelle outputs the epoch as the modification timestamp for all files. While unified format [spec](https://www.gnu.org/software/diffutils/manual/html_node/Detailed-Unified.html) does not really have any special meaning given to the epoch timestamp, GNU diff does give some special meaning to the epoch timestamp ([doc](https://www.gnu.org/software/diffutils/manual/html_node/Creating-and-Removing.html)). Given this treatment by GNU diff, `git apply` interprets the epoch timestamp as a file creation/deletion event ([source code](https://github.com/git/git/blob/69c786637d7a7fe3b2b8f7d989af095f5f49c3a8/apply.c#L895)). And so, to `git apply`, all files in the diff output of gazelle look like file creation events.

We could file this as a bug in git, but I doubt it will move anywhere because this behavior has been there for some time, and people may have depended on it. It might be more reasonable for gazelle to stop outputting epoch timestamps.

siddharthab:
**What type of PR is this?**

> Feature

> cmd/gazelle

Allows one to use the diff output of gazelle in `git apply`.

Fixes #1528

**Other notes for review**

Please let me know if you would like to have an integration test with `git`, but it may be an overkill.


cmd/gazelle/diff.go
```diff
@@ -36,7 +36,10 @@ func diffFile(c *config.Config, f *rule.File) error {
 	}
 	rel = filepath.ToSlash(rel)
 
-	date := "1970-01-01 00:00:00.000000000 +0000"
+	// Use a dummy timestamp that is not the epoch, so use one ns past the epoch.
```
fmeum, line 39: Could you mention here that we do this since the epoch is treated specially by some tools?

fmeum (approved):

fmeum (approved):

PR #1553, opened 2023-05-30
Add default directives for github.com/envoyproxy/protoc-gen-validate

mortenmj:
**What type of PR is this?**

Bug fix

Default directives

Sets default build file generation and default directives for `github.com/envoyproxy/protoc-gen-validate`

Fixes an issue when importing `github.com/envoyproxy/protoc-gen-validate` as a go module. This project is built with Bazel. The current (pre-fix) behaviour is for the WORKSPACE file in this project to be unaltered, which causes an error when it is used as a pure go module.

**Other notes for review**

The test case feels a bit light, but it does validate the fix

### DIFF
--- a/internal/bzlmod/default_gazelle_overrides.bzl
+++ b/internal/bzlmod/default_gazelle_overrides.bzl
@@ -1,12 +1,16 @@
 visibility("private")
 
 DEFAULT_BUILD_FILE_GENERATION_BY_PATH = {
+    "github.com/envoyproxy/protoc-gen-validate": "on",
     "github.com/google/safetext": "on",
 }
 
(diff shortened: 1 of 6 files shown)


fmeum (approved):

mortenmj:
> No objection to this at all, but for your own reference you might consider trying out https://github.com/bufbuild/protoc-gen-validate which is a maintained fork of that repo. I'll leave reviewing this to @fmeum but it looks good to me

Looks like https://github.com/envoyproxy/protoc-gen-validate redirects to https://github.com/bufbuild/protoc-gen-validate

Should we add both addresses?

fmeum:
What matters should be the module path and that hasn't changed, so I think we should be fine here.

PR #1554, opened 2023-05-31
fileinfo: fix not detecting 'unix' files to be OS specific

sluongng:
The build file generation logic works like this today:

```
GenerateRules()
  package.go->*goTarget.addFiles
  package.go->getPlatformStringsAddFunction()
  fileinfo.go->isOSArchSpecific()
```

Depending on whether the file has special Go build directives, Gazelle
would use a different logic to include/exclude the file into the 'srcs'
attribute of the Go targets.

In #1512, we added support for 'unix' build directive but did not tell
Gazelle to treat files with 'unix' file as OS specific file. This caused
Gazelle to mismatch the OS in a later stage and exclude the needed file
instead.

Fix that logic and provide some additional tests to reinforce the
fileinfo logic.

Also added a small knob to print stdout of Gazelle when it's run under
go_repository with `debug_mode` attribute set to True.  Most Gazelle's
logs are printed to stdout and not stderr.

**What type of PR is this?**

Bug fix

language/go

go_repository

Replaces #1551

Fix #1512

**Other notes for review**


sluongng:
cc @connyay @fmeum

fmeum (approved):
Great, thanks for the detective work!

connyay:
👍 This patch is working for my usage - sorry for the fire drill!

tempoz (commented):
LGTM, thanks for submitting that PR to gazelle!

sluongng:
another ping to request for a minor release

linzhp:
No objection for minor release, but this weekend is fully booked for me. Can @fmeum or @achew22 do it?

linzhp:
Released: https://github.com/bazelbuild/bazel-gazelle/releases/tag/v0.31.1

PR #1555, opened 2023-06-01
language/go: Emit apparent repo name of rules_go in select keys

fmeum:
**What type of PR is this?**

Bug fix

language/go

The constraint labels emitted as `select` keys now use the apparent name of the `rules_go` repository when using Bzlmod, with a fallback to `io_bazel_rules_go` if the `rules_go` module is not a declared `bazel_dep`.

### DIFF
--- a/cmd/gazelle/BUILD.bazel
+++ b/cmd/gazelle/BUILD.bazel
@@ -56,6 +56,7 @@ go_test(
         "//config",
         "//internal/wspace",
         "//testtools",
+        "@com_github_google_go_cmp//cmp",
         "@io_bazel_rules_go//go/tools/bazel:go_default_library",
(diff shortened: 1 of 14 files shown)


fmeum:
@jayconrod @tyler-french Could you take a look?

jayconrod (approved):
LGTM. Definitely nice to get rid of that `@io_bazel_`.

What does migration look like? If someone is using `io_bazel_rules_go` from `WORKSPACE`, then switches to `rules_go` from bzlmod, will Gazelle figure that out correctly and fix the `selects`? Might be worth adding a test for that if it's not covered somewhere already.

fmeum:
Migration just works, added an integration test to verify this.

PR #1556, opened 2023-06-12
Make known imports package vars public

pcj:
Gazelle has a "known imports" feature that operates during the resolve phase. First, the overrides are checked.  If an override is not found, known imports are checked.  If a known import is not found, the index is checked.

These "known import" maps are hardcoded using internal codegen tools that read @com_google_protobuf and @go_googleapis.

There is no way to opt-out of the known imports for "well known" types from @com_google_protobuf and @go_googleapis in most indexing modes.  

This makes it not possible to use gazelle to index the protos in those repositories (in file mode for example) because the resolve data should come from the index. In that case, override rules don't make sense (one would need to pre-index the repo being indexed!); the hardcoded ones correspond to -mode=package so they are not appropriate for file mode (or other modes).

In theory the unsafe package with go:linkname would make it technically possible to override these private variables from other packages, but apparently the compiler is inlining these variables because go:linkname does not work. 

This PR makes the known imports related variable public so gazelle extension authors have full control over the known imports and can set it at initialization time. In the case described above, an empty known import map is the correct one to resolve from the index.

**What type of PR is this?**

>
> Bug fix
> Feature
> Documentation
> Other

> For example:
>
> language/go
> cmd/gazelle
> go_repository
> all

Fixes #

**Other notes for review**

### DIFF
--- a/language/proto/BUILD.bazel
+++ b/language/proto/BUILD.bazel
@@ -10,7 +10,7 @@ known_imports(
     key = 0,
     package = "proto",
     value = 1,
-    var = "knownImports",
+    var = "KnownImports",
 )
 
 known_imports(
@@ -20,7 +20,7 @@ known_imports(
     key = 0,
     package = "proto",
     value = 3,
-    var = "knownProtoImports",
+    var = "KnownProtoImports",
 )
 
 known_imports(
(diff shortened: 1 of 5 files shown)


pcj:
Closing this for now, may have other workaround!

PR #1561, opened 2023-06-20
Decouple Google APIs

linzhp:
**What type of PR is this?**
> Other

language/proto

This PR stops Gazelle from mapping generated Go code by Google APIs to `go_proto_library` targets in `go_googleapi` repository. The generated Go code will be treated as regular Go packages. Gazelle still keeps the mapping for well-known types.

Fixes #

**Other notes for review**


.bazelrc
```diff
@@ -7,3 +7,5 @@ build:ci --spawn_strategy=standalone
 build:ci --genrule_strategy=standalone
 test:ci --test_strategy=standalone
 test:ci --test_output=errors
```
fmeum, line 9: This line is now redundant.

fmeum (approved):

.bazelrc
linzhp: moved to #1562

sluongng:
> After this PR, users will have to manually resolve Google APIs proto, if they want to import those proto files in their proto files.

It would be nice if we could provide some instructions for users to upgrade.

PR #1564, opened 2023-06-27
Update README.rst for v0.31.1 release

jonshea:
Update the version number and SHA in the "Setup" section of README.rst for the v0.31.1 release.

**What type of PR is this?**
Documentation


jonshea:
It also seems like `MODULE.bazel` was not updated, but I don’t think that can be fixed without a new release.

PR #1570, opened 2023-07-05
Fixes directives table

iamricard:
At some point a `|` character got dropped and the README wasn't rendering the directives table in full.


fmeum:
Fixed in https://github.com/bazelbuild/bazel-gazelle/pull/1571

PR #1572, opened 2023-07-05
Fix directives table display issue

fmeum:


fmeum:
@sluongng

fmeum:
Fixed in https://github.com/bazelbuild/bazel-gazelle/pull/1571

PR #1576, opened 2023-07-10
Preparing 0.32 release

linzhp:
Update Gazelle and rules_go version number for 0.32 release. This release intentionally avoid updating other dependencies to avoid disruptions to an already disruptive release.


fmeum:
@linzhp I pushed a commit that fixes CI. We can revert it as soon as both rules_go and gazelle have been added to the BCR together, which I will take care of after the gazelle 0.32.0 release.

fmeum:
@linzhp Yes, I think we can. After Gazelle has been released, I will merge the two BCR PRs into one and patch the version requirements to complement each other, which should work.

PR #1584, opened 2023-07-21
Add support for isolated `go_deps` usages

### LINKED ISSUE #1545: bzlmod: Isolation between two different modules using rules_go/gazelle

gazelle = 0.30.0

rules_go = 0.39.1

bazel = 6.2.0

Yes

MacOS/AARCH64

### What did you do?

Imagine we have a Bazel module that contains tools we use during building. These tools are implemented in Go and might make use of a variety of dependencies where the specific version does matter. Common examples might be modules not following a Go1 promise or where the generated output changes in non-backward compatible ways. An example of this might be github.com/prometheus/common and github.com/prometheus/prometheus.

In the main application I am importing this module but also have dependencies the same Go modules. A very basic example of this might be:

```
$ cat project1/MODULE.bazel
module(name = "project1", version = "1.0")

bazel_dep(name = "rules_go", version = "0.39.1", repo_name="rules_go")
bazel_dep(name = "gazelle", version = "0.30.0", repo_name="gazelle")

go_sdk = use_extension("@rules_go//go:extensions.bzl", "go_sdk")
go_sdk.download(version = "1.20.3")

use_repo(
    go_deps,
    "com_github_prometheus_prometheus",
)

go_deps.gazelle_override(
    directives = [
        "gazelle:proto disable",
    ],
    path = "github.com/prometheus/prometheus",
)

$ cat project1/go.mod
...
require github.com/prometheus/prometheus v0.44.0
...

$ cat project2/MODULE.bazel
module(name = "project2", version = "1.0")

bazel_dep(name = "rules_go", version = "0.39.1")
bazel_dep(name = "gazelle", version = "0.30.0")

go_sdk = use_extension("@rules_go//go:extensions.bzl", "go_sdk")
go_sdk.download(version = "1.20.3")

use_repo(
    go_deps,
    "com_github_prometheus_prometheus",
)

go_deps.gazelle_override(
    directives = [
        "gazelle:proto disable",
    ],
    path = "github.com/prometheus/prometheus",
)

$ cat project2/go.mod
...
require github.com/prometheus/prometheus v0.43.0
...

$ cat project_combine/MODULE.bazel
module(name = "foo", version = "1.0")

local_path_override(module_name = "project1", path = "../project1")
local_path_override(module_name = "project2", path = "../project2")

bazel_dep(name = "project1", version = "1.0")
bazel_dep(name = "project2", version = "1.0")

$ bazel build  --enable_bzlmod @project1//:project @project2//:project

I expect to build both versions and one being linked against one version of prometheus and one against the other version.

### What did you see instead?

```
        File "/private/var/tmp/_bazel_ich/586a70483d7911528e745954ae0c71df/external/gazelle~0.30.0/internal/bzlmod/go_deps.bzl", line 161, column 36, in _go_deps_impl
                _fail_on_non_root_overrides(module, "gazelle_override")
        File "/private/var/tmp/_bazel_ich/586a70483d7911528e745954ae0c71df/external/gazelle~0.30.0/internal/bzlmod/go_deps.bzl", line 71, column 17, in _fail_on_non_root_overrides
                fail(_report_forbidden_override(module, tag_class))
Error in fail: Using the "go_deps.gazelle_override" tag in a non-root Bazel module is forbidden, but module "project1" requests it.

If you need this override for a Bazel module that will be available in a public registry (such as the Bazel Central Registry), please file an issue at https://github.com/bazelbuild/bazel-gazelle/issues/new or submit a PR adding the required directives to the "directives.bzl" file at https://github.com/bazelbuild/bazel-gazelle/tree/master/internal/bzlmod/directives.bzl.
```

It fails as neither project is allowed to apply the necessary directives. Even after removing several of these checks, it appears to define a single repository and is linking against it.


My understanding of the visibility of repositories within a module sounded that something like this ought to be possible. The hackiest approach might be to produce a suffix for each repo dependency within a module?

### PR DESCRIPTION
**Pull request by fmeum**
Since isolated extension usages have their own sets of tags and repos, overrides can be allowed for such usages even in non-root modules. These usages also have to use their own dedicated `build_config` file for Gazelle, which requires a hack (see comment in `go_repository.bzl`).

Fixes #1545

### CODE REVIEW

**Comment by fmeum** [comment]
This needs a way to reference an extension-generated repo from another one. I will see whether I can come up with a workaround, otherwise will add functionality to Bazel 6.4.0 to support this.

**Comment by fmeum** [comment]
@tyler-french Please test at Uber, especially important due to the hack.

**Comment by fmeum** [comment]
Tests are passing now that Bazel 6.3.1 is out.

File: internal/go_repository.bzl [hunk 1]
```diff
@@ -280,6 +280,20 @@ def _go_repository_impl(ctx):
...
+        # if this repository is generated by a module extension rather than an invocation in
+        # WORKSPACE.
+        if "~" in ctx.attr.name:
+            # TODO: In Bazel 6.3.0 and earlier, there is no way to obtain a label referencing a repo
+            # generated by an extension from within that extension. We thus have to manually
+            # construct such a label pointing to the sibling `_go_repository_config` repo created by
+            # the `go_deps` extension. All extension-generated repos have names of the form
+            # `<prefix>~<name set by the extension>`.
+            extension_repo_prefix = ctx.attr.name.rpartition("~")[0] + "~"
+            repo_config = ctx.path(Label("@@" + extension_repo_prefix + "bazel_gazelle_go_repository_config//:WORKSPACE"))
```
tyler-french, line 294: When testing this I am getting `repo_config` set to `gazelle~override~_go_deps~~~go_deps~bazel_gazelle_go_repository_config/WORKSPACE`, is the "~~~" intentional?

tyler-french:
Tested at Uber. Everything builds correctly both on WORKSPACE and Bzlmod. I can also add an isolated `go_deps` myself. 
Just added one small question with label naming

internal/go_repository.bzl
fmeum: Yes, that's the magic that makes the isolation work: a more complicated repository name. The `~~~` you see is actually `~<name of the root module>~<version of the root module>~`, it just so happens that both the name and the version of the root module are always empty strings.

fmeum:
Thanks for testing!

PR #1596, opened 2023-08-01
Fix and let `# gazelle:follow` accept a glob

### LINKED ISSUE #1593: language/go/generate_test.go TestGenerateResult spuriously passing

commit fa6d6eab293df1f27ab6e97e3f6a6821e95306d2

v0.39.1

bazel 6.3.0

Yes

n/a

### What did you do?

1. Modify `language/go/testdata/bin/BUILD.want` to introduce an erroneous `srcs`
2. Run `bazel test //languge/go:go_test`

A failure.

### What did you see instead?

Success.

Digging in a bit more, the test harness loops through `regularFiles` provided by `walk.Walk`, but all of the files in the runfiles subdirectory are symlinks, thus skipped by `Walk` (as documented).  The end result is a spurious failure.

shahms:
**What type of PR is this?**
 Bug fix

 all (due to the implementation of the `gazelle:follow` directive)

This serves as a proposed resolution for #1593 but also actually implements the `follow` directive to do so.

Fixes #1593 

**Other notes for review**

### DIFF
--- a/walk/config.go
+++ b/walk/config.go
@@ -55,21 +55,11 @@ func (wc *walkConfig) isExcluded(rel, base string) bool {
 	if base == ".git" {
 		return true
 	}
-	f := path.Join(rel, base)
-	for _, x := range wc.excludes {
-		matched, err := doublestar.Match(x, f)
-		if err != nil {
-			// doublestar.Match returns only one possible error, and only if the
-			// pattern is not valid. During the configuration of the walker (see
-			// Configure below), we discard any invalid pattern and thus an error
-			// here should not be possible.
-			log.Panicf("error during doublestar.Match. This should not happen, please file an issue https://github.com/bazelbuild/bazel-gazelle/issues/new: %s", err)
-		}
-		if matched {
-			return true
-		}
-	}
-	return false
+	return matchAnyGlob(wc.excludes, path.Join(rel, base))
+}
+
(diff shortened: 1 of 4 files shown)


walk/config.go
```diff
@@ -108,6 +98,10 @@ func (cr *Configurer) Configure(c *config.Config, rel string, f *rule.File) {
 				}
 				wcCopy.excludes = append(wcCopy.excludes, path.Join(rel, d.Value))
 			case "follow":
+				if err := checkPathMatchPattern(path.Join(rel, d.Value)); err != nil {
```
fmeum, line 101: Am I right to say that this has very minor potential for breaking existing uses, namely only if a file path that was previously interpreted literally contains characters with special meaning for `doublestar.Match`? That would seem acceptable, but we should probably mention the set of these characters and how to escape them in the docs for `# gazelle:follow`.

fmeum (approved):
@linzhp Are you using `# gazelle:follow` at Uber?

walk/config.go
shahms: I should note that prior to this change `follow` was set on `walkConfig` but never actually read or used, so once potentially source of breakage is extant uses of `follow` that now become active.

linzhp:
> @linzhp Are you using `# gazelle:follow` at Uber?

No we don't

fmeum (approved):
Thanks!

PR #1597, opened 2023-08-02
add a go_test directive to enable generating go_test targets per _test.go file

### LINKED ISSUE #1594: Enable generating go_test targets per _test.go file in a directory
Similar to #1417, I'd like to add support for generating a distinct go_test target per _test.go file in a directory/package.  This would facilitate, for example, separating tests based on size, execution requirements, etc. without having to manually maintain dependencies or the file as a whole.

I've actually prepared a local patch implementing this behavior for Kythe and am hoping to be able to contribute it upstream.

shahms:
**What type of PR is this?**
Feature

 language/go

This adds a new directive controlling how go_test targets are generated.  Specifically, it adds a new `file` mode, which will generate a go_test target for each _test.go file in a package, rather than a single go_test target containing all such files.

Fixes #1594

**Other notes for review**


language/go/generate.go
```diff
@@ -343,6 +342,7 @@ func buildPackages(c *config.Config, dir, rel string, hasTestdata bool, er *embe
 				hasTestdata: hasTestdata,
 			}
 		}
+		fmt.Printf("adding %v to package %s\n", f, f.packageName)
```
fmeum, line 345: Is this a leftover debug print?

language/go/package.go
```diff
@@ -107,9 +108,14 @@ func (pkg *goPackage) addFile(c *config.Config, er *embedResolver, info fileInfo
 		if info.isCgo {
 			return fmt.Errorf("%s: use of cgo in test not supported", info.path)
 		}
-		pkg.test.addFile(c, er, info)
+		var test *goTarget
+		if getGoConfig(c).testMode == fileTestMode || len(pkg.tests) == 0 {
+			pkg.tests = append(pkg.tests, goTarget{})
+		}
+		test = &pkg.tests[len(pkg.tests)-1]
```
fmeum, line 115: Could you explain why that is the correct index to use?

language/go/testdata/tests_per_file/BUILD.want
```diff
@@ -0,0 +1,19 @@
+load("@io_bazel_rules_go//go:def.bzl", "go_test")
```
fmeum, line 1: Could you make this test more realistic by including a library that would need to be added as a dep to each test?

shahms: Sure, but deps aren't resolved by `TestGenerateRules` so they'll just show up in `_gazelle_imports`, not `deps`.

language/go/generate.go
shahms: Ugh, yeah, thanks.

language/go/package.go
shahms: Comment added. In per-file mode we'll always want to use the most-recently appended target. In defaultMode we always want to use the *only* target (which will also be the last).

shahms:
Any further comments here?

fmeum:
This looks good to me now, but I don't know the gazelle codebase well enough to be sure that this will behave correctly on updates (such as when switching test modes or adding new test files). 

Could you perhaps add tests for these cases?

shahms:
> This looks good to me now, but I don't know the gazelle codebase well enough to be sure that this will behave correctly on updates (such as when switching test modes or adding new test files).
> 
> Could you perhaps add tests for these cases?

Done. Thanks!

fmeum (approved):
Thanks for the great test coverage!

fmeum:
@shahms Could you resolve the merge conflicts and ping me afterwards?

language/go/config.go
```diff
@@ -30,6 +29,8 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/bazelbuild/bazel-gazelle/internal/module"
+
 	"github.com/bazelbuild/bazel-gazelle/config"
```
tyler-french, line 34:
```suggestion
	"github.com/bazelbuild/bazel-gazelle/internal/module"
	"github.com/bazelbuild/bazel-gazelle/config"
```
nit

shahms:
> @shahms Could you resolve the merge conflicts and ping me afterwards?

It looks like you took care of that? Thanks!

tyler-french:
Something interesting is happening. When I use `mode files` to test (from the PR), I am getting `nogo` failures from a generated `testmain.go` from rules_go, which I think should be excluded by some convention (FYI @fmeum).

Trying to investigate why this is happening, but my guess is there needs to be an update to the logic `nogo` uses to skip this `testmain.go` file.

As this is an opt in, that doesn't necessarily mean this situation is release blocking.

fmeum:
@sluongng Do you happen to have an idea why nogo triggers on `testmain.go` in this case?

sluongng:
I think we are only excluding the content of the generated coverage code and empty files right now. Not `testmain.go`.

There are 2 ways to go about fixing it: 
- Make `testmain.go` template in-compliance with the linter
- Adjust compilepkg.go to exclude `testmain.go`

I suspect the former would be easier, but I don't know the details of the analyzer @tyler-french is running in his repo.

If you choose the latter route, take a look at this variable usage to see how exclusion could be implemented. https://github.com/search?q=repo%3Abazelbuild%2Frules_go%20%2FnogoSrcsOrigin%2F&type=code

---

Regardless, @tyler-french it would be nice if you could help us add a test for this in either rules_go / gazelle or both. It would help us detect future regression earlier. Right now this PR is completely green and could be a very attractive target for merging. I know that @hawkingrei has been asking for something like this since they ran into Bazel shard count limitation (25?).

hawkingrei:
> I think we are only excluding the content of the generated coverage code and empty files right now. Not `testmain.go`.
> 
> There are 2 ways to go about fixing it:
> 
>     * Make `testmain.go` template in-compliance with the linter
> 
>     * Adjust compilepkg.go to exclude `testmain.go`
> 
> 
> I suspect the former would be easier, but I don't know the details of the analyzer @tyler-french is running in his repo.
> 
> If you choose the latter route, take a look at this variable usage to see how exclusion could be implemented. https://github.com/search?q=repo%3Abazelbuild%2Frules_go%20%2FnogoSrcsOrigin%2F&type=code
> 
> Regardless, @tyler-french it would be nice if you could help us add a test for this in either rules_go / gazelle or both. It would help us detect future regression earlier. Right now this PR is completely green and could be a very attractive target for merging. I know that @hawkingrei has been asking for something like this since they ran into Bazel shard count limitation (25?).

Yes, we have too many test case in the bazel shard. It is a good way to help us to solve this problem.

BTW, we have many goleak in the ```main_test.go``` so we should add this file into go_test per _test.go file.

language/go/package.go
```diff
@@ -34,6 +34,7 @@ import (
 type goPackage struct {
 	name, dir, rel        string
 	library, binary, test goTarget
+  tests                 []goTarget
 	proto                 protoTarget
 	hasTestdata           bool
```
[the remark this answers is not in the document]
tyler-french: Oops sorry didn't see that this was already merged. I tested and this works great, I think the `maintest.go` issues is safe because this feature is guarded by a directive anyway. Nice work!

tyler-french:
@fmeum @shahms Oops sorry didn't see that this was already merged. I agree, and think the `maintest.go` issue anyway is safe because this feature is guarded by a directive anyway. Nice work!

fmeum:
@shahms If possible, could you address @tyler-french's comments in a follow-up PR?

PR #1602, opened 2023-08-11
go_repository: fix missing default GOPROXY

sluongng:
This helps unblock the Go 1.21.0 upgrade

**What type of PR is this?**

>
> Bug fix
> Feature
> Documentation
> Other

> For example:
>
> language/go
> cmd/gazelle
> go_repository
> all

Fixes #

**Other notes for review**

### DIFF
--- a/internal/go_repository.bzl
+++ b/internal/go_repository.bzl
@@ -115,6 +115,9 @@ Authorization: Bearer RANDOM-TOKEN
 # We can't disable timeouts on Bazel, but we can set them to large values.
 _GO_REPOSITORY_TIMEOUT = 86400
 
+# Default value from `go env GOPROXY`
+_GOPROXY_DEFAULT = "https://proxy.golang.org,direct"
+
 def _go_repository_impl(ctx):
     # TODO(#549): vcs repositories are not cached and still need to be fetched.
     # Download the repository or module.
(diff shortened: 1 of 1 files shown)


fmeum:
Do you happen to have a reference for the requirement to have `GOPROXY` set? The error in CI reads `GOPROXY list is not the empty string, but contains no entries`, which to me indicates that we could get by without specifying a default if we ensure that the variable is set to the empty string.

sluongng:
I think in general, passing an empty string should be discouraged. It's hard to tell in the code sometimes because the empty string is also evaluated as False in Starlark

```
$> "a" if "" else "b"
"b"
```

I think the default value should either be `https://proxy.golang.org,direct`, which is what was set by Go toolchain by default, OR `direct`, which could be slow and not contain some dependencies versions.

In my personal testing, some public dependencies *coughAWScough* that folks depend on no longer exist via `direct` because Github tag deletion/rename, but they could still be found on `proxy.golang.org`. So I favor the former choice over the latter.

And this goes without saying but: this should not affect enterprises who wish to override the default GOPROXY with their own internal proxy.

fmeum:
Thanks for the additional explanation, that makes sense!

fmeum (approved):

PR #1605, opened 2023-08-14
Make `go_deps.from_file` compatible with Go 1.21

fmeum:
**What type of PR is this?**

Feature

`go_deps`

Ignores the new `toolchain` directive without errors and also supports fine-grained `go` versions.

### DIFF
--- a/internal/bzlmod/go_mod.bzl
+++ b/internal/bzlmod/go_mod.bzl
@@ -57,7 +57,7 @@ def parse_go_mod(content, path):
             continue
 
         if not current_directive:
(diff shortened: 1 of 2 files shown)


fmeum:
CC @sluongng

sluongng (approved):
Lgtm.

Nit: I think it's worth keeping the patch version and leave it unused.

fmeum:
As it may or may not be there, parsing out the patch version would complicate the code and tests. I would prefer to wait for a use case before we implement this, `go_mod.bzl` is only used internally.

PR #1606, opened 2023-08-14
fix: patch support for dir named workspace

michaellzc:
**What type of PR is this?**

Bug fix

all

fix https://github.com/bazelbuild/bazel-gazelle/pull/1498 to actually support dir named `workspace`

**Notes**

Upon upgrading to the latest gazelle that includes https://github.com/bazelbuild/bazel-gazelle/pull/1498, we're still seeing

```
gazelle: read /private/var/tmp/_bazel_michael/ddd60c3e827a4aa3b925c14f76ccd5bf/external/com_github_sourcegraph_controller_cdktf_gen_tfe/WORKSPACE: is a directory
```

In https://github.com/bazelbuild/bazel-gazelle/pull/1498, we missed that the `FindWORKSPACEFile` will return `/some/path/WORKSPACE` regardless if one can be found. See the final `return` statement.

https://github.com/bazelbuild/bazel-gazelle/blob/adce430fa38848f0b3805e05f9520e419a41dfd7/internal/wspace/finder.go#L38-L46

hence when we attempt to `rule.LoadWorkspaceFile`, we end up getting a `is a directory` error from `os.ReadFile`

this PR added handling on error return from `LoadWorkspaceFile` to ignore `IsDir` error as well.

https://github.com/michaellzc/bazel-gazelle/blob/86a4c72015d911a3bbe18ad57d473684300d8f1c/cmd/gazelle/fix-update.go#L143-L154

tested locally in a project:

```
| tfe/
| go.mod
| - workspace/
|   - workspace.go
```

```
$ gazelle fix -go_prefix somerpefx -repo_root .
# confirmed BUILD file is generated
$ cat workspace/BUILD.bazel

go_library(
    # ...
)
```

### DIFF
--- a/cmd/gazelle/fix-update.go
+++ b/cmd/gazelle/fix-update.go
@@ -18,6 +18,7 @@ package main
 import (
 	"bytes"
 	"context"
+	"errors"
 	"flag"
 	"fmt"
 	"io/ioutil"
@@ -26,6 +27,7 @@ import (
 	"path/filepath"
 	"sort"
 	"strings"
+	"syscall"
 
 	"github.com/bazelbuild/bazel-gazelle/config"
 	gzflag "github.com/bazelbuild/bazel-gazelle/flag"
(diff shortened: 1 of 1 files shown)


michaellzc:
CLA has been signed

alexeagle:
LGTM

jbedard (approved):
👍

fmeum (approved):
It looks like this should also work on Windows. Thanks for sending the fix!

PR #1607, opened 2023-08-16
Do not pin stable Bazel version in BCR test module

fmeum:
Match what we do for Gazelle itself, stable versions should be free of regressions and if they aren't, we should notice immediately.


linzhp:
Would this mean the test become less reproducible? For example, the test's behavior may become dependent on the bazel version available on a specific machine

fmeum:
With bazelisk as Bazel, it would always use the latest stable release. We don't pin the Bazel version for rules_go and Gazelle and only started doing that here as the latest stable version didn't support Bzlmod very well back then.

I'm also open to pinning all version, but then we would need to update all pins whenever a new patch release comes out.

linzhp:
I am worried that when a new Bazel version contains a breaking change either by design or accidentally, the CI would suddenly fail for irrelevant PRs.

fmeum:
Okay, we can look into pinning more versions than instead if pinning less.

PR #1611, opened 2023-08-17
Upgrade `golang.org/x/mod` Vendor to Support Go 1.21

alan910127:
**What type of PR is this?**

Bug fix

`go_deps`

This PR builds on the work done in PR #1605, which aims to add support for Go 1.21's updated version scheme.
However, the current `golang.org/x/mod` in the `vendor/` directory is not upgraded and is still using the old regex which does not support patch and prerelease versions.

Without this upgrade, users might get into the following error when running `bazel run //:gazelle`:

```shell
gazelle: reading module paths from /path/to/go.mod: /path/to/go.mod:3: invalid go version '1.21.0': must match format 1.23
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -9,7 +9,7 @@ require (
 	github.com/fsnotify/fsnotify v1.6.0
 	github.com/google/go-cmp v0.5.9
 	github.com/pmezard/go-difflib v1.0.0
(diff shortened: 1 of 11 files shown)


fmeum (approved):
Thanks!

PR #1613, opened 2023-08-17
Fix kinds and stmt

### LINKED ISSUE #1612: mismatched rules kind & stmt
[v0.32.0 ](https://github.com/bazelbuild/bazel-gazelle/releases/tag/v0.32.0)

[v0.41.0](https://github.com/bazelbuild/rules_go/releases/tag/v0.41.0)

bazel 6.3.2

Yes

Ubuntu 20.04.6 LTS x86_64

### What did you do?
`BUILD.bazel` contains rule packed into struct:
```bzl
selects.config_setting_group(
    name = "a",
    match_any = [
        "//:config_a",
        "//:config_b",
    ],
)
```

Parsed rule from the input by [ruleFromExpr()](https://github.com/bazelbuild/bazel-gazelle/blob/master/rule/rule.go#L757) function sets it's _kind_ to a parsed expression:
```go
kind := call.X
```
which in this case is:
```go
DotExpr{
    X: bzl.Ident{Name: "selects"},
    Name: "config_setting_group"
}
```

However when one wants to generate such rule with [NewRule()](https://github.com/bazelbuild/bazel-gazelle/blob/master/rule/rule.go#L723) 
```go
r := rule.NewRule("selects.config_setting_group", "config_group_name")
```
the kind will be set to 
```go
kindIdent := &bzl.Ident{Name: kind}
```

This changed the behavior e.g. for fixing loads where different statement is parsed. 

Both ways of creating rules should allow instantiating conforming rules to each other.
Kind should be exactly the same to be matched with type or by [bzl.FormatString()](https://github.com/bazelbuild/bazel-gazelle/blob/master/vendor/github.com/bazelbuild/buildtools/build/print.go#L56)
Identifier to search for load symbols should be also the same for both rules (parsed & generated).

### What did you see instead?
for loads:
```go
var loads = []rule.LoadInfo{
	{
		Name: "@bazel_skylib//lib:selects.bzl",
		Symbols: []string{
			"selects",
		},
	},
}
```
and an empty `BUILD.bazel` file (no `selects.config_setting_group` rules inside) load statement is not resolved correctly therefore it is missed in the final output.

birunts:
**What type of PR is this?**
Bug fix

internal/language
rule
tests

Bazel allows to pack rules or macros within a struct, e.g. convenience interfaces for select()
(see https://github.com/bazelbuild/bazel-skylib/blob/main/lib/selects.bzl)

Generated rule of `selects.config_setting_group` by [rule.NewRule()](https://github.com/bazelbuild/bazel-gazelle/blob/master/rule/rule.go#L723) is different with the rule parsed from the input file i.e. [ruleFromExpr()](https://github.com/bazelbuild/bazel-gazelle/blob/master/rule/rule.go#L757) in a way of different _kind_ and identifier when fixing loads ([FixLoads()](https://github.com/bazelbuild/bazel-gazelle/blob/master/merger/fix.go#L31)).

Newly generated rule:
```go
r := rule.NewRule("selects.config_setting_group", "config_group_name")
```
together with loads:
```go
var loads = []rule.LoadInfo{
	{
		Name: "@bazel_skylib//lib:selects.bzl",
		Symbols: []string{
			"selects",
		},
	},
}
```
should produce in the `BUILD.bazel` output::
```bzl
load("@bazel_skylib//lib:selects.bzl", "selects")

selects.config_setting_group(
    name = "config_group_name",
    ...
)
```

Fixes #1612

**Other notes for review**

### DIFF
--- a/rule/rule.go
+++ b/rule/rule.go
@@ -730,8 +730,9 @@ type attrValue struct {
 
 // NewRule creates a new, empty rule with the given kind and name.
 func NewRule(kind, name string) *Rule {
-	kindIdent := &bzl.Ident{Name: kind}
+	kindIdent := createDotExpr(kind)
 	call := &bzl.CallExpr{X: kindIdent}
+
 	r := &Rule{
 		stmt:    stmt{expr: call},
 		kind:    kindIdent,
@@ -750,6 +751,47 @@ func NewRule(kind, name string) *Rule {
 		r.attrs["name"] = nameAttr
 	}
 	return r
+}
+
+// Creates `bzl.Expr` for a kind which
+// is either `*bzl.DotExpr` or `*bzl.Ident`.
+//
+// For `myKind` kind it returns:
+//  &bzl.Ident{
+//      Name: "myKind"
+//  }
+//
+// For `myKind.inner` kind it returns:
+//  &bzl.DotExpr{
+//      Name: "inner",
+//      X: &bzl.Ident {
+//          Name: "myKind"
+//      }
+//  }
+func createDotExpr(kind string) bzl.Expr {
+	var expr bzl.Expr
+	parts := strings.Split(kind, ".")
+
+	if len(parts) > 1 {
+		// last `parts` element is the main bzl.DotExpr
+		var dotExpr *bzl.DotExpr = &bzl.DotExpr{Name: parts[len(parts)-1]}
+
+		_pDot := dotExpr
+
+		for idx := len(parts) - 2; idx > 0; idx-- {
+			d := &bzl.DotExpr{Name: parts[idx]}
+			_pDot.X = d
+			_pDot = d
+		}
+
+		// first `parts` element is the identifier
+		_pDot.X = &bzl.Ident{Name: parts[0]}
+		expr = dotExpr
+	} else {
+		expr = &bzl.Ident{Name: kind}
+	}
+
+	return expr
 }
 
 func isNestedDotOrIdent(expr bzl.Expr) bool {
@@ -999,7 +1041,9 @@ func (r *Rule) sync() {
 	}
 
 	call := r.expr.(*bzl.CallExpr)
-	call.X = r.kind
+
+	// update `call.X` (e.g.: "# gazelle:map_kind")
+	call.X = createDotExpr(r.Kind())
 
 	if len(r.attrs) > 1 {
 		call.ForceMultiLine = true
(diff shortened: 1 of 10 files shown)


rule/rule.go
```diff
@@ -721,8 +721,20 @@ type Rule struct {
...
 func NewRule(kind, name string) *Rule {
+	var call *bzl.CallExpr
 	kindIdent := &bzl.Ident{Name: kind}
-	call := &bzl.CallExpr{X: kindIdent}
+
+	parts := strings.Split(kind, ".")
+	if len(parts) > 1 {
+		// Create the same `CallExpr` like parser does
+		call = &bzl.CallExpr{X: &bzl.DotExpr{
+			Name: strings.Join(parts[1:], "."),
```
fmeum, line 731: I would have expected multiple `.`s to result in nested `DotExpr`s. Which existing part of Gazelle doesn't do this?

birunts: Yeah, you're right. I should have checked it more carefully. I'll correct that.

rule/rule.go
```diff
@@ -730,8 +730,16 @@ type attrValue struct {
...
+	var call *bzl.CallExpr
 	kindIdent := &bzl.Ident{Name: kind}
-	call := &bzl.CallExpr{X: kindIdent}
+
+	parts := strings.Split(kind, ".")
+	if len(parts) > 1 {
+		call = &bzl.CallExpr{X: createDotExpr(parts)}
+	} else {
+		call = &bzl.CallExpr{X: kindIdent}
+	}
```
fmeum, line 741: Could you extract out this part into `createDotExpr`? It can return a generic `Expr` that is either a `DotExpr` or an `Ident`.

fmeum (approved):
Looks good now, with one more comment.

Please mention me after making your changes as I don't get notifications for pushed on PRs.

rule/rule.go
birunts: Sure, changes updated.

rule/rule.go
```diff
@@ -730,8 +730,11 @@ type attrValue struct {
 
 // NewRule creates a new, empty rule with the given kind and name.
 func NewRule(kind, name string) *Rule {
+	var call *bzl.CallExpr
 	kindIdent := &bzl.Ident{Name: kind}
```
fmeum, line 734: Thanks for the update, just one quick question: Why do we still need the `kindIdent`? Couldn't we set `Rule.kind` to the `DotExpr` or would that break other parts of Gazelle? Having two different AST representations in this structure makes me a bit worried.

birunts: Actually I'm not sure to be honest, I don't know all Gazelle specifics, but afair `kind` needs to conform to one defined in `rule.KindInfo` map, e.g. for
```
var kinds = map[string]rule.KindInfo{
	"selects.config_setting_group": {
		...
	},
}
```
with `&bzl.Ident{Name: kind}` is matched to `selects.config_setting_group` string. If the same kind can be grabbed form the `DotExpr` then it will work as you're saying.

birunts: `Rule.Kind()` string is made of `bzl.createDotExpr()` so it should work :)

fmeum (approved):
Looks great, thanks!

PR #1617, opened 2023-08-25
bzlmod: Verify that the lockfile is platform-independent

fmeum:
By committing a single lockfile and running CI with `--lockfile_mode=error`, we can ensure that the same lockfile works across OSes and architectures.

Since the lockfile can get pretty large, we exclude it from the release archive and hence also don't check the lockfile in the BCR presubmit.


fmeum:
CC @tyler-french @linzhp

linzhp (approved):

PR #1618, opened 2023-08-26
bug fix: don't generate invalid go_binary rules for empty main packages

andyscott:
**What type of PR is this?**

Bug fix

language/go

Presently, gazelle generates `go_binary` targets for main packages with library code even if
the library doesn't have a main function. The no-main-function `go_binary` targets don't compile.

This PR updates gazelle to skip generation of `go_binary` for main packages where a main
function isn't present.

**Other notes for review**

The first commits add a test showing the undesirable behavior. The tests pass because we aren't compiling the generated BUILD files -- but if we did compile them, it would fail.

The subsequent commits fix the issue and update the tests.

### DIFF
--- a/cmd/gazelle/fix_test.go
+++ b/cmd/gazelle/fix_test.go
@@ -172,7 +172,7 @@ func TestNoChanges(t *testing.T) {
 	defer os.RemoveAll(dir)
 
 	goFile := filepath.Join(dir, "main.go")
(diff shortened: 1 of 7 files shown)


fmeum (approved):

fmeum:
Thanks!

andyscott:
Sure thing-- Thanks for the quick review too!

PR #1623, opened 2023-08-30
Add an entry for "github.com/googleapis/gax-go/v2" to default_gazelle_overrides.bzl

andrewmbenton:
**What type of PR is this?**

Other

bzlmod default config

When updating to v0.32.0 anyone with a transient dependency on `github.com/googleapis/gax-go/v2` will need to add an override `gazelle:proto disable` directive. See here for discussion: https://github.com/bazelbuild/rules_go/issues/3625#issuecomment-1674275131

Adding the directive as a default should prevent a bunch of manual effort.

### DIFF
--- a/internal/bzlmod/default_gazelle_overrides.bzl
+++ b/internal/bzlmod/default_gazelle_overrides.bzl
@@ -23,6 +23,9 @@ DEFAULT_DIRECTIVES_BY_PATH = {
         "gazelle:build_file_name BUILD.bazel",
         "gazelle:build_file_proto_mode disable_global",
     ],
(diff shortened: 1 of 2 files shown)


fmeum:
@andrewmbenton You can update the lockfile by running `bazel mod graph` in `tests/bcr` (or I can do it for you before the merge).

@tyler-french @linzhp Since I don't have personal experience with this package, could you check that this is a safe default?

linzhp (approved):
LGTM. We may also consider having `gazelle:proto disable` for all external repos by default

PR #1632, opened 2023-09-13
Set runfiles environment variables for Gazelle

fmeum:
This ensures that Gazelle itself or a language linked into it can discover runfiles no matter whether it needs `RUNFILES_MANIFEST_FILE` or `RUNFILES_DIR`. The latter is especially relevant for Python, which is run as part of the Gazelle Python plugin.


fmeum:
FYI @linzhp

PR #1665, opened 2023-11-13
feat(bzlmod): allow patches in `archive_override`s

tyler-french:
There are times when we need to specify custom `urls`, custom `directives` and custom `patches` to an archive. 

We should allow users to apply `patches` to the `archive_override`, similar to https://docs.bazel.build/versions/5.0.0/skylark/lib/globals.html#archive_override

### DIFF
--- a/internal/bzlmod/go_deps.bzl
+++ b/internal/bzlmod/go_deps.bzl
@@ -152,6 +152,8 @@ def _process_archive_override(archive_override_tag):
         urls = archive_override_tag.urls,
         sha256 = archive_override_tag.sha256,
         strip_prefix = archive_override_tag.strip_prefix,
+        patches = archive_override_tag.patches,
+        patch_strip = archive_override_tag.patch_strip,
     )
 
 def _extension_metadata(module_ctx, *, root_module_direct_deps, root_module_direct_dev_deps):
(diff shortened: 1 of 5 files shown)


internal/bzlmod/go_deps.bzl
```diff
@@ -220,8 +214,8 @@ def _go_deps_impl(module_ctx):
                 outdated_direct_dep_printer = fail
 
         _process_overrides(module_ctx, module, "gazelle_override", gazelle_overrides, _process_gazelle_override)
-        _process_overrides(module_ctx, module, "module_override", module_overrides, _process_module_override, archive_overrides)
-        _process_overrides(module_ctx, module, "archive_override", archive_overrides, _process_archive_override, module_overrides)
+        _process_overrides(module_ctx, module, "module_override", module_overrides, _process_module_override)
+        _process_overrides(module_ctx, module, "archive_override", archive_overrides, _process_archive_override)
```
fmeum, line 218: What's a use case for mixing `module_override` and `archive_override`? That logically seems wrong to me, maybe we should rather duplicate some of the attrs?

tyler-french: I have a Go module that's downloaded from a specific `url` (not through GoProxy) and then I need to add some gazelle directives and apply some patches to make it build correctly with Bazel.

fmeum: I see why you need to combine `archive_override` with `gazelle_override` in that case, but why the need for `module_override`? I think that we should add any of its capabilities you need here to `archive_override` instead: `archive_override` and `module_override` both allow specifying a different source for the Go modules, having both could result in breakages or confusing behavior.

tyler-french: I see what you mean. I updated this to better follow the design of the normal `archive_override` from Bzlmod

fmeum (approved):

PR #1677, opened 2023-11-30
feat(tools): add a tool to automate the generation of go_deps overrides

### LINKED ISSUE #1675: feat(bzlmod): script to generate `go_deps.{}_override`s from `go_repository` rules in a `.bzl` file
For migration assistance, we should have an automatic way to read the repos loaded from a `.bzl` files and automatically translate them to `go_deps` overrides.

This would probably look something like this:

```
import (
	"github.com/bazelbuild/bazel-gazelle/repo"
)

func run() error (
	w, err := rule.LoadWorkspaceFile("/repo-root/WORKSPACE", "")
	if err != nil {
		return nil, err
	}
	repos, fileMap, err := repo.ListRepositories(w)
	if err != nil {
		return nil, err
	}
	// (maybe) sort the repos.

	for _, r := range repos {
		if r.Kind() == "go_repository" {
			for _, k := range r.AttrKeys() {
				// Analyze the keys and print some overrides for "patches", "urls", "build_directives", etc.
			}
		}
	}
```
These could even be output as `bulldozer` commands to apply directly to

tyler-french:
This PR creates a tool that converts either a repository macro file or a WORKSPACE into a set of `go_deps.{}_overrides`. This is supposed to ease the migration to Bzlmod.

For example, if you were to run on this `go_repository` rule:

```
    go_repository(
        name = "com_github_bazelbuild_bazel_watcher",
        build_extra_args = ["-go_naming_convention_external=go_default_library"],
        build_file_generation = "off",  # keep
        build_file_proto_mode = "disable",
        importpath = "github.com/bazelbuild/bazel-watcher",
        patch_args = ["-p1"],
        patches = [
            # Remove it after they release this PR https://github.com/bazelbuild/bazel-watcher/pull/627
            "//patches:com_github_bazelbuild_bazel_watcher-go-embed.patch",
        ],
        sum = "h1:EfJzkMxJuNBGMVdEvkhiW7pAMwhaegbmAMaFCjLjyTw=",
        version = "v0.23.7",
    )
```

It would output
```
go_deps = use_extension("@bazel_gazelle//:extensions.bzl", "go_deps")

go_deps.gazelle_override(
    build_extra_args = ["-go_naming_convention_external=go_default_library"],
    build_file_generation = "off",
    directives = ["gazelle:proto disable"],
    path = "github.com/bazelbuild/bazel-watcher",
)

go_deps.module_override(
    patch_strip = 1,
    patches = [
        # Remove it after they release this PR https://github.com/bazelbuild/bazel-watcher/pull/627
        "//patches:com_github_bazelbuild_bazel_watcher-go-embed.patch",
    ],
    path = "github.com/bazelbuild/bazel-watcher",
)
```

Fixes #1675

### DIFF
--- a/tools/BUILD.bazel
+++ b/tools/BUILD.bazel
@@ -3,6 +3,7 @@ filegroup(
     testonly = True,
     srcs = [
         "BUILD.bazel",
+        "//tools/override-generator:all_files",
         "//tools/releaser:all_files",
     ],
     visibility = ["//visibility:public"],
--- a/tools/override-generator/README.md
+++ b/tools/override-generator/README.md
@@ -1,0 +1,25 @@
+# Bazel Go Repository to Gazelle Overrides Converter
+
+## WARNING: In development. This tool may functionally change and an experimental project.
+
+## Description
+This script converts `go_repository` rules to Gazelle `go_deps` overrides to assist in the migration to Bzlmod.
+
+## Usage
+Run the script with the following flags:
+
+- `--macro`: Path to a macro file to translate to overrides.
+- `--def_name`: Name of the macro's function name that loads the `go_repository` rules.
+- `--workspace`: Path to the workspace file, to load translate all rules loaded from the workspace.
+- `--output`: Path to the output file.
+- `--help`: Show help message.
+
+Only one of `--macro` or `--workspace` should be specified. The `--def_name` is required when `--macro` is specified.
+
+Example:
+```
+go run main.go --workspace /path/to/WORKSPACE --output /path/to/output.bzl
+```
+
+## Output
+The script outputs a file containing `go_deps` overrides based on the provided `go_repository` rules.
--- a/tools/override-generator/main.go
+++ b/tools/override-generator/main.go
@@ -1,0 +1,291 @@
+package main
+
+import (
+	"errors"
+	"flag"
+	"fmt"
+	"io"
+	"log"
+	"os"
+	"sort"
+	"strconv"
+	"strings"
+
+	"github.com/bazelbuild/bazel-gazelle/repo"
+	"github.com/bazelbuild/bazel-gazelle/rule"
+	"github.com/bazelbuild/buildtools/build"
+)
+
+const (
+	_name  = "override-generator"
+	_usage = "usage: This script converts `go_repository` rules to Gazelle `go_deps` overrides to assist in the migration to Bzlmod."
+)
+
+// override kinds.
+const (
+	_goDepsf         = `go_deps = use_extension("%s//:extensions.bzl", "go_deps")`
+	_gazelleOverride = "go_deps.gazelle_override"
+	_archiveOverride = "go_deps.archive_override"
+	_moduleOverride  = "go_deps.module_override"
+)
+
+// attribute constants that are used multiple times.
+const (
+	_buildFileGenerationAttr = "build_file_generation"
+	_buildFileProtoModeAttr  = "build_file_proto_mode"
+	_patchArgsAttr           = "patch_args"
+	_buildDirectivesAttr     = "build_directives"
+	_directivesAttr          = "directives"
+)
+
+var _defaultValues = map[string]string{
+	_buildFileGenerationAttr: "auto",
+	_buildFileProtoModeAttr:  "default",
+}
+
+var _mapAttrToOverride = map[string]string{
+	_buildDirectivesAttr:     _gazelleOverride,
+	_buildFileGenerationAttr: _gazelleOverride,
+	_patchArgsAttr:           _moduleOverride,
+	"patches":                _moduleOverride,
+	"build_extra_args":       _gazelleOverride,
+	"urls":                   _archiveOverride,
+	"strip_prefix":           _archiveOverride,
+	"sha256":                 _archiveOverride,
+}
+
+var _attrOverrideKeys = map[string]string{
+	_buildDirectivesAttr: _directivesAttr,
+}
+
+type overrideSet map[string]*rule.Rule
+
+type mainArgs struct {
+	macroPath       string
+	workspace       string
+	defName         string
+	outputFile      string
+	gazelleRepoName string
+}
+
+func main() {
+	a, err := parseArgs(os.Stderr, os.Args[1:])
+	if err != nil && !errors.Is(err, flag.ErrHelp) {
+		log.Fatalf("%+v", err)
+	}
+	if err := run(*a, os.Stderr); err != nil {
+		log.Fatal(err)
+	}
+}
+
+func parseArgs(stderr io.Writer, osArgs []string) (*mainArgs, error) {
+	a := &mainArgs{}
+	flag := flag.NewFlagSet(_name, flag.ContinueOnError)
+	flag.SetOutput(stderr)
+	flag.Usage = func() {
+		fmt.Fprintf(flag.Output(), _usage)
+		flag.PrintDefaults()
+	}
+	flag.StringVar(&a.macroPath, "macro", "", "path to the macro file")
+	flag.StringVar(&a.workspace, "workspace", "", "path to workspace")
+	flag.StringVar(&a.defName, "def_name", "", "name of the macro definition")
+	flag.StringVar(&a.outputFile, "output", "", "path to the output file")
+	flag.StringVar(&a.gazelleRepoName, "gazelle_repo_name", "@bazel_gazelle", "name of the gazelle repo to load go_deps, (default: @bazel_gazelle)")
+	flag.Parse(osArgs)
+
+	if a.macroPath != "" && a.workspace != "" {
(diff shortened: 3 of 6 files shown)


tools/BUILD.bazel
```diff
@@ -3,6 +3,7 @@ filegroup(
     testonly = True,
     srcs = [
         "BUILD.bazel",
+        "//tools/override-generator:all_files",
```
linzhp, line 6: The user-facing tools are under `cmd` and maintainer-facing tools are under `tools`.

tyler-french: I see. I am not sure how much we want to open this up to be a public CLI and maintain it indefinitely. What do you think?

tools/override-generator/main.go
```diff
@@ -0,0 +1,286 @@
...
+	if o, ok := set[kind]; ok {
+		override = o
+	}
+	if list := override.Attr(_directivesAttr); list != nil {
+		if listExpr, ok := list.(*build.ListExpr); ok {
+			listExpr.List = append(listExpr.List, &build.StringExpr{Value: directive})
+		}
+		return
+	}
+	override.SetAttr(_directivesAttr, []string{directive})
```
linzhp, line 251:
```suggestion
	directives = override.AttrStrings(_directivesAttr)
	directives = append(directives, directive)
	override.SetAttr(_directivesAttr, directives)
```

a bit simpler

tools/override-generator/main.go
```diff
@@ -0,0 +1,286 @@
...
+	for _, li := range v.List {
+		if liStr, ok := li.(*build.StringExpr); ok {
+			if strings.HasPrefix(liStr.Value, "-p") {
+				numStr := strings.TrimPrefix(liStr.Value, "-p")
+				if num, err := strconv.Atoi(numStr); err == nil {
+					override.SetAttr("patch_strip", num)
+					return
+				}
+			}
+		}
```
linzhp, line 265:
```suggestion

	patch_args = override.AttrStrings("patch_args")
	for _, arg := range patch_args {
		if !strings.HasPrefix(arg, "-p") {
			continue
		}
		numStr := strings.TrimPrefix(arg, "-p")
		if num, err := strconv.Atoi(numStr); err == nil {
			override.SetAttr("patch_strip", num)
			return
		}	
	}
```
simpler and less nesting

tools/override-generator/main.go
```diff
@@ -0,0 +1,286 @@
...
+	}
+
+	// Since "build_file_proto_mode" is added to the "directives", we need
+	// to run it after the fact to make sure that "directives" is set.
+	applyBuildFileProtoMode(r, set)
+	return set
+}
+
+func applyBuildFileProtoMode(r *rule.Rule, set overrideSet) {
+	a := _buildFileProtoModeAttr
```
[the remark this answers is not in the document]
tyler-french: Yep, this was a product of moving code around to make the logic work. Good catch.

tools/override-generator/main.go
```diff
@@ -0,0 +1,286 @@
...
+
+var _defaultValues = map[string]string{
+	_buildFileGenerationAttr: "on",
+	_buildFileProtoModeAttr:  "default",
+}
+
+var _mapAttrToOverride = map[string]overrideKeys{
+	_patchesAttr: {
+		kind:  _moduleOverride,
+		attrs: []string{_patchesAttr, _patchArgsAttr},
```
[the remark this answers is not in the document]
tyler-french: I adjusted this to be more efficient and only index into the map once.

tools/override-generator/main.go
tyler-french: Oh nice! `AttrStrings` is much nicer

tools/override-generator/main.go
tyler-french: Yep! I actually need to get the `AttrStrings` outside of this function but this works nicely.

tyler-french:
@fmeum Can you also take a look to make sure all my test cases are accurate, and I'm not missing any situations we want to test.

tools/override-generator/main_test.go
```diff
@@ -0,0 +1,202 @@
...
+                build_extra_args = ["-go_naming_convention_external=go_default_library"],
+                build_file_generation = "off",
+                directives = ["gazelle:proto disable"],
+                path = "github.com/bazelbuild/bazel-watcher",
+            )
+
+            go_deps.module_override(
+                patch_strip = 1,
+                patches = [
+                    #RemoveitaftertheyreleasethisPRhttps://github.com/bazelbuild/bazel-watcher/pull/627
```
fmeum, line 79: Are we losing spaces in comments?

tools/override-generator/main_test.go
```diff
@@ -0,0 +1,202 @@
...
+                build_file_proto_mode = "disable",
+                importpath = "github.com/ClickHouse/clickhouse-go/v2",
+                sum = "h1:Nbl/NZwoM6LGJm7smNBgvtdr/rxjlIssSW3eG/Nmb9E=",
+                version = "v2.0.12",
+            )`,
+            want: `go_deps.gazelle_override(
+                build_extra_args = ["-go_naming_convention_external=go_default_library"],
+                build_file_generation = "on",
+                directives = [
+                    "gazelle:resolvegogithub.com/ClickHouse/clickhouse-go/v2/external@com_github_clickhouse_clickhouse_go_v2//external",
```
fmeum, line 103: This is also missing spaces.

tyler-french: This was an issue with my test helper to try to compare the rules without spaces. I guess spaces are important in some situations, so I changed the helper to only trim leading and trailing spaces on each line.

tools/override-generator/main_test.go
tyler-french: This was an issue with my test helper to try to compare the rules without spaces. I guess spaces are important in some situations, so I changed the helper to only trim leading and trailing spaces on each line.

tools/override-generator/README.md
```diff
@@ -0,0 +1,23 @@
...
+- `--def_name`: Name of the macro's function name that loads the `go_repository` rules.
+- `--workspace`: Path to the workspace file, to load translate all rules loaded from the workspace.
+- `--output`: Path to the output file.
+- `--help`: Show help message.
+
+Only one of `--macro` or `--workspace` should be specified. The `--def_name` is required when `--macro` is specified.
+
+Example:
+```
+go run main.go --workspace /path/to/WORKSPACE --output /path/to/output.bzl
```
linzhp, line 19: How are we going to use this output.bzl? Manually copying and merging it to MODULE.bazel?

tools/override-generator/main.go
```diff
@@ -0,0 +1,280 @@
...
+var _defaultValues = map[string]string{
+	_buildFileGenerationAttr: "auto",
+	_buildFileProtoModeAttr:  "default",
+}
+
+var _mapAttrToOverride = map[string]string{
+	_buildDirectivesAttr: _gazelleOverride,
+	_buildFileGenerationAttr: _gazelleOverride,
+	_patchArgsAttr: _moduleOverride,
+	"patches": _moduleOverride,
```
linzhp, line 49: Should we map `patch_args` and `patches` to archive_override so we don't need to merge module_override with archive_override later?

tyler-french: If there's no `urls` archive specified, we want to use the `module_override` (rather than the `archive_override`. Naming the go_deps override `archive_override` was explicitly chosen to mirror the Bzlmod `archive_override`, meaning that we are explicitly loading this from a different archive URL.

If we don't want to do this, then we would need to adjust the implementation of the overrides and documentation.

tools/override-generator/README.md
tyler-french: This is the only way I can think of... I guess we could output to `os.Stdout` instead? 

I guess we could output a `buildozer` command somehow, but I am not too familiar with the changes needed to do this, and it would probably lose the comments. 

Once the MODULE.bazel file allows segmentation (cc @fmeum), maybe one of the segments can be `go_deps_overrides.bzl` and the core `MODULE.bazel` loads this file. Then, a repo can automatically sync Bzlmod and workspace deps.

linzhp: We can keep it simple like this for now. But if we start to run into scenarios when people edits the go_deps overrides in MODULE.bazel, we may want to call github.com/bazelbuild/buildtools/edit package to merge the changes

tyler-french:
Converting to draft to update docs to reference this tool

tyler-french:
> Converting to draft to update docs to reference this tool

Looks like these docs are in the `rules_go` repo.

fmeum (approved):

tyler-french:
Add more parameters to specify defaults, rather than hard coding them

fmeum (approved):
Looks great!

PR #1679, opened 2023-12-03
bzlmod: support go.mod replace directives with a version qualifier on the left

andyscott:
**What type of PR is this?**

Bug fix / Feature

bzlmod support for Go

This adds support for `go.mod` replace directives with a version qualifier as described by https://go.dev/ref/mod#go-mod-file-replace.

Fixes #

**Other notes for review**

### DIFF
--- a/internal/bzlmod/go_mod.bzl
+++ b/internal/bzlmod/go_mod.bzl
@@ -141,7 +141,7 @@ def _parse_directive(state, directive, tokens, comment, path, line_no):
     elif directive == "replace":
         # A replace directive might use a local file path beginning with ./ or ../
         # These are not supported with gazelle~go_deps.
-        if len(tokens) == 3 and tokens[2][0] == ".":
+        if (len(tokens) == 3 and tokens[2][0] == ".") or (len(tokens) > 3 and tokens[3][0] == "."):
             fail("{}:{}: local file path not supported in replace directive: '{}'".format(path, line_no, tokens[2]))
 
         # pattern: replace from_path => to_path to_version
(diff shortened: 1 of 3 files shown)


fmeum (approved):
Thanks!

@tyler-french Could you test this?

internal/bzlmod/go_mod.bzl
```diff
@@ -144,13 +144,26 @@ def _parse_directive(state, directive, tokens, comment, path, line_no):
         if len(tokens) == 3 and tokens[2][0] == ".":
```
tyler-french, line 144: We probably also need to check here that if a `from_version` is specified it is not being replaced with a local path
```suggestion
        if (len(tokens) == 3 and tokens[2][0] == ".") or len(tokens) > 3 and tokens[3][0] == ".":
```

andyscott: Good catch! Fixed.

internal/bzlmod/go_mod.bzl
```diff
@@ -144,13 +144,26 @@ def _parse_directive(state, directive, tokens, comment, path, line_no):
...
+                to_path = tokens[2],
+                version = _canonicalize_raw_version(tokens[3]),
+            )
+        # pattern: replace from_path from_version => to_path to_version
+        elif len(tokens) == 5 and tokens[2] == "=>":
+            state["replace"][tokens[0]] = struct(
+                from_version = _canonicalize_raw_version(tokens[1]),
+                to_path = tokens[3],
+                version = _canonicalize_raw_version(tokens[4]),
+            )
```
tyler-french, line 160: optional: I liked assigning the `from_path` beforehand so that it's clear to a user that this is the key to the state structure, but this is definitely opinionated.
```suggestion
        from_path = tokens[0]
        # pattern: replace from_path => to_path to_version
        if len(tokens) == 4 and tokens[1] == "=>":
            state["replace"][from_path] = struct(
                from_version = None,
                to_path = tokens[2],
                version = _canonicalize_raw_version(tokens[3]),
            )
        # pattern: replace from_path from_version => to_path to_version
        elif len(tokens) == 5 and tokens[2] == "=>":
            state["replace"][from_path] = struct(
                from_version = _canonicalize_raw_version(tokens[1]),
                to_path = tokens[3],
                version = _canonicalize_raw_version(tokens[4]),
            )
```

andyscott: Done

tyler-french (approved):
LGTM

PR #1680, opened 2023-12-06
Fix partial execution with mapped kinds

HALtheWise:
Previously, if mapped kinds were in use (i.e. `go_library` -> `go_custom_library`), and `gazelle update` was being run on just one directory, Gazelle's indexing system would fail to correctly index build rules of the mapped-to kind in directories that were not being re-built.

This is because the logic to handle mapped kinds was triggered only when Generate() was run, not for Imports().

This PR solves the issue with the smallest fix. The whole kind-mapping system is honestly kinda messy, and I didn't want to propose a major refactor.

**What type of PR is this?**

Bug fix

cmd/gazelle

**Other notes for review**

### DIFF
--- a/cmd/gazelle/fix-update.go
+++ b/cmd/gazelle/fix-update.go
@@ -327,6 +327,9 @@ func runFixUpdate(wd string, cmd command, args []string) (err error) {
 		// directory, just index the build file and move on.
 		if !update {
 			if c.IndexLibraries && f != nil {
+				for _, repl := range c.KindMap {
+					mrslv.MappedKind(rel, repl)
+				}
 				for _, r := range f.Rules {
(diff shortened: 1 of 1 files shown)


fmeum:
Do you see a way to add a test case for this behavior? I agree that a refactoring would be very difficult, but a test would probably be even more valuable and potentially easier to realize.

eric-skydio:
Unfortunately I didn't see an easy place to add a test, although you're probably more familiar with the codebase if you have suggestions. The problem is that the primary testing harnesses (generationtest, etc) don't exercise the partial-execution case, and it didn't look easy to hack in.

For what it's worth, we've been running this on our (large) repo for a couple weeks now and it seems to work.

PR #1683, opened 2023-12-06
fix(rule): IsEmpty should be false `# keep` on `NonEmptyAttrs`

tyler-french:
Rules like this should not be deleted and marked as empty:
```
go_test(
    name = "rule_test",
    srcs = [
        "@something//some_file", // # keep
    ],
    embed = [":rule"],
)
```
If an attr is a part of the `NonEmptyAttrs`, a user's explicit `# keep` on the attribute should prevent the rule from being removed.

### DIFF
--- a/rule/rule.go
+++ b/rule/rule.go
@@ -1025,6 +1025,17 @@ func (r *Rule) IsEmpty(info KindInfo) bool {
 	return true
 }
 
+// ShouldKeepAttrs checks if any attributes in the info.NonEmptyAttrs
+// have a `# keep` statement.
+func (r *Rule) ShouldKeepAttrs(info KindInfo) bool {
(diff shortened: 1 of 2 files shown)


rule/rule.go
```diff
@@ -1018,7 +1018,7 @@ func (r *Rule) IsEmpty(info KindInfo) bool {
 		return false
 	}
 	for k := range info.NonEmptyAttrs {
-		if _, ok := r.attrs[k]; ok {
+		if _, ok := r.attrs[k]; ok && !ShouldKeep(r.attrs[k].expr) {
```
fmeum, line 1021: Doesn't this just make it harder for the function to return `false`, whereas the aim is to make it return `false` more often?

PR #1684, opened 2023-12-06
bzlmod: support global directives with gazelle_override

andyscott:
**What type of PR is this?**

Feature

Go + bzlmod

Adds the ability to configure global directives that apply to all Go modules loaded with `go_deps.from_file` in `MODULE.bazel` files. E.g., to turn off protobuf:

```
go_deps.gazelle_override(
    directives = [
        "gazelle:proto disable_global",
    ],
)
```

Previously `gazelle_override` required the `"path"` attribute. I've made this attribute optional: when the value is unset, the override should apply globally. Note that as it stands, only directives are supported in the global override.

Fixes #

**Other notes for review**

I'm not sure if there's a good place to add tests for this functionality. I've tested it locally with a relatively large Go repo and it works fine.

### DIFF
--- a/internal/bzlmod/go_deps.bzl
+++ b/internal/bzlmod/go_deps.bzl
@@ -65,7 +65,7 @@ def _fail_on_duplicate_overrides(path, module_name, overrides):
         fail("Multiple overrides defined for Go module path \"{}\" in module \"{}\".".format(path, module_name))
 
 def _fail_on_unmatched_overrides(override_keys, resolutions, override_name):
-    unmatched_overrides = [path for path in override_keys if path not in resolutions]
+    unmatched_overrides = [path for path in override_keys if path and path not in resolutions]
     if unmatched_overrides:
         fail("Some {} did not target a Go module with a matching path: {}".format(
             override_name, ", ".join(unmatched_overrides)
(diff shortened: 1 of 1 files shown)


fmeum:
This looks reasonable to me, but we should only allow "global" `gazelle_overrides` *before* any non-global ones, otherwise the precedence rules will be confusing.

I wonder whether this would be the right time to disable proto generation globally for `go_repository` - while a breaking change, with this new tool it would be relatively straightforward for users to maintain the current behavior.

CC @tyler-french @linzhp

linzhp:
Yes, it's the right time to disable proto by default.

andyscott:
Sounds good to me!

Can you elaborate on the following bit:
> we should only allow "global" gazelle_overrides before any non-global ones, otherwise the precedence rules will be confusing.

Are you referring to the ordering of any `go_deps.gazelle_override` declarations in `MODULE.bazel`?

tyler-french:
@fmeum I would prefer to have a separate tag, something like `go_deps.global_overrides`, or `go_deps.directives`, and I would prefer to **not** depend on the order of the MODULE.bazel file. I think that the individual overrides should always take precedence over the global overrides - and both are only allowed in the root module. I would like to avoid depending on order as much as possible here.

fmeum:
@tyler-french I also dislike "order matters", I'm just not sure about the right way to enforce it. I would find a `MODULE.bazel` file with:
```starlark
go_deps.gazelle_override(...)
go_deps.default_gazelle_override(...)
go_deps.gazelle_override(...)
```
almost as confusing as one where we reuse the `gazelle_override` tag - simply because it isn't immediately obvious what the exact semantics of the "default" are. I would thus like to make this a hard failure, but as far as I can tell, we can only reason about the ordering of tags within a single tag class. That's the advantage I see in reusing the tag.

Issue #1114, opened 2021-10-12
Parallel walk?


pcj:
It seems like gazelle should be theoretically parallelizable when walking the filesystem, as separate subtrees are supposed to be independent.  However, it would require that things like config access and other operations are protected against concurrent writes, etc.

I assume this has been thought of, any particular design decisions on this? cc @jayconrod

robfig:
I can't think of any reason that this wouldn't work, other than being some significant implementation work.

achew22:
Go has a very robust system for [profiling](https://pkg.go.dev/runtime/pprof) where you add (copied from their docs) something like

```go
var cpuprofile = flag.String("cpuprofile", "", "write cpu profile to `file`")
var memprofile = flag.String("memprofile", "", "write memory profile to `file`")

func main() {
    flag.Parse()
    if *cpuprofile != "" {
        f, err := os.Create(*cpuprofile)
        if err != nil {
            log.Fatal("could not create CPU profile: ", err)
        }
        defer f.Close() // error handling omitted for example
        if err := pprof.StartCPUProfile(f); err != nil {
            log.Fatal("could not start CPU profile: ", err)
        }
        defer pprof.StopCPUProfile()
    }

    // ... rest of the program ...
```

Have we tried adding something like that to the [commands you want to profile](https://github.com/bazelbuild/bazel-gazelle/blob/master/cmd/gazelle/fix-update.go#L82)?

PR #1685, opened 2023-12-11
feat(fix-update): allow user to profile commands with `pprof`

tyler-french:
**What type of PR is this?**

Feature

`cmd/gazelle`

This adds two flags to the `fix` and the `update` command, `-cpuprofile` and `-memprofile`, which allow users to create output CPU and memory `pprof` files to better understand the performance of Gazelle. This is very helpful in debugging issues like #1688, which show up in certain repository setups.

The implementation mostly followed the instructions on https://pkg.go.dev/runtime/pprof#hdr-Profiling_a_Go_program to set up profiling options for the command in the `cmd/gazelle`.

Tests were added to ensure that the `newProfile` would return a `profile{}` that wouldn't panic, even if either provider is empty.

Relates to #1114


README.rst
```diff
@@ -575,6 +575,20 @@ The following flags are accepted:
 |                                                                                                            |
 | By default, all languages that this Gazelle was built with are processed.                                  |
 +-------------------------------------------------------------------+----------------------------------------+
+| :flag:`-cpuprofile filename`                                      | :value:`""`                            |
++-------------------------------------------------------------------+----------------------------------------+
+| If specified, gazelle uses [runtime/pprof](https://pkg.go.dev/runtime/pprof#StartCPUProfile) to collect    |
+| CPU profiling information from the command and save it to a file.                                          |
```
fmeum, line 581:
```suggestion
| CPU profiling information from the command and save it to the given file.                                          |
```

fmeum (approved):

linzhp (approved):

PR #1687, opened 2023-12-13
feat(resolve): optimize applying overrides to be efficient

### LINKED ISSUE #1688: `# gazelle:resolve` directives are very inefficient
Currently, we iterate through every single resolve override on every single import we try to resolve. If there are many resolve directives (for example, at Uber, we have 7000+), then we end up spending a huge amount of time on this.

We should optimize the way we apply resolve directives by using a `map` so that we don't need to iterate through all of the resolves for each import.

Unfortunately, for the regexp resolves, this is not possible, since we need to see if the regexp matches.

tyler-french:
**What type of PR is this?**

Feature

resolve

**Enhancements in Resolve Directive Handling**

This PR addresses the inefficiency of processing `# gazelle:resolve` directives. Previously, the system iterated over a slice for each import to find matching directives, which is very inefficient. This PR introduces two major improvements:

1. **Transition from Slice to Map**: We've shifted from using a slice to a map for storing override configurations. This change reduces lookup time complexity from O(n) to O(1) (or O(length of chain of enclosing dirs containing resolves)). 

2. **Parent-Child Config Structure**: To maintain directory-specific configurations without the overhead of copying maps (which is very inefficient), we've implemented a parent-child relationship in configurations. This structure allows each directory to have its unique overrides while inheriting unmodified settings from its parent. We can't use a single shared map, because this could mistakenly apply directives from one directory to another. `regexp` overrides are always copied and evaluated inline. 

Note on GC: to avoid a really complex object tree, the parent/child relationship skips all ancestors without `# gazelle:resolve` directives, greatly reducing the complexity.

Note of Regexp: although we need to iterate through these due to the matching logic, we could at least have a `map: lang -> []overrides` to improve the lookup some.

**Data from Uber**

We have ~millions of directories to traverse, and approximately 8000 resolve directives.

Profile before changes:
Screenshot 2023-12-13 at 4 20 56 PM

Profile using a full map copy (on small directory), showing the inefficiencies of copying a map each time:
Screenshot 2023-12-13 at 4 20 26 PM

Fixes #1688

### DIFF
--- a/resolve/config.go
+++ b/resolve/config.go
@@ -32,11 +32,8 @@ import (
 // returned first. If no override is found, label.NoLabel is returned.
 func FindRuleWithOverride(c *config.Config, imp ImportSpec, lang string) (label.Label, bool) {
 	rc := getResolveConfig(c)
-	for i := len(rc.overrides) - 1; i >= 0; i-- {
-		o := rc.overrides[i]
-		if o.matches(imp, lang) {
-			return o.dep, true
-		}
+	if dep, ok := rc.findOverride(imp, lang); ok {
+		return dep, true
 	}
 	for i := len(rc.regexpOverrides) - 1; i >= 0; i-- {
 		o := rc.regexpOverrides[i]
@@ -44,20 +41,12 @@ func FindRuleWithOverride(c *config.Config, imp ImportSpec, lang string) (label.
 			return o.dep, true
 		}
 	}
-
 	return label.NoLabel, false
 }
 
-type overrideSpec struct {
+type overrideKey struct {
 	imp  ImportSpec
 	lang string
-	dep  label.Label
-}
-
-func (o overrideSpec) matches(imp ImportSpec, lang string) bool {
-	return imp.Lang == o.imp.Lang &&
-		imp.Imp == o.imp.Imp &&
-		(o.lang == "" || o.lang == lang)
 }
 
 type regexpOverrideSpec struct {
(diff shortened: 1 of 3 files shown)


resolve/resolve_test.go
```diff
@@ -0,0 +1,126 @@
...
+	"github.com/bazelbuild/bazel-gazelle/rule"
+	"github.com/bazelbuild/bazel-gazelle/config"
+	"github.com/bazelbuild/bazel-gazelle/label"
+	"github.com/google/go-cmp/cmp"
+)
+
+func TestFindRuleWithOverride_ParentTraversal(t *testing.T) {
+	rootCfg := getConfig(t, "", []rule.Directive{
+		{Key: "resolve", Value: "go go github.com/root/repo @com_example//root:replacement"},
+		{Key: "resolve_regexp", Value: "go ^github.com/root/(.*)$ @com_example//regexp:replacement"},
```
fmeum, line 15: If need be, we could optimize these special types of (probably common) "prefix" regexp resolve directives further.

fmeum (approved):
Great stuff!

tyler-french:
> @fmeum Let me do some additional testing to make sure no unexpected behaviors are changed.

Ok, I ran a bunch of tests and they all seemed to behave as expected. Let me check to see if @linzhp has any hesitations here

resolve/config.go
```diff
@@ -97,78 +123,80 @@ func (*Configurer) KnownDirectives() []string {
...
+	var newOverrides map[overrideKey]label.Label
+	regexpOverrides := rc.regexpOverrides[:len(rc.regexpOverrides):len(rc.regexpOverrides)]
+
+	for _, d := range f.Directives {
+		if d.Key == "resolve" {
+			parts := strings.Fields(d.Value)
+			key := overrideKey{}
+			var lbl string
+			if len(parts) == 3 {
+				key.imp.Lang = parts[0]
```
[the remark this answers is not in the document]
tyler-french: Nice! This ended up saving more time!

resolve/resolve_test.go
tyler-french: True, this might be a bit complex logic. I think I'd prefer to wait and do this in a later PR.

fmeum: Yes, fully agree with that

PR #1692, opened 2023-12-14
disable lockfile by default

tyler-french:
Currently, there is no process or policy around the lock file in this module. Let's keep it disabled until we have a reason to use it.


linzhp (approved):

fmeum:
I would recommend `update` mode combined with an entry in `.gitignore`: You get the performance benefits on server restarts during local development, but don't pay the complexity cost in CI.

PR #1693, opened 2023-12-15
prepare release 0.35.0

tyler-french:
This PR prepares release `0.35.0` of Gazelle


linzhp (approved):

MODULE.bazel
```diff
@@ -1,12 +1,12 @@
...
     name = "gazelle",
-    version = "0.34.0",
+    version = "0.35.0",
     repo_name = "bazel_gazelle",
 )
 
 bazel_dep(name = "bazel_skylib", version = "1.5.0")
 bazel_dep(name = "protobuf", version = "3.19.6", repo_name = "com_google_protobuf")
-bazel_dep(name = "rules_go", version = "0.42.0", repo_name = "io_bazel_rules_go")
+bazel_dep(name = "rules_go", version = "0.44.0", repo_name = "io_bazel_rules_go")
```
fmeum, line 9: If tests pass without the update, we should not force it.

fmeum (approved):

MODULE.bazel
tyler-french: It does appear to fail without the update. Not entirely sure the root cause: https://buildkite.com/bazel/bazel-gazelle/builds/3360#_

fmeum: Not sure either, probably the change in which I made Gazelle aware of `bazel_dep`s that are also Go modules requires a change in the rules_go setup. That's fine.

PR #1696, opened 2023-12-23
Test with incompatible flags

fmeum:
Preparation for https://github.com/bazelbuild/bazel/issues/12821.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -9,3 +9,11 @@ test:ci --test_strategy=standalone
 
 common --lockfile_mode=update
 test --test_output=errors
+
+build:incompatible --incompatible_load_proto_rules_from_bzl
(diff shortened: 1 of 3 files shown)


fmeum:
A local test run points to this likely being resolved by updating the rules_go version to one including https://github.com/bazelbuild/rules_go/pull/3795.

.bazelrc
```diff
@@ -9,3 +9,12 @@ test:ci --test_strategy=standalone
 
 common --lockfile_mode=update
 test --test_output=errors
+
+build:incompatible --incompatible_load_proto_rules_from_bzl
+build:incompatible --incompatible_enable_cc_toolchain_resolution
```
linzhp, line 14: This is already flipped in Bazel 7 and will be removed

fmeum: Removed

PR #1700, opened 2023-12-27
rule: add accessor for attribute comments

### LINKED ISSUE #237: Allow some way to prioritise one file over another
The `# keep` option is great, but unfortunately it does not have any effect on the order of the item. It seems gazelle will reorder `srcs` values each time it is called. Problem is execution order sometimes matters in Go. I need a way to say "this file should be the first entry". At the moment the only option I see is to exclude the relevant directory from gazelle and manually define my rule.

Example:

```
    srcs = [
        "locale-packr.go",  # keep
        "locale.go",
    ],
```

After I run gazelle the result is:

```
    srcs = [
        "locale.go",
        "locale-packr.go", # keep
    ],
```

I need locale-packr.go to be first.

t-8ch:
**What type of PR is this?**

Feature

rule/rule.go

This allows user to attach comments like "# do not sort" to the assignment expressions maintained by rule.go.

rule.Format() calls into buildtools/build.Format() which in turn calls Rewrite(). Rewrite() by default sorts certain label arguments like "deps" and "hdrs". To preventing this sorting it is possible to provide the comment "# do not sort". However this comment needs to be attached to the assignment expression.
In rule.go the assignment expressions are not exposed to the user, making it impossible to attach the comment to it.

See https://github.com/bazelbuild/bazel-gazelle/issues/237#issuecomment-399476945

Fixes #237

### DIFF
--- a/rule/rule.go
+++ b/rule/rule.go
@@ -957,6 +957,16 @@ func (r *Rule) SetAttr(key string, value interface{}) {
 	r.updated = true
 }
 
+// GetAttrComments returns the comments for an attribute.
+// It can be used to attach comments like "do not sort".
+func (r *Rule) GetAttrComments(key string) *bzl.Comments {
+	attr, ok := r.attrs[key]
+	if !ok {
+		return nil
+	}
+	return attr.expr.Comment()
+}
+
 // PrivateAttrKeys returns a sorted list of private attribute names.
 func (r *Rule) PrivateAttrKeys() []string {
 	keys := make([]string, 0, len(r.private))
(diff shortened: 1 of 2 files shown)


fmeum (approved):
Thanks!

rule/rule.go
```diff
@@ -957,6 +957,14 @@ func (r *Rule) SetAttr(key string, value interface{}) {
 	r.updated = true
 }
 
+func (r *Rule) GetAttrComments(key string) *bzl.Comments {
```
fmeum, line 960: Could you add a doc comment?

t-8ch: Done

t-8ch:
I just noticed that there is also #1394 which does exactly the same.

fmeum:
@t-8ch While looking at the other PR, I noticed that the function should probably be called `AttrComments` rather than `GetAttrComments` for consistency. Could you change that in a follow-up PR?

PR #1701, opened 2023-12-27
adding more default gazelle overrides

linzhp:
**What type of PR is this?**
Feature

bzlmod

Adding more default gazelle overrides and change the override for k8s.io/apimachinery to generate proto targets but not go targets. This allows non-root Bazel modules to use proto files from k8s.io/apimachinery.

**Other notes for review**
Alternatives, we can ask people to download a separate k8s.io/apimachinery archive with when they want to import the proto files in a non-root module.


tyler-french (approved):

fmeum (commented):
How static do we expect these defaults to be? And a
could there be any existing working setups that could break due to this?

internal/bzlmod/default_gazelle_overrides.bzl
```diff
@@ -47,6 +50,9 @@ DEFAULT_DIRECTIVES_BY_PATH = {
     "github.com/googleapis/gnostic": [
         "gazelle:proto disable",
     ],
+    "github.com/pseudomuto/protoc-gen-doc": [
+        "gazelle:resolve go github.com/mwitkow/go-proto-validators @com_github_mwitkow_go_proto_validators//:validators_gogo",
```
linzhp, line 54: The reason for this override is github.com/mwitkow/go-proto-validators comes with its build files and the naming convention is not recognized by Gazelle. It will be as stable as the target names in github.com/mwitkow/go-proto-validators.

internal/bzlmod/default_gazelle_overrides.bzl
```diff
@@ -57,7 +63,8 @@ DEFAULT_DIRECTIVES_BY_PATH = {
         "gazelle:proto disable",
     ],
     "k8s.io/apimachinery": [
-        "gazelle:proto disable",
+        "gazelle:go_generate_proto false",
+        "gazelle:proto_import_prefix k8s.io/apimachinery",
```
linzhp, line 67: This should be quite stable, because any change to proto import prefix requires changing import statements in proto files

linzhp (commented):
@fmeum explained the stability inline. I also remove all `gazelle:resolve` from k8s.io/apimachinery. It turned out that Gazelle is able to resolve the proto deps within that repo automatically.

fmeum (approved):

PR #1704, opened 2023-12-28
Remove some usage of deprecated ioutil

dzbarsky:
**What type of PR is this?**
> Other
Optimization and removal of deprecated API usage

> all

`ioutil` is deprecated, usage should be migrated to `os`.
In particular, `os.Readdir` is faster than `ioutil.Readdir` because it does not `stat` all the files to gather extra data which we don't need. On a fairly large repo, a no-op run of Gazelle drops from 2.2s to 1.7s with this patch.

I only changes the `ioutil` usages in the files I was touching anyway, but can change the rest in this PR if it's preferred.
 
A TODO in the code

### DIFF
--- a/rule/rule.go
+++ b/rule/rule.go
@@ -27,7 +27,7 @@ package rule
 
 import (
 	"fmt"
-	"io/ioutil"
+	"io/fs"
 	"os"
 	"path/filepath"
 	"sort"
@@ -95,7 +95,7 @@ func EmptyFile(path, pkg string) *File {
 // This function returns I/O and parse errors without modification. It's safe
 // to use os.IsNotExist and similar predicates.
 func LoadFile(path, pkg string) (*File, error) {
-	data, err := ioutil.ReadFile(path)
+	data, err := os.ReadFile(path)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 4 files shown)


rule/rule.go
```diff
@@ -257,10 +257,10 @@ func scanExprs(defName string, stmt []bzl.Expr) (rules []*Rule, loads []*Load, f
 // If there is at least one matching file, a path will be returned by joining
 // dir and the first matching name. If there are no matching files, the
 // empty string is returned.
-func MatchBuildFileName(dir string, names []string, files []os.FileInfo) string {
+func MatchBuildFileName(dir string, names []string, ents []fs.DirEntry) string {
```
fmeum, line 260: Unfortunately this is a breaking change to public API. Can we add a new function and mark this one as deprecated? Or do you see a different way to prevent this while still reaping the performance benefits?

walk/walk.go
```diff
@@ -333,5 +332,5 @@ func resolveFileInfo(wc *walkConfig, dir, rel string, fi fs.FileInfo) fs.FileInf
 		// A symlink, but not one we could resolve.
 		return nil
 	}
-	return fi
+	return fs.FileInfoToDirEntry(fi)
```
fmeum, line 335: Are we actually using the extra metadata?

fmeum (changes requested):

walk/walk.go
dzbarsky: No, I think we `stat` here to resolve the symlink. We could use `os.Readlink` instead but then we won't know if we landed on a file or a dir, so it seems we'd need a `stat` either way. But also symlinks in source repos should be rare so I'm not worried about trying to optimize this path.

rule/rule.go
dzbarsky: oh good point, yeah let me add a new version instead

fmeum (approved):

PR #1705, opened 2023-12-29
rule: allow usage of Label struct with Rule APIs

t-8ch:
**What type of PR is this?**

Feature

rule

The Label is a convenient structure for custom logic. Allow its use with the Rule APIs so the user can directly feed structured data back into Gazelle.

### DIFF
--- a/rule/value.go
+++ b/rule/value.go
@@ -22,6 +22,7 @@ import (
 	"sort"
 
 	bzl "github.com/bazelbuild/buildtools/build"
+	"github.com/bazelbuild/bazel-gazelle/label"
 )
 
 // KeyValue represents a key-value pair. This gets converted into a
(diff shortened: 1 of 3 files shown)


rule/value.go
```diff
@@ -17,6 +17,7 @@ package rule
 
 import (
 	"fmt"
+	"github.com/bazelbuild/bazel-gazelle/label"
```
fmeum, line 20: I can't remember whether we have tooling for that, but please move this import down into the separate block below.

rule/value_test.go
```diff
@@ -16,6 +16,7 @@ limitations under the License.
 package rule
 
 import (
+	"github.com/bazelbuild/bazel-gazelle/label"
```
fmeum, line 19: Also move this down.

fmeum (approved):

t-8ch:
An alternative would be to introduce an interface to make the conversion pluggable.

fmeum:
> An alternative would be to introduce an interface to make the conversion pluggable.

We could look into that if more types come up. This particular one seems clear enough since `Label(...)` isn't available in BUILD files.

t-8ch:
> An alternative would be to introduce an interface to make the conversion pluggable.

Turns out there is already
```go
// BzlExprValue is implemented by types that have custom translations
// to Starlark values.
type BzlExprValue interface {
	BzlExpr() bzl.Expr
}
```

fmeum:
Do you want to submit a follow-up PR that implements this interface for Label instead?

PR #1711, opened 2024-01-04
Prefer go_grpc_library

### LINKED ISSUE #1709: gazelle generates and recreates grpc targets that rules_go complains about

0.35

0.44.2

6.4.0

Yes for rules_go and gazelle, but I'm not able to test on bazel 7.0.0

macOS 14.2.1

### What did you do?

Gazelle 0.35 and rules_go (both 0.44.1 and 0.44.2) disagree on how to create grpc protobuf targets. If you use gazelle to generate targets for a proto3 proto file that contains `service` to create a GRPC service, it'll generate a `go_proto_library` target that sets `compilers = ["@io_bazel_rules_go//proto:go_grpc"],`.

However, `rules_go` will warn about that:

```
WARNING: /path/to/proto/BUILD.bazel:12:17: in go_proto_library rule //proto/proto:mytarget: target '//proto/mytarget_go_proto' depends on deprecated target '@io_bazel_rules_go//proto:go_grpc': Migrate to //proto:go_grpc_v2 compiler (which you'll get automatically if you use the go_grpc_library() rule).
```

Which implies that you're supposed to use `go_grpc_library` as the macro (from `load("@io_bazel_rules_go//proto:def.bzl", "go_grpc_library")`).

The unfortunate thing here is that if you fix this by hand, gazelle will remove your `go_grpc_library` call and the `load` statement and replace it with the `go_proto_library` target with the compilers set on it.

`rules_go` seems to have decided on this new pattern [a few weeks ago](https://github.com/bazelbuild/rules_go/pull/3761).

Gazelle creating grpc targets without warnings

### What did you see instead?

Gazelle creating grpc targets with warnings

mering:
**What type of PR is this?**

Bug fix

language/go

See issue.

Fixes #1709

**Other notes for review**


language/go/fix.go
```diff
@@ -171,15 +171,15 @@ func migrateLibraryEmbed(c *config.Config, f *rule.File) {
...
 }
 
-// migrateGrpcCompilers converts "go_grpc_library" rules into "go_proto_library"
-// rules with a "compilers" attribute.
+// migrateGrpcCompilers converts "go_proto_library" rules with a "compilers"
+// attribute containing exactly oldGrpcCompilerLabel into "go_grpc_library" rules.
 func migrateGrpcCompilers(c *config.Config, f *rule.File) {
 	for _, r := range f.Rules {
-		if r.Kind() != "go_grpc_library" || r.ShouldKeep() || r.Attr("compilers") != nil {
+		if r.Kind() != "go_proto_library" || r.ShouldKeep() || r.Attr("compilers") == nil || len(r.AttrStrings("compilers")) != 1 || !strListAttrContains(r, "compilers", oldGrpcCompilerLabel) {
```
mering, line 178: How can I check if the corresponding proto `.hasServices()` instead of checking for the old compiler name?

repository.md
mering, line 1: I tried updating the docs via `bazel run //docs:update`  but the tests keep failing on GitHub (the docs tests were successful on my local machine though).

ryanpbrewster:
Is there a way to configure the behavior of `gazelle` depending on what version of `rules_go` it is being used with?

mering:
> Is there a way to configure the behavior of `gazelle` depending on what version of `rules_go` it is being used with?

I don't know but I would avoid such things whenever possible as they just introduce a lot of additional complexity and things one has to check when debugging.

fmeum:
I would also avoid this. With Bzlmod, each version of Gazelle can declare which minimum version of rules_go it is compatible with. Without Bzlmod, version resolution is non-existent anyway, so I'm fine with "users have to update Gazelle and rules_go in lockstep" since that's the most promising approach to solve any kind of WORKSPACE issue anyway.

repository.md
fmeum: I can look into that when the other issues are resolved, don't worry.

fmeum:
I submitted https://github.com/bazelbuild/rules_go/pull/3818 to temporarily undo the deprecation until we figure out the Gazelle migration.

language/go/fix_test.go
```diff
@@ -564,8 +564,9 @@ proto_library(
     visibility = ["//visibility:public"],
 )
 
-go_grpc_library(
+go_proto_library(
     name = "foo_go_proto",
+    compilers = ["@io_bazel_rules_go//proto:go_grpc"],
     importpath = "example.com/repo",
```
sluongng, line 570: It might be helpful to add a test here demonstrating how a `go_proto_library` with custom, non-default compilers would behave. I.e. 

- 1 case with `go_grpc` + `my_custom_compiler_foo`.
- 1 case with `my_custom_compiler_foo`.

language/go/fix.go
mering: I removed this fixer to unblock this PR as we do no longer need to convert in this direction. Existing rules will just be kept as is which should continue working. Adding a fixer to convert the opposite direction might be added in a future PR.

mering:
Tests are passing locally now. This should be enough to unblock this PR and adding back the deprecation in rules_go. As my company does not need any such Gazelle changes, unfortunately I can't spend more time on this.

fmeum (approved):
@linzhp Could you review and test this as well?

linzhp (changes requested):
I prefer generating `go_proto_library` and set `compilers` attributes, which is more flexible and more consistent with Gazelle's philosophy of "no hidden magic" (Yes, there are codebases not using Gazelle and extensively use macros to make magic happen, so people don't need to write too much build files). This is the direction we are moving to, unless there is something I am missing. 

It's OK for Gazelle to generate the new v2 grpc plugin by default when `# gazelle:go_grpc_compilers` is not set to encourage new codebase to use the new plugin. But for existing codebase, they can decide their timeline of migration by setting `# gazelle:go_grpc_compilers` at different levels of directory.

PR #1713, opened 2024-01-04
Make attribute value sorting configurable

t-8ch:
**What type of PR is this?**

Feature

rule

The existing hardcoded sorting logic can get in the way. Provide a mechanism for users to opt-in or opt-out of it.
The default behavior is unchanged.

### DIFF
--- a/rule/rule.go
+++ b/rule/rule.go
@@ -1059,8 +1059,8 @@ func (r *Rule) sync() {
 
 	for _, k := range []string{"srcs", "deps"} {
 		attr, ok := r.attrs[k]
-		_, isUnsorted := attr.val.(UnsortedStrings)
-		if ok && !isUnsorted {
+		maybeSorted, isMaybeSorted := attr.val.(MaybeSorted)
+		if ok && (!isMaybeSorted || maybeSorted.ShouldSort(k)) {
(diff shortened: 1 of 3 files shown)


fmeum (approved):
Thanks, I like how this makes `UnsortedStrings` less of a special case.

t-8ch:
We are using a custom struct that handles combinations of lists of plain strings and `select()`s and that implements `BzlExprValue`.

If no `select()`s are present and only strings are emitted, the default sorting logic of gazelle kicks in, which is undesirable.
Instead of having to add a new step that detects this case and emits now an `UnsortedStrings` it is easier with this interface.

fmeum:
Hmm, to me this sounds as if this isn't really about the type of the attribute value but really about the name of the attribute. Could we instead just make the set of attributes to sort configurable on the rule?

t-8ch:
That would be fine, too.
I only need a way to disable the sorting, and I wanted to keep it close to the current `UnsortedStrings` mechanism.

t-8ch:
I'm also a bit confused about the need for a custom sorting logic in gazelle in general.
Everything will go through buildifier anyways and that will just apply its own logic.

fmeum:
Let's go with https://github.com/bazelbuild/bazel-gazelle/pull/1719 instead.

PR #1716, opened 2024-01-08
feat(bzlmod): allow `default_gazelle_attributes` for `go_deps` extension

tyler-french:
This PR creates a new tag class `default_gazelle_attributes` which allow the root module to provide default values for the gazelle attributes used to create the `go_deps` repos.

This includes the attributes: `"build_file_generation"`, `"build_extra_args"`, and `"directives"`. 

Default attributes will be applied to an attribute only if there is no `gazelle_override` value. 

If `default_gazelle_attributes` is provided, it will override all path-specific overrides created in `internal/bzlmod/default_gazelle_overrides.bzl`, so users may need to reapply these with `gazelle_override`s.

If a repo has either a `gazelle_override` or there is a `default_gazelle_attributes`, the `"build_file_generation"` attribute will be set to `"on"` unless explicitly set to a different value. This is because `"build_extra_args"`, and `"directives"` anticipate build file generation being enabled.

If any value is provided for attrs in a `gazelle_override`, it will override (and not append to) any default attributes. Users must re-add any `"directives"` that were in `default_gazelle_attributes`

Additional refactoring was done to remove duplication of the handling of these attributes.

### DIFF
--- a/internal/bzlmod/go_deps.bzl
+++ b/internal/bzlmod/go_deps.bzl
@@ -128,7 +128,7 @@ def _get_override_or_default(specific_overrides, gazelle_default_attributes, def
     # 2nd. Check for default attributes provided by the user. This must be done before checking for
     # gazelle's defaults path overrides to prevent Gazelle from overriding a user-specified flag.
     #
-    # This will also cause "build_file_generation" to default to "on" if default attirbutes are found.
+    # This will also cause "build_file_generation" to default to "on" if default attributes are found.
     global_override_value = getattr(gazelle_default_attributes, attribute_name, None)
     if global_override_value:
         return global_override_value
--- a/internal/bzlmod/go_deps.bzl
+++ b/internal/bzlmod/go_deps.bzl
@@ -44,6 +44,44 @@ the required directives to the "default_gazelle_overrides.bzl" file at \
 https://github.com/bazelbuild/bazel-gazelle/tree/master/internal/bzlmod/default_gazelle_overrides.bzl.
 """
 
+_GAZELLE_ATTRS = {
+    "build_file_generation": attr.string(
+        default = "on",
+        doc = """One of `"auto"`, `"on"` (default), `"off"`.
+
+        Whether Gazelle should generate build files for the Go module.
+
+        Although "auto" is the default globally for build_file_generation,
+        if a `"gazelle_override"` or `"gazelle_default_attributes"` tag is present
+        for a Go module, the `"build_file_generation"` attribute will default to "on"
+        since these tags indicate the presence of `"directives"` or `"build_extra_args"`.
+
+        In `"auto"` mode, Gazelle will run if there is no build file in the Go
+        module's root directory.
+
+        """,
+        values = [
+            "auto",
+            "off",
+            "on",
+        ],
+    ),
+    "build_extra_args": attr.string_list(
+        default = [],
+        doc = """
+        A list of additional command line arguments to pass to Gazelle when generating build files.
+        """,
+    ),
+    "directives": attr.string_list(
+        doc = """Gazelle configuration directives to use for this Go module's external repository.
+
+        Each directive uses the same format as those that Gazelle
+        accepts as comments in Bazel source files, with the
+        directive name followed by optional arguments separated by
+        whitespace.""",
+    ),
+}
+
 def _fail_on_non_root_overrides(module_ctx, module, tag_class):
     if module.is_root:
         return
@@ -68,7 +106,8 @@ def _fail_on_unmatched_overrides(override_keys, resolutions, override_name):
     unmatched_overrides = [path for path in override_keys if path not in resolutions]
     if unmatched_overrides:
         fail("Some {} did not target a Go module with a matching path: {}".format(
-            override_name, ", ".join(unmatched_overrides)
+            override_name,
+            ", ".join(unmatched_overrides),
         ))
 
 def _check_directive(directive):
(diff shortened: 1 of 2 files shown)


internal/bzlmod/go_deps.bzl
```diff
@@ -124,6 +158,32 @@ def _is_dev_dependency(module_ctx, tag):
...
+    for module in module_ctx.modules:
+        _fail_on_non_root_overrides(module_ctx, module, "gazelle_default_attributes")
+
+    for module in module_ctx.modules:
+        tags = module.tags.gazelle_default_attributes
+        if not tags:
+            continue
+
+        if len(tags) > 1:
+            fail("Multiple 'go_deps.global_override' tags in module '{}': {}".format(
```
fmeum, line 173: The name needs to be updated and `tag.path` isn't available. You can use this pattern to refer to individual tags including line numbers: https://github.com/bazelbuild/rules_go/blob/55ea579a70958d66f24a907433978b21edd36b29/go/private/extensions.bzl#L125-L130

tests/bcr/MODULE.bazel
```diff
@@ -24,6 +24,11 @@ go_deps = use_extension("@gazelle//:extensions.bzl", "go_deps")
 
 # Validate a go.mod replace directive works.
 go_deps.from_file(go_mod = "//:go.mod")
+go_deps.gazelle_default_attributes(
```
fmeum, line 27: I like the new name!

tyler-french: Thanks! Yeah I figured using the word "override" isn't semantically accurate, this is more of a fallback.

tyler-french:
@fmeum Please do not merge, I have not yet tested this at Uber

tyler-french:
> @fmeum Please do not merge, I have not yet tested this at Uber

I tested this and it works as expected.

fmeum:
The one thing I'm worried about is that the new default attributes are override by the per-module defaults provided by Gazelle. This makes adding a new default more breaking. But having it the other way round would remove most of the use of the Gazelle-provided defaults.
Not sure what's best yet and this may also change as we flip global defaults.

tyler-french:
> The one thing I'm worried about is that the new default attributes are override by the per-module defaults provided by Gazelle. This makes adding a new default more breaking. But having it the other way round would remove most of the use of the Gazelle-provided defaults. Not sure what's best yet and this may also change as we flip global defaults.

I agree, but I think this implementation provides the most user-friendly approach, given the following considerations:
1. Path-specific overrides that gazelle defines should **only** used for dependencies **required to build `gazelle` and/or `rules_go`**.
2. We document that defaults are "fallbacks", and may be overridden. Explicit `gazelle_override` will always override path-specific overrides and fallbacks.

Two changes that should be added in this PR (which I will add), are:
1. Remove anything from [default_gazelle_overrides.bzl](https://github.com/bazelbuild/bazel-gazelle/blob/master/internal/bzlmod/default_gazelle_overrides.bzl) that isn't explicitly needed to build `rules_go` and `gazelle`.
2. Update documentation to include the overrides included in [default_gazelle_overrides.bzl](https://github.com/bazelbuild/bazel-gazelle/blob/master/internal/bzlmod/default_gazelle_overrides.bzl), the explicit extent of `gazelle_default_attributes`, and how users **should** use `gazelle_overrides` to override the small number deps in [default_gazelle_overrides.bzl](https://github.com/bazelbuild/bazel-gazelle/blob/master/internal/bzlmod/default_gazelle_overrides.bzl).

internal/bzlmod/default_gazelle_overrides.bzl
```diff
@@ -14,32 +14,27 @@
 
 visibility("private")
 
+# This file should ONLY specify default overrides for
+# repos explicitly required to build gazelle or rules_go.
+#
+# Otherwise, users should should specify these overrides themselves
```
fmeum, line 20: This is unfortunately not always possible: We introduced this file so that Bazel modules using these Go modules can be non-root modules and submitted to registries. While we want to restrict this list to the most common modules, we can't restrict it to just dependencies of rules_go and Gazelle.

fmeum:
I still think that the way you implemented precedence is the right one: Every module with Gazelle-provided defaults works if you don't specify a default override and if you do, you explicitly give up convenience for flexibility.

I think that's fine as long as we restore the list of defaults.

tests/bcr/pkg/pkg_test.go
```diff
@@ -14,8 +13,6 @@ import (
 	"github.com/fmeum/dep_on_gazelle"
 	"github.com/google/safetext/yamltemplate"
 	"github.com/stretchr/testify/require"
-
-	_ "github.com/envoyproxy/protoc-gen-validate/validate"
```
fmeum, line 18: Could you check what we added this for? If it still provides test coverage, let's add a comment instead of removing it.

tests/bcr/pkg/pkg_test.go
```diff
@@ -37,10 +34,6 @@ func TestBuildFileGeneration(t *testing.T) {
 	yamltemplate.HTMLEscapeString("<b>foo</b>")
 }
 
-func TestGeneratedFilesPreferredOverProtos(t *testing.T) {
```
fmeum, line 40: Same here, I think that we should try to keep the existing test coverage around.

fmeum:
The PR description is not up-to-date anymore, it still describes the previous precedence rule.

tests/bcr/MODULE.bazel
```diff
@@ -24,6 +24,26 @@ go_deps = use_extension("@gazelle//:extensions.bzl", "go_deps")
...
+    ],
+)
+
+# By defining `gazelle_default_attributes`, we also must individually
+# specify certain overrides from internal/bzlmod/default_gazelle_overrides.bzl
+go_deps.gazelle_override(
+
+    # Because this tag is specified, we must specify build_file_generation
+    # otherwise it will be set to "auto" by default.
+    build_file_generation = "on",
```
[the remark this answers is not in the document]
tyler-french: Yes, I like this. We can just change the default for the `attr` to `on` for Gazelle overrides.

internal/bzlmod/go_deps.bzl
```diff
@@ -44,6 +44,36 @@ the required directives to the "default_gazelle_overrides.bzl" file at \
 https://github.com/bazelbuild/bazel-gazelle/tree/master/internal/bzlmod/default_gazelle_overrides.bzl.
 """
 
+_GAZELLE_ATTRS = {
+    "build_file_generation": attr.string(
```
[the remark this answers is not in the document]
tyler-french: We can do this by just switching this `auto` to `on` in the `_GAZELLE_ATTRS`, which will only apply when a user supplies a `gazelle_override` or `gazelle_default_attributes`, otherwise it will be `auto`.

internal/bzlmod/go_deps.bzl
```diff
@@ -68,45 +98,51 @@ def _fail_on_unmatched_overrides(override_keys, resolutions, override_name):
...
         return
     fail("Invalid Gazelle directive: \"{}\". Gazelle directives must be of the form \"gazelle:key value\".".format(directive))
 
-def _get_build_file_generation(path, gazelle_overrides):
-    override = gazelle_overrides.get(path)
-    if override:
-        return override.build_file_generation
+def _get_override_or_default(specific_overrides, gazelle_default_attributes, default_path_overrides, path, default_value, attribute_name):
+    # 1st: Check for user-provided specific overrides. If a specific override is found,
+    # all of its attributes will be applied (even if left to the tag's default). This is to allow
```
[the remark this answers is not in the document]
tyler-french: Thanks!

internal/bzlmod/go_deps.bzl
```diff
@@ -553,45 +615,23 @@ _archive_override_tag = tag_class(
 )
 
 _gazelle_override_tag = tag_class(
-    attrs = {
+    attrs = dict({
```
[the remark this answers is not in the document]
tyler-french: Cool! I did not know this was possible

internal/bzlmod/go_deps.bzl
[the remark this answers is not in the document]
tyler-french: Let me know if you think this `dic` makes sense,

tests/bcr/MODULE.bazel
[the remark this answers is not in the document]
tyler-french: Resolving this. See reply to https://github.com/bazelbuild/bazel-gazelle/pull/1716/files#r1493331603

tyler-french:
> The PR description is not up-to-date anymore, it still describes the previous precedence rule.

Good point, updated.

internal/bzlmod/go_deps.bzl
```diff
@@ -68,45 +106,58 @@ def _fail_on_unmatched_overrides(override_keys, resolutions, override_name):
...
+    #
+    # This will also cause "build_file_generation" to default to "on" if a specific override is found.
+    specific_override = specific_overrides.get(path)
+    if specific_override and hasattr(specific_override, attribute_name):
+        return getattr(specific_override, attribute_name)
+
+    # 2nd. Check for default attributes provided by the user. This must be done before checking for
+    # gazelle's defaults path overrides to prevent Gazelle from overriding a user-specified flag.
+    #
+    # This will also cause "build_file_generation" to default to "on" if default attirbutes are found.
```
fmeum, line 131:
```suggestion
    # This will also cause "build_file_generation" to default to "on" if default attributes are found.
```

fmeum (approved):

PR #1721, opened 2024-01-12
Error if build_directives are specified but ignored

illicitonion:
**What type of PR is this?**

Bug fix

go_repository

Error if build_directives are specified but ignored

It can be frustrating to debug why a build directive isn't taking
effect, only to realise that the repo already has BUILD files checked in
so gazelle isn't going to apply them.

### DIFF
--- a/internal/go_repository.bzl
+++ b/internal/go_repository.bzl
@@ -333,6 +333,8 @@ def _go_repository_impl(ctx):
         if ctx.attr.debug_mode and result.stderr:
             print("%s gazelle.stdout: %s" % (ctx.name, result.stdout))
             print("%s gazelle.stderr: %s" % (ctx.name, result.stderr))
+    elif len(ctx.attr.build_directives) > 0:
+        fail("%s: build_directives were specified but are being ignored - perhaps the repo already has BUILD files present. Try setting build_file_generation = \"on\"" % ctx.name)
 
     # Apply patches if necessary.
     patch(ctx)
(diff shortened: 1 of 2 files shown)


fmeum (commented):
This would have saved me quite a bit of time, thanks!

I'm just a bit worried that the instructions only address the case where users directly use `go_repository`, but the error can also be shown if they use Bzlmod with `go_deps`. Could we improve the error in that case?

illicitonion:
What is the recommended remedy when using `go_deps`? Is there a module extension you'd need to call or something?

If you can share / link to a snippet of how you'd fix the issue, I can fix up the text.

fmeum:
`build_file_generation` is available on this tag: https://github.com/bazelbuild/rules_go/blob/master/docs/go/core/bzlmod.md#gazelle-directives

Which reminds me that we still haven't published stardoc docs for our module extensions.

@tyler-french I will try to find some time to do that. It requires a stardoc update, which did break something last time I tried.

illicitonion:
Oh interesting, AFAICT there's no bzlmod way to provide a value for `build_file_generation`? Is that an oversight, or am I missing it?

fmeum (approved):
@tyler-french Could you test this at Uber? Curious to see how many silent failures there are.

fmeum:
@illicitonion Could you fix the issue this caught in the test module?

fmeum:
@tyler-french and I discovered that https://github.com/bazelbuild/bazel-gazelle/pull/1716 conflicts with this PR and want to flesh out the interaction before merging either. 

Concretely, when you are able to set a default for `directives` but don't set one for `build_file_generation`, you are pretty likely to run into errors. Instead, we are considering forcing `build_file_generation` to always be `on` if default directives are used.

illicitonion:
@fmeum did you reach any conclusions? Should I close this, rebase it, or make something different?

fmeum:
With https://github.com/bazelbuild/bazel-gazelle/pull/1716, `build_file_generation` now defaults to `on` if any directives are set (with Bzlmod). Maybe that makes this PR obsolete? Do you have other cases in mind where this would help?

illicitonion:
Sounds good - thanks!

PR #1722, opened 2024-01-22
Apply map_kind to args as well as rule kinds

illicitonion:
**What type of PR is this?**

Bug fix

cmd/gazelle

https://github.com/bazelbuild/bazel-gazelle/pull/1310 added support for adding args from Gazelle, so you can express things like:

```
maybe(
    java_test,
    ...
)
```

But `map_kind` doesn't currently get applied to these args.

This PR adds that application.


cmd/gazelle/fix-update.go
```diff
@@ -380,17 +381,36 @@ func runFixUpdate(wd string, cmd command, args []string) (err error) {
...
+		}
+
+		for _, r := range allRules {
+			if replacementName, err := maybeRecordReplacement(r.Kind()); err != nil {
+				errorsFromWalk = append(errorsFromWalk, fmt.Errorf("looking up mapped kind: %w", err))
+			} else if replacementName != nil {
+				r.SetKind(*replacementName)
+			}
+
+			for i, arg := range r.Args() {
```
fmeum, line 406: I am slightly worried about false positives. What do you think of restricting this to the first argument? Also, could we check that the referenced identifier comes from a `load`, not a global variable?

illicitonion: Updated to both just apply to the first arg, and to only apply to symbols registered as being loaded by one of the enabled plugins. I don't think it's easy for us to check that it _was_ loaded, but it's easy to check whether it's something we _would_ load.

fmeum (approved):
@tyler-french Could you test this?

rule/rule.go
```diff
@@ -1016,6 +1016,15 @@ func (r *Rule) AddArg(value bzl.Expr) {
 	r.updated = true
 }
 
+// UpdateArg replaces an existing arg with a new value.
+func (r *Rule) UpdateArg(index int, value bzl.Expr) {
+	if len(r.args) < index {
+		panic(fmt.Sprintf("Can't update argument at index %d, only have %d args", index, len(r.args)))
```
tyler-french, line 1022: Can we return an error instead of panicking? Some references: https://github.com/uber-go/guide/blob/master/style.md#dont-panic

tyler-french:
> @tyler-french Could you test this?

Tested and it works as expected @fmeum

fmeum:
@illicitonion Friendly ping

rule/rule.go
illicitonion: Updated to return an error here, but the only caller is in a non-error-returning context so it still ends up panicking directly

illicitonion:
@fmeum I think we're good to go here? Thanks!

Issue #1549, opened 2023-05-29
go.work file support


don-azlok:
https://github.com/bazelbuild/bazel-gazelle/blob/3a1235c298b7dc11fa7ba994db4f92587c1aca79/internal/bzlmod/go_deps.bzl#L447

sluongng:
It's not impossible.

We need to rewrite https://github.com/bazelbuild/bazel-gazelle/blob/3a1235c298b7dc11fa7ba994db4f92587c1aca79/internal/bzlmod/go_mod.bzl to instead parsing `go.mod` and `go.sum` file as-is, to instead:
- execute `go mod download -json`
- parse the json outputs

Essentially recreating `deps_from_go_mod` and `sum_from_go_mod` with `go.work`

malt3:
> It's not impossible.

Do you mean it is actually impossible? Or is it instead a lot of extra work?
I think go.work OR support for multiple go.mod files would be great for large monorepos that also tend to use Bazel.
Would the maintainers welcome support for correctly handling a go.work file for dependency management?

sluongng:
Not impossible == possible.

And yes it will require a decent amount of work to get it done. Probably not "a lot", but may be 1-2 weeks worth of effort from start until merged (very rough est).

fmeum:
I haven't used Go workspaces yet, but if `go.work` really only consists of `use` and `go` directives, parsing it in Starlark may actually be easier than getting a Go SDK and running `go mod download -json`.

What I would like to see first is a clear specification of what supporting `go.work` actually means, considering that each Bazel module in the dependency graph could have its own `go.work` file and we have to reconcile those (or decide that only the root module should get to use this feature).

malt3:
> It's also not clear to me whether supporting go.work would also require making Gazelle of all the Go modules in the workspace (e.g. to generate correct importpaths).

I think `use` directives in go.work are conceptually similar to `replace` directives with local paths in go.mod files.
That probably means Gazelle has to be aware of all Go modules that are `use`d in a go.work file.
This way Gazelle could use the local module instead of generating labels for external repositories.

PR #1731, opened 2024-02-06
[Feature] bzlmod & go.work

### LINKED ISSUE #1720: Support Golang Workspaces in Gazelle
Like mentioned in this [issue](https://github.com/bazelbuild/rules_go/issues/3823) in the rules_go repository, here is the issue in the gazelle repository.

My question is when Gazelle will support [Golang workspaces](https://go.dev/ref/mod#workspaces) or more precisely when it will be possible to use ````go.work```` files with gazelle.

stefanpenner:
**What type of PR is this?**
Feature

language/go

Explores a possible approach for adding go.work support to the go bzlmod extension.

1. adds `go_deps.from_file(go_work = "//:go.work")`
2. adds ability to op-out of version conflict failures `go_deps.from_file(go_work = "//:go.work, fail_on_version_conflict = False)`
2. parses `go.work` files to discover used go modules and replace statements
3. generates repos for all deps specified in go.mod's referenced by the `//:go.work`
4. handle replace statements (both at the go.work and the go.mod level)
5. Errors with actionable error message if duplicate dependencies with differing versions arise (Question: can we do better?)

Fixes #1720 #1549

**Other notes for review**

Although this does work, it should be considered a prototype intended for discussion and feedback. If we decide to pursue this or something similar, expect this PR to evolve rapidly. Although it has some tests, it will require more. Additionally, some architectural changes may improve readability.

----

An additional consideration, a complementary PR to teach gazelle about go.work + go.mods would help mitigate the dependency bleed problem this PR has. We should discuss this as well.


stefanpenner:
Interesting that the CLA bot was angry, when I went to go sign-up it pointed out I signed the google CLA in 2014...

MODULE.bazel
```diff
@@ -48,15 +48,11 @@ use_repo(
     "org_golang_x_tools_go_vcs",
     # Referenced by `gazelle_binary`.
     "bazel_gazelle_go_repository_config",
-    # Dependencies of rules_go that require special handling in go_deps.
```
stefanpenner, line 51: probably should be dropped

internal/bzlmod/go_deps.bzl
```diff
@@ -194,6 +202,55 @@ _go_repository_config = repository_rule(
     },
 )
 
+
+def fail_if_duplicate_modules_have_different_versions(version, previous, module_tag, module_name_to_go_dot_mod_label, go_works):
```
stefanpenner, line 206: this has proven brittle, add tests

internal/bzlmod/go_deps.bzl
```diff
@@ -194,6 +202,55 @@ _go_repository_config = repository_rule(
...
+    if previous.version[0] == version[0] or str(current_label).endswith("go.work") or str(previous_label).endswith("go.work"):
+        corrective_measure =  default_corrective_mesasure
+    else:
+        label = module_name_to_go_dot_mod_label.get(module_tag.path)
+
+        if label:
+            # if the label is present that means the module_tag is of a go.mod file, which means the correct action may be different.
+            
+            # if the duplicate module in question is provided by go.work use statement then only manual intervention can fix it
+            # from_file_tags on go_work represents use statements in the go.work file
```
stefanpenner, line 244: TODO: from_file_tags is a list, and 246 needs to enumerate that list aswell

stefanpenner:
TODO (from call with team):

- [ ] add additional unit test support
- [ ] add one integration tests (search for go_bazel_test to find examples, also look at tests/bcr/<something>)
- [ ] only support `from_file(go_work = ...)` from root modules. (check is_root property of the module)
- [ ] double check that multiple from_file calls is disallowed
- [ ] mandate that from_file is called with only one go_work, or go_mod attribute, and not none.
- [ ] [Optional CLI Tool] auto-fix dependency divergence too
   - [ ] optionally add the ability to strictify transitive go dependencies. to lock down the native go experience similarly to the bazel experience.
- [ ] explore new `bazel mod tidy` command & implementation https://github.com/bazelbuild/bazel/pull/20483 - auto-updates use_repo
----

stefanpenner:
- [ ] Debugging BCR tests, I realized this also needs to work properly with https://github.com/bazelbuild/bazel-gazelle/pull/1526/ ... so Adding a todo.

.bazelci/presubmit.yml
```diff
@@ -41,6 +41,7 @@ tasks:
     # Regenerate the BUILD file for the test module using Gazelle.
     # Also verify -repo_config are generated correctly in gazelle.bash
     - rm pkg/BUILD.bazel
+    - touch pkg/BUILD.bazel # ensure empty pkg/BUILD.bazel exists
```
stefanpenner, line 44: Unless we refactor `tests/bcr` more significantly, this may be required as the existing of `pkg/BUILD.bazel` is required given that go.work "points" to the `pkg` package.

That being said, if there is an alternative, I'm open to suggestions.

.bazelversion
```diff
@@ -0,0 +1 @@
+6.4.0
```
stefanpenner, line 1: I noticed that CI was using 6.4.0 so I mimic that here. We can drop this, or land this separately... or keep this here?

internal/bzlmod/go_deps.bzl
```diff
@@ -194,6 +201,79 @@ _go_repository_config = repository_rule(
     },
 )
 
+def fail_on_version_conflict(version, previous, module_tag, module_name_to_go_dot_mod_label, go_works, fail_or_warn):
```
stefanpenner, line 204: TODO: at unit tests

internal/bzlmod/go_deps.bzl
```diff
@@ -194,6 +201,79 @@ _go_repository_config = repository_rule(
...
+
+    if not previous or version == previous.version:
+        # version is the same, skip because we won't error
+        return
+
+    # When using go.work, duplicate depenency versions are possible.
+    # This can cause issues, so we fail with a hopefully actionable error.
+    current_label = module_tag.parent_label
+
+    # TODO why? is this needed
```
stefanpenner, line 229: todo

internal/bzlmod/go_deps.bzl
```diff
@@ -194,6 +201,79 @@ _go_repository_config = repository_rule(
...
+      if fail_or_warn:
+        fail(message)
+      else:
+        print(message)
+
+def _fail_if_not_root(module, from_file_tag):
+    if module.is_root != True:
+        fail("go_deps.from_file(go_work = '{}') tag can only be used from a root module but: '{}' is not a root module.".format(from_file_tag.go_work, module.name))
+
+def _fail_if_invalid_from_file_usage(from_file_tag):
```
stefanpenner, line 270: todo tests

internal/bzlmod/go_mod.bzl
```diff
@@ -240,16 +322,37 @@ def sums_from_go_mod(module_ctx, go_mod_label):
     """
     _check_go_mod_name(go_mod_label.name)
 
-    # We go through a Label so that the module extension is restarted if go.sum
+    return parse_sumfile(module_ctx, go_mod_label, "go.sum")
+
+def sums_from_go_work(module_ctx, go_work_label):
```
stefanpenner, line 327: I believe the existence of a `go.work.sum`  signals that a go.mod is lacking a `go mod tidy` invocation. We could be strict, and error. Or we could warn or we could do nothing...

For correctness, we likely want to error, and I'm leaving towards this. But I'm curious what go engineers would like in practice

internal/bzlmod/go_deps.bzl
```diff
@@ -194,6 +201,78 @@ _go_repository_config = repository_rule(
...
+            for from_file_tags in [go_work.from_file_tags for go_work in go_works]:
+                for from_file_tag in from_file_tags:
+                    if from_file_tag.go_mod == label:
+                        corrective_measure = default_corrective_mesasure
+                        break
+        elif previous.module_tag.indirect or module_tag.indirect:
+            # if the dependency indirect, the user will need to manually update go.mod, run go mod tidy in that directory and then run go work sync
+            corrective_measure = default_corrective_mesasure
+        else:
+            # TODO: if the version are v0.8.0 and v0.17.0 go work sync wont work
```
stefanpenner, line 254: Just noticed this TODO: I'll tackle this shortly

.bazelci/presubmit.yml
```diff
@@ -41,6 +41,7 @@ tasks:
     # Regenerate the BUILD files for the test module using Gazelle.
     # Also verify -repo_config are generated correctly in gazelle.bash
     - rm pkg/BUILD.bazel proto/BUILD.bazel
+    - touch pkg/BUILD.bazel proto/BUILD.bazel
```
stefanpenner, line 44: Unless we refactor tests/bcr more significantly, this may be required as the existing of pkg/BUILD.bazel is required given that go.work "points" to the pkg package.

That being said, if there is an alternative, I'm open to suggestions.

fmeum: This is fine, gazelle generating BUILD files at all is covered by other tests pretty well.

.bazelversion
fmeum: This is fine, we can and should bump it separately.

tests/bzlmod/semver_test.bzl
```diff
@@ -77,8 +77,23 @@ def _semver_test_impl(ctx):
 
 semver_test = unittest.make(_semver_test_impl)
 
+def _humanize_test_impl(ctx):
+    env = unittest.begin(ctx)
+
+    # remove everything after and including the frist + as it is not preserved in a comparable
```
fmeum, line 83:
```suggestion
    # remove everything after and including the first + as it is not preserved in a comparable
```

internal/bzlmod/semver.bzl
```diff
@@ -78,3 +78,28 @@ def _semver_to_comparable(v, *, relaxed = False):
...
+        version: The comparable version object.
+
+    Returns:
+        A human-readable representation of the version.
+    """
+    core = [_flatten_segment(segment) for segment in version[0]]
+    meta = [_flatten_segment(segment) for segment in version[1] if segment[0] != "{"]
+    result = ".".join(core)
+
+    if (len(meta) > 0):
```
fmeum, line 100:
```suggestion
    if meta:
```

internal/bzlmod/semver.bzl
```diff
@@ -78,3 +78,28 @@ def _semver_to_comparable(v, *, relaxed = False):
 semver = struct(
     to_comparable = _semver_to_comparable,
 )
+
+def humanize_comparable_version(version):
```
fmeum, line 82: Let's export this as a field of the `semver` struct instead and call it "humanize_comparable".

internal/bzlmod/semver.bzl
```diff
@@ -78,3 +78,28 @@ def _semver_to_comparable(v, *, relaxed = False):
...
+    Please note, this is lossy as + meta segments are replaced with sentinal value to preserve expected sorting
+
+    Args:
+        version: The comparable version object.
+
+    Returns:
+        A human-readable representation of the version.
+    """
+    core = [_flatten_segment(segment) for segment in version[0]]
+    meta = [_flatten_segment(segment) for segment in version[1] if segment[0] != "{"]
```
fmeum, line 97: Can we extract the sentinel string into a shared constant?

tests/bzlmod/semver_test.bzl
```diff
@@ -77,8 +77,23 @@ def _semver_test_impl(ctx):
...
+    RESTORABLE_VERSIONS = [version.split("+")[0] for version in _SORTED_TEST_VERSIONS]
+
+    for version in RESTORABLE_VERSIONS:
+        humanized = humanize_comparable_version(semver.to_comparable(version, relaxed = RESTORABLE_VERSIONS))
```
fmeum, line 87: Why this value for `relaxed`?

repository.md
```diff
@@ -91,7 +91,7 @@ git_repository(
...
 | <a id="git_repository-remote"></a>remote |  The remote repository to download.   | String | required |  |
-| <a id="git_repository-repo_mapping"></a>repo_mapping |  In `WORKSPACE` context only: a dictionary from local repository name to global repository name. This allows controls over workspace dependency resolution for dependencies of this repository.<br><br>For example, an entry `"@foo": "@bar"` declares that, for any time this repository depends on `@foo` (such as a dependency on `@foo//some:target`, it should actually resolve that dependency within globally-declared `@bar` (`@bar//some:target`).<br><br>This attribute is _not_ supported in `MODULE.bazel` context (when invoking a repository rule inside a module extension's implementation function).   | <a href="https://bazel.build/rules/lib/dict">Dictionary: String -> String</a> | optional |  |
```
fmeum, line 94: This comes from the downgrade to 6.4.0. Let's use 7.0.2 instead.

internal/bzlmod/go_mod.bzl
```diff
@@ -240,16 +328,37 @@ def sums_from_go_mod(module_ctx, go_mod_label):
...
+    Returns:
+        A Dict[(string, string) -> (string)] is retruned where each entry
+        is defined by a Go Module's sum:
+            (path, version) -> (sum)
+    """
+    _check_go_work_name(go_work_label.name)
+
+    return parse_sumfile(module_ctx, go_work_label, "go.work.sum")
+
+def parse_sumfile(module_ctx, label, somefile):
```
fmeum, line 351:
```suggestion
def parse_sumfile(module_ctx, label, sumfile):
```

internal/bzlmod/go_mod.bzl
```diff
@@ -240,16 +328,37 @@ def sums_from_go_mod(module_ctx, go_mod_label):
...
+    """Loads the entries from a go.work.sum file given a go.workLabel.
+
+    Args:
+        module_ctx: a https://bazel.build/rules/lib/module_ctx object
+            passed from the MODULE.bazel call.
+        go_work_label: a Label for a `go.work` file. This label is used
+            to find the associated `go.work.sum` file.
+
+    Returns:
+        A Dict[(string, string) -> (string)] is retruned where each entry
```
fmeum, line 343:
```suggestion
        A Dict[(string, string) -> (string)] is returned where each entry
```

internal/bzlmod/go_mod.bzl
```diff
@@ -240,16 +328,37 @@ def sums_from_go_mod(module_ctx, go_mod_label):
     """
     _check_go_mod_name(go_mod_label.name)
 
-    # We go through a Label so that the module extension is restarted if go.sum
+    return parse_sumfile(module_ctx, go_mod_label, "go.sum")
+
+def sums_from_go_work(module_ctx, go_work_label):
+    """Loads the entries from a go.work.sum file given a go.workLabel.
```
fmeum, line 334:
```suggestion
    """Loads the entries from a go.work.sum file given a go.work label.
```

internal/bzlmod/go_deps.bzl
```diff
@@ -592,6 +712,10 @@ _module_tag = tag_class(
         ),
         "build_naming_convention": attr.string(doc = """Removed, do not use""", default = ""),
         "build_file_proto_mode": attr.string(doc = """Removed, do not use""", default = ""),
+        "parent_label": attr.label(
```
fmeum, line 715: If we only need this on the synthetic tags generated by `from_file`, we should not make it a public attribute.

internal/bzlmod/go_deps.bzl
```diff
@@ -577,7 +695,9 @@ _config_tag = tag_class(
 
 _from_file_tag = tag_class(
     attrs = {
-        "go_mod": attr.label(mandatory = True),
+        "go_mod": attr.label(mandatory = False),
+        "go_work": attr.label(mandatory = False),
+        "fail_on_version_conflict": attr.bool(default = True),
```
fmeum, line 700: Let's hardcode this to `True` for now.

internal/bzlmod/go_deps.bzl
```diff
@@ -553,13 +670,14 @@ def _get_sum_from_module(path, module, sums):
         entry = (module.replace, module.raw_version)
 
     if entry not in sums:
+        # TODO: if no sum exist, this is probably because a go mod tidy was missed
```
fmeum, line 673: Let's suggest running `bazel run @rules_go//go -- mod tidy`.

repository.md
stefanpenner: is this from https://github.com/bazelbuild/bazel-gazelle/pull/1731/files#r1503290363 or a different 6.4.0?

internal/bzlmod/go_deps.bzl
stefanpenner: @fmeum we had an internal request to make this configurable during larger migrations. Is it OK if we keep it?

repository.md
stefanpenner: When i drop this change, I end up with the following failures:

* `//cmd/gazelle:gazelle_test`  -> `could not locate go tool`
* `//internal:bazel_test` -> `error: unknown runfile: ./.bazelrc`

Downgrading to whats run in CI is required to not fail.

.bazelversion
stefanpenner: Now I remember why I added this. The repo doesn't yet support bazel 7.x.x, so I mimic'd what CI is using locally to ensure tests pass locally.

The failures:

* `//cmd/gazelle:gazelle_test` fails with: `could not locate go tool`
* `//internal:bazel_test` fails with: `error: unknown runfile: ./.bazelrc`

stefanpenner: I could go and fix these issues, but I suspect that should be a seperate PR (I don't think it's related to this one) But let me know.

tests/bzlmod/semver_test.bzl
stefanpenner: The existing test data provided by `_SORTED_TEST_VERSIONS` includes versions such as `0.1-a`, which trigger the following faillure:
 
> Semantic version strings must have exactly three dot-separated components, got: 0.1-a

I figure it's worth testing those as well, so I used relaxed for those versions. I can gladly take a different approach, please advise.

internal/bzlmod/go_deps.bzl
```diff
@@ -271,6 +277,77 @@ _go_repository_config = repository_rule(
...
+
+        if label:
+            # if the label is present that means the module_tag is of a go.mod file, which means the correct action may be different.
+
+            # if the duplicate module in question is provided by go.work use statement then only manual intervention can fix it
+            # from_file_tags on go_work represents use statements in the go.work file
+            for from_file_tags in [go_work.from_file_tags for go_work in go_works]:
+                for from_file_tag in from_file_tags:
+                    if from_file_tag.go_mod == label:
+                        corrective_measure = default_corrective_mesasure
```
fmeum, line 323: We are assigning this in many places. Maybe assign it once and only set again if using a non-default message?

internal/bzlmod/go_deps.bzl
```diff
@@ -271,6 +277,77 @@ _go_repository_config = repository_rule(
...
+    else:
+        print(message)
+
+def _fail_if_not_root(module, from_file_tag):
+    if module.is_root != True:
+        fail("go_deps.from_file(go_work = '{}') tag can only be used from a root module but: '{}' is not a root module.".format(from_file_tag.go_work, module.name))
+
+def _fail_if_invalid_from_file_usage(from_file_tag):
+    if (
+        (from_file_tag.go_work == None and from_file_tag.go_mod == None) and
```
fmeum, line 346: `bool(from_file_tag.go_work) == bool(from_file_tag.go_mod)` could be considered simpler (your call)

internal/bzlmod/go_deps.bzl
```diff
@@ -329,9 +406,40 @@ def _go_deps_impl(module_ctx):
...
-            module_path, module_tags_from_go_mod, go_mod_replace_map = deps_from_go_mod(module_ctx, from_file_tag.go_mod)
+            _fail_if_invalid_from_file_usage(from_file_tag)
+
+            if from_file_tag.go_mod:
+                from_file_tags.append(from_file_tag)
+            elif from_file_tag.go_work:
+                _fail_if_not_root(module, from_file_tag)
+
+                go_work = go_work_from_label(module_ctx, from_file_tag.go_work)
+                go_works.append(go_work)
```
fmeum, line 424: Can we ever have multiple of those? I would rather disallow that in the beginning.

internal/bzlmod/go_deps.bzl
```diff
@@ -382,12 +490,12 @@ def _go_deps_impl(module_ctx):
...
         paths = {}
+
         for module_tag in module.tags.module + additional_module_tags:
-            if module_tag.path in paths:
-                fail("Duplicate Go module path \"{}\" in module \"{}\".".format(module_tag.path, module.name))
+            if not module_tag.path in paths:
+                paths[module_tag.path] = None
             if module_tag.path in bazel_deps:
                 continue
-            paths[module_tag.path] = None
```
fmeum, line 390: Can we keep this unconditional assignment or would that interfere with new logic?

internal/bzlmod/go_deps.bzl
```diff
@@ -403,6 +511,14 @@ def _go_deps_impl(module_ctx):
                     root_module_direct_deps[_repo_name(module_tag.path)] = None
 
             version = semver.to_comparable(raw_version)
+            previous = paths.get(module_tag.path)
+
+            fail_or_warn = len([x for x in module.tags.from_file if x.fail_on_version_conflict == True]) > 0
```
fmeum, line 516:
```suggestion
            fail_or_warn = any([x.fail_on_version_conflict for x in module.tags.from_file])
```

internal/bzlmod/go_deps.bzl
```diff
@@ -553,13 +669,14 @@ def _get_sum_from_module(path, module, sums):
         entry = (module.replace, module.raw_version)
 
     if entry not in sums:
-        fail("No sum for {}@{} found".format(path, module.raw_version))
+        # TODO: if no sum exist, this is probably because a go mod tidy was missed
+        fail("No sum for {}@{} found. You may need to run: bazel run @rules_go//go -- mod tid".format(path, module.raw_version))
```
fmeum, line 673:
```suggestion
        fail("No sum for {}@{} found. You may need to run: bazel run @rules_go//go -- mod tidy".format(path, module.raw_version))
```

internal/bzlmod/go_deps.bzl
```diff
@@ -577,7 +694,9 @@ _config_tag = tag_class(
 
 _from_file_tag = tag_class(
     attrs = {
-        "go_mod": attr.label(mandatory = True),
+        "go_mod": attr.label(mandatory = False),
+        "go_work": attr.label(mandatory = False),
+        "fail_on_version_conflict": attr.bool(default = True),
```
fmeum, line 699: Let's replace this with a constant `True` for now as it gives us more flexibility to experiment with tolerating this situation in the future (e.g. MVS).

internal/bzlmod/go_deps.bzl
```diff
@@ -592,6 +711,10 @@ _module_tag = tag_class(
         ),
         "build_naming_convention": attr.string(doc = """Removed, do not use""", default = ""),
         "build_file_proto_mode": attr.string(doc = """Removed, do not use""", default = ""),
+        "_parent_label": attr.label(
```
fmeum, line 714: Instead of putting this on all tags, could you make it a property of the synthetic tags only that we test for with `getattr`? I don't know how Bzlmod handles tag attributes with a `_` prefix - they may still be settable.

fmeum (commented):
I went over the entire PR and quite like the state of it. Some more comments and then we'll likely be ready for a final review.

tests/bzlmod/semver_test.bzl
```diff
@@ -77,8 +77,23 @@ def _semver_test_impl(ctx):
...
 semver_test = unittest.make(_semver_test_impl)
 
+def _humanize_test_impl(ctx):
+    env = unittest.begin(ctx)
+
+    # remove everything after and including the first + as it is not preserved in a comparable
+    RESTORABLE_VERSIONS = [version.split("+")[0] for version in _SORTED_TEST_VERSIONS]
+
+    for version in RESTORABLE_VERSIONS:
+        humanized = semver.humanize_comparable(semver.to_comparable(version, relaxed = RESTORABLE_VERSIONS))
```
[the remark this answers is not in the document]
stefanpenner: yes, as the data being tested includes values that require the relaxed property. We could remove the relaxed behavior, but I think it wouldn't be testing the full set of possible user inputs. But I am open to changing if requirested

internal/bzlmod/semver.bzl
```diff
@@ -75,6 +75,32 @@ def _semver_to_comparable(v, *, relaxed = False):
         tuple(prerelease),
     )
 
+def _humanize_comparable(version):
+    """
+    Returns a human-readable representation of a comparable version object.
+
+    This is useful for providing more helpful error messages.
+
+    Please note, this is lossy as + meta segments are replaced with sentinal value to preserve expected sorting
```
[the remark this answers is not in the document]
stefanpenner: Potentially with more refactoring, I can explore that. If that refactoring ends up being tedious, would you be OK with this?

internal/bzlmod/go_deps.bzl
stefanpenner: Internally we have relied on this being configurable. This has been quite helpful to our engineerings when dealing with larger migrations to the new form. Would it be OK to continue to support that use-case?

internal/bzlmod/go_deps.bzl
stefanpenner: I believe this is also used for non-synthetic tags. But I need to refresh my memory.... if it is used for non-senthetic, could we keep it?

internal/bzlmod/go_deps.bzl
stefanpenner: Ya, updated.

I ended up removing the extra complexity here, as although the error message could be more concise. The cost of maintaining the logic doesn't IMHO justify the small improvement it provides

internal/bzlmod/go_deps.bzl
stefanpenner: That is much nicer!

internal/bzlmod/go_deps.bzl
stefanpenner: I believe we can unconditionally assign without changing the logic.

internal/bzlmod/semver.bzl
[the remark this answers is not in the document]
stefanpenner: For some reason, I didn't notice that the original versions were available via `module_tag.version` and `previous.module_tag.version`🤦. Given those are already present, I was able to remove all the humanization logic. Much cleaner, and better output.

tests/bzlmod/semver_test.bzl
stefanpenner: After poking around a little more, I realized the original versions were actually present and all this code could be safely removed.

internal/bzlmod/go_deps.bzl
```diff
@@ -571,13 +656,13 @@ def _get_sum_from_module(path, module, sums):
         entry = (module.replace, module.raw_version)
 
     if entry not in sums:
-        fail("No sum for {}@{} found".format(path, module.raw_version))
+        fail("No sum for {}@{} found. You may need to run: bazel run @rules_go//go -- mod tidy".format(path, module.raw_version))
```
stefanpenner, line 659: TODO: Explore if we can also tell them what directory to run this in, as that won't be clear from the error message

internal/bzlmod/go_deps.bzl
ThePatelCode: `fail_on_version_conflict: False` lists all version conflicts at once rather than discovering them incrementally, making it less frustrating for developers. Also, having the option to defer version conflicts in the 25-30 range while addressing more pressing issues is a great benefit of this optional flag when migrating big projects.

tests/bcr/go.mod
```diff
@@ -1,25 +1,3 @@
-// This will stop go mod from descending into this directory.
```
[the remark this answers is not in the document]
stefanpenner: Looking at this one now.

hunshcn:
Is now ready?

stefanpenner:
@hunshcn this should be close, I believe we are pending a final round of code-reviews from the maintainers. I believe some of them have been on DTO.

internal/bzlmod/go_mod.bzl
```diff
@@ -16,6 +16,94 @@ visibility([
...
 
+def _validate_go_version(path, state, tokens, line_no):
+    if len(tokens) == 1:
+        fail("{}:{}: expected another token after 'go'".format(path, line_no))
+    if state["go"] != None:
+        fail("{}:{}: unexpected second 'go' directive".format(path, line_no))
+    if len(tokens) > 2:
+        fail("{}:{}: unexpected token '{}' after '{}'".format(path, line_no, tokens[2], tokens[1]))
+
+def use_spec_to_label(workspace_name, use_directive):
```
fmeum, line 27: Let's call the parameter `repo_name`. `Label#workspace_name` is deprecated and the replacement is also called `repo_name`.

internal/bzlmod/go_deps.bzl
```diff
@@ -276,6 +282,49 @@ _go_repository_config = repository_rule(
...
+        previous: The previous module object.
+        module_tag: The module tag.
+        module_name_to_go_dot_mod_label: A dictionary mapping module paths to go.mod labels.
+        previous: The previous module object.
+    """
+
+    if not previous:
+        # no previous module, so no possible error
+        return
+
```
fmeum, line 300: Duplicated with the condition below
```suggestion
```

internal/bzlmod/go_deps.bzl
```diff
@@ -276,6 +282,49 @@ _go_repository_config = repository_rule(
...
+        # version is the same, skip because we won't error
+        return
+
+    # When using go.work, duplicate dependency versions are possible.
+    # This can cause issues, so we fail with a hopefully actionable error.
+    current_label = module_tag._parent_label
+
+    previous_label = previous.module_tag._parent_label
+
+    corrective_measure = "To correct this:\n 1. ensure that '{}' in all go.mod files is the same version.\n 2. in the folders where you made changes to run: go mod tidy\n 3. at the workspace root run: go work sync.".format(module_tag.path)
```
fmeum, line 311: You can use `"""..."""` string literals instead, which can contain literal line breaks, to make this more readable.

```suggestion
    corrective_measure = "To correct this:\n 1. ensure that '{}' in all go.mod files is the same version.\n 2. in the folders where you made changes run: bazel run @rules_go//go -- mod tidy\n 3. at the workspace root run: bazel run @rules_go//go -- work sync.".format(module_tag.path)
```

internal/bzlmod/go_deps.bzl
```diff
@@ -276,6 +282,49 @@ _go_repository_config = repository_rule(
...
+def fail_on_version_conflict(version, previous, module_tag, module_name_to_go_dot_mod_label, fail_or_warn):
+    """
+    Check if duplicate modules have different versions, and fail with a useful error message if they do.
+
+    Args:
+        version: The version of the module.
+        previous: The previous module object.
+        module_tag: The module tag.
+        module_name_to_go_dot_mod_label: A dictionary mapping module paths to go.mod labels.
+        previous: The previous module object.
```
fmeum, line 294: Out of date

internal/bzlmod/go_deps.bzl
```diff
@@ -276,6 +282,49 @@ _go_repository_config = repository_rule(
     },
 )
 
+def fail_on_version_conflict(version, previous, module_tag, module_name_to_go_dot_mod_label, fail_or_warn):
```
fmeum, line 285: "fail_or_warn" makes it hard to figure out what to pass for what behavior.

Instead, how about:
* call this function `check_for_version_conflict`
* call the last argument `conflict_printer` and pass in the `print` or `fail` function at the call site

internal/bzlmod/go_deps.bzl
```diff
@@ -276,6 +282,49 @@ _go_repository_config = repository_rule(
...
+
+    message = "Multiple versions of {} found:\n - {} contains: {}\n - {} contains {}.\n{}".format(module_tag.path, current_label, module_tag.version, previous_label, previous.module_tag.version, corrective_measure)
+
+    if fail_or_warn:
+        fail(message)
+    else:
+        print(message)
+
+def _fail_if_not_root(module, from_file_tag):
+    if module.is_root != True:
```
fmeum, line 321:
```suggestion
    if not module.is_root:
```

internal/bzlmod/go_deps.bzl
```diff
@@ -344,18 +393,46 @@ def _go_deps_impl(module_ctx):
...
+            _fail_if_invalid_from_file_usage(from_file_tag)
+
+            if from_file_tag.go_mod:
+                from_file_tags.append(from_file_tag)
+            elif from_file_tag.go_work:
+                _fail_if_not_root(module, from_file_tag)
+
+                go_work = go_work_from_label(module_ctx, from_file_tag.go_work)
+
+                # this ensures go.work replacements as considered
```
fmeum, line 411:
```suggestion
                # this ensures go.work replacements are considered
```

internal/bzlmod/go_deps.bzl
```diff
@@ -607,7 +691,9 @@ _config_tag = tag_class(
 
 _from_file_tag = tag_class(
     attrs = {
-        "go_mod": attr.label(mandatory = True),
+        "go_mod": attr.label(mandatory = False),
+        "go_work": attr.label(mandatory = False),
+        "fail_on_version_conflict": attr.bool(default = True),
```
fmeum, line 696: Could you add a `doc` comment to this new attribute?

internal/bzlmod/go_deps.bzl
```diff
@@ -580,13 +664,13 @@ def _get_sum_from_module(path, module, sums):
...
     if entry not in sums:
-        fail("No sum for {}@{} found".format(path, module.raw_version))
+        fail("No sum for {}@{} found. You may need to run: bazel run @rules_go//go -- mod tidy".format(path, module.raw_version))
 
     return sums[entry]
 
 def _safe_insert_sum(sums, entry, new_sum):
     if entry in sums and new_sum != sums[entry]:
-        fail("Multiple mismatching sums for {}@{} found.".format(entry[0], entry[1]))
+        fail("Multiple mismatching sums for {}@{} found. {} vs {}".format(entry[0], entry[1], new_sum, sums[entry]))
```
fmeum, line 673:
```suggestion
        fail("Multiple mismatching sums for {}@{} found: {} vs {}".format(entry[0], entry[1], new_sum, sums[entry]))
```

internal/bzlmod/go_deps.bzl
```diff
@@ -276,6 +282,49 @@ _go_repository_config = repository_rule(
...
+    corrective_measure = "To correct this:\n 1. ensure that '{}' in all go.mod files is the same version.\n 2. in the folders where you made changes to run: go mod tidy\n 3. at the workspace root run: go work sync.".format(module_tag.path)
+
+    message = "Multiple versions of {} found:\n - {} contains: {}\n - {} contains {}.\n{}".format(module_tag.path, current_label, module_tag.version, previous_label, previous.module_tag.version, corrective_measure)
+
+    if fail_or_warn:
+        fail(message)
+    else:
+        print(message)
+
+def _fail_if_not_root(module, from_file_tag):
```
fmeum, line 320: Let's inline this as it is only used once and the message is specific to `go_work`.

internal/bzlmod/go_deps.bzl
```diff
@@ -276,6 +282,49 @@ _go_repository_config = repository_rule(
...
+    if module.is_root != True:
+        fail("go_deps.from_file(go_work = '{}') tag can only be used from a root module but: '{}' is not a root module.".format(from_file_tag.go_work, module.name))
+
+def _fail_if_invalid_from_file_usage(from_file_tag):
```
fmeum, line 324: Let's also inline this as it is only used once and very short.

internal/bzlmod/go_deps.bzl
```diff
@@ -427,6 +503,14 @@ def _go_deps_impl(module_ctx):
                     root_module_direct_deps[_repo_name(module_tag.path)] = None
 
             version = semver.to_comparable(raw_version)
+            previous = paths.get(module_tag.path)
+
+            fail_or_warn = any([x.fail_on_version_conflict for x in module.tags.from_file])
+
+            # rather then failing, we could do MVS here, or some other heuristic
```
fmeum, line 510:
```suggestion
            # TODO: Rather than failing, we could do MVS here, or some other heuristic
```

fmeum (approved):
This looks good to me and I expect this to have been the final round of review. I mostly suggested local changes.

internal/bzlmod/go_mod.bzl
```diff
@@ -16,6 +16,94 @@ visibility([
...
+    if use_directive.find("..") >= 0:
+        fail("go.work use directive: '{}' contains '..' which is not currently supported.".format(use_directive))
+
+    if use_directive.startswith("./"):
+        use_directive = use_directive[2:]
+
+    if use_directive.startswith("."):
```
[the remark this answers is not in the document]
stefanpenner: @fmeum I believe so, as the `use_directive` in the label we are generated bellow can't have a `.`.

The following my understanding, but feel free to correct me if I'm missing something:

❌  label to the go.mod that use directive refers to with the dot: `@@bar//.foo:go.mod`       <-- invalid label
✅  label to the go.mod that use directive refers to without the dot: `@@bar//foo:go.mod`  <-- valid label

.bazelci/presubmit.yml
```diff
@@ -4,6 +4,9 @@ matrix:
     - ubuntu2204
     - macos
   bazel: [6.x, 7.x]
+  bcr_variant:
+    - tests/bcr/go_work
+    - tests/bcr/go_mod
```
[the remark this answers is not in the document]
stefanpenner: I'm not 100% sure I understood, but is this the change you wanted: https://github.com/bazelbuild/bazel-gazelle/pull/1731/files#diff-013d9ca829fec7936ff875c095e37317d86488040d8b746f048f25066601650eR2 ?

internal/bzlmod/go_deps.bzl
stefanpenner: I like it!

internal/bzlmod/go_mod.bzl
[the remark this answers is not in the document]
fmeum: Why is the first label invalid? While it would be very unusual to have a `go.mod` in a hidden directory, it doesn't seem impossible.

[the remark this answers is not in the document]
stefanpenner: My bad, you right, I don't think 34 is needed. I confused it with 31 for some reason. Let me drop this scenario.

stefanpenner:
@fmeum I believe I've addressed all the outstanding feedback, but let me know if I missed anything critical and I'll jump right on it.

stefanpenner:
rebased

stefanpenner:
Follow up PR that adds FilePath ReplaceDirective support can be found here: https://github.com/bazelbuild/bazel-gazelle/pull/1776

internal/bzlmod/go_deps.bzl
```diff
@@ -276,6 +282,40 @@ _go_repository_config = repository_rule(
...
+    previous_label = previous.module_tag._parent_label
+
+    corrective_measure = """To correct this:
+    1. ensure that '{}' in all go.mod files is the same version.
+    2. in the folders where you made changes run: bazel run @rules_go//go -- mod tidy
+    3. at the workspace root run: bazel run @rules_go//go -- work sync.""".format(module_tag.path)
+
+    message = """Multiple versions of {} found:
+    - {} contains: {}
+    - {} contains {}.{}""".format(module_tag.path, current_label, module_tag.version, previous_label, previous.module_tag.version, corrective_measure)
```
fmeum, line 315:
```suggestion
    - {} contains: {}
{}""".format(module_tag.path, current_label, module_tag.version, previous_label, previous.module_tag.version, corrective_measure)
```

internal/bzlmod/go_deps.bzl
```diff
@@ -344,12 +384,41 @@ def _go_deps_impl(module_ctx):
...
+                from_file_tags = from_file_tags + go_work.from_file_tags
+            else:
+                fail("Either \"go_mod\" or \"go_work\" must be specified in \"go_deps.from_file\" tags.")
+
+        for from_file_tag in from_file_tags:
+            module_path, module_tags_from_go_mod, go_mod_replace_map, module_name = deps_from_go_mod(module_ctx, from_file_tag.go_mod)
+            module_name_to_go_dot_mod_label[module_name] = from_file_tag.go_mod
 
-            # Collect the relative path of the root module's go.mod file if it lives in the main
-            # repository.
```
fmeum, line 352: Keep this comment.

internal/bzlmod/go_mod.bzl
```diff
@@ -240,16 +326,41 @@ def sums_from_go_mod(module_ctx, go_mod_label):
...
+    # into the module that provides the sumfile
+    sum_label = Label("@@{}//{}:{}".format(
+        label.workspace_name,
+        label.package,
+        sumfile,
     ))
-    go_sum_content = module_ctx.read(go_sum_label)
-    return parse_go_sum(go_sum_content)
+
+    if module_ctx.path(sum_label).exists:
```
fmeum, line 359: Does this work if the file doesn't exist? Maybe delete the `go.work.sum` file from the test module as it existing should be the less likely case.

fmeum (approved):

internal/bzlmod/go_deps.bzl
stefanpenner: Good catch!

internal/bzlmod/go_mod.bzl
stefanpenner: @fmeum you are right that doesn't work as I expected. I'm actually not 100% sure how to handle this case now, beyond creating an executable and shelling out to test the files existence which seems funky.

It is really unfortunate to not support both cases \w go.work.sum and without go.work.sum... any guidance?

fmeum: `ctx.path(label)` fails if the file doesn't exist, but `ctx.path(string)` doesn't. So what you can do is:
1. Use `ctx.path()` to resolve the label of go.work to a path. It always exists, so that's fine.
2. Manipulate the path string to be that of `go.work.sum` and turn that into a path, then test for existence.
3. If it exists, also `ctx.path` the label of `go.work.sum` to let Bazel track its contents.
4. If `ctx.watch` is available, also watch the (non-existent) path so that the rule is invalidated if the file shows up. I don't think that's possible without `watch`, but maybe that's fine because the file can't change without any of the `*.mod` files changing?

stefanpenner: @fmeum oh let me give that a try.

internal/bzlmod/go_mod.bzl
```diff
@@ -240,16 +326,55 @@ def sums_from_go_mod(module_ctx, go_mod_label):
...
+    if maybe_go_work_sum_path.exists:
+        return parse_sumfile(module_ctx, go_work_label, "go.work.sum")
+    else:
+        # 4. if go.work.sum does not exist, we should watch it incase it appears in the future
+        if hasattr(module_ctx, "watch_tree"):
+            # module_ctx.watch_tree is only available in bazel >= 7.1
+            module_ctx.watch_tree(maybe_go_work_sum_path)
+
+        # 5. return an empty dict as no sum file was found
+        return {}
```
stefanpenner, line 365: @fmeum here is my, seemingly working, attempt at implementing https://github.com/bazelbuild/bazel-gazelle/pull/1731#discussion_r1564615266 

I've tested this manually a little, with and without go.work.sum files, and it appears to work. But it could use another set of eyes to sanity check.

internal/bzlmod/go_mod.bzl
```diff
@@ -240,16 +326,55 @@ def sums_from_go_mod(module_ctx, go_mod_label):
...
+    maybe_go_work_sum_path = go_work_path.get_child("../go.work.sum")
+
+    # 3. check for it's existence
+    if maybe_go_work_sum_path.exists:
+        return parse_sumfile(module_ctx, go_work_label, "go.work.sum")
+    else:
+        # 4. if go.work.sum does not exist, we should watch it incase it appears in the future
+        if hasattr(module_ctx, "watch_tree"):
+            # module_ctx.watch_tree is only available in bazel >= 7.1
+            module_ctx.watch_tree(maybe_go_work_sum_path)
```
stefanpenner, line 362: @fmeum here is my, seemingly working, attempt at implementing https://github.com/bazelbuild/bazel-gazelle/pull/1731#discussion_r1564615266

I've tested this manually a little, with and without go.work.sum files, and it appears to work. But it could use another set of eyes to sanity check.

internal/bzlmod/go_mod.bzl
stefanpenner: the attempt -> https://github.com/bazelbuild/bazel-gazelle/pull/1731

tyler-french (approved):
Tested at Uber

internal/bzlmod/go_mod.bzl
```diff
@@ -240,16 +326,55 @@ def sums_from_go_mod(module_ctx, go_mod_label):
...
+    """
+    _check_go_work_name(go_work_label.name)
+
+    # next we need to test if the go.work.sum file exists, this is a little tricky so we use an indirect approach:
+
+    # 1. convert go_work_label into a path
+    go_work_path = module_ctx.path(go_work_label)
+
+    # 2. use the go_work_path to create a path for the heisen go.work.sum file
+    maybe_go_work_sum_path = go_work_path.get_child("../go.work.sum")
```
fmeum, line 353:
```suggestion
    maybe_go_work_sum_path = go_work_path.dirname.get_child("go.work.sum")
```

internal/bzlmod/go_mod.bzl
```diff
@@ -240,16 +326,55 @@ def sums_from_go_mod(module_ctx, go_mod_label):
...
+
+    # 3. check for it's existence
+    if maybe_go_work_sum_path.exists:
+        return parse_sumfile(module_ctx, go_work_label, "go.work.sum")
+    else:
+        # 4. if go.work.sum does not exist, we should watch it incase it appears in the future
+        if hasattr(module_ctx, "watch_tree"):
```
fmeum, line 360: It's a single file, `watch` should be sufficient.

internal/bzlmod/go_mod.bzl
```diff
@@ -240,16 +326,55 @@ def sums_from_go_mod(module_ctx, go_mod_label):
...
+    go_work_path = module_ctx.path(go_work_label)
+
+    # 2. use the go_work_path to create a path for the heisen go.work.sum file
+    maybe_go_work_sum_path = go_work_path.get_child("../go.work.sum")
+
+    # 3. check for it's existence
+    if maybe_go_work_sum_path.exists:
+        return parse_sumfile(module_ctx, go_work_label, "go.work.sum")
+    else:
+        # 4. if go.work.sum does not exist, we should watch it incase it appears in the future
```
fmeum, line 359:
```suggestion
        # 4. if go.work.sum does not exist, we should watch it in case it appears in the future
```

internal/bzlmod/go_mod.bzl
```diff
@@ -240,16 +326,55 @@ def sums_from_go_mod(module_ctx, go_mod_label):
...
+
+    # next we need to test if the go.work.sum file exists, this is a little tricky so we use an indirect approach:
+
+    # 1. convert go_work_label into a path
+    go_work_path = module_ctx.path(go_work_label)
+
+    # 2. use the go_work_path to create a path for the heisen go.work.sum file
+    maybe_go_work_sum_path = go_work_path.get_child("../go.work.sum")
+
+    # 3. check for it's existence
```
fmeum, line 355:
```suggestion
    # 3. check for its existence
```

stefanpenner:
I've been thinking about how the resolutions work, and I want to confirm a few more more advanced scenarios. Let's hold off on merging for the moment.

I should be able to finish the testing by tomorrow, and report back. :)

internal/bzlmod/go_mod.bzl
fmeum, line 362: Same as above, use `watch`

fmeum (approved):

fmeum:
Sounds good, remaining comments are all just nits in changed code.

stefanpenner:
I believe I have a reasonable resolution to the issue I noticed, working to implement the fix now. Let's see how it goes :)

stefanpenner:
> I believe I have a reasonable resolution to the issue I noticed, working to implement the fix now. Let's see how it goes :)

TL;DR: Within a go.work, if module A depends on an external module B (not in go.work), and B in turn depends on module C which is included in the go.work, then the version of C referenced by B will differ from the version of C that is directly included in the go.work.

Our team has been describing this the "the sandwich problem".

In general, a well formed workspace must not contain the _the sandwich problem_ as even in a perfectly functioning system this scenario will be a source of headaches and confusion.

Given this, my recommendation is:

1. proceed with this PR
2. separately explore how to enable this capability within gazelle and then add this capability in the future.

fmeum (approved):

fmeum:
Agreed, let's merge this and as a follow-up:
1. Resolve the problem you mentioned.
2. Add docs to our Bzlmod guide.

fmeum:
Thanks for working on this large project!

Issue #1738, opened 2024-02-22
Gazaelle not recognising math/rand/v2 as stdlib


udf2457:
### What version of gazelle are you using?

0.35.0

### What version of rules_go are you using?

0.45.1

### What version of Bazel are you using?

7.0.2

### What operating system and processor architecture are you using?

MacOS Apple Silicon 

### What did you do?

Gazelle appears not to recognise `math/rand/v2` is part of stdlib.

I have other code in my repo that uses other parts of go stdlib and I have no issues, its only this new code I've added that uses `math/rand/v2` that causes issues.

```
gazelle: finding module path for import math/rand/v2: go: malformed module path "math/rand/v2": missing dot in first path element
ERROR: /Users/<REDACTED>/inet/web/BUILD.bazel:19:8: GoCompilePkg internal/inet/web/web_test.internal.a failed: (Exit 1): builder failed: error executing GoCompilePkg command (from target //internal/inet/web:web_test) bazel-out/darwin_arm64-opt-exec-ST-13d3ddad9198/bin/external/rules_go~0.45.1~go_sdk~go_default_sdk/builder_reset/builder compilepkg -sdk external/rules_go~0.45.1~go_sdk~go_default_sdk -installsuffix ... (remaining 39 arguments skipped)
```

udf2457:
For anyone who finds this at a later date, the problem was becasue default `go_sdk` download is `1.21.1`, so I had to explicitly name the current version in `MODULE.bazel`, i.e. `go_sdk.download(version = "1.22.0")`

Clearly the error message provided by `gazelle` is incredibly unhelpful, verging on useless though !

PR #1740, opened 2024-02-23
add new std_packages from go 1.22

tyler-french:
New packages added in: https://tip.golang.org/doc/go1.22#go/version

Resolves some of #1738


rabbbit (approved):

fmeum (approved):
Could we also improve the error message when referring to a package of this form (no `.`) but it isn't found? That would help on future additions to the stdlib.

tyler-french:
> Could we also improve the error message when referring to a package of this form (no `.`) but it isn't found? That would help on future additions to the stdlib.

Sure! We can keep the issue open, but I think this can be a separate PR.

PR #1753, opened 2024-03-13
Warn if integrity not specified for archive_override

scaiper:
**What type of PR is this?**

Feature

go_repository

This PR adds a warning for not specified sha256 for archive in the go_repository rule which provides the correct checksum.
It simplifies finding and fixing such cases. It is important for reproducible builds and caching.

It is a small change, not sure if issue is needed.


fmeum (approved):
Thanks!

PR #1755, opened 2024-03-13
Use host module cache whithout build cache and verify it

### LINKED ISSUE #1754: Use host module cache without build cache

0.35.0

0.45.1

7.0.2

### Description

We would like to have somewhat persistent Go module cache on our CI workers. We've looked at GO_REPOSITORY_USE_HOST_CACHE but it overrides GOPATH and GOCACHE also, while we only need to change GOMODCACHE. Secondly, `go mod download` do not verify a module in case of a case hit, which is not ideal.

scaiper:
**What type of PR is this?**

Feature

go_repository_cache

This PR adds an environment variable GO_REPOSITORY_USE_HOST_MODCACHE, which enables use of only GOMODCACHE without affecting GOPATH and GOCACHE.
And adds a sum verification to fetch_repo tool, since `go mod download` do not verify modules taken from cache.

Fixes #1754

### DIFF
--- a/cmd/fetch_repo/BUILD.bazel
+++ b/cmd/fetch_repo/BUILD.bazel
@@ -9,7 +9,10 @@ go_library(
     ],
     importpath = "github.com/bazelbuild/bazel-gazelle/cmd/fetch_repo",
     visibility = ["//visibility:private"],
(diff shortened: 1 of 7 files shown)


scaiper:
Thanks for the fixes. Missed it.

fmeum (approved):

PR #1765, opened 2024-03-25
Make `# gazelle:proto file` work without needing to set different `option go_package` in .proto files

### LINKED ISSUE #1724: #gazelle:proto file” directive isn’t usable unless option go_package is explicitly set in each proto_library
0.29.0

0.40.1

7.0.0

yes

macOS Sonoma 14.0, arm64

Context: We maintain a large golang repository with gazelle, and often have multiple .proto files in a single directory. Due to some tooling requirements, we want each .proto file to have its own proto_library rule.

I tried using the “#gazelle:proto file” directive, but this also caused separate go_proto_library targets to be generated, with the same importpaths, which seems to cause build failures (similar issue [here](https://github.com/bazelbuild/rules_go/issues/1841)). It’s not really an option for us to enforce specification of unique “option go_package” lines in the .proto files. Ideally, we’d have some directive that works with “#gazelle:proto file” – maybe something like “#gazelle:go_proto importpath” so that we can generate a proto_library target for each .proto file, but only have one go_proto_library target per “inferred importpath” BUILD file. This would mean setting the “protos” attribute of the go_proto_library, which currently doesn’t seem to be supported. We’re thinking of creating a patch for our purposes, but was curious if there’s a better way to accomplish what we want, or if something like this could be an upstream change.

jeromep-stripe:
**What type of PR is this?**
Feature

language/go

This changes the gazelle's `go_proto_library` generation behavior.
Previously, in .proto file mode, gazelle would generate a separate `go_proto_library` target for each `proto_library` target, and these would have the same values for the importpath attribute. This caused build errors in certain situations.

This change makes Gazelle group `proto_library` targets which would otherwise have the same `go_proto_library` importpath, into a single `go_proto_library` using the `protos` attribute.

Fixes #1724

**Other notes for review**
- While testing this PR, I ran `bazelisk test //...` and saw two tests fail 
```
FAIL: //cmd/gazelle:gazelle_test (see /private/var/tmp/_bazel/7faad61fac7f6d421e59e31d0146c79e/execroot/_main/bazel-out/darwin_arm64-fastbuild/testlogs/cmd/gazelle/gazelle_test/test.log)
INFO: From Testing //cmd/gazelle:gazelle_test:
==================== Test output for //cmd/gazelle:gazelle_test:
could not locate go tool
================================================================================
FAIL: //internal:bazel_test (see /private/var/tmp/_bazel/7faad61fac7f6d421e59e31d0146c79e/execroot/_main/bazel-out/darwin_arm64-fastbuild/testlogs/internal/bazel_test/test.log)
INFO: From Testing //internal:bazel_test:
==================== Test output for //internal:bazel_test:
error: unknown runfile: ./.bazelrc
================================================================================
```
This looks like an environment issue. Looks like CI is green though, so I think we can ignore this.

- When generating a single `go_proto_library` target for multiple `proto_library` targets, the change creates a `combinedImports` [platformStringsBuilder variable](https://github.com/bazelbuild/bazel-gazelle/pull/1765/files#diff-f616e692f90c14eb97101f4943cec5fce1d3bade9e81e76e67ba82cdd55c2025R562-R586). I'm not sure if my general approach to combining the dependencies is correct, and would appreciate guidance.


ouguoc2-stripe:
@linzhp FYI -- let @jeromep-stripe know what you think.

ouguoc2-stripe:
Ping on this @linzhp -- let us know what we can do to help.

linzhp:
Oh, sorry, I missed the previous ping. Let me take a look later this week.

language/go/generate.go
```diff
@@ -427,53 +478,125 @@ func (g *generator) generateProto(mode proto.Mode, target protoTarget, importPat
...
+		for _, target := range targets {
+			builtImports := target.imports.build()
+			// handle Generics
+			for _, genericString := range builtImports.Generic {
+				// if the generic string is not in the set of generic src strings, add it
+				if _, ok := protoSources[genericString]; !ok {
+					combinedImports.addGenericString(genericString)
+				}
+			}
+			//// handle OS
```
linzhp, line 572: Do we need OS/Arch/Platform strings? I guess proto imports are all independent of OS/Arch/Platform?

language/go/generate.go
```diff
@@ -427,53 +478,125 @@ func (g *generator) generateProto(mode proto.Mode, target protoTarget, importPat
...
 		goProtoLibrary.SetAttr("compilers", gc.goProtoCompilers)
 	}
 	if g.shouldSetVisibility {
 		goProtoLibrary.SetAttr("visibility", visibility)
 	}
-	goProtoLibrary.SetPrivateAttr(config.GazelleImportsKey, target.imports.build())
+	if len(targets) == 1 {
+		goProtoLibrary.SetPrivateAttr(config.GazelleImportsKey, targets[0].imports.build())
+	} else {
+		protoSources := make(map[string]bool)
```
linzhp, line 552: nit: prefer `map[string]struct{}`, because each entry in map[string]bool has 3 states: not exist, true, false, but we only need to know whether it exists or not.

linzhp:
Some minor comments. LGTM overall. I also tested it in Uber's Go repo.

language/go/generate.go
jeromep-stripe: Unsure. I was pattern matching and just trying to combine everything into a single platformStringsBuilder

language/go/generate.go
jeromep-stripe: I don't think I understand. `protoSources` is a set being used to dedupe proto src files, which I implemented as a map. Why would there need to be 3 states then?

jeromep-stripe:
@linzhp ty for the review, I addressed your comments!

language/go/config.go
```diff
@@ -28,9 +28,10 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/bazelbuild/bazel-gazelle/internal/module"
```
linzhp: bring this back to the group below

linzhp (approved):

PR #1767, opened 2024-03-28
Ignore `purego` build constraint

mattyclarkson:
`purego` is a proposed[1] de-facto build constraint to denote building _only_ Go code.

We must ignore this constraint to make sure that files that toggle on `purego` are added to the Bazel target `srcs` and let the compiler decide if they should be compiled or not.

[1]: https://github.com/golang/go/issues/23172

This is required to land bazelbuild/rules_go#3901 which will enable the de-facto `purego` build constraint to be picked up in the Bazel Go ecosystem automatically when `pure = "on"` is defined in a `go_binary`, etc.

### DIFF
--- a/language/go/build_constraints.go
+++ b/language/go/build_constraints.go
@@ -307,13 +307,13 @@ func matchAuto(tokens []string) (*buildTags, error) {
 	return newBuildTags(x)
 }
 
-// isIgnoredTag returns whether the tag is "cgo" or is a release tag.
+// isIgnoredTag returns whether the tag is "cgo", "purego", "race", "msan"  or is a release tag.
 // Release tags match the pattern "go[0-9]\.[0-9]+".
 // Gazelle won't consider whether an ignored tag is satisfied when evaluating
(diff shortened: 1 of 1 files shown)


language/go/build_constraints.go
```diff
@@ -313,7 +313,7 @@ func matchAuto(tokens []string) (*buildTags, error) {
 // build constraints for a file and will instead defer to the compiler at compile
 // time.
 func isIgnoredTag(tag string) bool {
-	if tag == "cgo" || tag == "race" || tag == "msan" {
+	if tag == "cgo" || tag == "purego" || tag == "race" || tag == "msan" {
```
sluongng, line 316: Might need to fix the doc as well

> isIgnoredTag returns whether the tag is "cgo" or is a release tag.

fmeum (approved):
Looks good to me, just update the comment.

language/go/build_constraints.go
mattyclarkson: Updated

PR #1799, opened 2024-05-07
[Fixes #1736] disable proto generation for external go modules by default

### LINKED ISSUE #1736: Proto generation is not yet disabled in go.mods provided by go_deps in bzlmod

Latest (0.35)

latest (0.44)

Latest (7.0.0)

yes

all

### What did you do?

Add a go.mod which contains both `.proto` and `.pb.go` files, via go_deps.

No protoc generation, generated BUILD files should reference the `pb.go` file.

### What did you see instead?

When gazelle encounters `.proto` files it generates BUILD rules that compile the protos. This creates both a unexpected dependency on protoc, slower setup, and build experiences.

stefanpenner:
[Fixes #1736] disable proto generation for external go modules by default

**What type of PR is this?**

>

Bug fix

> Feature
> Documentation
> Other

> For example:
>
> language/go
> cmd/gazelle

go_repository

> all

By convention go modules with proto files are to generate proto.go bindings at publish time, this PR ensures we don't regenerate those bindings for external dependencies, as doing so is quite wasteful as it introduces the full protoc toolchain which itself also needs to be built from scratch.

For folks that want the old behavior, they can add an override for the go mod in question by doing the following:

```starlark
go_deps.gazelle_override(
    directives = [ ],
    path = "<go module id>",
)
```

Fixes #1736

**Other notes for review**


internal/bzlmod/default_gazelle_overrides.bzl
```diff
@@ -22,60 +22,29 @@ DEFAULT_BUILD_FILE_GENERATION_BY_PATH = {
...
-    ],
-    "github.com/cockroachdb/errors": [
-        "gazelle:proto disable",
-    ],
-    "github.com/gogo/googleapis": [
-        "gazelle:proto disable",
-    ],
-    "github.com/gogo/protobuf": [
-        "gazelle:proto disable",
+        "gazelle:proto disable_global",
```
stefanpenner, line 27: Context for this line (and the others that re-introduce `gazelle:proto disable_global`)

given how overrides work, when an override is provided it will override the default behavior. This means existing overrides that wish to have the new default behavior, will need to now include `gazelle:proto disable_global` themselves. ()

This feels a little odd, so we should discuss. As far as I see we have the following options:

a. do what this PR does
b. don't bother introducing `gazelle:proto disable_global` to this DEFAULT_DIRECTIVES unless the mod has proto files already
c. invest in more advanced merging so that the defaults are not lost during merging. This would then also require the ability selectively override existing defaults

internal/bzlmod/go_deps.bzl
```diff
@@ -144,7 +144,7 @@ def _get_override_or_default(specific_overrides, gazelle_default_attributes, def
     return default_value
 
 def _get_directives(path, gazelle_overrides, gazelle_default_attributes):
-    return _get_override_or_default(gazelle_overrides, gazelle_default_attributes, DEFAULT_DIRECTIVES_BY_PATH, path, [], "directives")
+    return _get_override_or_default(gazelle_overrides, gazelle_default_attributes, DEFAULT_DIRECTIVES_BY_PATH, path, ["gazelle:proto disable_global"], "directives")
```
stefanpenner, line 147: Do we want this to be `disable` or `disable_global` ? (I'll poke around to understand the difference better)

internal/bzlmod/default_gazelle_overrides.bzl
```diff
@@ -22,60 +22,29 @@ DEFAULT_BUILD_FILE_GENERATION_BY_PATH = {
...
-        "gazelle:proto disable",
-    ],
-    "k8s.io/apiextensions-apiserver": [
-        "gazelle:proto disable",
+        "gazelle:proto disable_global",
     ],
     "k8s.io/apimachinery": [
         "gazelle:go_generate_proto false",
         "gazelle:proto_import_prefix k8s.io/apimachinery",
+        "gazelle:proto disable_global",
```
linzhp, line 47: Let's don't disable proto here. People will need the proto targets from this module, otherwise they will have to download it again with `http_archive`.

internal/bzlmod/default_gazelle_overrides.bzl
linzhp: I think we should do b: if those modules don't have `gazelle:proto disable` already, it either means they don't have proto files or people still need the proto files.

internal/bzlmod/go_deps.bzl
linzhp: For a Go module without any proto file, `disable` and `disable_global` no longer have a difference. I slightly prefer "disable" just because it's shorter and doesn't let people wonder what "global" means

linzhp (approved):

hunshcn:
Any update?

hunshcn:
ping @stefanpenner 
I think this work is very meaningful, and it would be great if it could be merged.

stefanpenner:
Looks Like im working on bazel stuff again, so I will try to carve out some time to help Shepard this forward.

fmeum:
There's also https://github.com/bazelbuild/bazel-gazelle/pull/1888, so @TvdW and you might want to align.

TvdW:
> There's also #1888, so @TvdW and you might want to align.

Indeed, I created mine after noticing that this one was not moving forward anymore. They do the same thing. Mine covers both build files and proto files, so if we land this one first it reduces the diff on mine. That said, I'm happy to push them both forward at the same time in my PR as well. Your call, @stefanpenner!

PR #1800, opened 2024-05-11
Fix go.work use ROOT moddir

hunshcn:
Fix go work use ROOT moddir

It works on https://github.com/stefanpenner/bazel-gazelle/commit/acaa88e691debcbccf87a996e7b9bd00d3e104f2, but fail at master.

cc @stefanpenner

### DIFF
--- a/tests/bcr/go_work/go.work
+++ b/tests/bcr/go_work/go.work
@@ -1,9 +1,10 @@
 go 1.21
 
 use (
+	.
 	./pkg
-	./tools
 	./proto
+	./tools
 )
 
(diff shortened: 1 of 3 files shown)


fmeum (approved):
Thanks for sending the fix!

stefanpenner:
@hunshcn thank you!

tests/bcr/go_work/go.work
```diff
@@ -1,9 +1,10 @@
 go 1.21
 
 use (
+	.
```
stefanpenner, line 4: this is a scenario I wasn't aware of, thanks for adding it!

PR #1806, opened 2024-05-19
Add support for `debug_mode` option to `go_deps`

### LINKED ISSUE #1805: Passing debug_mode via go_deps to go_repository is not supported
It appears that there is no way to pass in `debug_mode` to `go_repository` via `go_deps` with bzlmod. Unless I'm missing something, this makes it difficult to see when things go wrong during BUILD file generation because errors are not reported directly from the `gazelle` command.

https://github.com/bazelbuild/bazel-gazelle/blob/f93e45f142de1a32389f1b8798880e5cf903c432/internal/go_repository.bzl#L372-L374

davidbyttow:
**What type of PR is this?**
Bug fix

go_deps

This allows debug_mode to be passed to each `go_repository` as a global option. It could be further extended to allow on a per-module basis later.

Fixes #1805


fmeum (approved):

fmeum:
Thanks for resolving this yourself!

PR #1820, opened 2024-05-31
Optimize Gazelle's filesystem traversal

### LINKED ISSUE #1819: Optimize Gazelle's filesystem walk
As discussed in https://github.com/bazelbuild/bazel-gazelle/pull/1737#issuecomment-2016614858 and the previous comments on that PR

dzbarsky:
**What type of PR is this?**

Performance

All

This PR improves Gazelle's filesystem walk in 2 ways:
1. By using a parallel walk, resolving a long-standing TODO
2. By adding an optional interface that plugins can implement to receive the directory contents, thus saving them from issuing duplicative syscalls to gather the same info

Fixes #1819

**Other notes for review**

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -52,6 +52,8 @@ use_repo(
     # Dependencies of rules_go that require special handling in go_deps.
     "com_github_golang_protobuf",
     "org_golang_google_protobuf",
+    # fastwalk
+    "com_github_charlievieth_fastwalk",
 )
 
 bazel_dep(name = "bazel_skylib_gazelle_plugin", version = "1.4.1", dev_dependency = True)
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -90,6 +90,7 @@ register_unittest_toolchains()
 # gazelle:repository go_repository name=com_github_bmatcuk_doublestar_v4 importpath=github.com/bmatcuk/doublestar/v4
 # gazelle:repository go_repository name=com_github_burntsushi_toml importpath=github.com/BurntSushi/toml
 # gazelle:repository go_repository name=com_github_census_instrumentation_opencensus_proto importpath=github.com/census-instrumentation/opencensus-proto
+# gazelle:repository go_repository name=com_github_charlievieth_fastwalk importpath=github.com/charlievieth/fastwalk
 # gazelle:repository go_repository name=com_github_chzyer_logex importpath=github.com/chzyer/logex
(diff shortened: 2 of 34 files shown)


dzbarsky:
@linzhp please take a look?

uhthomas:
Why not simple use [filepath.WalkDir](https://pkg.go.dev/path/filepath#WalkDir) instead? Doesn't need a new dependency and achieves the same goal of being more efficient by not calling `os.Lstat` for each file or directory.

DavidZbarsky-at:
> Why not simply use [filepath.WalkDir](https://pkg.go.dev/path/filepath#WalkDir) instead? Doesn't need a new dependency and achieves the same goal of being more efficient by not calling `os.Lstat` for each file or directory.

Because it doesn't parallelize the walk. This is faster 

See also https://github.com/bazelbuild/bazel-gazelle/pull/1737

uhthomas:
>Because it doesn't parallelize the walk. This is faster

I understand that, though I imagine a lot of the time savings will be from not calling `os.Lstat`. Do you have some figures to demonstrate the difference parallelisation makes here? Such, if `WalkDir` provides a significant improvement over `Walk`, but the difference between `WalkDir` and `fastwalk` is not as significant, then the maintenance burden of an external package may not be worth it.

DavidZbarsky-at:
> > Because it doesn't parallelize the walk. This is faster
> 
> I understand that, though I imagine a lot of the time savings will be from not calling `os.Lstat`. Do you have some figures to demonstrate the difference parallelisation makes here? Such, if `WalkDir` provides a significant improvement over `Walk`, but the difference between `WalkDir` and `fastwalk` is not as significant, then the maintenance burden of an external package may not be worth it.

I don't have exact numbers offhand anymore; we wrote these patches months ago. FWIW `WalkDir` on its own is not trivial to use due to the complex iteration; that's part of the purpose of the Trie. Also in our testing the parallelization definitely made a significant different; between that and the Trie we reduced the runtime from 10s to 4.5s.

We do run gazelle as part of pre-commit hook so we are fairly perf-sensitive on this. I recognize that perhaps performance is not a primary motivation for the project; feel free to close out the PR if you aren't interested.

(BTW are you sure the os.Lstat is still an issue? I believe os.Readdir is already avoiding it since Go 1.16, perhaps you were thinking of ioutil.Readdir? I did measure some speedups when I contributed https://github.com/bazelbuild/bazel-gazelle/pull/1704/files)

linzhp:
I agreed with @uhthomas that we should avoid another external library before comparing it with filepath.WalkDir.

PR #1822, opened 2024-06-05
Support label using regexp in directive `gazelle:resolve_regexp`

### LINKED ISSUE #1821: gazelle:resolve_regexp should support regex in label

0.33.0

0.41.0

7.0.1
Yes

macOS Sonoma 14.0, arm64

### Issue

Context: We maintain a large golang repository with gazelle, and we would like gazelle to manage resolution for importpaths that point to targets in an external repository that we have. Although the `gazelle:repository` directive exists, since it operates heuristically, the importpath doesn't match the expected target name that gazelle attempts to resolve it to. 

We would like to use the `gazelle:resolve_regexp` to solve this. `gazelle:resolve_regexp` has an argument for the import-string-regexp and the label it matches to. We'd like to support the label also being a regexp, and populate that regexp with matches from the import-string-regexp. 

I have a change that would do this, which I will PR after creating this issue, but would like some feedback on the possibility of supporting this in bazel-gazelle, as folks on our team have expressed that this seems intuitive change to the existing directive and would allow us the possibility to use this for many upcoming cases where we need to manipulate gazelle resolves.

lkassar-stripe:
**What type of PR is this?**

Feature

resolve/config.go

Supports regexp in the resolution label for directive `gazelle:resolve_regexp`, to match pieces of the import statement and use them to populate the path where gazelle resolves a target to. 

Fixes #1821

**Other notes for review**


tyler-french:
@lkassar-stripe Could you please add some tests with example use cases?

Also, if there's associated documentation for `gazelle:resolve_regexp` we should probably update that too.

lkassar-stripe:
@tyler-french 
Hello! Apologies for the delay here. This should now be ready for review.

ouguoc2-stripe:
@tyler-french gentle ping here. :) Any chance we can get your thoughts?

tyler-french (approved):

PR #1828, opened 2024-06-19
fix: support leading ./ in .bazelignore

jbedard:
**What type of PR is this?**
Bug fix

all

Bazel supports leading ./ in .bazelignore, so should gazelle

Fixes #

**Other notes for review**

### DIFF
--- a/walk/config.go
+++ b/walk/config.go
@@ -135,8 +135,12 @@ func (c *Configurer) loadBazelIgnore(repoRoot string, wc *walkConfig) error {
 			log.Printf("the .bazelignore exclusion pattern must not be a glob %s", ignore)
 			continue
 		}
-		// Ensure we remove trailing slashes or the exclude matching won't work correctly
-		wc.excludes = append(wc.excludes, strings.TrimSuffix(ignore, "/"))
+
+		// Clean the path to remove any extra '.', './' etc otherwise
+		// the exclude matching won't work correctly.
+		ignore = path.Clean(ignore)
+
+		wc.excludes = append(wc.excludes, ignore)
 	}
 	return nil
 }
--- a/walk/walk_test.go
+++ b/walk/walk_test.go
@@ -205,6 +205,10 @@ foo/*
 
 # Random comment followed by a line
 a.file
+
+# Paths can have a ./ prefix
+./b.file
+././blah/../ugly/c.file
 `,
 		},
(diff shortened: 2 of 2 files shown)


walk/config.go
```diff
@@ -135,8 +135,12 @@ func (c *Configurer) loadBazelIgnore(repoRoot string, wc *walkConfig) error {
 			log.Printf("the .bazelignore exclusion pattern must not be a glob %s", ignore)
 			continue
 		}
-		// Ensure we remove trailing slashes or the exclude matching won't work correctly
-		wc.excludes = append(wc.excludes, strings.TrimSuffix(ignore, "/"))
+
+		// Ensure we remove leading "./" and trailing "/" or the exclude matching won't work correctly
+		ignore = strings.TrimPrefix(ignore, "./")
```
fmeum, line 140: Should we also strip this prefix if it is repeated?

jbedard: I think that's unlikely enough it's not worth the extra code 🤷

jbedard: Actually I'm quite wrong there, using `filepath.Clean` is simpler and I think more "correct" then the use of `TrimPrefix` we had before.

WDYT? See the latest and new test case with the ugly path in it.

walk/config.go
```diff
@@ -24,6 +24,7 @@ import (
 	"log"
 	"os"
 	"path"
+	"path/filepath"
```
fmeum, line 27:
```suggestion
```

walk/config.go
```diff
@@ -135,8 +136,12 @@ func (c *Configurer) loadBazelIgnore(repoRoot string, wc *walkConfig) error {
 			log.Printf("the .bazelignore exclusion pattern must not be a glob %s", ignore)
 			continue
 		}
-		// Ensure we remove trailing slashes or the exclude matching won't work correctly
-		wc.excludes = append(wc.excludes, strings.TrimSuffix(ignore, "/"))
+
+		// Clean the path to remove any extra '.', './' etc otherwise
+		// the exclude matching won't work correctly.
+		ignore = filepath.Clean(ignore)
```
fmeum, line 142:
```suggestion
		ignore = path.Clean(ignore)
```
for consistent behavior across platforms.

fmeum (approved):

fmeum (approved):

PR #1833, opened 2024-06-21
Restore compatibility with Go 1.18

### LINKED ISSUE #1826: `rules_go` should work with Go versions older than 1.20

0.37.0

0.48.0

7.2.0

Yes

macOS

### What did you do?

Attempted to use an older version of Go (1.18) to build a project

Everything should Just Work

### What did you see instead?

`go_mod_download.go` uses `errors.Join` which was introduced in Go 1.20. Because of this, it could not be compiled, which causes runs of gazelle to fail.

fmeum:
**What type of PR is this?**

Bug fix

cmd/gazelle

Also test compatibility of fetch_repo and gazelle.

Fixes #1826

**Other notes for review**


fmeum:
FYI @tyler-french

PR #1835, opened 2024-06-24
Remove reliance on specific canonical repo name scheme

fmeum:
**What type of PR is this?**

Bug fix

go_deps

Instead of relying on a particular separator char such as `~`, we instead only rely on the guaranteed fact that the apparent name of an extension repo is the last component of its canonical name.

Prepares for https://github.com/bazelbuild/bazel/issues/22865

**Other notes for review**


fmeum:
Going to merge this now for Bazel presubmit.

PR #1837, opened 2024-07-03
update-repos use pruned graph

sluongng:


sluongng:
@fmeum this is failing with 
```
(19:31:14) ERROR: error loading package '@@io_bazel_rules_go//proto': at /var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/ec321eb2cc2d0f8f91b676b6d4c66c29/external/io_bazel_rules_go/proto/compiler.bzl:20:5: cannot load '@@rules_proto//proto:proto_common.bzl': no such file
```

which seems to be a rules_go issue? `rules_proto` needs to be part of `go_rules_dependencies`? 🤔

fmeum:
rules_go only depends on rules_proto if you use proto functionality, which is why it was decided at some point to not provide this dep implicitly. You may have to update to 6.0.0 manually.

sluongng:
I think that's a bit strange to make a non-dev dep "optional".
The deps are usually declared with `maybe` and will only be lazily downloaded when used.

So I don't see a reason why rules_proto needed to be excluded here.

tyler-french (changes requested):
@sluongng Could we separate out all of the dependency changes to a different PR which we merge first?

This is a bit difficult to review, and might cause unexpected downstream behavior.

I would also like to avoid making any upgrades in `go.mod` file that aren't absolutely necessary.

tyler-french:
> @sluongng Could we separate out all of the dependency changes to a different PR which we merge first?
> 
> This is a bit difficult to review, and might cause unexpected downstream behavior.
> 
> I would also like to avoid making any upgrades in `go.mod` file that aren't absolutely necessary.

Oh, I just saw https://github.com/bazelbuild/bazel-gazelle/pull/1838, let me take a look there. It seems like there's some CI issues

sluongng:
Yeah I have separated it to https://github.com/bazelbuild/bazel-gazelle/pull/1838. Let's discuss it there.

PR #1841, opened 2024-07-16
[Gazelle] Fix Duplicate Load Bug

ckilian867:
If two extensions both import Kinds from the same file, and no load statement for those Kinds is already in the BUILD.bazel file, then Gazelle will add the load statement twice.

In this commit, we fix the issue by updating the tracked loads when a new load is inserted into the index. This will cause future iterations to update the existing load instead of creating a duplicate.

**What type of PR is this?**

>
Bug fix

cmd/gazelle

This PR fixes a bug where is two extensions each load a statement from the same file, and the file has no load statements at all, duplicate load statements will be added to the top of the file.

None, its a small bug fix and the above directions say one isn't needed.

**Other notes for review**

None

### DIFF
--- a/merger/fix.go
+++ b/merger/fix.go
@@ -123,6 +123,7 @@ func FixLoads(f *rule.File, knownLoads []rule.LoadInfo) {
 			if load != nil {
 				index := newLoadIndex(f, known.After)
 				load.Insert(f, index)
+				loads = append(loads, load)
 			}
 		}
 	}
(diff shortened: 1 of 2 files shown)


tyler-french (approved):

fmeum:
@ckilian867 The CI failures is a flake, but you still need to sign the CLA.

ckilian867:
> @ckilian867 The CI failures is a flake, but you still need to sign the CLA.

I'm working with our OSS team to get this resolved

fmeum:
Thanks, just ping me when it's done.

ckilian867:
@fmeum Check is passing now

Issue #1843, opened 2024-07-19
Feature Request: Record Service and Message names in Proto's fileInfo


ckilian867:
Currently, while the proto extension parses the proto file, it [uses regex](https://github.com/bazelbuild/bazel-gazelle/blob/030dba3acbd63cc2d16292c4f1dc06e66b7c1b1a/language/proto/fileinfo.go#L98) to find certain information about the file, including package, options, and imports. 

One of the things being parsed is [services](https://github.com/bazelbuild/bazel-gazelle/blob/030dba3acbd63cc2d16292c4f1dc06e66b7c1b1a/language/proto/fileinfo.go#L98). However, this is currently just to track a boolean on [whether there are any services](https://github.com/bazelbuild/bazel-gazelle/blob/030dba3acbd63cc2d16292c4f1dc06e66b7c1b1a/language/proto/fileinfo.go#L38) in the current file. 

For determining Java imports, we'd need to know the Java package option ([already available](https://github.com/bazelbuild/bazel-gazelle/blob/030dba3acbd63cc2d16292c4f1dc06e66b7c1b1a/language/proto/fileinfo.go#L35)) as well as the names of the messages and services being generated. 

This feature request would be to add two new fields to [FileInfo](https://github.com/bazelbuild/bazel-gazelle/blob/030dba3acbd63cc2d16292c4f1dc06e66b7c1b1a/language/proto/fileinfo.go#L35): `ServiceNames` and `MessageNames` (or something similar). In addition to the current information parsed from the proto file, we'd record this information and return it to the caller for later use. 

We'd be happy to implement this and raise a PR, but wanted to understand any concerns with this approach before spending cycles on implementation.

PR #1844, opened 2024-07-25
[Proto] Keep track of the names of Services, Messages, and Enums

ckilian867:
In order to construct a fully qualified Java import from the proto file, we need to know the values of the "java_package" option as well as the services, messages, and enums in the proto file. The "java_pacakge" option is already recorded. In this commit, we add support for parsing the names of the services, message, and enums of a proto file and tracking them in the FileInfo struct.

**What type of PR is this?**
Feature

language/proto

This feature tracks additional information that is parsed from the proto file and tracked in the FileInfo object. This includes the names of all Services, Messages, and Enums in the proto file. This will be helpful for constructing the imports for fully qualified imports for the Java language. 

https://github.com/bazelbuild/bazel-gazelle/issues/1843

**Other notes for review**

### DIFF
--- a/language/proto/fileinfo.go
+++ b/language/proto/fileinfo.go
@@ -36,6 +36,10 @@ type FileInfo struct {
 	Imports []string
 
 	HasServices bool
+
+	Services []string
+	Messages []string
+	Enums []string
 }
 
 // Option represents a top-level option statement in a .proto file. Only
@@ -77,6 +81,32 @@ func protoFileInfo(dir, name string) FileInfo {
 		case match[serviceSubexpIndex] != nil:
 			info.HasServices = true
 
+			// match is of the format "service ServiceName {".
+			// extract just the service name
+			fullMatch := string(match[serviceSubexpIndex])
+			if serviceName, ok := extractObjectName(fullMatch); ok {
+				info.Services = append(info.Services, serviceName)
+			}
+
+		case match[messageSubexpIndex] != nil:
+			// match is of the format "message MessageName {".
+			// extract just the message name
+			fullMatch := string(match[messageSubexpIndex])
+			if messageName, ok := extractObjectName(fullMatch); ok {
+				info.Messages = append(info.Messages, messageName)
+			}
+
+
+
+		case match[enumSubexpIndex] != nil:
+			// match is of the format "enum EnumName {".
+			// extract just the enum name
+			fullMatch := string(match[enumSubexpIndex])
+			if enumName, ok := extractObjectName(fullMatch); ok {
+				info.Enums = append(info.Enums, enumName)
+			}
+
+
 		default:
 			// Comment matched. Nothing to extract.
 		}
@@ -92,6 +122,8 @@ const (
 	optkeySubexpIndex  = 3
 	optvalSubexpIndex  = 4
 	serviceSubexpIndex = 5
+	messageSubexpIndex = 6
+	enumSubexpIndex = 7
 )
 
 // Based on https://developers.google.com/protocol-buffers/docs/reference/proto3-spec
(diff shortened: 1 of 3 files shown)


language/proto/fileinfo.go
```diff
@@ -77,6 +81,26 @@ func protoFileInfo(dir, name string) FileInfo {
 		case match[serviceSubexpIndex] != nil:
 			info.HasServices = true
 
+			// match is of the format "service ServiceName {".
+			// extract just the service name
+			fullMatch := string(match[serviceSubexpIndex])
+			serviceName := strings.Fields(fullMatch)[1]
```
tyler-french, line 87: This will `panic` if (for some reason) there's no space. Also, is the trailing space required?

I think this is valid:
```
service ServiceName{
```

Because of. this complexity, I think it would make sense to add this `extractObjectName()` to a function:

The function could then extract based on the situation:
```
func extractObject(fullMatch) {
  fields := fullMatch.Fields
  if len(fields) < 2 {
    // handle the error
  }
  return strings.TrimSuffix(fields[1], "{"
}
```

language/proto/fileinfo_test.go
```diff
@@ -109,13 +111,52 @@ import "second.proto";`,
...
 			want: FileInfo{
 				HasServices: false,
+				Messages: []string{"ServiceAccount"},
+			},
+		},{
+			desc: "multiple service names",
+			name:  "service.proto",
+			proto: `service ServiceA { string service = 1; }
+
+			service ServiceB { string service = 1; }
```
tyler-french, line 129: Can we add this test case:
```
service ServiceB{ string service = 1; }
```

language/proto/fileinfo.go
ckilian867: That's a good call. The way the service regex is currently written, the space on either side is optional. This means that `service ABC{` would get picked up as a service, but also `message serviceABC {`  would also get picked up (tested this out and it did).

I think we should update all the regexes to require a space between service/message/enum and then trim the suffix of any potential leftover "{" like in your comment

tyler-french (approved):

language/proto/fileinfo.go
ckilian867: Updated in newest revision

language/proto/fileinfo_test.go
ckilian867: Updated in newest revision

language/proto/fileinfo.go
```diff
@@ -77,6 +81,36 @@ func protoFileInfo(dir, name string) FileInfo {
...
 			info.HasServices = true
 
+			// match is of the format "service ServiceName {".
+			// extract just the service name
+			fullMatch := string(match[serviceSubexpIndex])
+			serviceName, ok := extractObjectName(fullMatch)
+
+			if ok {
+				info.Services = append(info.Services, serviceName)
+			}
```
tyler-french, line 91:
```suggestion
			if serviceName, ok := extractObjectName(fullMatch); if ok {
				info.Services = append(info.Services, serviceName)
			}
```

tyler-french (approved):

language/proto/fileinfo.go
```diff
@@ -77,6 +81,32 @@ func protoFileInfo(dir, name string) FileInfo {
 		case match[serviceSubexpIndex] != nil:
 			info.HasServices = true
 
+			// match is of the format "service ServiceName {".
+			// extract just the service name
+			fullMatch := string(match[serviceSubexpIndex])
+			serviceName, ok := extractObjectName(fullMatch); if ok {
+				info.Services = append(info.Services, serviceName)
+			}
```
tyler-french, line 89:
```suggestion
			if serviceName, ok := extractObjectName(fullMatch); ok {
				info.Services = append(info.Services, serviceName)
			}
```

language/proto/fileinfo.go
```diff
@@ -77,6 +81,32 @@ func protoFileInfo(dir, name string) FileInfo {
...
+
+		case match[messageSubexpIndex] != nil:
+			// match is of the format "message MessageName {".
+			// extract just the message name
+			fullMatch := string(match[messageSubexpIndex])
+			messageName, ok := extractObjectName(fullMatch); if ok {
```
tyler-french, line 95:
```suggestion
			if messageName, ok := extractObjectName(fullMatch); ok {
```

language/proto/fileinfo.go
```diff
@@ -77,6 +81,32 @@ func protoFileInfo(dir, name string) FileInfo {
...
+				info.Messages = append(info.Messages, messageName)
+			}
+
+
+
+		case match[enumSubexpIndex] != nil:
+			// match is of the format "enum EnumName {".
+			// extract just the enum name
+			fullMatch := string(match[enumSubexpIndex])
+			enumName, ok := extractObjectName(fullMatch); if ok {
```
tyler-french, line 105:
```suggestion
			if enumName, ok := extractObjectName(fullMatch); ok {
```

PR #1845, opened 2024-07-25
[Proto] Require space between 'service' and service name in regex matching

ckilian867:
Prior to this commit, the regex for determining whether a file had services did not require a space between the word "service" and the service name. This could lead to false alarms, like "message serviceABC". In this commit, we update the regex to only match if there is at least one space between service and the service name.

**What type of PR is this?**

Bug fix

language/proto

Fixes a bug where the HasServices regex could incorrectly match messages that have "service" at the beginning of its name.

None, small bug fix

**Other notes for review**

### DIFF
--- a/language/proto/fileinfo_test.go
+++ b/language/proto/fileinfo_test.go
@@ -110,10 +110,35 @@ import "second.proto";`,
 			want: FileInfo{
 				HasServices: true,
 			},
-		}, {
+		},
+		{
+			desc:  "service multiple spaces",
+			name:  "service.proto",
+			proto: `service      ChatService   {}`,
+			want: FileInfo{
+				HasServices: true,
+			},
+		},
+		{
+			desc:  "service no space for bracket after service name",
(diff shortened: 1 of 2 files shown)


ckilian867 (commented):
Note: this test fails after this change without the corresponding source code change

language/proto/fileinfo_test.go
```diff
@@ -110,10 +110,35 @@ import "second.proto";`,
...
+			proto: `serviceChatService {}`,
+			want: FileInfo{
+				HasServices: false,
+			},
+		},
+		{
 			desc:  "service as name",
 			name:  "service.proto",
-			proto: `message ServiceAccount { string service = 1; }`,
+			proto: `message serviceAccount { string service = 1; }`,
```
ckilian867, line 141: Note: this test fails after this change without the corresponding source code change

tyler-french (approved):

PR #1849, opened 2024-07-30
prepare release 0.38

tyler-french:
This PR prepares a new minor release of Gazelle. Many dependencies are not upgraded to prevent backward incompatibility.


MODULE.bazel
```diff
@@ -1,13 +1,13 @@
 module(
     name = "gazelle",
-    version = "0.37.0",
+    version = "0.38.0",
     repo_name = "bazel_gazelle",
 )
 
-bazel_dep(name = "bazel_features", version = "1.9.1")
-bazel_dep(name = "bazel_skylib", version = "1.5.0")
+bazel_dep(name = "bazel_features", version = "1.11.0")
```
fmeum, line 7: We shouldn't bump `bazel_dep` versions unless we actually need features from the new versions (keep the rules_go bump to get rid of the override)

fmeum (approved):

MODULE.bazel
```diff
@@ -58,7 +58,7 @@ bazel_dep(name = "bazel_skylib_gazelle_plugin", version = "1.4.1", dev_dependenc
 bazel_dep(name = "stardoc", version = "0.6.2", dev_dependency = True, repo_name = "io_bazel_stardoc")
 
 go_sdk_dev = use_extension("@io_bazel_rules_go//go:extensions.bzl", "go_sdk", dev_dependency = True)
-go_sdk_dev.download(version = "1.22.4")
+go_sdk_dev.download(version = "1.22.5")
```
tyler-french, line 61: @fmeum This one is fine though right? because we want to verify as large a _range_ of versions as possible with the dev dependency.

fmeum: Yes, that's a good idea.

fmeum (approved):

PR #1863, opened 2024-08-13
Preserve canonical labels as such

### LINKED ISSUE #1862: Can't resolve root module from external module
v0.38.0

v0.47.1
7.1.1

Yes

Linux x86

### What did you do?
Add a `gazelle:resolve` to resolve a name to a target in the local root module like so:
```
go_deps.gazelle_override(
  path = "cel.dev/expr",
  build_file_generation = "on",
  directives = ["gazelle:resolve proto google/rpc/status.proto @@//:googleapis_rpc_status"]
)
```
I expected this to successfully resolve, as `@@//...` should be visible to external modules.

### What did you see instead?
It did not resolve. Inspecting the generated BUILD.bazel file shows that the double `@@` was replaced with a single `@`. 

Generated BUILD.bazel:
```
proto_library(
    name = "expr_proto",
    srcs = [...],
    deps = [
        ...,
        "@//:googleapis_rpc_status"
    ],
)
```

fmeum:
**What type of PR is this?**

Bug fix

label

Previously, parsing and then stringifying a label would turn `@@canonical~name//pkg:target` into `@canonical~name//pkg:target`, which is no longer valid.

The new field on `Label` defaults to `false` if not set, which preserves the current behavior.

Fixes #1862

**Other notes for review**

### DIFF
--- a/label/label.go
+++ b/label/label.go
@@ -51,6 +51,12 @@ type Label struct {
 	// Relative indicates whether the label refers to a target in the current
 	// package. Relative is true if and only if Repo and Pkg are both omitted.
 	Relative bool
+
+	// Canonical indicates whether the repository name is canonical. If true,
+	// then the label will be stringified with an extra "@" prefix if it is
+	// absolute.
+	// Note: Label does not apply any kind of repo mapping.
+	Canonical bool
 }
 
 // New constructs a new label from components.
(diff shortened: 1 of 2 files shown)


tyler-french (approved):

label/label.go
```diff
@@ -83,10 +89,11 @@ func Parse(s string) (Label, error) {
 	origStr := s
 
 	relative := true
+	canonical := false
 	var repo string
-	// if target name begins @@ drop the first @
 	if strings.HasPrefix(s, "@@") {
 		s = s[len("@"):]
+		canonical = true
```
tyler-french, line 96: optional:
```
canonical := strings.HasPrefix(s, "@@")
if canonical {
		s = s[len("@"):]
}
```

PR #1866, opened 2024-08-14
Add default directives for github.com/containerd

mering:
**What type of PR is this?**

Other

internal/bzlmod

Disable github.com/containerd proto generation in default directives as [suggested in `rules_go` documentation](https://github.com/bazelbuild/rules_go/blob/master/docs/go/core/bzlmod.md#gazelle-directives).

**Other notes for review**


fmeum (approved):

mering:
@fmeum Thanks for the quick merge. Is it possible to create a bugfix release including this change such that we can use it soon?

PR #1868, opened 2024-08-16
refactor: compute transitive embeds during indexing

jbedard:
**What type of PR is this?**
refactor

all

Refactor working toward serializing+caching of the rule index.

The `ruleRecord` should be as close as possible to what a language returns. All computed data such as that for indexing should be computed from the `ruleRecord` and persisted elsewhere.

Ref: #1181
Some ideas are similar to #1182 while trying to reduce the scope of the change and diff.

**Other notes for review**

IMO this is a reasonable cleanup on its own, without even taking serializing+caching into account 🤷

### DIFF
--- a/resolve/index.go
+++ b/resolve/index.go
@@ -78,6 +78,20 @@ type RuleIndex struct {
 	importMap      map[ImportSpec][]*ruleRecord
 	mrslv          func(r *rule.Rule, pkgRel string) Resolver
 	crossResolvers []CrossResolver
+
+	// embedded indicates whether another rule of the same language embeds this
+	// the rule. Embedded rules should not be indexed.
+	embedded map[label.Label]struct{}
+
+	// embeds is the transitive closure of labels for rules that each rule embeds
+	// (as determined by the Embeds method). This only includes rules in the same
+	// language (i.e., it includes a go_library embedding a go_proto_library, but
+	// not a go_proto_library embedding a proto_library).
+	embeds map[label.Label][]label.Label
+
+	// imports is the transitive closure of all labels for rules that each rule embeds
+	// (as determined by the Embeds method). This includes rules in other languages.
+	imports map[label.Label][]ImportSpec
 }
 
 // ruleRecord contains information about a rule relevant to import indexing.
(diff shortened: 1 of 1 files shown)


resolve/index.go
```diff
@@ -169,30 +182,32 @@ func (ix *RuleIndex) AddRule(c *config.Config, r *rule.Rule, f *rule.File) {
...
-		ix.collectEmbeds(er)
+		ix.collectEmbeds(er, didCollectEmbeds)
 		erResolver := ix.mrslv(er.rule, er.file.Pkg)
 		if resolver.Name() == erResolver.Name() {
-			er.embedded = true
-			r.embeds = append(r.embeds, er.embeds...)
+			ix.embedded[er.label] = true
+			ix.embeds[r.label] = append(ix.embeds[r.label], ix.embeds[er.label]...)
 		}
 		r.importedAs = append(r.importedAs, er.importedAs...)
```
jbedard, line 212: This one is next, if I can't easily+quickly do it in this PR then a followup...

resolve/index.go
```diff
@@ -78,6 +78,20 @@ type RuleIndex struct {
 	importMap      map[ImportSpec][]*ruleRecord
 	mrslv          func(r *rule.Rule, pkgRel string) Resolver
 	crossResolvers []CrossResolver
+
+	// embedded indicates whether another rule of the same language embeds this
+	// the rule. Embedded rules should not be indexed.
+	embedded map[label.Label]bool
```
fmeum, line 84: Could this be a `map[label.Label]struct{}` to save memory?

resolve/index.go
```diff
@@ -91,18 +105,10 @@ type ruleRecord struct {
 	// Used to build a map from ImportSpecs to ruleRecords.
 	importedAs []ImportSpec
 
-	// embeds is the transitive closure of labels for rules that this rule embeds
-	// (as determined by the Embeds method). This only includes rules in the same
-	// language (i.e., it includes a go_library embedding a go_proto_library, but
-	// not a go_proto_library embedding a proto_library).
+	// Embeds is the set of labels (of any language) that this rule directly Embeds,
```
fmeum, line 108:
```suggestion
	// Embeds is the set of labels (of any language) that this rule directly embeds,
```

fmeum, line 108: Is it true that this covers all languages? The logic below looks this would only contain labels for the same language.

resolve/index.go
jbedard: You mean combining the 3 maps? I think that should work 🤔

fmeum: I just meant that you don't use the tri-state of `bool` + key existence here, so empty structs as values would save memory. But if you can combine the maps, that's of course even better 😅

jbedard: Interesting, TIL!

I'll switch to `struct{}` but leaving it as 3 maps for now. Maybe that could be a followup 🤔

resolve/index.go
jbedard: This field is whatever `Resolver.Embeds` returns (normalized a bit so the labels are always absolute).

We really have no idea what the label is, what `ImportSpec`s that label will produce or what language those `ImportSpec`s will be.

I find a lot of this terminology confusing so let me know if you can think of better phrasing for any of these comments while we're editing them.

jbedard:
I've updated the comments on a bunch of the internal struct fields, see the last commit.

fmeum (approved):

PR #1870, opened 2024-08-16
Support rules map_kind in go and proto languages

lbcjbb:
**What type of PR is this?**

`Bug fix`

- `language/go`
- `language/proto`

This PR takes into account the rule mapping defined with the `# gazelle:map_kind` instructions for all Go and Proto rules.


fmeum (approved):
Could you add a test case?

lbcjbb:
I've added some tests, but I needed to move the `map_kind` logic from `//cmd/gazelle` to a new shareable internal package (9e4702a81d45d9c302a5e8c424754bd21a5923d0), to be used in go and proto languages tests (8a8d69a5bf365fa0e104d6ae2d51cbe340cd0822).

PR #1872, opened 2024-08-19
refactor: compute labelMap apon indexing

jbedard:
**What type of PR is this?**
Other

all

Separate the real state vs indexing of that state, working toward the ability to load/save the state and re-indexing from that.


resolve/index.go
```diff
@@ -168,12 +178,7 @@ func (ix *RuleIndex) AddRule(c *config.Config, r *rule.Rule, f *rule.File) {
 		embeds:     embeds,
 		lang:       lang,
 	}
-	if _, ok := ix.labelMap[record.label]; ok {
```
jbedard, line 171: This is the one change in this PR... now duplicates will be persisted in the `ix.rules` and only resolved when indexing. I think this is rare enough the extra data being persisted is worth cleanly separating the "source of truth" and indexed data 🤷

fmeum (approved):

PR #1873, opened 2024-08-19
refactor: remove reference to rule.Rule in mapResolver interface

jbedard:
**What type of PR is this?**
Other

all

Ensuring all fields in `ruleRecord` are primitive or easily serializable.

### DIFF
--- a/resolve/index.go
+++ b/resolve/index.go
@@ -73,7 +73,7 @@ type CrossResolver interface {
 // RuleIndex is a table of rules in a workspace, indexed by label and by
 // import path. Used by Resolver to map import paths to labels.
 type RuleIndex struct {
-	mrslv          func(r *rule.Rule, pkgRel string) Resolver
+	mrslv          func(ruleKind, pkgRel string) Resolver
 	crossResolvers []CrossResolver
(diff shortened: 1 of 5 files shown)


resolve/index.go
```diff
@@ -118,7 +118,7 @@ type ruleRecord struct {
 //
 // kindToResolver is a map from rule kinds (for example, "go_library") to
 // Resolvers that support those kinds.
-func NewRuleIndex(mrslv func(r *rule.Rule, pkgRel string) Resolver, exts ...interface{}) *RuleIndex {
+func NewRuleIndex(mrslv func(ruleKind, pkgRel string) Resolver, exts ...interface{}) *RuleIndex {
```
fmeum, line 121: This is technically public API, although I don't see how a language implementation would use it. Can we keep its signature unchanged somehow?

jbedard: I was assuming no tests failing meant it wasn't public API 😢

fmeum: I don't. Let's ship and find out.

fmeum (approved):

PR #1878, opened 2024-08-20
refactor: make ruleRecord serializable

jbedard:
Using `json.[Un]Marchal`  is one way of serializing the data and the easiest one to start with. This change just does the ugly diff renaming everything to work with those marshalling methods. Do we want to rename any of these fields while we have to change them anyway?

**What type of PR is this?**
Other

all

Make the `ruleRecord` serializable to json in preparation for caching it.

### DIFF
--- a/resolve/index.go
+++ b/resolve/index.go
@@ -110,22 +110,22 @@ type RuleIndex struct {
 
 // ruleRecord contains information about a rule relevant to import indexing.
 type ruleRecord struct {
-	kind  string
-	label label.Label
+	Kind  string      `json:"kind"`
+	Label label.Label `json:"label"`
 
-	pkg string
+	Pkg string `json:"pkg"`
 
 	// A list of ImportSpecs by which this rule may be imported.
-	importedAs []ImportSpec
+	ImportedAs []ImportSpec `json:"importedAs"`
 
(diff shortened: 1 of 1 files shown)


resolve/index.go
```diff
@@ -110,22 +110,22 @@ type RuleIndex struct {
 
 // ruleRecord contains information about a rule relevant to import indexing.
 type ruleRecord struct {
-	kind  string
-	label label.Label
+	Kind  string      `json:"kind"`
```
fmeum, line 113: Since you probably know this part of the code better than me by now: Is it correct that we don't expose any fields or methods that return a `ruleRecord`? Otherwise making the fields public could end up adding new public API.

jbedard: AFAIU `ruleRecord` is private to `RuleIndex` in this file. It is used to persist the data collected from languages and never directly exposed.

I think `FindRulesByImport` is a good example which searches the data (actually searches `importMap` which is computed from the `ruleRecord`s), then returns a completely different `FindResult` struct and never exposes `ruleRecord`s or `importMap`...

So `ruleRecord` is private, but it's fields are public so the json encoding works 🤷

fmeum: Sounds good, thanks for confirming!

PR #1879, opened 2024-08-20
Remove module extension repo name hack

fmeum:
**What type of PR is this?**

Cleanup

go_repository

This hack is not necessary as the name of the referenced repo is static and thus can be `use_repo`ed (in fact, it is already visible).


fmeum:
Doesn't work with isolated extensions.

PR #1880, opened 2024-08-20
feat: add option to persist rule index

jbedard:
**What type of PR is this?**
Feature

cmd/gazelle

Add the ability to persist the `RuleIndex` across runs to allow updating only the specified directories when indexing is required.

Ref (and possibly fixes?): #1181


cmd/gazelle/fix-update.go
```diff
@@ -311,15 +318,38 @@ func runFixUpdate(wd string, cmd command, args []string) (err error) {
...
-	walk.Walk(c, cexts, uc.dirs, uc.walkMode, func(dir, rel string, c *config.Config, update bool, f *rule.File, subdirs, regularFiles, genFiles []string) {
+	walk.Walk(c, cexts, uc.dirs, walkMode, func(dir, rel string, c *config.Config, update bool, f *rule.File, subdirs, regularFiles, genFiles []string) {
 		// If this file is ignored or if Gazelle was not asked to update this
 		// directory, just index the build file and move on.
 		if !update {
-			if c.IndexLibraries && f != nil {
-				for _, repl := range c.KindMap {
-					mrslv.MappedKind(rel, repl)
-				}
+			for _, repl := range c.KindMap {
```
jbedard, line 349: I think this should always have been outside the `if c.IndexLibraries` condition. Everywhere else `MappedKind(...)` is invoked is not conditional based on indexing.

walk/walk.go
```diff
@@ -115,7 +115,7 @@ func Walk(c *config.Config, cexts []config.Configurer, dirs []string, mode Mode,
 		}
 	}
 
-	updateRels := buildUpdateRelMap(c.RepoRoot, dirs)
+	updateRels := NewUpdateFilter(c.RepoRoot, dirs, mode)
```
jbedard, line 118: I could potentially pre-factor all this?

It's basically just wrapping all the util methods into a struct. That struct is only used internally but needs to be "public" (the capital `New`) because it's used in both `walk/` and `resolve/`.

jbedard: https://github.com/bazelbuild/bazel-gazelle/pull/1881 if we want to merge that first

cmd/gazelle/fix-update.go
jbedard: https://github.com/bazelbuild/bazel-gazelle/pull/1882

resolve/index.go
```diff
@@ -145,6 +148,66 @@ func NewRuleIndex(mrslv func(ruleKind, pkgRel string) Resolver, exts ...interfac
...
+	indexDbFile, err := os.Open(indexDbPath)
+	if err != nil {
+		if os.IsNotExist(err) {
+			return false, nil
+		}
+		return false, err
+	}
+	defer indexDbFile.Close()
+
+	indexDbContent, err := io.ReadAll(indexDbFile)
```
fmeum, line 165: How large does this file get? I'm wondering whether `json.NewDecoder` could be more appropriate here. Doesn't matter for the first iteration though.

resolve/index.go
```diff
@@ -145,6 +148,66 @@ func NewRuleIndex(mrslv func(ruleKind, pkgRel string) Resolver, exts ...interfac
...
+
+func (ix *RuleIndex) SaveIndex(indexDbPath string) error {
+	// TODO: write index version
+
+	indexDbContent, err := json.Marshal(ix.rules)
+	if err != nil {
+		return err
+	}
+
+	indexDbFile, err := os.OpenFile(indexDbPath, os.O_RDWR|os.O_CREATE, 0666)
```
fmeum, line 197: This should probably use `os.Create` instead so that the file is truncated if it exists.

resolve/index.go
jbedard: I assume this could get fairly large and I've been wondering if we'll want something more compact then json. But I figured if this is EXPERIMENTAL and we can break it whenever we want then plain json is a good place to start?

cmd/gazelle/fix-update.go
```diff
@@ -96,6 +99,7 @@ func (ucr *updateConfigurer) RegisterFlags(fs *flag.FlagSet, cmd string, c *conf
 	fs.StringVar(&ucr.memProfile, "memprofile", "", "write memory profile to `file`")
 	fs.Var(&gzflag.MultiFlag{Values: &ucr.knownImports}, "known_import", "import path for which external resolution is skipped (can specify multiple times)")
 	fs.StringVar(&ucr.repoConfigPath, "repo_config", "", "file where Gazelle should load repository configuration. Defaults to WORKSPACE.")
+	fs.StringVar(&uc.ruleIndexFile, "indexdb", "", "EXPERIMENTAL: file to cache the rule index")
```
jbedard, line 102: Suggestions for the naming of the cli arg? I used "indexdb" because this is caching the `RuleIndex` data 🤷 
Maybe we should use a `--experimental` prefix?

jbedard:
Just to re-iterate a few points and limitations of this:

This is EXPERIMENTAL with no api guarantees. 99% chance there will be breaking changes. You should never commit this file to source-control.

When this EXPERIMENTAL feature is used:
* the `resolver.Imports/Embeds()` calls for the entire workspace (the "rule index") is written to disk 
* when specifying `dirs` (default: `.`) on the command line only those directories are updated in the "rule index"
* only the `dirs` specified on the command line will be traversed, therefore:
  * only those `dirs` will have `Imports/Embeds()` invoked
  * only those `dirs` will have `Fix()`, `GenerateRules()` invoked
  * only those `dirs` will have BUILDs updated, warnings generated etc.
  * only those `dirs` will have things like resolve warnings outputted

If `resolver.Imports/Embeds()` have side-effects then this won't work very well when they are cached and not invoked. Don't do that.

There is no cache invalidation here. It is entirely up to the user to pass the correct list of `dirs` that require reprocessing.

FUTURE: if those updated `dirs` changed what rules they generate and the `Imports/Embeds()` of those rules change it is possible that other rules outside `dirs` are now referencing something incorrect. I think if `Imports/Embeds()` of those `dirs` change we might want to re-run `Fix` + `GenerateRules` for all directories.

jbedard:
@fmeum did you have any opinions on this going forward? Is this something you think should live in gazelle?

For large repos even simply walking the fs can be slow (even after the changes making it concurrent) so any type of incremental BUILD-generation needs to run on a subset of the workspace which is the goal of this PR.

fmeum:
I could very well see this live in Gazelle. Do you think that any of the two recent proposals https://github.com/bazel-contrib/bazel-gazelle/issues/1891 and https://github.com/bazel-contrib/bazel-gazelle/issues/1907 could substantially improve the situation for your large repos? If so, I would prefer to explore this route first, mostly since serialization, even if experimental, often ends up constraining future efforts to change the persisted data structure. Not doing unnecessary work in the first place could save us the complexity of having to persist this work.

jbedard:
Avoiding any type of caching/persisting would always be ideal if we can do it 👍 

For persisting/serialization... I would prefer just thinking of it as caching, it should never be required. If we just throw a hash at the start of the serialized data then we can ignore any cached data from an older version of gazelle or different list of languages etc (hash = gazelle version + every language compiled into the gazelle binary maybe), then we should never have to deal with older data formats at least.

jbedard:
I'll try to leave comments on #1891 and #1907. I worry any major improvements will require breaking changes to truly help though, maybe completely replacing traversal+indexing+resolving with something lazy by default?

I think a lot of scenarios might still need to know the fs tree in order to do things based on convention or lazy indexing though, but if we can then do the actual indexing+resolving in a lazy fashion that would be great.

PR #1881, opened 2024-08-20
refactor: move walk util methods into struct

jbedard:
**What type of PR is this?**
Other

all

Prefactor for #1880

### DIFF
--- a/walk/walk.go
+++ b/walk/walk.go
@@ -115,7 +115,7 @@ func Walk(c *config.Config, cexts []config.Configurer, dirs []string, mode Mode,
 		}
 	}
 
-	updateRels := buildUpdateRelMap(c.RepoRoot, dirs)
+	updateRels := NewUpdateFilter(c.RepoRoot, dirs, mode)
 
 	var visit func(*config.Config, string, string, bool)
 	visit = func(c *config.Config, dir, rel string, updateParent bool) {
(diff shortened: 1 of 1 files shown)


walk/walk.go
```diff
@@ -162,32 +162,38 @@ func Walk(c *config.Config, cexts []config.Configurer, dirs []string, mode Mode,
...
 			genFiles := findGenFiles(wc, f)
 			wf(dir, rel, c, update, f, subdirs, regularFiles, genFiles)
 		}
 	}
 	visit(c, c.RepoRoot, "", false)
 }
 
-// buildUpdateRelMap builds a table of prefixes, used to determine which
+// An UpdateFilter tracks which directories need to be updated
+type UpdateFilter struct {
```
fmeum, line 182: Since we are moving these types to a public package, could you add a comment marking them as internal? Alternatively, could we carve out a new `internal` subpackage and move them there?

jbedard: For now I can actually make this struct private which I've done in this PR.

But for https://github.com/bazelbuild/bazel-gazelle/pull/1880 this logic of which directories are updated is needed when deciding what data to load vs re-compute.

walk/walk.go
```diff
@@ -162,32 +162,38 @@ func Walk(c *config.Config, cexts []config.Configurer, dirs []string, mode Mode,
...
+type updateFilter struct {
+	mode Mode
+
+	// map from slash-separated paths relative to the
+	// root directory ("" for the root itself) to a boolean indicating whether
+	// the directory should be updated.
+	updateRels map[string]bool
+}
+
+// NewUpdateFilter builds a table of prefixes, used to determine which
```
fmeum, line 191: Could you mark this as internal via a comment?

jbedard: I've added an INTERNAL note.

walk/walk.go
jbedard: I've reverted back to using `UpdateFilter` (making it go-lang public) so it can be used elsewhere in the next PR. I've added the INTERNAL comment here as well.

fmeum (approved):

PR #1884, opened 2024-08-21
refactor: remove nested walk function literal

jbedard:
This function literal confuses me everytime I see it and imo is nicer as a plain function.

**What type of PR is this?**
Other

all

Minor refactoring/cleanup.

### DIFF
--- a/walk/walk.go
+++ b/walk/walk.go
@@ -122,72 +122,72 @@ func Walk(c *config.Config, cexts []config.Configurer, dirs []string, mode Mode,
 		log.Printf("error loading .bazelignore: %v", err)
 	}
 
-	var visit func(*config.Config, string, string, bool)
-	visit = func(c *config.Config, dir, rel string, updateParent bool) {
-		haveError := false
+	visit(c, cexts, isBazelIgnored, knownDirectives, updateRels, wf, c.RepoRoot, "", false)
+}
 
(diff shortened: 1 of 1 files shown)


jbedard:
Make sure you look at the diff [ignoring whitespace](https://github.com/bazelbuild/bazel-gazelle/pull/1884/files?w=1)

walk/walk.go
```diff
@@ -122,72 +122,72 @@ func Walk(c *config.Config, cexts []config.Configurer, dirs []string, mode Mode,
...
-		if err != nil {
-			log.Print(err)
-			if c.Strict {
-				// TODO(https://github.com/bazelbuild/bazel-gazelle/issues/1029):
-				// Refactor to accumulate and propagate errors to main.
-				log.Fatal("Exit as strict mode is on")
-			}
-			haveError = true
-		}
+func visit(c *config.Config, cexts []config.Configurer, isBazelIgnored isIgnoredFunc, knownDirectives map[string]bool, updateRels *UpdateFilter, wf WalkFunc, dir, rel string, updateParent bool) {
```
jbedard, line 128: The number of args here might be making it worth wrapping some of these fields in a struct?

fmeum: I actually prefer it this way, structs always make me suspicious of something else modifying it after the initial struct literal.

fmeum (approved):
Didn't know about `?w=1` on GitHub, this is super useful!

PR #1889, opened 2024-08-23
Adjust gazelle for upcoming rules_go changes

dzbarsky:
**What type of PR is this?**
Code refactor/future-proofing

gazelle binary

rules_go is deprecating some attributes from go_context (such as `package_list`), so stop using it.
I am also considering revamping how runfiles collection works in rules_go, so we can just collect the runfiles from `languages` directly to insulate ourselves from future changes there.

Fixes #

**Other notes for review**


fmeum (commented):
Can we make it so that this works with both the current and the upcoming version of rules_go? It wouldn't matter for Bzlmod, but WORKSPACE users will have an easier time updating if that's the case.

dzbarsky:
> Can we make it so that this works with both the current and the upcoming version of rules_go? It wouldn't matter for Bzlmod, but WORKSPACE users will have an easier time updating if that's the case.

The rules_go changes don't break gazelle, we can land this future-proofing after the release is cut

PR #1892, opened 2024-08-26
WIP: Add a lazy indexing mode to gazelle

### LINKED ISSUE #1891: Proposal: Lazy indexing for faster subdirectory updates
### Problem
Currently, gazelle operates in indexing or non-indexing mode. When indexing is enabled, gazelle will walk the entire repo, using the `Imports` method of language extensions to build up a rule index. For large repos, this can lead to a substantial slowdown compared to operating in a non-indexing mode when trying to run `gazelle` in a subdirectory (~10s -> 0.8s in my case).

The time penalty from indexing is so great that there are ongoing discussions and work to allow gazelle to save and load its index to the disk - https://github.com/bazelbuild/bazel-gazelle/issues/1181

Not all languages are able to operate without indexing due to ambiguities in mapping source code import statements to the bazel target labels that provide those imports. For example, it is not possible to determine python target labels from import statements. As an example, how can you map the following imports to bazel target labels without an index?

```
# All of these could be //common/math:foo
import common.math.foo.some_func
import common.math.foo.MyFooClass
import common.math.foo

# Is this //common/math:foo or //common:math?
from common.math import foo

# Is this 'bar function' from //common/math:foo or 'bar module' //common/math/foo:bar
from common.math.foo import bar
```

Even though you cannot explicitly determine the target labels for each of those imports, you _can_ reasonably guess which packages they would be in (`common`, `common/math`, `common/math/foo`). Because of common conventions for directory structure matching import path structure, we don't need to index the entire repository to resolve these imports. We only need to index the packages that we _think_ the targets would be in.

### Proposal

I would like to add a new `-lazy_index` flag to gazelle. When this flag is enabled (with `-index=true`), gazelle will not visit all directories in the repo. Instead, it will only visit the directories to update. The `GenerateResult` struct would gain a `PkgsToIndex` field that language extensions should populate. Once gazelle is finished generating rules for a given subdir, it would look at all of the `PkgsToIndex`, then index them prior to finalizing the ruleIndex and calling language extension `Resolve` methods.

For our above python code example, this leads to a result like,
```go
PkgsToIndex: []string{"", "common", "common/math", "common/math/foo", "common/math/foo/some_func", "common/math/foo/MyFooClass", "common/math/foo/bar"}
```

Notably, it is _okay_ for rules to suggest indexing packages that don't exist. We can have gazelle ignore directories that don't exist in the workspace when an extension asks to index them.

### Benchmarking

I hacked together an implementation of lazy indexing with rules_python. In a repository with ~25k directories, it reduced the runtime of running gazelle in a single subdirectory from `5-10s` to `~1s` (depending on the directory and how much indexing was avoidable).

The gazelle changes can be seen in https://github.com/bazelbuild/bazel-gazelle/pull/1892.
The rules_python changes can be seen in https://github.com/alex-torok/rules_python/pull/1

### Questions / Challenges

- Is this something that the community is interested in? Would language extension authors be willing to implement the additional behavior to support lazy indexing?
- Are there languages where lazy indexing would not work or does not make any sense?
  - I'm mostly familiar with golang and python and would like feedback from people familiar with other gazelle language extensions.
- Does the suggested interface change to add an additional `PkgsToIndex` field to the `GenerateResult` sound good, or should this be an entirely separate method?
  - I feel like it makes sense on `GenerateResult`, as the language extension should have all of the context it needs to make the recommendation to gazelle. Alternatively, there could be a new interface that extensions implement where we pass in the GenerateResult and have it return the packages to index.

This solution does pose problems for repos that don't have a package directory structure that matches the language import path structure. For example, `a.b.c` could come from `//my/repo/is/wierd`. In this case, people could write a language extension that has a nearly-no-op `Generate` method that returns a fixed set of nonconventional packages to index. Alternatively, there could be a `-lazy_index_include=some/path` to force additional packages (and their subdirs) to be indexed.

alex-torok:
WIP: Example implementation of lazy indexing (See #1891). Will update docs & add tests once discussion in above issue is further along.

**What type of PR is this?**

Feature

cmd/gazelle

Add lazy indexing mode. TODO: fill out more context

Fixes #1891

**Other notes for review**

Currently WIP - please keep discussion in above issue for now.

### DIFF
--- a/cmd/gazelle/fix-update.go
+++ b/cmd/gazelle/fix-update.go
@@ -136,7 +136,11 @@ func (ucr *updateConfigurer) CheckFlags(fs *flag.FlagSet, c *config.Config) erro
 		uc.dirs[i] = dir
 	}
 
-	if ucr.recursive && c.IndexLibraries {
+	if ucr.recursive && c.LazyIndex {
+		uc.walkMode = walk.UpdateSubdirsMode
+	} else if c.LazyIndex {
+		uc.walkMode = walk.UpdateDirsMode
+	} else if ucr.recursive && c.IndexLibraries {
 		uc.walkMode = walk.VisitAllUpdateSubdirsMode
 	} else if c.IndexLibraries {
 		uc.walkMode = walk.VisitAllUpdateDirsMode
@@ -240,6 +244,8 @@ type visitRecord struct {
 	// empty is a list of empty Go rules that may be deleted.
 	empty []*rule.Rule
 
+	toIndex map[string]struct{}
+
 	// file is the build file being processed.
 	file *rule.File
 
@@ -311,8 +317,11 @@ func runFixUpdate(wd string, cmd command, args []string) (err error) {
 		}
 	}()
 
+	visitedPkgs := make(map[string]struct{})
+
 	var errorsFromWalk []error
 	walk.Walk(c, cexts, uc.dirs, uc.walkMode, func(dir, rel string, c *config.Config, update bool, f *rule.File, subdirs, regularFiles, genFiles []string) {
+		visitedPkgs[rel] = struct{}{}
 		// If this file is ignored or if Gazelle was not asked to update this
 		// directory, just index the build file and move on.
 		if !update {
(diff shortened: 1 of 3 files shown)


jbedard:
How can we handle `PkgsToIndex` that are not subdirectories?

alex-torok:
> How can we handle `PkgsToIndex` that are not subdirectories?

The `PkgsToIndex` are expected to be fully specified package paths from the root of the workspace. Language extensions would have to convert relative imports into package paths based off of `GenerateArgs.Rel`.

PR #1897, opened 2024-08-28
revert #1873

tyler-french:
We need access to `rule.Name()` as well as `rule.Kind()` in mapped resolver.

### DIFF
--- a/resolve/index.go
+++ b/resolve/index.go
@@ -73,7 +73,7 @@ type CrossResolver interface {
 // RuleIndex is a table of rules in a workspace, indexed by label and by
 // import path. Used by Resolver to map import paths to labels.
 type RuleIndex struct {
-	mrslv          func(ruleKind, pkgRel string) Resolver
+	mrslv          func(r *rule.Rule, pkgRel string) Resolver
 	crossResolvers []CrossResolver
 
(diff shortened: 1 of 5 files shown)


fmeum:
@jbedard We need to roll this change back as it breaks the public resolver API, which has a legitimate need to access other rule fields. Do you see a way to reland the desired behavior without touching public API?

fmeum (approved):

resolve/index.go
```diff
@@ -110,6 +110,8 @@ type RuleIndex struct {
 
 // ruleRecord contains information about a rule relevant to import indexing.
 type ruleRecord struct {
+	rule  *rule.Rule
```
tyler-french, line 113: This is still serializable since https://pkg.go.dev/encoding/json#Marshal only acts on exported fields of a structure.

jbedard: The problem will be unmarshalling though, won't it?

jbedard:
I'll look at it again if we must revert it. 

Is this actually a publicly documented API though?

tyler-french:
> I'll look at it again if we must revert it.
> 
> Is this actually a publicly documented API though?

https://pkg.go.dev/github.com/bazelbuild/bazel-gazelle@v0.38.0/resolve#NewRuleIndex

PR #1900, opened 2024-08-28
WIP: create an interface to allow for convention checking in gazelle

tyler-french:
Currently, at Uber, we do not use an index. Instead, we use conventions and a set of `# gazelle:resolve` statements in the root BUILD.bazel file.

This is beneficial because it means we can run Gazelle `O(size change)` instead of `O(size repo)'


tyler-french:
@fmeum @linzhp still a draft, but let me know if you have any preliminary thoughts. Especially on how to make it more ubiquitous

tyler-french:
Will be resolved by Gazelle 2.0 interface.

PR #1906, opened 2024-08-29
Writing resolved Go modules as K:V JSON for external bookkeeping

### LINKED ISSUE #1905: go_deps should expose resolved versions for all Go modules it Bazelizes, for external bookkeeping
`v0.38.0`

`v0.49.0`

`7.1.2`

Yes

macOS Sonoma 14.5, arm64 (Apple M1 Max)

This isn't a bug but a feature request. I'd like the `go_deps` module extension to expose the mapping between every Go module it bazelizes and its resolved version. This would be extremely useful for external bookkeeping (for e.g. we can easily communicate to CI systems that rely on native package manager metadata, precisely what version of each Go module was used in the build).

I'll shortly create a sample PR demonstrating what we might need although I'm not claiming that it would the optimal mechanism to surface this data.

Buzz-Lightyear:
**What type of PR is this?**
Feature

go_deps extension

I'd like the go_deps module extension to expose the mapping between every Go module it bazelizes and its resolved version. This would be extremely useful for external bookkeeping (for e.g. we can easily communicate to CI systems that rely on native package manager metadata, precisely what version of each Go module was used in the build).

Fixes #1905


internal/bzlmod/go_deps.bzl
```diff
@@ -643,8 +644,12 @@ def _go_deps_impl(module_ctx):
 
             go_repository_args.update(repo_args)
 
+        if "version" in go_repository_args:
+            resolved_go_modules[go_repository_args["importpath"]] = go_repository_args["version"]
+
         go_repository(**go_repository_args)
 
+    module_ctx.file("./resolved_go_modules.json", content=json.encode_indent(resolved_go_modules))
```
fmeum, line 652: Files written by module extensions are emitted into an unspecified temporary directory that should be removed after execution. We would need to instantiate a repo with this info or accept an absolute path via an environment variable.

Buzz-Lightyear: Makes sense, I'm propagating this similar to config.json used by rules_go. I was able to verify that I can access this file as `@bazel_gazelle_go_repository_config//:resolved_deps.json`. Would this approach work?

Buzz-Lightyear: I can also expose this via a separate repo rule if that's preferable. On a related note, while I can pull this into my workspace as `@bazel_gazelle_go_repository_config//:resolved_deps.json`, `bazel mod tidy` still thinks this repo is an indirect dependency and removes it from `MODULE.bazel` (I'm explicitly passing it in as inputs to both a build rule and a `sh_binary` rule). Is that perhaps a bug? If not, can I add a directive to inform `bazel mod tidy` that I intend to keep that use_repo call?

Buzz-Lightyear: Unsure why Windows build of rules_go is failing but interestingly I see Bazel warn that the repo is an indirect dependency in Gazelle's own MODULE.bazel file:
```
(16:03:24) WARNING: C:/b/bk-windows-xmds/bazel/bazel-gazelle/tests/bcr/go_work/MODULE.bazel:27:24: The module extension go_deps defined in @gazelle//:extensions.bzl reported incorrect imports of repositories via use_repo():
--
  |  
  | Imported, but reported as indirect dependencies by the extension:
  | bazel_gazelle_go_repository_config, org_golang_google_protobuf
```

Buzz-Lightyear: Looks like the failure was intermittent and unrelated to the change

fmeum:
The implementation looks reasonable, I'm just not sure whether this is the right way to export this information and thus don't want to commit to it yet.

Could you perhaps create a discussion on this repo and use it to explain your use case in more detail? I would like to gather more feedback from the community so that we can make more informed design decisions.

Buzz-Lightyear:
Sounds good @fmeum , started https://github.com/bazelbuild/bazel-gazelle/discussions/1918

PR #1908, opened 2024-09-04
feat: add support for .gitignore

### LINKED ISSUE #1166: Exclude files/folders that are .gitignore'd
This is a major usability hurdle at one of Aspect's clients. Gazelle fails because they have some junk in their repo which is covered by a `.gitignore` so they don't realize they have an unclean state. The error messages when gazelle descends into these files is cryptic to them. They then ask a support question where it's not obvious that `git clean -ffdx` is the solution.

Gazelle should simply ignore things that are .gitignore'd (probably with a directive to opt-in/out)

jbedard:
**What type of PR is this?**
> Feature

> all

Allow users to opt-in to respecting the gitignore when collecting files for gazelle extensions to consume.

While bazel can access git-ignored files many repositories manually exclude git-ignored content by not including it in source file lists or globs in BUILDs or by duplicating the git-ignore entries as `# gazelle:exclude`.

Ideally users would put all bazel ignored content in `.bazelignore` but without glob support that is often not done.

This should remain opt-in for repositories wanting it but by default bazel can access git-ignored content so gazelle should as well.

Fixes #1166

**Other notes for review**


jbedard:
@fmeum can you confirm if adding this to the gazelle core is something we'd like to do? If so are there preferences for which library to use vs manual?

tyler-french:
How is this different from using 'gazelle:exclude' directives?

jbedard:
People won't want to copy+paste gitignore files into BUILD files...?

jbedard:
Functionally gitignore also works a bit different with things like `!` operators etc. That is one reason to use a library such as `go-git` instead of doing it ourselves but we could also potentially do it ourselves but don't support `!`...

fmeum:
I kind of like this as it reduces the amount of Gazelle-specific configuration.

I don't know which library is best, but we should definitely use one rather than roll our own.

jbedard:
> I don't know which library is best, but we should definitely use one rather than roll our own.

I agree with should use a real library and not try to re-implement gitignore logic, otherwise we'll just end up with a never-ending stream "it does work like git" bugs.

The `go-git` library is what I've used in all the aspect gazelle extensions and I've never had any issues so I think that's a good place to start. If we add appropriate tests then we should be able to change the implementation later.

jbedard:
FWIW I've also tried other libraries which all had issues in the aspect gazelle extensions:
* https://github.com/sabhiram/go-gitignore - too slow, don't recall if there were other bugs
* https://github.com/denormal/go-gitignore - had multiple bugs (https://github.com/aspect-build/aspect-cli/issues/657)

Those both also seem to be unmaintained.

Others I have not tried seem to also be unmaintained and many look too simple or slow:
* https://github.com/zabawaba99/go-gitignore
* https://github.com/src-d/go-git
* https://github.com/zabawaba99/go-gitignore

walk/config.go
```diff
@@ -151,6 +182,47 @@ func loadBazelIgnore(repoRoot string) (isIgnoredFunc, error) {
...
+	parent *gitignoreEntry
+}
+
+func createGitIgnorer(parent *gitignoreEntry, rel string, ignoreReader io.Reader) *gitignoreEntry {
+	domain := []string{}
+	if rel != "" {
+		domain = strings.Split(rel, "/")
+	}
+
+	matcherPatterns := make([]gitignore.Pattern, 0)
```
dzbarsky, line 197: nit: `var matcherPatterns []gitignore.Pattern` saves an alloc :)

walk/config.go
```diff
@@ -151,6 +182,47 @@ func loadBazelIgnore(repoRoot string) (isIgnoredFunc, error) {
...
 }
 
+type gitignoreEntry struct {
+	i      *gitignore.Matcher
+	base   string
+	parent *gitignoreEntry
+}
+
+func createGitIgnorer(parent *gitignoreEntry, rel string, ignoreReader io.Reader) *gitignoreEntry {
+	domain := []string{}
```
dzbarsky, line 192: `var domain []string`

walk/config.go
```diff
@@ -56,6 +61,17 @@ func (wc *walkConfig) isExcluded(rel, base string) bool {
 	return matchAnyGlob(wc.excludes, path.Join(rel, base))
 }
 
+func (wc *walkConfig) isGitIgnored(rel, base string) bool {
+	// .gitignore
+	if wc.gitignoreEnabled && wc.gitignore != nil {
+		if wc.gitignore.isIgnored(strings.Split(path.Join(rel, base), "/")) {
```
dzbarsky, line 67: is it possible to make this work by iterating segments instead of needing to materialize them with a `Split`?

jbedard: You mean maintaining an array as we recurse instead of having to do Split+Join over and over?

jbedard: I think I've got that going...

tests/BUILD.bazel
```diff
@@ -35,16 +35,40 @@ gazelle_binary(
...
+TEST_GEN_FILES_OUT = [[d.replace(".test-", ".") for d in data] for data in TEST_GEN_FILES_IN]
+
+# genrules renaming files with ".test-" prefix to "." prefix.
+[
+    [
+        genrule(
+            name = "rename_%s_%s" % (TESTS[i], j),
+            srcs = [TEST_GEN_FILES_IN[i][j]],
+            outs = [TEST_GEN_FILES_OUT[i][j]],
+            cmd = "cp $< $@",
```
jbedard, line 58: Something here is not working on windows where this genrule is either not being generated or is not dropping the `.test-` from the path, I assume because of slashes.

Anyone see anything wrong here? Or have a windows box to debug it more?

walk/config.go
DavidZbarsky-at: Nice!

walk/walk.go
```diff
@@ -122,16 +122,15 @@ func Walk(c *config.Config, cexts []config.Configurer, dirs []string, mode Mode,
 		log.Printf("error loading .bazelignore: %v", err)
 	}
 
-	visit(c, cexts, isBazelIgnored, knownDirectives, updateRels, wf, c.RepoRoot, "", false)
+	visit(c, cexts, isBazelIgnored, knownDirectives, updateRels, wf, []string{}, "", false)
```
dzbarsky, line 125: nit: `[]string{}` =-> `nil`

walk/walk.go
```diff
@@ -122,16 +122,15 @@ func Walk(c *config.Config, cexts []config.Configurer, dirs []string, mode Mode,
...
-func visit(c *config.Config, cexts []config.Configurer, isBazelIgnored isIgnoredFunc, knownDirectives map[string]bool, updateRels *UpdateFilter, wf WalkFunc, dir, rel string, updateParent bool) {
-	if isBazelIgnored(rel) {
-		return
-	}
-
+func visit(c *config.Config, cexts []config.Configurer, isBazelIgnored isIgnoredFunc, knownDirectives map[string]bool, updateRels *UpdateFilter, wf WalkFunc, relParts []string, rel string, updateParent bool) {
 	haveError := false
 
+	// Absolute path to the directory being visited
+	dir := path.Join(c.RepoRoot, filepath.FromSlash(rel))
```
dzbarsky, line 132: push this down closer to the loop in case we will early exit?

walk/walk.go
```diff
@@ -155,31 +154,42 @@ func visit(c *config.Config, cexts []config.Configurer, isBazelIgnored isIgnored
...
+		entRel := path.Join(rel, name)
+
+		// Skip blank entry names, .git directories, ignored and excluded.
+		if name == "" || name == ".git" || isBazelIgnored(entRel) || wc.isExcluded(entRel) {
 			continue
 		}
-		ent := resolveFileInfo(wc, dir, rel, ent)
+
+		// Skip git-ignored when the gitignore option is enabled.
+		if wc.gitignoreEnabled && wc.isGitIgnored(append(relParts, name), ent.IsDir()) {
```
dzbarsky, line 173: I think you want to do `relParts = append(relParts, name)` and then `relParts = relParts[:len(relParts)-1]` after the call. Otherwise when `relParts` is full you are going to realloc it for every ent

walk/config.go
```diff
@@ -49,15 +54,22 @@ func getWalkConfig(c *config.Config) *walkConfig {
...
+}
+
+func (wc *walkConfig) shouldFollow(p string) bool {
+	return matchAnyGlob(wc.follow, p)
 }
 
-func (wc *walkConfig) shouldFollow(rel, base string) bool {
-	return matchAnyGlob(wc.follow, path.Join(rel, base))
+func (wc *walkConfig) isGitIgnored(p []string, isDir bool) bool {
+	if wc.gitignoreEnabled && wc.gitignore != nil {
```
dzbarsky, line 66: do you actually need the `wc.gitignore != nil` check? It looks like a nil receiver will be handled correctly? https://github.com/go-git/go-git/blob/cef892e5701bb6ae7c740e088ad4c19894f2fd60/plumbing/format/gitignore/matcher.go#L18-L30

walk/walk.go
jbedard: It's used on the next line to do the `ReadDir`...

walk/walk.go
jbedard: Good idea for this loop.

Note for the recursive loop below I don't think we should though. Technically we probably can atm but that might get confusing.

walk/config.go
jbedard: It looks like you're correct, so we can make this method 1 line 👍

tests/BUILD.bazel
jbedard: I gave up and ignored this test on windows for now. I confirmed it is renaming the files but the `.test-*` files are still visible. I assume this is a windows+sandboxing issue?

walk/config.go
```diff
@@ -151,6 +172,50 @@ func loadBazelIgnore(repoRoot string) (isIgnoredFunc, error) {
...
+		// Trim, ignore empty lines and comments
+		i := strings.TrimSpace(reader.Text())
+		if i == "" || strings.HasPrefix(i, "#") {
+			continue
+		}
+
+		matcherPatterns = append(matcherPatterns, gitignore.ParsePattern(i, domain))
+	}
+
+	// Fallback to the parent if no patterns are found
```
dzbarsky, line 199: so this is the case when you have an empty `.gitignore`? That seems rare enough that I would replace this check and the `if` below with just `matcherPatterns = append(parent.patterns, matcherPatterns...)` - I think this should work even when `parent` is nil

walk/walk.go
```diff
@@ -155,31 +154,50 @@ func visit(c *config.Config, cexts []config.Configurer, isBazelIgnored isIgnored
...
-			regularFiles = append(regularFiles, base)
+			regularFiles = append(regularFiles, name)
 		}
 	}
 
 	shouldUpdate := updateRels.shouldUpdate(rel, updateParent)
 	for _, sub := range subdirs {
 		if subRel := path.Join(rel, sub); updateRels.shouldVisit(subRel, shouldUpdate) {
-			visit(c, cexts, isBazelIgnored, knownDirectives, updateRels, wf, filepath.Join(dir, sub), subRel, shouldUpdate)
+			// Create a new slice to allow the nested 'visit' call to modify the slice without affecting the parent.
```
dzbarsky, line 197: do we expect the existing elements to get mutated? Or only append new ones? Asking because if `relParts` has capacity for an extra element, then `subParts` will be a new slice but they will share the same backing store, so mutating it's 0..n-1 index elements will also mutate `relParts`

jbedard: The only mutation is of `entParts` which is local to this method, however `entParts = append(relParts, ".")`.

We might have to think about this more when the loop becomes concurrent though.

DavidZbarsky-at:
This looks great to me, @fmeum wdyt?

DavidZbarsky-at (approved):

walk/walk.go
```diff
@@ -161,17 +161,37 @@ func visit(c *config.Config, cexts []config.Configurer, knownDirectives map[stri
...
 		base := ent.Name()
 		entRel := path.Join(rel, base)
 		if wc.isExcluded(entRel) {
 			continue
 		}
+
+		entParts[len(entParts)-1] = base
+
+		// Skip git-ignored when the gitignore option is enabled.
+		if wc.isGitIgnored(entParts, ent.IsDir()) {
```
jbedard, line 191: Note this is its own conditional and not embedded within `isExcluded` for a few reasons:
* I don't think .gitignore can exclude the parent directory it is within the way `# gazelle:exclude *` can (that is the reason we must invoke `isExcluded` both above and within this `for` loop).
* `isGitIgnored` requires the path as `[]string` and needs to know `IsDir()`, I don't think we should override `isExcluded` with more params

fmeum:
Based on https://github.com/aspect-build/aspect-cli/issues/755, what do you expect to be the use case for this if we can't enable it by default due to performance issues? It's surprising to me that `.gitignore` processing should be costly - do you know what the hotspots are in the current implementation?

jbedard:
I was suggesting it should be disabled by default because that is the ideal use case for both "correctness" and performance (although performance is only really a concern in very large projects globbing on hundreds of thousands of files).

However enabling it by default is probably a better initial experience for most projects so I'm fine with that as well.

Longer explanation, imo...

The ideal use case is when bazelignore specifies everything bazel should ignore, and gitignore specifies everything git should ignore. Those are not always 1-1... sometimes bazel might need to process things that are git ignored, very often things such as IDE config are in git but should be ignored by bazel etc.

In reality the more common use case is projects use gitignore and then (partially) duplicate that in `glob([exclude])` + `# gazelle:exclude`, and bazelignore is almost unused. This is done either due to lack of awareness of bazelignore or the fact that bazelignore does not support globs so it is too much of a hassle or sometimes not even possible.

This feature would remove the need for many of those `glob([exclude])` + `# gazelle:exclude` which I think makes this feature worth it for most projects. I essentially agree with everything in this comment: https://github.com/bazel-contrib/bazel-gazelle/issues/1166#issuecomment-1024595950

jbedard:
Rebased. @fmeum any updates with getting this reviewed/merged?

jbedard:
Rebased again, and updated the `go-git` library.

jbedard:
@fmeum has your opinion changed now that bazel8 has [ignore_directories](https://bazel.build/versions/8.0.0/rules/lib/globals/repo#ignore_directories)? See https://github.com/bazel-contrib/bazel-gazelle/issues/1994

We could drop this PR and tell people to use `ignore_directories` instead which really is the "correct" solution - gazelle should ignore the same things bazel ignores (anything extra should be a gazelle api such as `# gazelle:exclude`).

The `ignore_directories` is just a bazel `glob()` like in BUILD files so it is much simpler then gitignore, we can just use `doublestar` like we already do for `# gazelle:excludes` instead of this `go-git` library in this PR etc.

fmeum:
I appreciate the effort you put into this, but yes, I think that we should support `ignore_directories` instead. As you remarked, it's simpler but still expressive enough to be generally useful. We could also enable it by default without any concerns about that excluding too much or too little.

jbedard:
Here's a quick `ignore_directories` implementation: https://github.com/bazel-contrib/bazel-gazelle/pull/2039

PR #1910, opened 2024-09-05
override-generator: allow users to specify default values

tyler-french:
If a repo specifies `gazelle_default_attributes` to be something
different than the `gazelle_override` tag defaults,
the user should be able to pass this to the override generator
so that the output takes this into consideration, and skips
outputing redundant information.

This PR allows the user to specify a default `build_file_generation`
and `build_file_proto_mode` that is used in their repo, and generate appropriate overrides.

These two tags need to be handled manually after the other tags because of the
non-trivial logic in them.

### DIFF
--- a/tools/override-generator/main.go
+++ b/tools/override-generator/main.go
@@ -31,17 +31,14 @@ const (
 
 // attribute constants that are used multiple times.
 const (
-	_buildFileGenerationAttr = "build_file_generation"
-	_buildFileProtoModeAttr  = "build_file_proto_mode"
-	_patchArgsAttr           = "patch_args"
-	_buildDirectivesAttr     = "build_directives"
-	_directivesAttr          = "directives"
+	_buildFileGenerationAttr        = "build_file_generation"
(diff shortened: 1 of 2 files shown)


tools/override-generator/main.go
```diff
@@ -131,13 +131,17 @@ func run(a mainArgs, stderr io.Writer) error {
...
+		_buildFileProtoModeAttr:  a.defaultBuildFileProtoMode,
+	}
 
 	// Iterate over all repositories and convert them to override rules
 	// The repos are ordered by "name", and the sets are sorted, so the output
 	// will be deterministic.
 	for _, r := range repos {
 		if r.Kind() == "go_repository" {
-			repoOverrides := goRepositoryToOverrideSet(r)
+			repoOverrides := goRepositoryToOverrideSet(r, defaultAttrs)
```
tyler-french, line 144: Actually need to refactor this a little. Since even though we don't want the override when it's just default attributes. We will need it if there's a `gazelle-override` created, since if there is, it falls back to `auto`

fmeum (approved):

PR #1914, opened 2024-09-06
bzlmod/go_work: support relative paths

rockwotj:
**What type of PR is this?**

Feature

go_deps.from_file for bzlmod

We have a use case where we have multiple go modules in a single repo,
but adding go.work breaks some existing tooling, if it's added to the
root. We want to add it to another location like: bazel/thirdparty
but then we need to support having entries in go.work that look like:
`../../src/go/rpk`. This patch allows leading parent segments (..) in
go.work files to support our use case.

Did not file an issue (I can if desired).

**Other notes for review**

This doesn't guard against relative paths that point outside the repo.
I think if desired we could calculate the maximum number of parent segments
and limit them. I also have no desire to do full path normalization...

### DIFF
--- a/internal/bzlmod/go_mod.bzl
+++ b/internal/bzlmod/go_mod.bzl
@@ -27,6 +27,17 @@ def _validate_go_version(path, state, tokens, line_no):
         fail("{}:{}: unexpected token '{}' after '{}'".format(path, line_no, tokens[2], tokens[1]))
 
 def use_spec_to_label(repo_name, use_directive):
+    # Strip all leading ../ assuming they point into the existing repo
+    if use_directive.startswith("../"):
+        segments = []
+        done = False
+        for segment in use_directive.split("/"):
+            if not done and segment == "..":
+                continue
+            segments.append(segment)
+            done = True
+        use_directive = "/".join(segments)
+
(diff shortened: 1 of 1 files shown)


internal/bzlmod/go_mod.bzl
```diff
@@ -27,6 +27,17 @@ def _validate_go_version(path, state, tokens, line_no):
         fail("{}:{}: unexpected token '{}' after '{}'".format(path, line_no, tokens[2], tokens[1]))
 
 def use_spec_to_label(repo_name, use_directive):
+    # Strip all leading ../ assuming they point into the existing repo
```
rockwotj, line 30: I guess we have access to bazel skylib, and I could use that...

rockwotj:
I found a work around and don't need this (which was kind of a hack to be honest)

fmeum:
Could you share the workaround for posterity?

rockwotj:
https://github.com/redpanda-data/redpanda/commit/5ee6285538fcef659f01a53faf7bfb82fe0ccf3a

Kind of a hack still, but when computing the label the go.work file's location is not accounted for so we can put it in a different directory and the labels stay the same. Fixing that would mean it should normalize paths which would also be the proper fix to allow ../ directives

PR #1917, opened 2024-09-10
feat: include package URL in package_info

alexeagle:
Requires upgrade to the 1.0.0 release of rules_license, which is more appropriate than a pre-release anyway.

Follow-up to #1852

Note that I'm still unable to gather the resulting data from an end-user point of view using the aspects provided by rules_license. Will look for some help on Slack to see if there's still missing plumbing.

### DIFF
--- a/internal/go_repository.bzl
+++ b/internal/go_repository.bzl
@@ -392,10 +392,10 @@ def _generate_package_info(*, importpath, version):
     # See specification:
     # https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#golang
     # scheme:type/namespace/name@version?qualifiers#subpath
-    purl = "pkg:golang/{namespace_and_name}{version}".format(
+    purl = "pkg:golang/{namespace_and_name}@{version}".format(
         namespace_and_name = importpath,
-        version = "@{}".format(version) if version else "",
-    )
+        version = version,
+    ) if version else None
 
     return """
 load("@rules_license//rules:package_info.bzl", "package_info")
--- a/internal/go_repository.bzl
+++ b/internal/go_repository.bzl
@@ -365,6 +365,7 @@ def _go_repository_impl(ctx):
         # Do not override a REPO.bazel patched in by users. This also provides a
         # way for users to opt out of Gazelle-generated package_info.
         repo_file = ctx.path("REPO.bazel")
+        print("REPO.bazel exists?", repo_file.exists)
         if not repo_file.exists:
             ctx.file("REPO.bazel", """\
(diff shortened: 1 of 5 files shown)


internal/go_repository.bzl
```diff
@@ -365,6 +365,7 @@ def _go_repository_impl(ctx):
         # Do not override a REPO.bazel patched in by users. This also provides a
         # way for users to opt out of Gazelle-generated package_info.
         repo_file = ctx.path("REPO.bazel")
+        print("REPO.bazel exists?", repo_file.exists)
```
fmeum, line 368:
```suggestion
```

internal/go_repository.bzl
```diff
@@ -387,8 +388,16 @@ def _generate_package_info(*, importpath, version):
     package_name = importpath
 
     # TODO: Consider adding support for custom remotes.
-    package_url = "https://" + importpath if version else None
+    package_url = "https://{}".format(importpath) if version else None
     package_version = version.removeprefix("v") if version else None
+    # See specification:
+    # https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#golang
+    # scheme:type/namespace/name@version?qualifiers#subpath
+    purl = "pkg:golang/{namespace_and_name}{version}".format(
```
fmeum, line 396: We should only set this if `version` is not `None` as if it isn't, we aren't using `go mod` to fetch the module but get it from wherever (say local path or mirror). I would rather not claim any metadata is correct in that case for now.

internal/go_repository.bzl
```diff
@@ -387,8 +388,16 @@ def _generate_package_info(*, importpath, version):
...
+        namespace_and_name = importpath,
+        version = "@{}".format(version) if version else "",
```
fmeum, line 398: The purl spec doesn't contain any information or examples on how version numbers are encoded. Do you know of any consumers that support this format?

internal/go_repository.bzl
alexeagle: done

internal/go_repository.bzl
alexeagle: Based on https://github.com/package-url/purl-spec/issues/63

- ORT, Syft, and SCTK seem to be some consumers that read it
- There are some thorny cases
- version as `@1.2.3` does not seem contentious

fmeum: Wish I hadn't started reading that thread, why can't we have nice (and simple) things for once?

But I agree that the current form is definitely the most pleasant one, so let's go with it.

fmeum (approved):

PR #1920, opened 2024-09-12
perf: Improve merger.Mergefile

dzbarsky:
**What type of PR is this?**
perf improvement

gazelle core

This slightly improve runtime perf due to avoiding allocations of the error object and reduced GC pressure
[Profile](https://private-user-images.githubusercontent.com/94403429/367062767-c15a6a5d-e237-463e-beaf-332aa0fe4634.png?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MjYxNzcwMjAsIm5iZiI6MTcyNjE3NjcyMCwicGF0aCI6Ii85NDQwMzQyOS8zNjcwNjI3NjctYzE1YTZhNWQtZTIzNy00NjNlLWJlYWYtMzMyYWEwZmU0NjM0LnBuZz9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNDA5MTIlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjQwOTEyVDIxMzIwMFomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPTRhNmM4MGY5YTI3OTNlNDc4YjI1YTNlYzUwMGNhMmE4ZDIyNjk3ZjUzZDg5NWE2NmEyZjM0NTQ5YmMyYzY2MjYmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JmFjdG9yX2lkPTAma2V5X2lkPTAmcmVwb19pZD0wIn0.kIYCIu8U3kE3P7bq6rhNe3T2WgYuDKmlyQlq_lfn--M)

Fixes #

**Other notes for review**

### DIFF
--- a/merger/merger.go
+++ b/merger/merger.go
@@ -117,7 +117,7 @@ func MergeFile(oldFile *rule.File, emptyRules, genRules []*rule.Rule, phase Phas
 
 	// Merge empty rules into the file and delete any rules which become empty.
 	for _, emptyRule := range emptyRules {
-		if oldRule, _ := Match(oldFile.Rules, emptyRule, kinds[emptyRule.Kind()]); oldRule != nil {
+		if oldRule, _ := match(oldFile.Rules, emptyRule, kinds[emptyRule.Kind()], false); oldRule != nil {
 			if oldRule.ShouldKeep() {
 				continue
 			}
@@ -211,6 +211,10 @@ func substituteRule(r *rule.Rule, substitutions map[string]string, info rule.Kin
 // order that attributes are listed). If disambiguation is successful,
 // the rule and nil are returned. Otherwise, nil and an error are returned.
 func Match(rules []*rule.Rule, x *rule.Rule, info rule.KindInfo) (*rule.Rule, error) {
+	return match(rules, x, info, true)
+}
+
+func match(rules []*rule.Rule, x *rule.Rule, info rule.KindInfo, wantError bool) (*rule.Rule, error) {
 	xname := x.Name()
 	xkind := x.Kind()
 	var nameMatches []*rule.Rule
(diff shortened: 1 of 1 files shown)


fmeum:
Do you have a lot of errors in your repo? The profile link doesn't work for me, so I'm not sure how much this actually saves.

dzbarsky:
It saves around 30ms for us, which isn't a huge amount but these little things add up :)

I think it points to a bit of a bigger issue though in terms of scaling # of languages/plugins. What is happening in our case is that we have a predominantly Typescript repo with some Go, Rust, and Java code sprinkled in. We make the default target in each package (`:foo` in package `foo`) a `ts_library` if there are any TS files in the package, which there are in the majority of packages. The Go plugin tries to do the same thing, which is what leads to the creation of these errors (go_library generated target has same name but different kind as existing ts_library target). We currently aren't using Rust or Java with Gazelle, but presumably if we were, they might suffer from the same issue of trying to use the same target names.

One way to solve this would be to restrict which paths apply which plugins, but we typically group our code by functionality rather than by language, so it would require a lot of tuning of inclusions/exclusions.

Perhaps this can be addressed with a bigger restructuring of the code that could allow to avoid more processing of the Go plugin if there are no Go files in the package?

fmeum:
Thanks for the context. I also don't see how we could easily improve the high-level situation. But this low-level optimization is simple enough that I'm now convinced we should merge it.

fmeum (approved):

PR #1921, opened 2024-09-12
feat: support only updating existing BUILD files

### LINKED ISSUE #1832: request: option to not generate in all directories
### Request

Provide a way to only invoke language callbacks on directories that already have a BUILD. Include subdirectory content in `language.GenerateArgs.{Subdirs, RegularFiles, GenFiles}` for subdirectories without a BUILD.

### Background

In rules_go and other rulesets it seems normal to generate BUILD in every directory. In a lot of languages/rulesets this is not always ideal and sometimes impossible (circular deps across directories is way too common in some ecosystems...).

In multiple gazelle languages now I've added directives such as `generation_mode: create|update` where "update" means only existing BUILDs are updated and no new BUILDs are created. This requires hacks because `language.GenerateArgs` only provides the list of direct child file/subdirs and not nested ones. The `Configure` and `GenerateRules` calls are then still called for every single directory and we must noop if there is no BUILD there, this causes extra overhead, extra fs walking overhead etc.

### Proposal

A cross-language directive to enable "update-only" mode, where no new BUILDs are created. When enabled:
* `Configure.Configure` is only be invoked on directories that have a BUILD
* `language.GenerateRules` is only be invoked on directories that have a BUILD
* `GenerateArgs.{RegularFiles, GenFiles}` includes files in subdirs that do not have BUILDs
* `GenerateArgs.Subdirs` only includes subdirs that *do* have BUILDs

jbedard:
Close #1832

**What type of PR is this?**

Feature

> Other

> all

Add an option to enable only updating existing BUILDs and not creating any new ones.

I think this is common among extension authors, normally implemented with a quick-exit within `Configure` if the `*rule.File` is `nil`. However all the subdirs then need to be manually traversed to find files in subdirectories with no BUILD.

I've always maintained a patch similar to this PR to avoid plugins having to do manual fs operations.

Fixes #1832

**Other notes for review**

### DIFF
--- a/walk/config.go
+++ b/walk/config.go
@@ -33,14 +33,27 @@ import (
 	gzflag "github.com/bazelbuild/bazel-gazelle/flag"
 )
 
+// generationModeType represents one of the generation modes.
+type generationModeType string
+
+// Generation modes
+const (
+	// Update: update and maintain existing BUILD files
+	generationModeUpdate generationModeType = "update_only"
+
+	// Create: create new and update existing BUILD files
+	generationModeCreate generationModeType = "create_and_update"
+)
+
 // TODO(#472): store location information to validate each exclude. They
 // may be set in one directory and used in another. Excludes work on
 // declared generated files, so we can't just stat.
 
 type walkConfig struct {
-	excludes []string
-	ignore   bool
-	follow   []string
+	updateOnly bool
+	excludes   []string
+	ignore     bool
+	follow     []string
 }
 
 const walkName = "_walk"
@@ -70,7 +83,7 @@ func (*Configurer) RegisterFlags(fs *flag.FlagSet, cmd string, c *config.Config)
 func (*Configurer) CheckFlags(fs *flag.FlagSet, c *config.Config) error { return nil }
 
 func (*Configurer) KnownDirectives() []string {
-	return []string{"exclude", "follow", "ignore"}
+	return []string{"generation_mode", "exclude", "follow", "ignore"}
 }
 
 func (cr *Configurer) Configure(c *config.Config, rel string, f *rule.File) {
@@ -82,6 +95,16 @@ func (cr *Configurer) Configure(c *config.Config, rel string, f *rule.File) {
 	if f != nil {
 		for _, d := range f.Directives {
 			switch d.Key {
+			case "generation_mode":
+				switch generationModeType(strings.TrimSpace(d.Value)) {
+				case generationModeUpdate:
+					wcCopy.updateOnly = true
+				case generationModeCreate:
+					wcCopy.updateOnly = false
+				default:
+					log.Fatalf("unknown generation_mode %q in //%s", d.Value, f.Pkg)
+					continue
+				}
 			case "exclude":
 				if err := checkPathMatchPattern(path.Join(rel, d.Value)); err != nil {
 					log.Printf("the exclusion pattern is not valid %q: %s", path.Join(rel, d.Value), err)
--- a/walk/walk.go
+++ b/walk/walk.go
@@ -140,7 +140,7 @@ func Walk(c *config.Config, cexts []config.Configurer, dirs []string, mode Mode,
 //
 // Traversal may skip subtrees or files based on the config.Config exclude/ignore/follow options
 // as well as the UpdateFilter callbacks.
-func visit(c *config.Config, cexts []config.Configurer, knownDirectives map[string]bool, updateRels *UpdateFilter, trie *pathTrie, wf WalkFunc, rel string, updateParent bool) {
+func visit(c *config.Config, cexts []config.Configurer, knownDirectives map[string]bool, updateRels *UpdateFilter, trie *pathTrie, wf WalkFunc, rel string, updateParent bool) ([]string, bool) {
 	haveError := false
 
 	// Absolute path to the directory being visited
@@ -157,11 +157,16 @@ func visit(c *config.Config, cexts []config.Configurer, knownDirectives map[stri
 		haveError = true
 	}
 
-	configure(cexts, knownDirectives, c, rel, f)
-	wc := getWalkConfig(c)
+	collectionOnly := f == nil && getWalkConfig(c).updateOnly
 
+	// Configure the current directory if not only collecting files
+	if !collectionOnly {
+		configure(cexts, knownDirectives, c, rel, f)
+	}
+
+	wc := getWalkConfig(c)
 	if wc.isExcluded(rel) {
-		return
+		return nil, false
 	}
 
 	// Filter and collect files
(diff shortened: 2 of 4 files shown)


walk/config.go
```diff
@@ -73,7 +86,7 @@ func (*Configurer) RegisterFlags(fs *flag.FlagSet, cmd string, c *config.Config)
 func (*Configurer) CheckFlags(fs *flag.FlagSet, c *config.Config) error { return nil }
 
 func (*Configurer) KnownDirectives() []string {
-	return []string{"exclude", "follow", "ignore"}
+	return []string{"generation_mode", "exclude", "follow", "ignore"}
```
jbedard, line 89: This is the directive named used by the Aspect plugins today. I think we need something better here that doesn't include the word "mode" that already has a meaning within gazelle.

Ideas?

dzbarsky: build_file_action?
create_build_files yes/no?

walk/walk.go
```diff
@@ -125,9 +125,9 @@ func Walk(c *config.Config, cexts []config.Configurer, dirs []string, mode Mode,
 	visit(c, cexts, isBazelIgnored, knownDirectives, updateRels, wf, c.RepoRoot, "", false)
 }
 
-func visit(c *config.Config, cexts []config.Configurer, isBazelIgnored isIgnoredFunc, knownDirectives map[string]bool, updateRels *UpdateFilter, wf WalkFunc, dir, rel string, updateParent bool) {
+func visit(c *config.Config, cexts []config.Configurer, isBazelIgnored isIgnoredFunc, knownDirectives map[string]bool, updateRels *UpdateFilter, wf WalkFunc, dir, rel string, updateParent bool) ([]string, bool) {
```
dzbarsky, line 128: nit: add a comment explaining what this returns?

walk/walk.go
```diff
@@ -170,24 +178,39 @@ func visit(c *config.Config, cexts []config.Configurer, isBazelIgnored isIgnored
...
 
+	var subdirs []string
+
+	// Visit subdirectories dividing them into `regularFiles` and `subdirs``
 	shouldUpdate := updateRels.shouldUpdate(rel, updateParent)
-	for _, sub := range subdirs {
+	for _, sub := range dirs {
 		if subRel := path.Join(rel, sub); updateRels.shouldVisit(subRel, shouldUpdate) {
-			visit(c, cexts, isBazelIgnored, knownDirectives, updateRels, wf, filepath.Join(dir, sub), subRel, shouldUpdate)
+			subFiles, mergeFiles := visit(c, cexts, isBazelIgnored, knownDirectives, updateRels, wf, filepath.Join(dir, sub), subRel, shouldUpdate)
```
dzbarsky, line 193: nit: maybe rename mergeFiles to `shouldMerge` or `isInParent` or something that makes it more obviously a boolean?

walk/config.go
DavidZbarsky-at: I am going to try to this patch at Airtable. I am thinking instead of manually specifying `generation_mode` to isntead do something like (pseudocode) `collectionOnly := f == nil !any(file.endswith('.go') for file in ents)`, since we have go code spread throughout the tree

jbedard: I think different repos will have different preferences and we can't really make any assumptions based on which files exist.

fmeum: Assuming you are in David's situation, could you still leverage `generation_mode: update` to simplify your language implementation or would you need to resort to `create` with the same hack mentioned in the linked issue? 

I'm asking because adding a new directive comes with a complexity cost that may not be justified if it doesn't address many use cases. Do you see a way to simplify what a language would need to do to get this behavior without adding a new directive?

jbedard: I've changed my hack described in the issue to [a patch](https://github.com/aspect-build/aspect-cli/blob/main/patches/bazelbuild_bazel-gazelle_aspect-walk-subdir.patch) that's essentially the same as this PR. The hack required manually walking the fs (in addition to gazelle walking it) which is bad in many ways, performance being a big one.

I think no matter what a new API is needed so gazelle can generate the `language.GenerateArgs` with the correct files. It could be an API in the `Configure` phase, but today there are no APIs there that change gazelle's behaviour and adding such an API sounds far more complex then adding a directive.

jbedard:
I've rebased this and added a test.

@fmeum any ideas or feedback here? WDYT?

walk/config.go
```diff
@@ -33,14 +33,27 @@ import (
...
+// GenerationModeType represents one of the generation modes.
+type GenerationModeType string
+
+// Generation modes
+const (
+	// Update: update and maintain existing BUILD files
+	GenerationModeUpdate GenerationModeType = "update"
+
+	// Create: create new and update existing BUILD files
+	GenerationModeCreate GenerationModeType = "create"
```
fmeum, line 45: What do you think about `update_only` and `create_and_update`? Slightly more verbose, but also more descriptive.

walk/config.go
```diff
@@ -33,14 +33,27 @@ import (
 	gzflag "github.com/bazelbuild/bazel-gazelle/flag"
 )
 
+// GenerationModeType represents one of the generation modes.
+type GenerationModeType string
```
fmeum, line 37: Can we make this private to the package?

walk/config.go
```diff
@@ -70,7 +83,7 @@ func (*Configurer) RegisterFlags(fs *flag.FlagSet, cmd string, c *config.Config)
 func (*Configurer) CheckFlags(fs *flag.FlagSet, c *config.Config) error { return nil }
 
 func (*Configurer) KnownDirectives() []string {
-	return []string{"exclude", "follow", "ignore"}
+	return []string{"generation_mode", "exclude", "follow", "ignore"}
```
fmeum, line 86: Please add docs for the new directive.

I don't have an idea, but would prefer a more descriptive term if we can find one. We already have a `build_file_mode`.

fmeum (approved):
Sorry for the delay. I find this useful and it doesn't introduce a lot of complexity, so I'm happy to merge after the minor comments have been resolved.

walk/config.go
jbedard: We already have `build_file_mode`?

walk/config.go
jbedard: Those sound good to me 👍

walk/config.go
```diff
@@ -33,14 +33,27 @@ import (
...
+// generationModeType represents one of the generation modes.
+type generationModeType string
+
+// Generation modes
+const (
+	// Update: update and maintain existing BUILD files
+	generationModeUpdate generationModeType = "update_only"
+
+	// Create: create new and update existing BUILD files
+	generationModeCreate generationModeType = "create"
```
fmeum, line 45: I think we discussed `create_and_update` for this.

walk/config.go
fmeum: You are right, sounds okay

fmeum (approved):
This is still missing an entry in the docs table on directives.

walk/config.go
jbedard: Opps, updated.

walk/config.go
```diff
@@ -82,6 +95,16 @@ func (cr *Configurer) Configure(c *config.Config, rel string, f *rule.File) {
...
 		for _, d := range f.Directives {
 			switch d.Key {
+			case "generation_mode":
+				switch generationModeType(strings.TrimSpace(d.Value)) {
+				case generationModeUpdate:
+					wcCopy.updateOnly = true
+				case generationModeCreate:
+					wcCopy.updateOnly = false
+				default:
+					log.Printf("unknown generation_mode %q in //%s", d.Value, f.Pkg)
```
jbedard, line 105: @fmeum I added this print+continue since you last reviewed, so now it ignores invalid values...

jbedard:
I think I addressed all the feedback here. Look ok?

walk/config.go
fmeum: Can we fail here instead? It's often hard to spot these "warnings" in the output.

jbedard: Done 👍

fmeum (approved):

PR #1928, opened 2024-09-18
perf: load existing BUILD files concurrently

jbedard:
Now that the fs is walked concurrently we can also load BUILDs concurrently at the same time, that is primarily the `rule.LoadFile` method.

If `ReadBuildFilesDir` is set then this may have a more significant performance gain because the extra `os.ReadDir` call will also be run concurrently now.

**What type of PR is this?**

> Other

> all

Performance

### DIFF
--- a/walk/config.go
+++ b/walk/config.go
@@ -75,7 +75,6 @@ type Configurer struct {
 	// Excludes and BUILD filenames specified on the command line.
 	// May be extending with BUILD directives.
 	cliExcludes       []string
-	cliFollow         []string
 	cliBuildFileNames string
 
 	// Alternate BUILD read/write directories
@@ -86,7 +85,6 @@ func (wc *Configurer) RegisterFlags(fs *flag.FlagSet, cmd string, c *config.Conf
 	c.Exts[walkConfigurerName] = wc
 
 	fs.Var(&gzflag.MultiFlag{Values: &wc.cliExcludes}, "exclude", "pattern that should be ignored (may be repeated)")
-	fs.Var(&gzflag.MultiFlag{Values: &wc.cliFollow}, "follow", "pattern that should be followed (may be repeated)")
 	fs.StringVar(&wc.cliBuildFileNames, "build_file_name", strings.Join(config.DefaultValidBuildFileNames, ","), "comma-separated list of valid build file names.\nThe first element of the list is the name of output build files to generate.")
 	fs.StringVar(&wc.readBuildFilesDir, "experimental_read_build_files_dir", "", "path to a directory where build files should be read from (instead of -repo_root)")
 	fs.StringVar(&wc.writeBuildFilesDir, "experimental_write_build_files_dir", "", "path to a directory where build files should be written to (instead of -repo_root)")
--- a/walk/config.go
+++ b/walk/config.go
@@ -51,17 +51,14 @@ const (
 // declared generated files, so we can't just stat.
 
 type walkConfig struct {
-	updateOnly bool
-	excludes   []string
-	ignore     bool
-	follow     []string
+	updateOnly          bool
+	excludes            []string
+	ignore              bool
+	follow              []string
+	validBuildFileNames []string
 }
 
-const walkName = "_walk"
-
-func getWalkConfig(c *config.Config) *walkConfig {
-	return c.Exts[walkName].(*walkConfig)
-}
+const walkConfigurerName = "_walk"
 
 func (wc *walkConfig) isExcluded(p string) bool {
 	return matchAnyGlob(wc.excludes, p)
@@ -73,35 +70,44 @@ func (wc *walkConfig) shouldFollow(p string) bool {
 
 var _ config.Configurer = (*Configurer)(nil)
 
-type Configurer struct{}
+type Configurer struct {
+	cliExcludes []string
+	cliFollow   []string
+}
 
-func (*Configurer) RegisterFlags(fs *flag.FlagSet, cmd string, c *config.Config) {
-	wc := &walkConfig{}
-	c.Exts[walkName] = wc
-	fs.Var(&gzflag.MultiFlag{Values: &wc.excludes}, "exclude", "pattern that should be ignored (may be repeated)")
+func (cc *Configurer) RegisterFlags(fs *flag.FlagSet, cmd string, c *config.Config) {
+	c.Exts[walkConfigurerName] = cc
+	fs.Var(&gzflag.MultiFlag{Values: &cc.cliExcludes}, "exclude", "pattern that should be ignored (may be repeated)")
+	fs.Var(&gzflag.MultiFlag{Values: &cc.cliFollow}, "follow", "pattern that should be followed (may be repeated)")
 }
 
 func (*Configurer) CheckFlags(fs *flag.FlagSet, c *config.Config) error { return nil }
 
 func (*Configurer) KnownDirectives() []string {
-	return []string{"generation_mode", "exclude", "follow", "ignore"}
+	return []string{"build_file_name", "generation_mode", "exclude", "follow", "ignore"}
 }
 
-func (cr *Configurer) Configure(c *config.Config, rel string, f *rule.File) {
-	wc := getWalkConfig(c)
+func (cr *Configurer) Configure(c *config.Config, rel string, f *rule.File) {}
+
+func (wc *walkConfig) newChild() *walkConfig {
 	wcCopy := &walkConfig{}
 	*wcCopy = *wc
 	wcCopy.ignore = false
+	return wcCopy
+}
 
+func (wc *walkConfig) readConfig(rel string, f *rule.File) {
 	if f != nil {
 		for _, d := range f.Directives {
 			switch d.Key {
+			case "build_file_name":
+				wc.validBuildFileNames = strings.Split(strings.TrimSpace(d.Value), ",")
 			case "generation_mode":
 				switch generationModeType(strings.TrimSpace(d.Value)) {
 				case generationModeUpdate:
-					wcCopy.updateOnly = true
+					wc.updateOnly = true
 				case generationModeCreate:
-					wcCopy.updateOnly = false
+					wc.updateOnly = false
 				default:
 					log.Fatalf("unknown generation_mode %q in //%s", d.Value, f.Pkg)
 					continue
@@ -111,23 +117,21 @@ func (cr *Configurer) Configure(c *config.Config, rel string, f *rule.File) {
 					log.Printf("the exclusion pattern is not valid %q: %s", path.Join(rel, d.Value), err)
 					continue
 				}
-				wcCopy.excludes = append(wcCopy.excludes, path.Join(rel, d.Value))
+				wc.excludes = append(wc.excludes, path.Join(rel, d.Value))
 			case "follow":
 				if err := checkPathMatchPattern(path.Join(rel, d.Value)); err != nil {
 					log.Printf("the follow pattern is not valid %q: %s", path.Join(rel, d.Value), err)
 					continue
 				}
-				wcCopy.follow = append(wcCopy.follow, path.Join(rel, d.Value))
+				wc.follow = append(wc.follow, path.Join(rel, d.Value))
 			case "ignore":
 				if d.Value != "" {
 					log.Printf("the ignore directive does not take any arguments. Did you mean to use gazelle:exclude instead? in //%s '# gazelle:ignore %s'", f.Pkg, d.Value)
 				}
-				wcCopy.ignore = true
+				wc.ignore = true
 			}
 		}
 	}
-
-	c.Exts[walkName] = wcCopy
 }
 
 type isIgnoredFunc = func(string) bool
--- a/walk/config.go
+++ b/walk/config.go
@@ -28,6 +28,7 @@ import (
 
 	"github.com/bazelbuild/bazel-gazelle/config"
 	"github.com/bazelbuild/bazel-gazelle/rule"
+	bzl "github.com/bazelbuild/buildtools/build"
 	"github.com/bmatcuk/doublestar/v4"
 
 	gzflag "github.com/bazelbuild/bazel-gazelle/flag"
@@ -166,12 +167,75 @@ func loadBazelIgnore(repoRoot string) (isIgnoredFunc, error) {
 		excludes[ignore] = struct{}{}
 	}
 
-	isIgnored := func(p string) bool {
+	isBazelIgnored := func(p string) bool {
 		_, ok := excludes[p]
 		return ok
 	}
 
-	return isIgnored, nil
+	return isBazelIgnored, nil
+}
+
+func loadRepoDirectoryIgnore(repoRoot string) (isIgnoredFunc, error) {
+	repoFilePath := path.Join(repoRoot, "REPO.bazel")
+	repoFileContent, err := os.ReadFile(repoFilePath)
+	if errors.Is(err, fs.ErrNotExist) {
+		return nothingIgnored, nil
+	}
+	if err != nil {
+		return nothingIgnored, fmt.Errorf("REPO.bazel exists but couldn't be read: %v", err)
+	}
+
+	ast, err := bzl.Parse(repoRoot, repoFileContent)
+	if err != nil {
+		return nothingIgnored, fmt.Errorf("failed to parse REPO.bazel: %v", err)
+	}
+
+	var ignoreDirectories []string
+
+	// Search for ignore_directories([...ignore strings...])
+	for _, expr := range ast.Stmt {
+		if call, isCall := expr.(*bzl.CallExpr); isCall {
+			if inv, isIdentCall := call.X.(*bzl.Ident); isIdentCall && inv.Name == "ignore_directories" {
+				if len(call.List) != 1 {
+					return nothingIgnored, fmt.Errorf("REPO.bazel ignore_directories() expects one argument")
+				}
+
+				list, isList := call.List[0].(*bzl.ListExpr)
+				if !isList {
+					return nothingIgnored, fmt.Errorf("REPO.bazel ignore_directories() unexpected argument type: %T", call.List[0])
+				}
+
+				for _, item := range list.List {
+					if strExpr, isStr := item.(*bzl.StringExpr); isStr {
+						if err := checkPathMatchPattern(strExpr.Value); err != nil {
+							log.Printf("the ignore_directories() pattern %q is not valid: %s", strExpr.Value, err)
+							continue
+						}
+
+						ignoreDirectories = append(ignoreDirectories, strExpr.Value)
+					}
+				}
+
+				// Only a single ignore_directories() is supported in REPO.bazel and searching can stop.
+				break
+			}
+		}
+	}
+
+	if len(ignoreDirectories) == 0 {
+		return nothingIgnored, nil
+	}
+
+	isRepoIgnored := func(p string) bool {
+		for _, ignore := range ignoreDirectories {
+			if doublestar.MatchUnvalidated(ignore, p) {
+				return true
+			}
+		}
+		return false
+	}
+
+	return isRepoIgnored, nil
 }
 
 func checkPathMatchPattern(pattern string) error {
--- a/walk/config.go
+++ b/walk/config.go
@@ -82,12 +82,12 @@ type Configurer struct {
 }
 
 func (wc *Configurer) RegisterFlags(fs *flag.FlagSet, cmd string, c *config.Config) {
-	c.Exts[walkConfigurerName] = wc
-
 	fs.Var(&gzflag.MultiFlag{Values: &wc.cliExcludes}, "exclude", "pattern that should be ignored (may be repeated)")
 	fs.StringVar(&wc.cliBuildFileNames, "build_file_name", strings.Join(config.DefaultValidBuildFileNames, ","), "comma-separated list of valid build file names.\nThe first element of the list is the name of output build files to generate.")
 	fs.StringVar(&wc.readBuildFilesDir, "experimental_read_build_files_dir", "", "path to a directory where build files should be read from (instead of -repo_root)")
 	fs.StringVar(&wc.writeBuildFilesDir, "experimental_write_build_files_dir", "", "path to a directory where build files should be written to (instead of -repo_root)")
+
+	c.Exts[walkConfigurerName] = wc
 }
 
 func (wc *Configurer) CheckFlags(_ *flag.FlagSet, c *config.Config) error {
--- a/walk/config.go
+++ b/walk/config.go
@@ -116,14 +116,20 @@ func (*Configurer) KnownDirectives() []string {
 
 func (cr *Configurer) Configure(c *config.Config, rel string, f *rule.File) {}
 
-func (wc *walkConfig) newChild() *walkConfig {
-	wcCopy := &walkConfig{}
-	*wcCopy = *wc
-	wcCopy.ignore = false
-	return wcCopy
-}
+func (wcp *walkConfig) newChild(rel string, f *rule.File) *walkConfig {
+	wc := &walkConfig{
+		// Always false by default.
+		ignore: false,
 
-func (wc *walkConfig) readConfig(rel string, f *rule.File) {
+		// Ineritied, potentially overwritten with directives.
+		updateOnly:          wcp.updateOnly,
+		validBuildFileNames: wcp.validBuildFileNames,
+
+		// Will potentially be extended with directives, must fully clone.
+		excludes: append([]string(nil), wcp.excludes...),
+		follow:   append([]string(nil), wcp.follow...),
+	}
+
 	if f != nil {
 		for _, d := range f.Directives {
 			switch d.Key {
@@ -159,6 +165,8 @@ func (wc *walkConfig) readConfig(rel string, f *rule.File) {
 			}
 		}
 	}
+
+	return wc
 }
 
 type ignoreFilter struct {
--- a/walk/config.go
+++ b/walk/config.go
@@ -163,9 +163,39 @@ func (wc *walkConfig) readConfig(rel string, f *rule.File) {
 	}
 }
 
+type ignoreFilter struct {
+	isBazelIgnored         isIgnoredFunc
+	isRepoDirectoryIgnored isIgnoredFunc
+}
+
 type isIgnoredFunc = func(string) bool
 
 var nothingIgnored isIgnoredFunc = func(string) bool { return false }
+
+func newIgnoreFilter(repoRoot string) *ignoreFilter {
+	isBazelIgnored, err := loadBazelIgnore(repoRoot)
+	if err != nil {
+		log.Printf("error loading .bazelignore: %v", err)
+	}
+
+	isRepoDirectoryIgnored, err := loadRepoDirectoryIgnore(repoRoot)
+	if err != nil {
+		log.Printf("error loading REPO.bazel ignore_directories(): %v", err)
+	}
+
+	return &ignoreFilter{
+		isBazelIgnored:         isBazelIgnored,
+		isRepoDirectoryIgnored: isRepoDirectoryIgnored,
+	}
+}
+
+func (f *ignoreFilter) isDirectoryIgnored(p string) bool {
+	return f.isBazelIgnored(p) || f.isRepoDirectoryIgnored(p)
+}
+
+func (f *ignoreFilter) isFileIgnored(p string) bool {
+	return f.isBazelIgnored(p)
+}
 
 func loadBazelIgnore(repoRoot string) (isIgnoredFunc, error) {
 	ignorePath := path.Join(repoRoot, ".bazelignore")
--- a/walk/config.go
+++ b/walk/config.go
@@ -117,6 +117,8 @@ func (*Configurer) KnownDirectives() []string {
 func (cr *Configurer) Configure(c *config.Config, rel string, f *rule.File) {}
 
 func (wcp *walkConfig) newChild(rel string, f *rule.File) *walkConfig {
+	// MUST BE THREAD SAFE - will be invoked concurrently while fs-walking and BUILD-parsing.
+
 	wc := &walkConfig{
 		// Always false by default.
 		ignore: false,
--- a/walk/config.go
+++ b/walk/config.go
@@ -164,47 +164,47 @@ func (wc *walkConfig) readConfig(rel string, f *rule.File) {
 }
 
 type ignoreFilter struct {
-	isBazelIgnored         isIgnoredFunc
-	isRepoDirectoryIgnored isIgnoredFunc
+	ignoreDirectoryGlobs []string
+	ignorePaths          map[string]struct{}
 }
 
-type isIgnoredFunc = func(string) bool
-
-var nothingIgnored isIgnoredFunc = func(string) bool { return false }
-
 func newIgnoreFilter(repoRoot string) *ignoreFilter {
-	isBazelIgnored, err := loadBazelIgnore(repoRoot)
+	bazelignorePaths, err := loadBazelIgnore(repoRoot)
 	if err != nil {
 		log.Printf("error loading .bazelignore: %v", err)
 	}
 
-	isRepoDirectoryIgnored, err := loadRepoDirectoryIgnore(repoRoot)
+	repoDirectoryIgnores, err := loadRepoDirectoryIgnore(repoRoot)
 	if err != nil {
 		log.Printf("error loading REPO.bazel ignore_directories(): %v", err)
 	}
 
 	return &ignoreFilter{
-		isBazelIgnored:         isBazelIgnored,
-		isRepoDirectoryIgnored: isRepoDirectoryIgnored,
+		ignorePaths:          bazelignorePaths,
+		ignoreDirectoryGlobs: repoDirectoryIgnores,
 	}
 }
 
 func (f *ignoreFilter) isDirectoryIgnored(p string) bool {
-	return f.isBazelIgnored(p) || f.isRepoDirectoryIgnored(p)
+	if _, ok := f.ignorePaths[p]; ok {
+		return true
+	}
+	return matchAnyGlob(f.ignoreDirectoryGlobs, p)
 }
 
 func (f *ignoreFilter) isFileIgnored(p string) bool {
-	return f.isBazelIgnored(p)
+	_, ok := f.ignorePaths[p]
+	return ok
 }
 
-func loadBazelIgnore(repoRoot string) (isIgnoredFunc, error) {
+func loadBazelIgnore(repoRoot string) (map[string]struct{}, error) {
 	ignorePath := path.Join(repoRoot, ".bazelignore")
 	file, err := os.Open(ignorePath)
 	if errors.Is(err, fs.ErrNotExist) {
-		return nothingIgnored, nil
+		return nil, nil
 	}
 	if err != nil {
-		return nothingIgnored, fmt.Errorf(".bazelignore exists but couldn't be read: %v", err)
+		return nil, fmt.Errorf(".bazelignore exists but couldn't be read: %v", err)
 	}
 	defer file.Close()
 
@@ -230,27 +230,22 @@ func loadBazelIgnore(repoRoot string) (isIgnoredFunc, error) {
 		excludes[ignore] = struct{}{}
 	}
 
-	isBazelIgnored := func(p string) bool {
-		_, ok := excludes[p]
-		return ok
-	}
-
-	return isBazelIgnored, nil
+	return excludes, nil
 }
 
-func loadRepoDirectoryIgnore(repoRoot string) (isIgnoredFunc, error) {
+func loadRepoDirectoryIgnore(repoRoot string) ([]string, error) {
 	repoFilePath := path.Join(repoRoot, "REPO.bazel")
 	repoFileContent, err := os.ReadFile(repoFilePath)
 	if errors.Is(err, fs.ErrNotExist) {
-		return nothingIgnored, nil
+		return nil, nil
 	}
 	if err != nil {
-		return nothingIgnored, fmt.Errorf("REPO.bazel exists but couldn't be read: %v", err)
+		return nil, fmt.Errorf("REPO.bazel exists but couldn't be read: %v", err)
 	}
 
 	ast, err := bzl.Parse(repoRoot, repoFileContent)
 	if err != nil {
-		return nothingIgnored, fmt.Errorf("failed to parse REPO.bazel: %v", err)
+		return nil, fmt.Errorf("failed to parse REPO.bazel: %v", err)
 	}
 
 	var ignoreDirectories []string
@@ -260,12 +255,12 @@ func loadRepoDirectoryIgnore(repoRoot string) (isIgnoredFunc, error) {
 		if call, isCall := expr.(*bzl.CallExpr); isCall {
 			if inv, isIdentCall := call.X.(*bzl.Ident); isIdentCall && inv.Name == "ignore_directories" {
 				if len(call.List) != 1 {
-					return nothingIgnored, fmt.Errorf("REPO.bazel ignore_directories() expects one argument")
+					return nil, fmt.Errorf("REPO.bazel ignore_directories() expects one argument")
 				}
 
 				list, isList := call.List[0].(*bzl.ListExpr)
 				if !isList {
-					return nothingIgnored, fmt.Errorf("REPO.bazel ignore_directories() unexpected argument type: %T", call.List[0])
+					return nil, fmt.Errorf("REPO.bazel ignore_directories() unexpected argument type: %T", call.List[0])
 				}
 
 				for _, item := range list.List {
@@ -285,20 +280,7 @@ func loadRepoDirectoryIgnore(repoRoot string) (isIgnoredFunc, error) {
 		}
 	}
 
-	if len(ignoreDirectories) == 0 {
-		return nothingIgnored, nil
-	}
-
-	isRepoIgnored := func(p string) bool {
-		for _, ignore := range ignoreDirectories {
-			if doublestar.MatchUnvalidated(ignore, p) {
-				return true
-			}
-		}
-		return false
-	}
-
-	return isRepoIgnored, nil
+	return ignoreDirectories, nil
 }
 
 func checkPathMatchPattern(pattern string) error {
--- a/walk/config.go
+++ b/walk/config.go
@@ -24,6 +24,7 @@ import (
 	"log"
 	"os"
 	"path"
+	"path/filepath"
 	"strings"
 
 	"github.com/bazelbuild/bazel-gazelle/config"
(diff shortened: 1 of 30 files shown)


walk/walk.go
```diff
@@ -363,6 +363,9 @@ func resolveFileInfo(wc *walkConfig, dir, rel string, ent fs.DirEntry) fs.DirEnt
 type pathTrie struct {
 	children map[string]*pathTrie
 	entry    *fs.DirEntry
+
+	buildFileErr error
```
jbedard, line 367: This could probably be done different or better, but persisting this `error` here keeps the scope of this PR minimal. We can think about changing it more in the future.

walk/walk_test.go
```diff
@@ -167,7 +167,7 @@ func TestCustomBuildName(t *testing.T) {
 		}
 	})
 	want := []string{
-		"sub/BUILD.test",
+		"sub/BUILD.bazel",
```
jbedard, line 170: Why would this happen? 🤔

jbedard: Oh, because `ValidBuildFileNames` can be changed with `# gazelle:build_file_name` in addition to via the `-build_file_name` cli arg 😢

Maybe that means this change is not possible?

jbedard: This diff is what solves this: https://github.com/bazel-contrib/bazel-gazelle/compare/24249c096f17345a49645f307e4a76908ef36963..f34ba22506f31a59cb305230a133aa32cf15d79e

walk/walk.go
```diff
@@ -384,22 +387,37 @@ func buildTrie(c *config.Config, updateRels *UpdateFilter, isIgnored isIgnoredFu
...
+		// Update the `validBuildFileNames` config while traversing the tree.
+		for _, d := range trie.build.Directives {
+			if d.Key == "build_file_name" {
+				validBuildFileNames = strings.Split(d.Value, ",")
+				break
+			}
+		}
+	}
+
 	for _, entry := range entries {
```
jbedard, line 421: With a bit more work we can now respect `# gazelle:exclude` here to skip any file/dirs that are excluded. Hopefully it would be an `isExcluded(entryPath)` alongside the `isIgnored(entryPath)` a few lines down within this loop.

jbedard: Here's a poc of that: https://github.com/jbedard/bazel-gazelle/compare/loadbuild-concurrent...jbedard:bazel-gazelle:fs-exclude

walk/walk.go
```diff
@@ -399,22 +402,37 @@ func buildTrie(c *config.Config, updateRels *UpdateFilter, isIgnored isIgnoredFu
...
 	if err != nil {
 		return err
 	}
 
+	trie.build, trie.buildFileErr = loadBuildFile(c, validBuildFileNames, rel, dir, entries)
+
+	if trie.build != nil {
+		// Update the `validBuildFileNames` config while traversing the tree.
+		for _, d := range trie.build.Directives {
+			if d.Key == "build_file_name" {
```
jbedard, line 429: This is an example of directives that effect the fs-walk. Those directives must be read while we actually do the concurrent fs-walk and later during `Configure`.

IMO this shows how some builtin "core options" are different then "language directives". The [poc moving exclude to fs-walk](https://github.com/bazel-contrib/bazel-gazelle/pull/1928/files#r1962409426) highlights this distinction even more.

jayconrod: This duplication of logic is unfortunate. It makes me wonder how much performance parsing BUILD files in parallel gives us and whether it's worth it. It should be an I/O and memory bound process, not CPU, but there's still probably some speedup there.

Instead of passing in `validBuildFileNames`, how about passing down `*walk.Configurer` as an argument and applying that to the configuration in each directory? Then we can use the real code. We should be able to get `*walk.Configurer` out of `c.Exts`. This requires more allocation (for cloning the `*config.Config` in each directory). Without a benchmark, it's hard to say what the performance impact is though.

jbedard: I have not properly measured the performance yet. Parsing concurrently is probably negligible, the I/O concurrently might make a minor difference, the only significant thing might be when `# gazelle:exclude`s exclude a significant amount of directories which will reduce CPU + I/O.

Performance aside... I think your idea [changing the trie to a concurrent cache](https://github.com/bazel-contrib/bazel-gazelle/issues/1891#issuecomment-2666792101) would require something similar to this PR? I was imagining that cache would essentially make all this `buildTrie` + `loadBuildFile` lazy+cached, but those two steps need to be merged together before they can be loaded+cached together...

jayconrod: Since we're not yet measuring performance, and the allocation cost is likely to be small, how about pulling in `*walkConfig` here? I see you have it in the `# gazelle:exclude` PoC. Let's not do all of that yet in this PR, but let's move in that direction instead of duplicating the configuration code.

jayconrod (commented):
Side note: as a reviewer, I'd appreciate if you avoid force-pushing the PR branch. GitHub handles this really poorly because comment threads are attached to commits, and old commits are lost.

jbedard:
> Side note: as a reviewer, I'd appreciate if you avoid force-pushing the PR branch. GitHub handles this really poorly because comment threads are attached to commits, and old commits are lost.

Sorry I'll try making fixup commits instead. I thought the github UI handled this decently now though, don't the comments stick around even when the commit is no longer in the PR?

walk/walk.go
jbedard: I agree `walkConfig` should move here. It will just require a bit of refactoring that I was hoping to avoid for now, I might try doing that in a separate PR that can land first... 🤔

jbedard: I think this is a good prefactor? https://github.com/bazel-contrib/bazel-gazelle/pull/2044

jbedard: > how about pulling in *walkConfig here

I've added a commit doing that, and rebased [the branch moving exclude](https://github.com/jbedard/bazel-gazelle/compare/loadbuild-concurrent...jbedard:bazel-gazelle:fs-exclude)

jayconrod:
> Sorry I'll try making fixup commits instead. I thought the github UI handled this decently now though, don't the comments stick around even when the commit is no longer in the PR?

Sometimes, not always. If I go to Conversations > Unresolved, there are a couple threads marked Outdated. If I click on one of those, it just takes me to an error page. I don't think there's a way to compare against the current branch or see what's changed since the last review since the old commit is gone.

jbedard:
I added more to this PR...
* Pulled latest
* Pulled #2043 in because it's approved and this PR can use that refactor

Important changes:
1. moved `isExcluded` into the fs traversal
2. moved the `resolveFileInfo` calls into the fs traversal
3. moved the `generation_mode` logic into the fs traversal

The `isExcluded` and `resolveFileInfo` being moved means both `exclude` and `follow` globs are now invoked concurrently, excluded files are never even put into the trie and excluded directories are never traversed at all.

Moving the `generation_mode` logic into fs traversal means the `visit` calls simply walk the trie and no longer has any logic about merging different trie nodes. Basically the trie is 1-1 with `Configure` + `GenerateRules` recursion now.

walk/config.go
```diff
@@ -78,14 +75,18 @@ type Configurer struct {
...
 
 	// Alternate BUILD read/write directories
 	readBuildFilesDir, writeBuildFilesDir string
 }
 
 func (wc *Configurer) RegisterFlags(fs *flag.FlagSet, cmd string, c *config.Config) {
+	c.Exts[walkConfigurerName] = wc
+
 	fs.Var(&gzflag.MultiFlag{Values: &wc.cliExcludes}, "exclude", "pattern that should be ignored (may be repeated)")
+	fs.Var(&gzflag.MultiFlag{Values: &wc.cliFollow}, "follow", "pattern that should be followed (may be repeated)")
```
jayconrod, line 89: Is there a reason this needs to be added as a command-line flag? This PR seems like it could be done by refactoring only without changing the CLI. If this stays in, then the CLI documentation in README.rst needs to be updated, too.

(If we didn't mind the incompatible change, I wish we could remove nearly all of the CLI flags. Directives are a more powerful, consistent mechanism. And it's better not to have two ways to do the same thing.)

walk/config.go
```diff
@@ -144,29 +146,57 @@ func (cr *Configurer) Configure(c *config.Config, rel string, f *rule.File) {
...
+type ignoreFilter struct {
+	isBazelIgnored         isIgnoredFunc
+	isRepoDirectoryIgnored isIgnoredFunc
 }
 
 type isIgnoredFunc = func(string) bool
 
 var nothingIgnored isIgnoredFunc = func(string) bool { return false }
 
+func newIgnoreFilter(repoRoot string) *ignoreFilter {
```
jayconrod, line 175: Optional: this comment mostly applies to pre-existing code.

`ignoreFilter` seems too indirect to me. It's not possible from reading the code of its constructor or methods to understand what it does. I think `loadBazelIgnore` and `loadRepoDirectoryIgnore` should return the data needed to fully implement `isDirectoryIgnored` and `isFileIgnored` instead of returning closures.

walk/walk.go
```diff
@@ -163,63 +155,48 @@ func visit(c *config.Config, cexts []config.Configurer, knownDirectives map[stri
...
+		regularFiles = append(regularFiles, entSub)
 	}
 
-	shouldUpdate := updateRels.shouldUpdate(rel, updateParent)
+	shouldUpdate := updateRels.shouldUpdate(trie.rel, updateParent)
+
+	// Ensure a deterministic order for regular files and subdirectories.
+	// This includes both the params to the callback as well as the `visit` order.
+	sort.Strings(regularFiles)
+	sort.Slice(trie.children, func(i, j int) bool {
```
jayconrod, line 176: Nit: `slices.Sort` would be a little more concise than `sort.Slice`.

walk/config.go
jbedard: Yeah I want to drop this, it was needed to make a test pass for some reason. I'll have to look again and figure out another way to do it...

walk/config.go
jayconrod: Sounds good, I'll copy this comment there.

walk/walk.go
```diff
@@ -137,13 +138,13 @@ func Walk(c *config.Config, cexts []config.Configurer, dirs []string, mode Mode,
...
+func visit(c *config.Config, cexts []config.Configurer, knownDirectives map[string]bool, updateRels *UpdateFilter, trie *pathTrie, wf WalkFunc, updateParent bool) ([]string, bool) {
 	haveError := false
 
 	// Absolute path to the directory being visited
-	dir := filepath.Join(c.RepoRoot, rel)
+	dir := filepath.Join(c.RepoRoot, trie.rel)
 
-	f, err := loadBuildFile(c, rel, dir, trie.files)
+	f, err := trie.build, trie.buildFileErr
 	if err != nil {
```
jbedard, line 148: This error logging moving to `buildTrie` would remove the need for the `trie.buildFileError` field and (imo) clean everything up, but that would be a minor change in behaviour that we can discuss later.

That would introduce the assumption that `c.Strict` is global (which it is atm) and not modifiable per-BUILD like other `c.*` fields :/

walk/walk.go
```diff
@@ -154,63 +155,48 @@ func visit(c *config.Config, cexts []config.Configurer, knownDirectives map[stri
...
 	}
 
-	shouldUpdate := updateRels.shouldUpdate(rel, updateParent)
+	shouldUpdate := updateRels.shouldUpdate(trie.rel, updateParent)
+
+	// Ensure a deterministic order for regular files and subdirectories.
+	// This includes both the params to the callback as well as the `visit` order.
+	slices.Sort(regularFiles)
+
+	// TODO: don't sort content of trie, pre-sort it or sort only a local copy
```
jbedard, line 177: IMO both of the `.Sort*` calls being here isn't ideal. It would be nicer if the `trie` struct already had all the nodes sorted and readonly, but the need for a `map` to store the path makes that more difficult atm. Maybe another thing that can be cleaned up later?

jayconrod:
Generally LGTM, but test failures need resolution.

jbedard:
I'm going to need help with that failing go test. Do you understand what that test is trying to accomplish? It is setup very odd, manually invoking `.Configure` in a non-standard way etc.

jayconrod:
I took a brief look at the test, sorry I don't have much time to debug further.

It looks like the test runs `walk.Walk` over `go/testdata`. It's simulating part of what `runFixUpdate` would do but only checking `GenerateRules`. When it visits a directory, it considers that to be a test if there's a `BUILD.want` file. If there are no tests, it fails at the top level, since there must be something broken in `walk`.

I added a quick log statement, it looks like `regularFiles` is empty on each visit, so no `BUILD.want` file.

jbedard:
Ok, I think I got it, see https://github.com/bazel-contrib/bazel-gazelle/pull/1928/commits/8d68e7b976f3c851e1f7642ef52f9082a76eb957.

Previously the test was loading a `rule.File` from a string (not from disk) and then invoking `.Configure` assuming it would pickup the directive and respect it. After moving symlink following into the fs-walk that is no longer true. That `.Configure` call was incorrect anyway imo, because it was followed by `walk.Walk` which would invoke `.Configure` again.

jbedard:
However I think after making that test more realistic it is having other side-effects, so maybe we're not done yet. I don't have any failures locally anymore though? 🤔

jayconrod:
The change to `go/generate_test.go` seems reasonable to me. It shouldn't cause other test failures.

For most of the remaining test errors, it looks like when Gazelle is run as part of `go_repository`, it's failing to generate `BUILD` files in some directories. Maybe a race or some other failure? That would be unrelated to test changes. It's run as a repository rule, so try `bazel clean --expunge` first to make sure those get re-evaluated.

There's an error on Windows that's more verbose. It looks like there's a nil pointer dereference somewhere:

<details><pre>

(01:29:34) ERROR: An error occurred during the fetch of repository 'gazelle~override~go_deps~org_golang_x_tools':
--
  | Traceback (most recent call last):
  | File "C:/b/7fgneixv/external/gazelle~override/internal/go_repository.bzl", line 360, column 17, in _go_repository_impl
  | fail("failed to generate BUILD files for %s: %s" % (
  | Error in fail: failed to generate BUILD files for golang.org/x/tools: gazelle: C:\b\7fgneixv\external\gazelle~override~go_deps~org_golang_x_tools\cmd\fiximports\testdata\src\old.com\bad\bad.go: error reading go file: C:\b\7fgneixv\external\gazelle~override~go_deps~org_golang_x_tools\cmd\fiximports\testdata\src\old.com\bad\bad.go:2:43: expected 'package', found 'EOF'
  | gazelle: found packages testdata (callgraph_comma_maps.go) and test (issue63146.go) in C:\b\7fgneixv\external\gazelle~override~go_deps~org_golang_x_tools\go\callgraph\vta\testdata\src
  | gazelle: C:\b\7fgneixv\external\gazelle~override~go_deps~org_golang_x_tools\go\loader\testdata\badpkgdecl.go: error reading go file: C:\b\7fgneixv\external\gazelle~override~go_deps~org_golang_x_tools\go\loader\testdata\badpkgdecl.go:1:34: expected 'package', found 'EOF'
  | gazelle: C:\b\7fgneixv\external\gazelle~override~go_deps~org_golang_x_tools\go\packages\packagestest\testdata\groups\two\modules\example.com\extra\geez\help.go: error reading go file: C:\b\7fgneixv\external\gazelle~override~go_deps~org_golang_x_tools\go\packages\packagestest\testdata\groups\two\modules\example.com\extra\geez\help.go:1:16: expected ';', found '.'
  | gazelle: C:\b\7fgneixv\external\gazelle~override~go_deps~org_golang_x_tools\go\packages\packagestest\testdata\groups\two\modules\example.com\extra\v2\geez\help.go: error reading go file: C:\b\7fgneixv\external\gazelle~override~go_deps~org_golang_x_tools\go\packages\packagestest\testdata\groups\two\modules\example.com\extra\v2\geez\help.go:1:16: expected ';', found '.'
  | gazelle: C:\b\7fgneixv\external\gazelle~override~go_deps~org_golang_x_tools\go\packages\packagestest\testdata\groups\two\modules\example.com\extra\v2\me.go: error reading go file: C:\b\7fgneixv\external\gazelle~override~go_deps~org_golang_x_tools\go\packages\packagestest\testdata\groups\two\modules\example.com\extra\v2\me.go:1:16: expected ';', found '.'
  | gazelle: C:\b\7fgneixv\external\gazelle~override~go_deps~org_golang_x_tools\go\packages\packagestest\testdata\groups\two\modules\example.com\extra\yo.go: error reading go file: C:\b\7fgneixv\external\gazelle~override~go_deps~org_golang_x_tools\go\packages\packagestest\testdata\groups\two\modules\example.com\extra\yo.go:1:16: expected ';', found '.'
  | gazelle: C:\b\7fgneixv\external\gazelle~override~go_deps~org_golang_x_tools\go\packages\packagestest\testdata\groups\two\modules\example.com\tempmod\main.go: error reading go file: C:\b\7fgneixv\external\gazelle~override~go_deps~org_golang_x_tools\go\packages\packagestest\testdata\groups\two\modules\example.com\tempmod\main.go:1:16: expected ';', found '.'
  | gazelle: C:\b\7fgneixv\external\gazelle~override~go_deps~org_golang_x_tools\go\packages\packagestest\testdata\groups\two\modules\example.com\what@v1.0.0\main.go: error reading go file: C:\b\7fgneixv\external\gazelle~override~go_deps~org_golang_x_tools\go\packages\packagestest\testdata\groups\two\modules\example.com\what@v1.0.0\main.go:1:16: expected ';', found '.'
  | gazelle: C:\b\7fgneixv\external\gazelle~override~go_deps~org_golang_x_tools\go\packages\packagestest\testdata\groups\two\modules\example.com\what@v1.1.0\main.go: error reading go file: C:\b\7fgneixv\external\gazelle~override~go_deps~org_golang_x_tools\go\packages\packagestest\testdata\groups\two\modules\example.com\what@v1.1.0\main.go:1:16: expected ';', found '.'
  | panic: runtime error: invalid memory address or nil pointer dereference
  | [signal 0xc0000005 code=0x0 addr=0x8 pc=0x52bf12]
  |  
  | goroutine 1 [running]:
  | github.com/bazelbuild/bazel-gazelle/walk.visit.func1(0xc0004d7ec0?, 0xc00070b620?)
  | src/github.com/bazelbuild/bazel-gazelle/walk/walk.go:181 +0x12
  | slices.partitionCmpFunc[...]({0xc000744080?, 0xf, 0x0}, 0x0?, 0x0, 0x0?, 0x6a2ca0?)
  | C:/b/7fgneixv/external/rules_go~0.50.1~go_sdk~gazelle_bcr_go_mod_tests__download_0/src/slices/zsortanyfunc.go:157 +0x2e2
  | slices.pdqsortCmpFunc[...]({0xc000744080, 0xf, 0x10}, 0xc00101bd40, 0xc00001f380?, 0x1a?, 0x6a2ca0?)
  | C:/b/7fgneixv/external/rules_go~0.50.1~go_sdk~gazelle_bcr_go_mod_tests__download_0/src/slices/zsortanyfunc.go:114 +0x2f8
  | slices.SortFunc[...](...)
  | C:/b/7fgneixv/external/rules_go~0.50.1~go_sdk~gazelle_bcr_go_mod_tests__download_0/src/slices/sort.go:32
  | github.com/bazelbuild/bazel-gazelle/walk.visit(0xc00101bd40, {0xc0001a0120, 0x6, 0x6}, 0xc00073f2f8, 0xc0001136c0, 0xc00054ce40, 0xc00073fdf0, 0x1)
  | src/github.com/bazelbuild/bazel-gazelle/walk/walk.go:180 +0x4ef
  | github.com/bazelbuild/bazel-gazelle/walk.visit(0xc0006f5930, {0xc0001a0120, 0x6, 0x6}, 0xc00073f2f8, 0xc0001136c0, 0xc00054cd80, 0xc00073fdf0, 0x1)
  | src/github.com/bazelbuild/bazel-gazelle/walk/walk.go:193 +0x81b
  | github.com/bazelbuild/bazel-gazelle/walk.visit(0xc0006f56c0, {0xc0001a0120, 0x6, 0x6}, 0xc00073f2f8, 0xc0001136c0, 0xc00054ccc0, 0xc00073fdf0, 0x1)
  | src/github.com/bazelbuild/bazel-gazelle/walk/walk.go:193 +0x81b
  | github.com/bazelbuild/bazel-gazelle/walk.visit(0xc0006f5520, {0xc0001a0120, 0x6, 0x6}, 0xc00073f2f8, 0xc0001136c0, 0xc000697b00, 0xc00073fdf0, 0x1)
  | src/github.com/bazelbuild/bazel-gazelle/walk/walk.go:193 +0x81b
  | github.com/bazelbuild/bazel-gazelle/walk.visit(0xc00011d450, {0xc0001a0120, 0x6, 0x6}, 0xc00073f2f8, 0xc0001136c0, 0xc000430060, 0xc00073fdf0, 0x1)
  | src/github.com/bazelbuild/bazel-gazelle/walk/walk.go:193 +0x81b
  | github.com/bazelbuild/bazel-gazelle/walk.visit(0xc00011cf70, {0xc0001a0120, 0x6, 0x6}, 0xc00073f2f8, 0xc0001136c0, 0xc0001a01e0, 0xc00073fdf0, 0x0)
  | src/github.com/bazelbuild/bazel-gazelle/walk/walk.go:193 +0x81b
  | github.com/bazelbuild/bazel-gazelle/walk.Walk(0xc00011cf70, {0xc0001a0120, 0x6, 0x6}, {0xc000112eb0, 0x1, 0x1}, 0x0, 0xc000459df0)
  | src/github.com/bazelbuild/bazel-gazelle/walk/walk.go:129 +0x21c
  | main.runFixUpdate({0xc000148230, 0x42}, 0x0, {0xc00014e010, 0x12, 0x1f})
  | src/github.com/bazelbuild/bazel-gazelle/cmd/gazelle/fix-update.go:317 +0xcf4
  | main.run({0xc000148230?, 0x42?}, {0xc00014e010?, 0x12?, 0x1f?})
  | src/github.com/bazelbuild/bazel-gazelle/cmd/gazelle/main.go:95 +0x1dc
  | main.main()
  | src/github.com/bazelbuild/bazel-gazelle/cmd/gazelle/main.go:72 +0xf5
  | (01:29:34) ERROR: <builtin>: fetching go_repository rule //:gazelle~override~go_deps~org_golang_x_tools: Traceback (most recent call last):
  | File "C:/b/7fgneixv/external/gazelle~override/internal/go_repository.bzl", line 360, column 17, in _go_repository_impl
  | fail("failed to generate BUILD files for %s: %s" % (
  | Error in fail: failed to generate BUILD files for golang.org/x/tools: gazelle: C:\b\7fgneixv\external\gazelle~override~go_deps~org_golang_x_tools\cmd\fiximports\testdata\src\old.com\bad\bad.go: error reading go file: C:\b\7fgneixv\external\gazelle~override~go_deps~org_golang_x_tools\cmd\fiximports\testdata\src\old.com\bad\bad.go:2:43: expected 'package', found 'EOF'
  | gazelle: found packages testdata (callgraph_comma_maps.go) and test (issue63146.go) in C:\b\7fgneixv\external\gazelle~override~go_deps~org_golang_x_tools\go\callgraph\vta\testdata\src
  | gazelle: C:\b\7fgneixv\external\gazelle~override~go_deps~org_golang_x_tools\go\loader\testdata\badpkgdecl.go: error reading go file: C:\b\7fgneixv\external\gazelle~override~go_deps~org_golang_x_tools\go\loader\testdata\badpkgdecl.go:1:34: expected 'package', found 'EOF'
  | gazelle: C:\b\7fgneixv\external\gazelle~override~go_deps~org_golang_x_tools\go\packages\packagestest\testdata\groups\two\modules\example.com\extra\geez\help.go: error reading go file: C:\b\7fgneixv\external\gazelle~override~go_deps~org_golang_x_tools\go\packages\packagestest\testdata\groups\two\modules\example.com\extra\geez\help.go:1:16: expected ';', found '.'
  | gazelle: C:\b\7fgneixv\external\gazelle~override~go_deps~org_golang_x_tools\go\packages\packagestest\testdata\groups\two\modules\example.com\extra\v2\geez\help.go: error reading go file: C:\b\7fgneixv\external\gazelle~override~go_deps~org_golang_x_tools\go\packages\packagestest\testdata\groups\two\modules\example.com\extra\v2\geez\help.go:1:16: expected ';', found '.'
  | gazelle: C:\b\7fgneixv\external\gazelle~override~go_deps~org_golang_x_tools\go\packages\packagestest\testdata\groups\two\modules\example.com\extra\v2\me.go: error reading go file: C:\b\7fgneixv\external\gazelle~override~go_deps~org_golang_x_tools\go\packages\packagestest\testdata\groups\two\modules\example.com\extra\v2\me.go:1:16: expected ';', found '.'
  | gazelle: C:\b\7fgneixv\external\gazelle~override~go_deps~org_golang_x_tools\go\packages\packagestest\testdata\groups\two\modules\example.com\extra\yo.go: error reading go file: C:\b\7fgneixv\external\gazelle~override~go_deps~org_golang_x_tools\go\packages\packagestest\testdata\groups\two\modules\example.com\extra\yo.go:1:16: expected ';', found '.'
  | gazelle: C:\b\7fgneixv\external\gazelle~override~go_deps~org_golang_x_tools\go\packages\packagestest\testdata\groups\two\modules\example.com\tempmod\main.go: error reading go file: C:\b\7fgneixv\external\gazelle~override~go_deps~org_golang_x_tools\go\packages\packagestest\testdata\groups\two\modules\example.com\tempmod\main.go:1:16: expected ';', found '.'
  | gazelle: C:\b\7fgneixv\external\gazelle~override~go_deps~org_golang_x_tools\go\packages\packagestest\testdata\groups\two\modules\example.com\what@v1.0.0\main.go: error reading go file: C:\b\7fgneixv\external\gazelle~override~go_deps~org_golang_x_tools\go\packages\packagestest\testdata\groups\two\modules\example.com\what@v1.0.0\main.go:1:16: expected ';', found '.'
  | gazelle: C:\b\7fgneixv\external\gazelle~override~go_deps~org_golang_x_tools\go\packages\packagestest\testdata\groups\two\modules\example.com\what@v1.1.0\main.go: error reading go file: C:\b\7fgneixv\external\gazelle~override~go_deps~org_golang_x_tools\go\packages\packagestest\testdata\groups\two\modules\example.com\what@v1.1.0\main.go:1:16: expected ';', found '.'
  | panic: runtime error: invalid memory address or nil pointer dereference
  | [signal 0xc0000005 code=0x0 addr=0x8 pc=0x52bf12]
  |  
  | goroutine 1 [running]:
  | github.com/bazelbuild/bazel-gazelle/walk.visit.func1(0xc0004d7ec0?, 0xc00070b620?)
  | src/github.com/bazelbuild/bazel-gazelle/walk/walk.go:181 +0x12
  | slices.partitionCmpFunc[...]({0xc000744080?, 0xf, 0x0}, 0x0?, 0x0, 0x0?, 0x6a2ca0?)
  | C:/b/7fgneixv/external/rules_go~0.50.1~go_sdk~gazelle_bcr_go_mod_tests__download_0/src/slices/zsortanyfunc.go:157 +0x2e2
  | slices.pdqsortCmpFunc[...]({0xc000744080, 0xf, 0x10}, 0xc00101bd40, 0xc00001f380?, 0x1a?, 0x6a2ca0?)
  | C:/b/7fgneixv/external/rules_go~0.50.1~go_sdk~gazelle_bcr_go_mod_tests__download_0/src/slices/zsortanyfunc.go:114 +0x2f8
  | slices.SortFunc[...](...)
  | C:/b/7fgneixv/external/rules_go~0.50.1~go_sdk~gazelle_bcr_go_mod_tests__download_0/src/slices/sort.go:32
  | github.com/bazelbuild/bazel-gazelle/walk.visit(0xc00101bd40, {0xc0001a0120, 0x6, 0x6}, 0xc00073f2f8, 0xc0001136c0, 0xc00054ce40, 0xc00073fdf0, 0x1)
  | src/github.com/bazelbuild/bazel-gazelle/walk/walk.go:180 +0x4ef
  | github.com/bazelbuild/bazel-gazelle/walk.visit(0xc0006f5930, {0xc0001a0120, 0x6, 0x6}, 0xc00073f2f8, 0xc0001136c0, 0xc00054cd80, 0xc00073fdf0, 0x1)
  | src/github.com/bazelbuild/bazel-gazelle/walk/walk.go:193 +0x81b
  | github.com/bazelbuild/bazel-gazelle/walk.visit(0xc0006f56c0, {0xc0001a0120, 0x6, 0x6}, 0xc00073f2f8, 0xc0001136c0, 0xc00054ccc0, 0xc00073fdf0, 0x1)
  | src/github.com/bazelbuild/bazel-gazelle/walk/walk.go:193 +0x81b
  | github.com/bazelbuild/bazel-gazelle/walk.visit(0xc0006f5520, {0xc0001a0120, 0x6, 0x6}, 0xc00073f2f8, 0xc0001136c0, 0xc000697b00, 0xc00073fdf0, 0x1)
  | src/github.com/bazelbuild/bazel-gazelle/walk/walk.go:193 +0x81b
  | github.com/bazelbuild/bazel-gazelle/walk.visit(0xc00011d450, {0xc0001a0120, 0x6, 0x6}, 0xc00073f2f8, 0xc0001136c0, 0xc000430060, 0xc00073fdf0, 0x1)
  | src/github.com/bazelbuild/bazel-gazelle/walk/walk.go:193 +0x81b
  | github.com/bazelbuild/bazel-gazelle/walk.visit(0xc00011cf70, {0xc0001a0120, 0x6, 0x6}, 0xc00073f2f8, 0xc0001136c0, 0xc0001a01e0, 0xc00073fdf0, 0x0)
  | src/github.com/bazelbuild/bazel-gazelle/walk/walk.go:193 +0x81b
  | github.com/bazelbuild/bazel-gazelle/walk.Walk(0xc00011cf70, {0xc0001a0120, 0x6, 0x6}, {0xc000112eb0, 0x1, 0x1}, 0x0, 0xc000459df0)
  | src/github.com/bazelbuild/bazel-gazelle/walk/walk.go:129 +0x21c
  | main.runFixUpdate({0xc000148230, 0x42}, 0x0, {0xc00014e010, 0x12, 0x1f})
  | src/github.com/bazelbuild/bazel-gazelle/cmd/gazelle/fix-update.go:317 +0xcf4
  | main.run({0xc000148230?, 0x42?}, {0xc00014e010?, 0x12?, 0x1f?})
  | src/github.com/bazelbuild/bazel-gazelle/cmd/gazelle/main.go:95 +0x1dc
  | main.main()
  | src/github.com/bazelbuild/bazel-gazelle/cmd/gazelle/main.go:72 +0xf5
  | (01:29:34) ERROR: C:/b/7fgneixv/external/gazelle~override~go_deps~com_github_bazelbuild_buildtools/build/BUILD.bazel:5:8: @gazelle~override~go_deps~com_github_bazelbuild_buildtools//build:parse.y.go_yacc depends on @gazelle~override~go_deps~org_golang_x_tools//cmd/goyacc:goyacc in repository @gazelle~override~go_deps~org_golang_x_tools which failed to fetch. no such package '@gazelle~override~go_deps~org_golang_x_tools//cmd/goyacc': failed to generate BUILD files for golang.org/x/tools: gazelle: C:\b\7fgneixv\external\gazelle~override~go_deps~org_golang_x_tools\cmd\fiximports\testdata\src\old.com\bad\bad.go: error reading go file: C:\b\7fgneixv\external\gazelle~override~go_deps~org_golang_x_tools\cmd\fiximports\testdata\src\old.com\bad\bad.go:2:43: expected 'package', found 'EOF'
  | gazelle: found packages testdata (callgraph_comma_maps.go) and test (issue63146.go) in C:\b\7fgneixv\external\gazelle~override~go_deps~org_golang_x_tools\go\callgraph\vta\testdata\src
  | gazelle: C:\b\7fgneixv\external\gazelle~override~go_deps~org_golang_x_tools\go\loader\testdata\badpkgdecl.go: error reading go file: C:\b\7fgneixv\external\gazelle~override~go_deps~org_golang_x_tools\go\loader\testdata\badpkgdecl.go:1:34: expected 'package', found 'EOF'
  | gazelle: C:\b\7fgneixv\external\gazelle~override~go_deps~org_golang_x_tools\go\packages\packagestest\testdata\groups\two\modules\example.com\extra\geez\help.go: error reading go file: C:\b\7fgneixv\external\gazelle~override~go_deps~org_golang_x_tools\go\packages\packagestest\testdata\groups\two\modules\example.com\extra\geez\help.go:1:16: expected ';', found '.'
  | gazelle: C:\b\7fgneixv\external\gazelle~override~go_deps~org_golang_x_tools\go\packages\packagestest\testdata\groups\two\modules\example.com\extra\v2\geez\help.go: error reading go file: C:\b\7fgneixv\external\gazelle~override~go_deps~org_golang_x_tools\go\packages\packagestest\testdata\groups\two\modules\example.com\extra\v2\geez\help.go:1:16: expected ';', found '.'
  | gazelle: C:\b\7fgneixv\external\gazelle~override~go_deps~org_golang_x_tools\go\packages\packagestest\testdata\groups\two\modules\example.com\extra\v2\me.go: error reading go file: C:\b\7fgneixv\external\gazelle~override~go_deps~org_golang_x_tools\go\packages\packagestest\testdata\groups\two\modules\example.com\extra\v2\me.go:1:16: expected ';', found '.'
  | gazelle: C:\b\7fgneixv\external\gazelle~override~go_deps~org_golang_x_tools\go\packages\packagestest\testdata\groups\two\modules\example.com\extra\yo.go: error reading go file: C:\b\7fgneixv\external\gazelle~override~go_deps~org_golang_x_tools\go\packages\packagestest\testdata\groups\two\modules\example.com\extra\yo.go:1:16: expected ';', found '.'
  | gazelle: C:\b\7fgneixv\external\gazelle~override~go_deps~org_golang_x_tools\go\packages\packagestest\testdata\groups\two\modules\example.com\tempmod\main.go: error reading go file: C:\b\7fgneixv\external\gazelle~override~go_deps~org_golang_x_tools\go\packages\packagestest\testdata\groups\two\modules\example.com\tempmod\main.go:1:16: expected ';', found '.'
  | gazelle: C:\b\7fgneixv\external\gazelle~override~go_deps~org_golang_x_tools\go\packages\packagestest\testdata\groups\two\modules\example.com\what@v1.0.0\main.go: error reading go file: C:\b\7fgneixv\external\gazelle~override~go_deps~org_golang_x_tools\go\packages\packagestest\testdata\groups\two\modules\example.com\what@v1.0.0\main.go:1:16: expected ';', found '.'
  | gazelle: C:\b\7fgneixv\external\gazelle~override~go_deps~org_golang_x_tools\go\packages\packagestest\testdata\groups\two\modules\example.com\what@v1.1.0\main.go: error reading go file: C:\b\7fgneixv\external\gazelle~override~go_deps~org_golang_x_tools\go\packages\packagestest\testdata\groups\two\modules\example.com\what@v1.1.0\main.go:1:16: expected ';', found '.'
  | panic: runtime error: invalid memory address or nil pointer dereference
  | [signal 0xc0000005 code=0x0 addr=0x8 pc=0x52bf12]
  |  
  | goroutine 1 [running]:
  | github.com/bazelbuild/bazel-gazelle/walk.visit.func1(0xc0004d7ec0?, 0xc00070b620?)
  | src/github.com/bazelbuild/bazel-gazelle/walk/walk.go:181 +0x12
  | slices.partitionCmpFunc[...]({0xc000744080?, 0xf, 0x0}, 0x0?, 0x0, 0x0?, 0x6a2ca0?)
  | C:/b/7fgneixv/external/rules_go~0.50.1~go_sdk~gazelle_bcr_go_mod_tests__download_0/src/slices/zsortanyfunc.go:157 +0x2e2
  | slices.pdqsortCmpFunc[...]({0xc000744080, 0xf, 0x10}, 0xc00101bd40, 0xc00001f380?, 0x1a?, 0x6a2ca0?)
  | C:/b/7fgneixv/external/rules_go~0.50.1~go_sdk~gazelle_bcr_go_mod_tests__download_0/src/slices/zsortanyfunc.go:114 +0x2f8
  | slices.SortFunc[...](...)
  | C:/b/7fgneixv/external/rules_go~0.50.1~go_sdk~gazelle_bcr_go_mod_tests__download_0/src/slices/sort.go:32
  | github.com/bazelbuild/bazel-gazelle/walk.visit(0xc00101bd40, {0xc0001a0120, 0x6, 0x6}, 0xc00073f2f8, 0xc0001136c0, 0xc00054ce40, 0xc00073fdf0, 0x1)
  | src/github.com/bazelbuild/bazel-gazelle/walk/walk.go:180 +0x4ef
  | github.com/bazelbuild/bazel-gazelle/walk.visit(0xc0006f5930, {0xc0001a0120, 0x6, 0x6}, 0xc00073f2f8, 0xc0001136c0, 0xc00054cd80, 0xc00073fdf0, 0x1)
  | src/github.com/bazelbuild/bazel-gazelle/walk/walk.go:193 +0x81b
  | github.com/bazelbuild/bazel-gazelle/walk.visit(0xc0006f56c0, {0xc0001a0120, 0x6, 0x6}, 0xc00073f2f8, 0xc0001136c0, 0xc00054ccc0, 0xc00073fdf0, 0x1)
  | src/github.com/bazelbuild/bazel-gazelle/walk/walk.go:193 +0x81b
  | github.com/bazelbuild/bazel-gazelle/walk.visit(0xc0006f5520, {0xc0001a0120, 0x6, 0x6}, 0xc00073f2f8, 0xc0001136c0, 0xc000697b00, 0xc00073fdf0, 0x1)
  | src/github.com/bazelbuild/bazel-gazelle/walk/walk.go:193 +0x81b
  | github.com/bazelbuild/bazel-gazelle/walk.visit(0xc00011d450, {0xc0001a0120, 0x6, 0x6}, 0xc00073f2f8, 0xc0001136c0, 0xc000430060, 0xc00073fdf0, 0x1)
  | src/github.com/bazelbuild/bazel-gazelle/walk/walk.go:193 +0x81b
  | github.com/bazelbuild/bazel-gazelle/walk.visit(0xc00011cf70, {0xc0001a0120, 0x6, 0x6}, 0xc00073f2f8, 0xc0001136c0, 0xc0001a01e0, 0xc00073fdf0, 0x0)
  | src/github.com/bazelbuild/bazel-gazelle/walk/walk.go:193 +0x81b
  | github.com/bazelbuild/bazel-gazelle/walk.Walk(0xc00011cf70, {0xc0001a0120, 0x6, 0x6}, {0xc000112eb0, 0x1, 0x1}, 0x0, 0xc000459df0)
  | src/github.com/bazelbuild/bazel-gazelle/walk/walk.go:129 +0x21c
  | main.runFixUpdate({0xc000148230, 0x42}, 0x0, {0xc00014e010, 0x12, 0x1f})
  | src/github.com/bazelbuild/bazel-gazelle/cmd/gazelle/fix-update.go:317 +0xcf4
  | main.run({0xc000148230?, 0x42?}, {0xc00014e010?, 0x12?, 0x1f?})
  | src/github.com/bazelbuild/bazel-gazelle/cmd/gazelle/main.go:95 +0x1dc
  | main.main()
  | src/github.com/bazelbuild/bazel-gazelle/cmd/gazelle/main.go:72 +0xf5
  | (01:29:35) ERROR: Analysis of target '//:gazelle_test-runner' failed; build aborted:
  | (01:29:35) INFO: Elapsed time: 28.830s
  | (01:29:35) INFO: 0 processes.
  | (01:29:35) FAILED: Build did NOT complete successfully (263 packages loaded, 20656 targets configured)
  | Fetching repository @gazelle~override~_go_deps~~~go_tools~com_github_google_uuid; starting
  | Fetching repository @gazelle~override~_go_deps~~~go_tools~com_github_xrash_smetrics; starting
  | Fetching repository @gazelle~override~_go_deps~~~go_tools~com_github_cpuguy83_go_md2man_v2; starting
  | Fetching repository @gazelle~override~_go_deps~~~go_tools~com_github_vektah_gqlparser_v2; starting
  | Fetching repository @gazelle~override~_go_deps~~~go_tools~org_golang_x_text; starting
  | Fetching repository @gazelle~override~_go_deps~~~go_tools~org_golang_x_sys; starting
  | Fetching repository @gazelle~override~_go_deps~~~go_tools~org_golang_x_mod; starting
  | Fetching repository @gazelle~override~go_deps~in_gopkg_yaml_v2; starting
  | bazel build failed with exit code 1
  | 🚨 Error: The command exited with status 1
  | user command error: exit status 1

</pre></details>

walk/walk.go
```diff
@@ -366,28 +338,72 @@ func shouldFollow(wc *walkConfig, dir, rel string, ent fs.DirEntry) bool {
...
-	}
+func (trie *pathTrie) addChild(c *pathTrie) *pathTrie {
+	trie.rw.Lock()
+	defer trie.rw.Unlock()
+
+	trie.children = append(trie.children, c)
+	return c
+}
+
+func (trie *pathTrie) addFiles(p []string) {
```
jbedard, line 380: Now that the `updateOnly` logic is part of the concurrent fs-traversal the construction of `trie.{children,files}` may cross goroutines and needs these lock util methods.

Note that this one adding files accepts an array so we lock once per directory, not once per file within the directory.

jbedard:
Ok, I think this is finally ready to go. Thanks @mattem for testing this out in a larger repo and helping with some of the concurrency challenges.

Just to summarize things again...

Main change: BUILDs are loaded+parsed concurrently while traversing the fs

This allows some directives to be processed during that concurrent traversal as well, including:
* `# gazelle:follow` glob evaluations are now evaluated during the concurrent fs traversal
* `# gazelle:exclude` glob evaluations are now evaluated during the concurrent fs traversal
* ^ excluded and non-followed files/directories are no longer walked or persisted at all in `pathTrie`
* the `# gazelle:follow` `os.Stat` calls are invoked during the concurrent fs traversal
* the `# gazelle:generation_mode` can now be taken into account when building the `pathTrie` instead of while traversing it. This makes the traversal `visit` code simpler since it now has less data that needs modifying/restructured.

walk/config.go
```diff
@@ -108,33 +107,42 @@ func (wc *Configurer) CheckFlags(_ *flag.FlagSet, c *config.Config) error {
...
+
+	wc := &walkConfig{
+		// Always false by default.
+		ignore: false,
+
+		// Ineritied, potentially overwritten with directives.
+		updateOnly:          wcp.updateOnly,
+		validBuildFileNames: wcp.validBuildFileNames,
+
+		// Will potentially be extended with directives, must fully clone.
```
jayconrod, line 130: Optimization nit: you don't need to fully clone these. `wcp.excludes[:len(wcp.excludes):len(wcp.excludes)]` creates a slice that shares the same memory, but the capacity equals the length, so if anything's appended, it's guaranteed to copy the elements into a new backing array. Copy-on-append basically.

jayconrod (commented):
Quick review pass only, no real comments yet. I'd like some more time to pick through and think about how this will work with lazy indexing. (I expected to quickly review, but this has changed a lot since I last looked).

To be honest, I'm struggling to see `pathTrie` as the right data structure. I don't think there's a strong need to keep track of child nodes. It seems to add a lot of synchronization complexity with both a lock and wait group per directory.

The `cache` pattern seems better to me: store everything in a concurrent map for simpler concurrency and more random access. It's easy to populate with a concurrent walk but also possible to populate later on with `RelsToVisit`.

walk/config.go
jbedard: Great 👍

jbedard:
> I expected to quickly review, but this has changed a lot since I last looked

Yeah this has grown far beyond what I thought it would 😢 

I might try to pull some of it out but moving the BUILD parsing into the concurrent fs walk brought in a **lot** more then I thought it would.

> To be honest, I'm struggling to see pathTrie as the right data structure.

As long as we keep the concurrency for all I/O and things like BUILD-parsing, glob evaluation etc.

> The cache pattern seems better to me: store everything in a concurrent map for simpler concurrency and more random access. It's easy to populate with a concurrent walk but also possible to populate later on with RelsToVisit

Can we do that now without waiting for the full `RelsToVisit` implementation? Honestly I think most of this PR will be required for that anyway, even if put into a brand new struct 🤷

jbedard:
FWIW in one very large repo this is PR is cutting a fairly significant 25% off the wall time. That more then I predicted!

I think it's almost all because the `# gazelle:excludes` and [gitignore](https://github.com/bazel-contrib/bazel-gazelle/pull/1908) are now done concurrently as part of the fs-walk (yes I still have that gitignore patch which will probably be needed until the REPO `ignore_directories` is more widely used).

This repo has ~1500 BUILD files, the entire repo uses `generation_mode: update_only` so the number of directories is much higher around 30k.

With gazelle at HEAD (4 runs with no changed BUILDs):
```
  67.57s user 14.22s system 388% cpu 21.038 total
  69.18s user 28.45s system 486% cpu 20.066 total
  69.54s user 31.01s system 499% cpu 20.116 total
  69.38s user 31.45s system 498% cpu 20.222 total
```

With this PR:
```
  68.51s user 6.29s system 460% cpu 16.227 total
  68.61s user 5.61s system 494% cpu 14.996 total
  69.27s user 5.39s system 508% cpu 14.697 total
  68.65s user 6.59s system 478% cpu 15.717 total
```

jayconrod:
> As long as we keep the concurrency for all I/O and things like BUILD-parsing, glob evaluation etc.

Agreed, I'll make sure to preserve that.

> Can we do that now without waiting for the full RelsToVisit implementation? Honestly I think most of this PR will be required for that anyway, even if put into a brand new struct 🤷

My plan is to do that `cache` refactoring as a separate PR from the `RelsToVisit` implementation, though I'll probably hold off on merging most of that PR stack until I have some feedback that it's working as intended.

> FWIW in one very large repo this is PR is cutting a fairly significant 25% off the wall time. That more then I predicted!

Very impressive! We definitely need to speed up Gazelle for repos like this. I'm glad this helps so much.

------

I've had some more time to review, and this looks good. I think we should merge this now, and the `RelsToVisit`-related changes can come later. Approving now, but tagging @linzhp and @fmeum too in case they have more thoughts.

Some changes may be needed for `RelsToVisit`, but I don't want you all to wait for me being slow: I'm traveling this week and won't be able to work on it until next week. The performance improvements here are really nice, so I'll be careful to preserve the new behavior.

jayconrod (approved):

jbedard:
I guess this was replaced by https://github.com/bazel-contrib/bazel-gazelle/pull/2073

PR #1931, opened 2024-09-22
refactor: change pathTrie per-file to per-directory

jbedard:
Now the `pathTrie` aligns with the directory tree, not full fs tree which includes files. Aligning with the directory tree makes more sense to me, and I think will make it simpler to do things like load BUILDs concurrently while doing the fs-traversal (and persisting that in the `pathTrie`) etc.

**What type of PR is this?**

> Other

> all

Performance + simplifying things for future changes.

### DIFF
--- a/walk/walk.go
+++ b/walk/walk.go
@@ -23,7 +23,6 @@ import (
 	"os"
 	"path"
 	"path/filepath"
-	"sort"
 	"strings"
 
 	"github.com/bazelbuild/bazel-gazelle/config"
@@ -398,12 +397,6 @@ func walkDir(root, rel string, eg *errgroup.Group, limitCh chan struct{}, isIgno
 	if err != nil {
 		return err
 	}
-
-	// Sort trie entries once before processing to ensure
-	// a consistent order of traversal and deterministic output.
-	sort.Slice(entries, func(i, j int) bool {
-		return entries[i].Name() < entries[j].Name()
-	})
 
 	for _, entry := range entries {
 		entryName := entry.Name()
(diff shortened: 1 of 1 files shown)


walk/walk.go
```diff
@@ -165,28 +156,38 @@ func visit(c *config.Config, cexts []config.Configurer, knownDirectives map[stri
...
-		ent := resolveFileInfo(wc, dir, entRel, ent)
-		switch {
-		case ent == nil:
+		if ent := resolveFileInfo(wc, dir, entRel, ent); ent != nil {
+			regularFiles = append(regularFiles, base)
+		}
+	}
+
+	// Filter and collect subdirectories
+	var subdirs []string
```
fmeum, line 173: Do we need this separate slice? I think that it just contains `t.entry.Name()` for ever subtrie.

walk/walk.go
```diff
@@ -397,6 +399,12 @@ func walkDir(root, rel string, eg *errgroup.Group, limitCh chan struct{}, isIgno
 		return err
 	}
 
+	// Sort trie entries once before processing to ensure
```
fmeum, line 402: This should be unnecessary as `os.ReadDir` sorts the entries: https://cs.opensource.google/go/go/+/refs/tags/go1.23.2:src/os/dir.go;l=127

jbedard: I was just moving it to be done while iterating over the `ReadDir` instead of done later in `visit`. But if we can remove it then even better 👍

walk/walk.go
jbedard: This array gets passed into the `WalkFunc` API.

Although we can move construction of this into the `if updateRels.shouldCall` block. Or can merge this loop and the next so we only have to construct 1/2 of these arrays...

jbedard: See the fixup commit

fmeum (approved):

jbedard:
Windows is broken? 😒

fmeum:
Fixed it on master and rebased

PR #1932, opened 2024-09-22
refactor: simplify walk configuration, func signatures

jbedard:
**What type of PR is this?**

> Other

> all

Minor refactoring which will make further improvements easier.

### DIFF
--- a/walk/walk.go
+++ b/walk/walk.go
@@ -131,6 +131,15 @@ func Walk(c *config.Config, cexts []config.Configurer, dirs []string, mode Mode,
 	visit(c, cexts, knownDirectives, updateRels, trie, wf, "", false)
 }
 
+// Recursively traverse a trie to:
+//  1. configure (top-down)
+//  2. invoke the WalkFunc (bottom-up)
+//
+// Configuration includes building the config.Config for the directory
+// which is inherited by the child directories.
+//
+// Traversal may skip subtrees or files based on the config.Config exclude/ignore/follow options
+// as well as the UpdateFilter callbacks.
 func visit(c *config.Config, cexts []config.Configurer, knownDirectives map[string]bool, updateRels *UpdateFilter, trie *pathTrie, wf WalkFunc, rel string, updateParent bool) {
 	haveError := false
 
@@ -148,7 +157,7 @@ func visit(c *config.Config, cexts []config.Configurer, knownDirectives map[stri
 		haveError = true
 	}
 
-	c = configure(cexts, knownDirectives, c, rel, f)
+	configure(cexts, knownDirectives, c, rel, f)
 	wc := getWalkConfig(c)
 
 	if wc.isExcluded(rel) {
@@ -182,7 +191,7 @@ func visit(c *config.Config, cexts []config.Configurer, knownDirectives map[stri
 			subdirs = append(subdirs, base)
 
 			if updateRels.shouldVisit(entRel, shouldUpdate) {
-				visit(c, cexts, knownDirectives, updateRels, t, wf, entRel, shouldUpdate)
+				visit(c.Clone(), cexts, knownDirectives, updateRels, t, wf, entRel, shouldUpdate)
 			}
 		}
 	}
@@ -294,10 +303,7 @@ func loadBuildFile(c *config.Config, pkg, dir string, ents []fs.DirEntry) (*rule
 	return rule.LoadFile(path, pkg)
 }
 
-func configure(cexts []config.Configurer, knownDirectives map[string]bool, c *config.Config, rel string, f *rule.File) *config.Config {
-	if rel != "" {
-		c = c.Clone()
-	}
+func configure(cexts []config.Configurer, knownDirectives map[string]bool, c *config.Config, rel string, f *rule.File) {
 	if f != nil {
 		for _, d := range f.Directives {
 			if !knownDirectives[d.Key] {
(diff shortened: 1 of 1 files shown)


walk/walk.go
```diff
@@ -297,10 +295,7 @@ func loadBuildFile(c *config.Config, pkg, dir string, ents []fs.DirEntry) (*rule
 	return rule.LoadFile(path, pkg)
 }
 
-func configure(cexts []config.Configurer, knownDirectives map[string]bool, c *config.Config, rel string, f *rule.File) *config.Config {
-	if rel != "" {
-		c = c.Clone()
```
fmeum, line 302: Could you add a comment to this function to make it clear that `c` may be modified and should be cloned by the caller?

walk/walk.go
```diff
@@ -279,12 +277,12 @@ func (u *UpdateFilter) shouldVisit(rel string, updateParent bool) bool {
...
 
-func loadBuildFile(c *config.Config, pkg, dir string, ents []fs.DirEntry) (*rule.File, error) {
+func loadBuildFile(c *config.Config, pkg string, ents []fs.DirEntry) (*rule.File, error) {
 	var err error
-	readDir := dir
+	readDir := filepath.Join(c.RepoRoot, pkg)
 	readEnts := ents
 	if c.ReadBuildFilesDir != "" {
-		readDir = filepath.Join(c.ReadBuildFilesDir, filepath.FromSlash(pkg))
+		readDir = filepath.Join(c.ReadBuildFilesDir, pkg)
```
fmeum, line 285: Why is this safe? Does `pkg` contain OS-specific path separators at this point?T

walk/walk.go
```diff
@@ -186,12 +183,13 @@ func visit(c *config.Config, cexts []config.Configurer, knownDirectives map[stri
 	shouldUpdate := updateRels.shouldUpdate(rel, updateParent)
 	for _, sub := range subdirs {
 		if subRel := path.Join(rel, sub); updateRels.shouldVisit(subRel, shouldUpdate) {
-			visit(c, cexts, knownDirectives, updateRels, trie.children[sub], wf, subRel, shouldUpdate)
+			visit(c.Clone(), cexts, knownDirectives, updateRels, trie.children[sub], wf, subRel, shouldUpdate)
```
fmeum, line 186: It looks like we cloned only once (in `configure`) before this change, but now clone once for every subdir. Was that a bug (we should have cloned before) or why is this not a potential performance concern?

jbedard: Each `visit` is essentially just a `configure` though. However now I wonder (and forget) if this will conflict with https://github.com/bazel-contrib/bazel-gazelle/pull/1921 so I think we should focus on that PR first.

walk/walk.go
jbedard: We don't do it 3 lines up (which was moved from the `loadBuildFile` call site), so I don't think we should do it here either. We could do the opposite and do it in both, but I'd rather take the one with less code if it works...

jbedard:
Lets also do https://github.com/bazel-contrib/bazel-gazelle/pull/1931 first.

walk/walk.go
jbedard: I dropped this second commit for now.

walk/walk.go
jbedard: Done

jbedard:
> Lets also do https://github.com/bazel-contrib/bazel-gazelle/pull/1931 first.

I take that back. I think this is worth merging as-is for now 🤷

fmeum (approved):

walk/walk.go
```diff
@@ -182,7 +191,7 @@ func visit(c *config.Config, cexts []config.Configurer, knownDirectives map[stri
 			subdirs = append(subdirs, base)
 
 			if updateRels.shouldVisit(entRel, shouldUpdate) {
-				visit(c, cexts, knownDirectives, updateRels, t, wf, entRel, shouldUpdate)
+				visit(c.Clone(), cexts, knownDirectives, updateRels, t, wf, entRel, shouldUpdate)
```
fmeum, line 194: We would now clone once per child rather than once, right? That sounds worse in the abstract, have you measured the performance impact?

jbedard: It's the exact same, just less confusing imo. It creates a root `Config` and then clones once per visited child directory.

Previously:
1. construct the root `Config` and pass it to the root `visit` invocation
2. pass that config into `configure` which *might* clone it and return a new one
3. pass the `configure`-returned config into the recursive `visit`

In step 1 the config might be for the parent directory or the current directory which is super confusing.
From step 2 onward it is the config of the current directory due to the conditional cloning within `configure`.

IMO neither of those are clear and it confuses me every time (including right now trying to remember+describe what's going on).

Now:
1. construct the root `Config` and pass it to the root `visit` invocation
2. `configure` is passed the config of the directory being configured
3. clone the config for each child when recursing

The config passed to `visit` is for the directory being visited.
The config passed to `configure` is the config for the directory being configured.
No more confusion about the config being for the parent vs current directory, no more conditional cloning, no more switching from parent to current dir config in the middle of the method etc.

fmeum: Thanks for the explanation! I now understand why this is simpler.

My only concern is that "possibly cloned" can mean "may optimize out a clone", so I would like to see @tyler-french or @linzhp run this in a benchmark.

jbedard: When I say "possibly cloned" I actually mean "cloned if not the root". So we construct the root config, then clone per child. Same thing after my change except without all the confusing (imo) conditional logic.

fmeum: Ah, right, that makes this an easy choice.

PR #1937, opened 2024-09-26
fix(repo): load repositories should fail on duplicates

tyler-french:
We've recently had issues because we have multiple `go_repository` rules with the same name, at different versions.

If we run `go mod tidy` and `gazelle update-repos`, the rule that gets updated isn't always the one that ends up being used in the build.

To solve this, we are planning to add a check in CI to make sure no duplicate go_repository rules are loaded.

This change is needed in order to use `gazelle/repo` to enable this feature. This can also be potentially used to help aid others in migrating to Bzlmod or simply cleaning up their repos.


linzhp:
> If we run go mod tidy and gazelle update-repos, the rule that gets updated isn't always the one that ends up being used in the build.

This feels like a bug. I can't think of a scenario where this is desired. Can we simply fail when duplicates are found, instead of adding additional APIs to support strict mode?

@fmeum what do you think?

fmeum:
I agree with @linzhp here, let's make this strict behavior the default. Every issue like this that goes unnoticed just causes even more trouble down the line, with less of a clue about what's going on.

tyler-french:
As long as we do a good job announcing this change

PR #1938, opened 2024-09-27
Evaluate build tags as both true and false

### LINKED ISSUE #1262: `build_tags` does not generate conditional build files

`v0.24.0`

`v0.29.0`

`5.1.1`

Yes

macOS 12.3.1
Apple M1 Pro

### What did you do?

I have a project where a few things are compiled conditionally (based on the tag). There are different files for when the tag is there and not, eg:

```
file.go        # Compiled when +build !mytag
file_mytag.go  # Compiled when +build mytag
```

Gazelle generates the BUILD files at the time of running:

`bazel run //:gazelle`

It's required that I pass the `-build_tags=mytag` to generate the build file so I can later on compile with `--define gotags=mytag`. I there a way I can get build files with conditional where I don't have to generate a different build file for different tags.

Something that conditionally builds based on the tag provided later on by `--define gotags=...` like [here](https://github.com/bazelbuild/rules_go#how-do-i-cross-compile).

### What did you see instead?

No conditionals in build files, instead just static files based on the tag specified while generating them.

patrickmscott:
Gazelle already iterates over all supported os/arch combos to include go files. This is similar but expanded to custom tags. When adding a new build tag, we create a copy with the tag set to true and append to the original. This means that build tag evaluation will grow exponentially.

Fixes https://github.com/bazelbuild/bazel-gazelle/issues/1262

**What type of PR is this?**

Feature

language/go

It changes the behavior of `build_tags` to defer evaluation to build time. Instead of assuming the tag is _always_ true, it treats the tag as both true and false when evaluating tags for a file.

Fixes #1262

**Other notes for review**

### DIFF
--- a/language/go/config.go
+++ b/language/go/config.go
@@ -41,6 +41,16 @@ import (
 
 var minimumRulesGoVersion = version.Version{0, 29, 0}
 
+type tagSet map[string]bool
+
+func (ts tagSet) clone() tagSet {
+	c := make(tagSet)
+	for k, v := range ts {
+		c[k] = v
+	}
+	return c
+}
+
 // goConfig contains configuration values related to Go rules.
 type goConfig struct {
 	// The name under which the rules_go repository can be referenced from the
@@ -53,7 +63,7 @@ type goConfig struct {
 
 	// genericTags is a set of tags that Gazelle considers to be true. Set with
 	// -build_tags or # gazelle:build_tags. Some tags, like gc, are always on.
-	genericTags map[string]bool
+	genericTags []tagSet
 
 	// prefix is a prefix of an import path, used to generate importpath
 	// attributes. Set with -go_prefix or # gazelle:prefix.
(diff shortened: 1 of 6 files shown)


language/go/config.go
```diff
@@ -41,6 +41,16 @@ import (
 
 var minimumRulesGoVersion = version.Version{0, 29, 0}
 
+type tagSet map[string]bool
+
+func (ts tagSet) clone() tagSet {
+	c := make(tagSet)
```
fmeum, line 47:
```suggestion
	c := make(tagSet, len(ts))
```

language/go/config.go
```diff
@@ -41,6 +41,16 @@ import (
 
 var minimumRulesGoVersion = version.Version{0, 29, 0}
 
+type tagSet map[string]bool
```
fmeum, line 44: Are we actually using the "present, but mapped to false" state or should this be a `map[string]struct{}`?

fmeum (approved):
@tyler-french Could you test this at Uber? I would like to make sure that we don't see any accidental exponential blowup.

patrickmscott:
@tyler-french or @fmeum any update? I'm still running on a patched version of gazelle.

tyler-french (approved):

alopezz:
👋 This makes Gazelle unusable with a relatively high number of build tags. It causes it to never finish on a repo with 20+ tags.

I think I have an idea of how to implement the equivalent functionality in a more performant (linear) way, I'll try to put up a PR with it as soon as I can.

alopezz:
This would be more or less the idea: https://github.com/bazel-contrib/bazel-gazelle/pull/2058

PR #1853, opened 2024-08-01
Add QNX and OSX to known platforms

hofbi:
**What type of PR is this?**

> Uncomment one line below and remove others.
>
Bug fix

**What package or component does this PR mostly affect?**

language/go
cmd/gazelle

**What does this PR do? Why is it needed?**

Adding more known platforms to avoid issue reported in https://github.com/bazelbuild/bazel-gazelle/issues/1735.

**Which issues(s) does this PR fix?**

Fixes #1735

**Other notes for review**

PR #1940, opened 2024-09-30
update version of rules_go

francine-blanc:
**What type of PR is this?**

Bug fix

language/go
cmd/gazelle

This is a small PR to update the version of `rules_go` used by `gazelle` for bzlmod.

When more known platforms were added in this PR:  https://github.com/bazelbuild/bazel-gazelle/pull/1853, the version of `rules_go` in the `MODULE.bazel` was not updated to 0.50.1 (the earliest release the change appears in). This means that although `gazelle` depends on a newer version of `rules_go`, which pulls in all of these additional platforms, without the newer version of `rules_go` being referenced in the `MODULE.bazel`, errors such as the following are seen:
`no such target '@@rules_go~//go/platform:qnx': target 'qnx' not declared in package 'go/platform'` when attempting to build a binary with a go_dependency that refers to these platforms.

n/a

**Other notes for review**

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -1,6 +1,7 @@
 module(
     name = "gazelle",
-    version = "0.39.0",
+    # Updated by the Publish to BCR app.
+    version = "0.999.0",
     repo_name = "bazel_gazelle",
 )
 


fmeum:
Sorry, forgot about your PR.

Issue #1945, opened 2024-10-05
runtime.godebugDefault handling doesn't work with gazelle/rules_go


zecke:
### What version of gazelle are you using?

0.39.1

### What version of rules_go are you using?

0.51.0

### What version of Bazel are you using?

7.3.2

### Does this issue reproduce with the latest releases of all the above?

These should be the latest version

### What operating system and processor architecture are you using?

AARCH64 and macos

### What did you do?

I want to set the default value of a GODEBUG variable. According to the [documentation](https://go.dev/doc/godebug) there are three ways to do this:

1. Use `//go:debug somevalue=X` before the package declaration
2. Include it in the go.mod
3. (Include it in the go.work)

### What did you expect to see?

Trying first and second I expected the resulting binary to include the `godebugDefault`

### What did you see instead?

Both the directive and the go.mod file have no impact on compiling the go binary.

### More details and other observations

- Interestingly bazel/rules_go allow me to link against the `internal/godebug` package while `go build` forbids this
- `go build` would warn about unknown directive and when built with bazel it does not
- Use x_defs and setting runtime.godebugDefault is a working workaround.

### Reproducer set-up

```shell
$ cat MODULE.bazel
bazel_dep(name = "rules_go", version = "0.50.1")
bazel_dep(name = "gazelle", version = "0.39.1")


go_sdk = use_extension("@rules_go//go:extensions.bzl", "go_sdk")

# Download an SDK for the host OS & architecture as well as common remote execution platforms.
go_sdk.download(version = "1.23.1")

go_deps = use_extension("@gazelle//:extensions.bzl", "go_deps")
go_deps.from_file(go_mod = "//:go.mod")

$ cat BUILD.bazel
load("@gazelle//:def.bzl", "gazelle")
load("@rules_go//go:def.bzl", "go_binary", "go_library")

gazelle(name = "gazelle")

go_library(
    name = "godebug_lib",
    srcs = ["main.go"],
    importpath = "github.com/zecke/godebug",
    visibility = ["//visibility:private"],
)

go_binary(
    name = "godebug",
    embed = [":godebug_lib"],
    visibility = ["//visibility:public"],
    #x_defs = {
    #    "runtime.godebugDefault": "tlsrsakex=1",
    #},
)

$ cat main.go
//go:debug tlsrsakex=2
package main


import (
	"fmt"
	"internal/godebug"
)

func main() {
	g := godebug.New("tlsrsakex")
[...]

PR #1946, opened 2024-10-05
Skip over godebug lines in go.mod/go.work files

zecke:
Go 1.21 introduced support for godebug lines in go.mod and go.work files. Skip over these lines instead of failing

Related #1945

**What type of PR is this?**

>
> Bug fix
> Feature
> Documentation
> Other

> For example:
>
> language/go
> cmd/gazelle
> go_repository
> all

Fixes #

**Other notes for review**

### DIFF
--- a/internal/bzlmod/go_mod.bzl
+++ b/internal/bzlmod/go_mod.bzl
@@ -80,6 +80,8 @@ def parse_go_work(content, go_work_label):
                 state["use"].append(tokens[0])
             elif current_directive == "replace":
                 _parse_replace_directive(state, tokens, go_work_label.name, line_no)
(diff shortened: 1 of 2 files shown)


zecke:
What if we handle this like the go_embed directive and update the x_defs?

zecke:
> Yeah, I'm all for treating this exactly like Go itself does, I just don't know how it's wired up exactly. Happy to review a change that goes beyond skipping.

From what I can tell is that one of the challenges of rules_go is that the builder is not using `go build`. Which means we don't benefit from some of the checks (e.g. checking if the flag exists, if one can link an internal package).

PR #1948, opened 2024-10-06
Support MODULE.bazel test cases for gazelle_generation_test.

gonzojive:
**What type of PR is this?**

Feature (testing enhancement for extension developers)

gazelle_generation_test

gazelle_generation_test uses the existence of a WORKSPACE file to indicate a directory is a test case. This PR extends gazelle_generation_test to also look for directories with a MODULE.bazel file.

A test case is modified to check the logic.

Fixes # 1947

**Other notes for review**

Added as part of kotlin gazelle development.

### DIFF
--- a/internal/generationtest/generation_test.go
+++ b/internal/generationtest/generation_test.go
@@ -48,7 +48,7 @@ func TestFullGeneration(t *testing.T) {
 	if err != nil {
 		t.Fatalf("Could not convert gazelle binary path %s to absolute path. Error: %v", *gazelleBinaryPath, err)
 	}
-	testNames := map[string]bool{}
+	testNames := map[string]struct{}{}
 	for _, f := range runfiles {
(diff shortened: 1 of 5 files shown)


internal/generationtest/generation_test.go
```diff
@@ -48,9 +48,10 @@ func TestFullGeneration(t *testing.T) {
 	if err != nil {
 		t.Fatalf("Could not convert gazelle binary path %s to absolute path. Error: %v", *gazelleBinaryPath, err)
 	}
+	testNames := map[string]bool{}
```
fmeum, line 51: Could you make this a `map[string]struct{}` instead? That's a more idiomatic version of a set.

fmeum (approved):

internal/generationtest/generation_test.go
gonzojive: done.

fmeum (approved):

PR #1949, opened 2024-10-08
Don't hard fail if there's a second request to bazelize the same Go module

Buzz-Lightyear:
**What type of PR is this?**
Bug fix

> go_deps

It's possible that a go.mod file may contain a (module, version) pair listed both as a direct and indirect requirement. In such cases, the module extension attempts to create a second `go_repository` rule which fails with an error like this:
```
        File "/private/var/tmp/<some_path>/gazelle~/internal/bzlmod/go_deps.bzl", line 646, column 22, in _go_deps_impl
                go_repository(**go_repository_args)
Error in repository_rule: A repo named <some_repo> is already generated by this module extension at /private/var/tmp/<some_path>/gazelle~/internal/bzlmod/go_deps.bzl:646:22
```

This PR fixes such an error by simply keeping track of the modules that were already processed.

I'm happy to create an issue but the instructions said I could skip if it's a small fix

### DIFF
--- a/internal/bzlmod/go_deps.bzl
+++ b/internal/bzlmod/go_deps.bzl
@@ -592,6 +592,7 @@ def _go_deps_impl(module_ctx):
                 ),
             )
 
+    repos_processed = {}
     for path, module in module_resolutions.items():
         if hasattr(module, "module_name"):
             # Do not create a go_repository for a Go module provided by a bazel_dep.
@@ -602,6 +603,10 @@ def _go_deps_impl(module_ctx):
             # Do not create a go_repository for a dep shared with the non-isolated instance of
             # go_deps.
             continue
+        if module.repo_name in repos_processed:
+            continue
+
+        repos_processed[module.repo_name] = True
         go_repository_args = {
             "name": module.repo_name,
             # Compared to the name attribute, the content of this attribute does not go through repo


internal/bzlmod/go_deps.bzl
```diff
@@ -592,6 +592,7 @@ def _go_deps_impl(module_ctx):
                 ),
             )
 
+    repos_processed = {}
     for path, module in module_resolutions.items():
```
fmeum, line 596: This is a dict keyed by path, so I'm surprised to see duplicated modules. Could you share what `path` and `module` look like for a colliding pair?

stefanpenner: I'm also curious, would it be possible to add the failing scenario to the existing test suite? 
I would like to be sure that we aren't masking a upstream problem

stefanpenner: > It's possible that a go.mod file may contain a (module, version) pair listed

I see this is the scenario, do you think it would be possible to add it to something like https://github.com/bazelbuild/bazel-gazelle/blob/master/tests/bcr/go_mod/go.mod

Buzz-Lightyear: `path`: `github.com/shurcooL/githubv4`
`module`: `struct(local_path = None, raw_version = "0.0.0-20230424031643-6cea62ecd5a9", repo_name = "com_github_shurcool_githubv4", to_path = None, version = ((("", 0), ("", 0), ("", 0)), (("20230424031643-6cea62ecd5a9",),)))`

Buzz-Lightyear: @stefanpenner I'll defer to @fmeum here but I'd think [this might be the landing spot](https://github.com/bazelbuild/bazel-gazelle/tree/master/tests/bzlmod) for the test. Perhaps a new `go_deps_test.bzl` is required to test the Starlark logic in the module extension.

fmeum: These deps aren't identical though, they just happen to have the same import path ignoring case. They could be very different modules and I don't think we should just let the first definition win.

Not sure how to handle this situation. Can you try replacing one of these deps with the other via a `replace` directive?

We may have to come up with a better naming scheme for repos that involve upper-case characters.

@linzhp How are you dealing with such packages?

Buzz-Lightyear: Woah `shurcool ` vs `shurcooL`, nice catch @fmeum

Buzz-Lightyear: Replacing doesn't look like the right move here:
```
go: github.com/shurcool/githubv4@v0.0.0-20230424031643-6cea62ecd5a9 used for two different module paths (github.com/shurcooL/githubv4 and github.com/shurcool/githubv4)
```

Buzz-Lightyear: @fmeum @linzhp What do you think about https://github.com/bazelbuild/bazel-gazelle/pull/1949/commits/9ca14e9ec019b288ab8feff35c1fb21ea0517743?

stefanpenner: @Buzz-Lightyear I believe that encoding could also cause conflicts, if we choose to encode these differently, we should strive for something that doesn't also cause issues.

@fmeum Although these packages are technically different, how does that work for filesystems that are case insensitive like the default APFS configuration for macOS. If those packages get written to disk, does go itself disambiguate, if so what mechanism does it use?

image

----

After searching for other case sensitive issues related to go mods, all occurrences were in error and caused some serious issues. My gut is telling me we should normalize, and treat modules in a case insensitive manner, potentially having a manual escape hatch for when that was in error. Now if there are some examples against this, please do share so I better understand this space.

internal/bzlmod/utils.bzl
```diff
@@ -119,6 +119,17 @@ def with_replaced_or_new_fields(_struct, **replacements):
 
     return struct(**new_struct_assignments)
 
+def repo_name(importpath):
+    path_segments = importpath.split("/")
+    segments = reversed(path_segments[0].split(".")) + path_segments[1:]
+    candidate_name = "_".join(segments).replace("-", "_")
+
+    def _encode_case(c):
```
stefanpenner, line 127: changing the encoding isn't a bad idea, but I do worry that this encoding itself could cause more conflicts. Is there a more conflict resistant encoding? Also, I believe any encoding change would require both the extension and gazelle BUILD.bazel generation to be in proper sync.

stefanpenner: I'm not super convinced yet [that supportive case sensitivity is the right call](https://github.com/bazelbuild/bazel-gazelle/pull/1949/files#r1793727670) – obviously open to convincing, but sharing my current gut-feeling.

internal/bzlmod/go_deps.bzl
Buzz-Lightyear: I think `X` => `_x` is safe @stefanpenner because `_` isn't allowed in Go import paths. However, you bring up an interesting point re: case insensitive file systems and how the module cache is organized. For e.g. I see that this has arisen in the past (https://github.com/golang/go/issues/41646) and the module in question was `github.com/sirupsen/logrus` vs `github.com/Sirupsen/logrus`. There can only be one module cached in a case insensitive FS:

```
 > ls -la $GOMODCACHE/github.com/sirupsen
total 0
drwxr-xr-x   3 smuthu  wheel    96 Oct  9 10:39 .
drwxr-xr-x  16 smuthu  wheel   512 Oct  9 10:41 ..
dr-xr-xr-x  52 smuthu  wheel  1664 Oct  9 10:39 logrus@v1.9.3
```

This is a bit of a conundrum. Go permits case sensitive module names and in theory modules that only differ in case can in fact have different functionalities as @fmeum suggested but in practice, I can't think of a valid example to do this. Perhaps picking a winner on the module extension side while sticking with the current repo name format is the move here @fmeum ?

linzhp: Could you just update all your imports to use `github.com/shurcooL/githubv4`?

Buzz-Lightyear: @linzhp Thanks for chiming in and yes, that does work for our specific use case. However in the future it's possible that this won't be an option and it's ideal to solve this for the generic use case.

If the solution is simply that Gazelle won't distinguish modules that only differ in case, then it's best to document that explicitly here. Currently, the extension just fails because it won't create duplicate `go_repository` rules.

fmeum: Could you update this PR to instead fail with a clean error message mentioning the two colliding names and explaining that module names that differ only in casing and punctuation are not supported (yet)?

Buzz-Lightyear: Certainly @fmeum

internal/bzlmod/utils.bzl
fmeum: ! isn't valid in repo names though.

PR #1952, opened 2024-10-09
fix: overwrite attribute values when not mergeable

jbedard:
**What type of PR is this?**

>
> Bug fix
> Feature
> Documentation
> Other

> For example:
>
> language/go
> cmd/gazelle
> go_repository
> all

Fixes #

**Other notes for review**

### DIFF
--- a/rule/merge.go
+++ b/rule/merge.go
@@ -46,8 +46,9 @@ func MergeRules(src, dst *Rule, mergeable map[string]bool, filename string) {
 		return
 	}
 
-	// Process attributes that are in dst but not in src.
+	// Process mergeable attributes that are in dst but not in src.
 	for key, dstAttr := range dst.attrs {
+		// Skip if a new mergeable value is in src, it is not mergeable, or should keep the existing value
 		if _, ok := src.attrs[key]; ok || !mergeable[key] || ShouldKeep(dstAttr.expr) {
 			continue
 		}
@@ -64,8 +65,10 @@ func MergeRules(src, dst *Rule, mergeable map[string]bool, filename string) {
 	// Merge attributes from src into dst.
 	for key, srcAttr := range src.attrs {
 		if dstAttr, ok := dst.attrs[key]; !ok {
+			// Not in dst, set the new attribute value
 			dst.SetAttr(key, srcAttr.expr.RHS)
 		} else if mergeable[key] && !ShouldKeep(dstAttr.expr) {
+			// Mergeable attribute not tagged to keep, merge the attribute values.
 			if mergedValue, err := mergeAttrValues(&srcAttr, &dstAttr); err != nil {
 				start, end := dstAttr.expr.RHS.Span()
 				log.Printf("%s:%d.%d-%d.%d: could not merge expression", filename, start.Line, start.LineRune, end.Line, end.LineRune)
@@ -74,6 +77,11 @@ func MergeRules(src, dst *Rule, mergeable map[string]bool, filename string) {
 			} else {
 				dst.SetAttr(key, mergedValue)
 			}
+		} else if key != "name" && key != "visibility" {
+			// Not mergeable or tagged to keep. Copy the attribute and overwrite.
+			// TODO: why is "name" special?
+			// TODO: why is "visibility" special?
+			dst.SetAttr(key, srcAttr.expr.RHS)
 		}
 	}
 
@@ -86,14 +94,14 @@ func MergeRules(src, dst *Rule, mergeable map[string]bool, filename string) {
 //
 // The following kinds of expressions are recognized.
 //
-//   * nil
-//   * strings (can only be merged with strings)
-//   * lists of strings
-//   * a call to select with a dict argument. The dict keys must be strings,
+//   - nil
+//   - strings (can only be merged with strings)
+//   - lists of strings
+//   - a call to select with a dict argument. The dict keys must be strings,
 //     and the values must be lists of strings.
-//   * a list of strings combined with a select call using +. The list must
+//   - a list of strings combined with a select call using +. The list must
 //     be the left operand.
-//   * an attr value that implements the Merger interface.
+//   - an attr value that implements the Merger interface.
 //
 // An error is returned if the expressions can't be merged, for example
 // because they are not in one of the above formats.


merger/merger.go
```diff
@@ -107,21 +107,32 @@ const UnstableInsertIndexKey = "_gazelle_insert_index"
...
+			return kinds[r.Kind()].MergeableAttrs[attr]
 		} else {
-			return kinds[r.Kind()].ResolveAttrs
+			return kinds[r.Kind()].ResolveAttrs[attr]
 		}
 	}
 
+	isManaged := func(r *rule.Rule, attr string) bool {
+		// Name and visibility are uniquely managed by gazelle.
+		if attr == "name" || attr == "visibility" {
```
jbedard, line 121: I actually don't think `visibility` should be here but it causes too many changes in this repo atm. I think those changes are actually correct but they could be breaking such as reducing the visibility of `//internal/...`.

rule/merge.go
```diff
@@ -41,14 +43,14 @@ import (
 // marked with a "# keep" comment, values in the attribute not marked with
 // a "# keep" comment will be dropped. If the attribute is empty afterward,
 // it will be deleted.
-func MergeRules(src, dst *Rule, mergeable map[string]bool, filename string) {
+func MergeRules(src, dst *Rule, isMergeable, isManaged IsMergeable, filename string) {
```
fmeum, line 46: I agree that this is a bug to be fixed, but this changes the signature of a public function and is thus a breaking change. 

Since a `map[string]bool` actually track three states per entry (absent, present with false, present with true), we could potentially use that information to encode the `isManaged` bit in  a mostly backwards compatible way (ignoring that someone may already set a value to `false` explicitly, which would be very unidiomatic).

fmeum (changes requested):

rule/merge.go
jbedard: Is this public API? I've never understood what is public API vs just public so other packages within gazelle can use it. Should I consider anything listed on https://pkg.go.dev/github.com/bazelbuild/bazel-gazelle as public?

I can either do the `map[string]bool` `true|false` thing, or maybe create a new method and the old one can keep the current behaviour (while simply invoking the new one).

Do you have any preferences?

jbedard: I've updated with this map `false` value idea, see the fixup commit.

fmeum (approved):
@tyler-french Could you test this at Uber?

rule/merge.go
```diff
@@ -41,6 +41,9 @@ import (
 // marked with a "# keep" comment, values in the attribute not marked with
 // a "# keep" comment will be dropped. If the attribute is empty afterward,
 // it will be deleted.
+//
+// If src has an attribute not present in 'mergeable' and not marked with a
```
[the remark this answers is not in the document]
jbedard: I think this comment was correct as-is, although maybe confusing.

This is pointing out that attributes in `src` will overwrite those in `dst` if they are unknown and the value in `dst` is not marked as `# keep`.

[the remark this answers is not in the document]
jbedard: I've updated the comment a bit. See the fixup commits.

fmeum (approved):

tyler-french:
> @tyler-french Could you test this at Uber?

I'm seeing 1000s of build files behaving differently, need to dig a bit deeper. Please hold off on merging for now

jbedard:
@tyler-french what types of changes are you seeing? Are effected attributes in the `KindInfo` anywhere?

jbedard:
@tyler-french have you had a chance to look at this? I don't understand why you'd ever want `SetAttr` to not update the attribute?

stoiky:
@tyler-french any news on this? We've been waiting for this fix at Adobe.

Thank you @jbedard and @fmeum!

tyler-french:
The main issues we are seeing are:
- Things being removed from `data` attributes of `go_test` rules (specifically targets added are removed) and replaced with whatever `gazelle` applies.
- Non-deterministic changes to the ordering of fields, (specifically for **ordered**, non-sorted `string_list` attributes
- Changes to the behavior of the `self_import` attribute in the `gomock` rule (paths changing)
- Random new additions to certain fields

I will try to make reproductions with public code sometime soon

jbedard:
Are those changes all in rules_go targets or in custom languages you have?

jbedard:
@tyler-french any updates here? Even just ideas for tests I could write to reproduce the issues you're finding...

tyler-french:
The behavior that we depend on (for non-mergable attributes) is:
If an attribute doesn't exist, `SetAttr` sets the value of the attribute
If a user manually populates that attribute with their own value, then `SetAttr` cannot overwrite it (i.e. it is not mergeable), so we keep the previous value.

The issue here is that we have many attributes that are supplied by a user, and the understanding is that gazelle is used to generate these attributes if they don't exist, but once they do exist, gazelle doesn't have the "permission" to merge/overwrite them, since they are marked as not mergeable. 

This is particularly needed/required when the gazelle plugin is not necessarily deterministic (e.g. an unsorted list), that way, we don't have to worry about gazelle overwriting an attribute when the value already exists, if the attribute is not mergeable.

I think the solution here is that we should have a separate `OverwriteAttrs` in `rule.KindInfo` that allows users to specify attributes the new value should always overwrite, thereby preserving historic behavior.

@linzhp do you have any thoughts?

tyler-french:
So, I guess another option is we address the non-determinism in these extensions and also add `# keep` statements where appropriate.

So it's more of a question of either: is changing the accepted behavior of non-mergable attributes risky? and/or is there a way we can alleviate issues that arise from people who depend on this behavior.

tyler-french:
Would this PR override custom visibility attributes? We don't want to overwrite custom settings for this

linzhp:
There are 3 types of attributes:

1. unmanaged: Gazelle doesn't manage them at all. All edits in build files are preserved.
2. default values: if the attribute is not set (e.g., in newly generated rules), Gazelle would populate the default value. However, if people customize them in the build files, the edits are preserved. This is currently achieved by setting the attributes in the extensions without declaring them in `KindInfo`. This is a frequent use case for `visibility` and `go_test`'s `data`.
3. full control. Gazelle and its extensions are the only authoritative source. All user edits are overridden unless `# keep` is present. This is achieved by adding them in the `KindInfo`.

This PR removes the support of Type 2. As you found out, you have to make a special case for `visibility`, because it legitimately needs to be Type 2. Since what you need is Type 3 attributes, why don't you make them `MergeableAttrs` or `ResolveAttrs`?

jbedard:
My intention was to expand the "default values" scenario (point 2) to allow languages to decide the default vs override behaviour, where today gazelle forces the default behaviour.

> Since what you need is Type 3 attributes, why don't you make them `MergeableAttrs` or `ResolveAttrs`?

The primary scenario I had was gazelle only managing attributes if a user opts-in, and otherwise letting the user do it manually (without needing `# keep` 100% of the time). The main one being all the tsconfig related `ts_project` attributes. There's [a lot of them](https://github.com/aspect-build/aspect-cli/blob/2025.05/gazelle/js/generate.go#L62-L75) so sometimes it's nice for gazelle to automate it, but other times users prefer just doing it within a macro to avoid unnecessary clutter in BUILD files (especially when these attributes are the same across an entire repo). These attributes are all trivial strings or booleans, no "merging" or "resolving" is required so I feel like they don't belong in either of those categories.

The second use case I had was a feature where the attributes aren't known at compile time, maybe they come from directives or a config file of some type etc. If a user changes those directives or config files the attributes might need to change.

jbedard:
Regarding letting languages decide on default vs overwriting attributes... This PR allows languages to make that decision, but we could also add something such as `SetAttrDefault` to make that easier.

Or to avoid changing existing `SetAttr` behaviour we could instead add `OverwriteAttr`. IMO `SetAttr` setting the attribute and `SetAttrDefault` setting only a default is significantly clearer, but `OverwriteAttr` would be a new feature and not change any existing behaviour...

Note that `# keep` should always have priority in all these scenarios, like it does today.

linzhp:
According to the doc here: https://github.com/bazel-contrib/bazel-gazelle/blob/669939453cba5ac2874407245f02a0d281fd13fa/rule/merge.go#L35-L38

and the implementation here: https://github.com/bazel-contrib/bazel-gazelle/blob/669939453cba5ac2874407245f02a0d281fd13fa/rule/merge.go#L108-L110

string and boolean (and other "scalar" values) attributes are overridden by Gazelle if you declare them as `MergeableAttrs`. Yeah, the word "merge" is a bit confusing in this case, because there is no way to merge, say "True" into "False". We can interpret it as the attributes from "src" that Gazelle uses when "merging two rules".

Can you try declaring those attributes are `MergeableAttrs` as see if it works?

jbedard:
Yeah, I understand the gazelle "mergeable" doesn't necessarily mean it will be merged. I think of it more as being "managed by gazelle" (which includes merge logic for non-primitives, `# keep` etc).

> Can you try declaring those attributes are MergeableAttrs as see if it works?

That's what I have done to workaround the issue this PR is "fixing". It just means I need a workaround to [maintain the existing values](https://github.com/aspect-build/aspect-cli/blob/2025.05/gazelle/js/generate.go#L651-L659) when I don't want gazelle to be managing those attributes.

linzhp:
> That's what I have done to workaround the issue this PR is "fixing". It just means I need a workaround to [maintain the existing values](https://github.com/aspect-build/aspect-cli/blob/2025.05/gazelle/js/generate.go#L651-L659) when I don't want gazelle to be managing those attributes.

I don't think removing the support of Type 2 attributes is fixing an issue. You just need to make them Type 3, which is already supported. Like you noted above, you will have to find another workaround if you remove the Type 2 support in order to maintain existing values.

jbedard:
> I don't think removing the support of Type 2 attributes is fixing an issue.

This PR simply adds the ability to set an attribute without marking it as mergeable. This does not remove any functionality.

IMO if you want "default values" (type 2) then you should only set that default value if the attribute does not already have a value. Isn't that the definition of a default value?

jbedard:
I have a workaround to accomplish what I'm looking for, but this use case seemed standard enough I thought a workaround shouldn't be required.

If we don't want to change existing behaviour we could add this functionality another way.

If we don't want to add this functionality we can close this PR.

linzhp:
I still don't understand the hesitancy of making an attribute `MergeableAttrs`, when the behavior you want is what `MergeableAttrs` are designed to work. I am concerned that this PR is making unnecessary behavior change

jbedard:
> I still don't understand the hesitancy of making an attribute MergeableAttrs, when the behavior you want is what MergeableAttrs are designed to work.

I am using `MergeableAttrs` like you are suggesting, but that requires what imo is a lot of unnecessary code to [fetch the exist rule](https://github.com/aspect-build/aspect-cli/blob/2025.06/gazelle/js/generate.go#L527) then [re-apply existing attributes](https://github.com/aspect-build/aspect-cli/blob/2025.06/gazelle/js/generate.go#L651-L659) so they don't get deleted during merge if I don't set them.

I just want to conditionally set an attribute value, without it being deleted otherwise:
```
if generateAsdfEnabled {
   rule.SetAttr("asdf", someVal)
}
```

That is not possible today without all the extra code I linked to.

linzhp:
Thanks for providing the links. It helped a lot in understanding your situation.

It seems you want to take full control of `srcs` of `ts_project()` but only take full control of `ts_config` when `cfg.GetTsConfigGenerationEnabled()`, otherwise, `ts_config` is kept as it is.

With the current design, you can implement something like this:

```go
type overrideMe struct{}

func (overrideMe) Merge(other bzl.Expr) bzl.Expr {
  return other
}
``` 

So when `cfg.GetTsConfigGenerationEnabled() == false`, you can just

```go
sourceRule.SetAttr("tsconfig", overrideMe{})
```

Does that work?

jbedard:
> Does that work?

That should work and is better then my current solution 👍 

But that still seems far more complicated then it should be 🤷

jbedard:
It seems like that `overrideMe{}` idea requires `overrideMe` implementing both `rule.Merger` and `build.Expr` which seems to make it even worse (more code) then my existing workaround :/

linzhp:
If it helps, you can upstream `overrideMe` and change the merge logic to make a special case for it so you don't need to implement bzl.Expr

jbedard:
Do you think that would be worth maintaining? Something such as:
```
sourceRule.SetAttr("x", rule.KeepExisting)
```
?

linzhp:
Yeah, I think that sounds like a useful feature to maintain, than changing the existing behavior of non-mergeable attributes. @fmeum @jayconrod What do you think?

fmeum:
> Yeah, I think that sounds like a useful feature to maintain, than changing the existing behavior of non-mergeable attributes. @fmeum @jayconrod What do you think?

I agree that making this optional and explicit is better.

PR #1955, opened 2024-10-11
feat: warn when it looks like you're using ignore instead of exclude directive

alex-torok:
**What type of PR is this?**

Feature

all

`ignore` and `exclude` are somewhat confusing directive names, and are easily mixed up. Show a warning when using the `gazelle:ignore` directive with an argument. 

I've seen many places where people do something like `# gazelle:ignore testdata` and accidentally disable gazelle for the entire directory and subdirectories. This can be seen in [many public github repos](https://github.com/search?q=%2F%23+gazelle%3Aignore+.%2F+language%3AStarlark&type=code&l=Starlark).

**Other notes for review**

Add a new generated test directory to assert that the warning is shown correctly.


fmeum (approved):
Thanks, that's a neat warning!

PR #1960, opened 2024-10-18
Use rules_shell for shell targets

ted-xie:
**What type of PR is this?**

 Bug fix

all

Adds an explicit load (and dep) for sh_binary, which is no longer packaged inside Bazel 8. This is backwards-compatible with Bazel 7.

Fixes https://github.com/bazel-contrib/bazel-gazelle/issues/1959

**Other notes for review**


fmeum (commented):
This change unfortunately isn't fully backwards compatible unless we find a way to invoke `rules_shell`'s dependency macros from the existing WORKSPACE dependency macros of gazelle. We shouldn't have to change `WORKSPACE` to get this to work.

Could you give that a try?

PR #1966, opened 2024-10-23
Fix Bazel module resolution regression in go_deps

fmeum:
**What type of PR is this?**

Bug fix

`go_deps`

2a65ce83f2fc9cd7022139a621fe7fde8157a85f made it so that Bazel modules would no longer override Go modules if their versions don't match exactly.

Also do not report version mismatches for overridden Bazel modules.

Fixes #

**Other notes for review**

### DIFF
--- a/internal/bzlmod/go_deps.bzl
+++ b/internal/bzlmod/go_deps.bzl
@@ -566,13 +566,16 @@ def _go_deps_impl(module_ctx):
         if path in archive_overrides or path in gazelle_overrides or path in module_overrides or path in replace_map:
             continue
 
-        # Only use the Bazel module if it is at least as high as the required Go module version.
-        if path in module_resolutions and bazel_dep.version != module_resolutions[path].version:
+        # Do not report version mismatches for overridden Bazel modules.
+        if path in module_resolutions and bazel_dep.version != _HIGHEST_VERSION_SENTINEL and bazel_dep.version != module_resolutions[path].version:
             outdated_direct_dep_printer("\n\nMismatch between versions requested for module {module}\nBazel dependency version requested in MODULE.bazel: {bazel_dep_version}\nGo module version requested in go.mod: {go_module_version}\nPlease resolve this mismatch to prevent discrepancies between native Go and Bazel builds\n\n".format(
                 module = path,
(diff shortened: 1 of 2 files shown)


fmeum:
@Buzz-Lightyear Could you review this?

Buzz-Lightyear (approved):
LGTM

internal/bzlmod/go_deps.bzl
```diff
@@ -566,21 +566,27 @@ def _go_deps_impl(module_ctx):
         if path in archive_overrides or path in gazelle_overrides or path in module_overrides or path in replace_map:
             continue
 
-        # Only use the Bazel module if it is at least as high as the required Go module version.
-        if path in module_resolutions and bazel_dep.version != module_resolutions[path].version:
+        # Do not report version mismatches for overridden Bazel modules.
+        if path in module_resolutions and bazel_dep.version != _HIGHEST_VERSION_SENTINEL and bazel_dep.version != module_resolutions[path].version:
             outdated_direct_dep_printer("\n\nMismatch between versions requested for module {module}\nBazel dependency version requested in MODULE.bazel: {bazel_dep_version}\nGo module version requested in go.mod: {go_module_version}\nPlease resolve this mismatch to prevent discrepancies between native Go and Bazel builds\n\n".format(
```
honnix, line 571: Is it expected that this `DEBUG` print also happens when the dep in `go.mod` is indirect?

To give a concrete example, we have `cel.dev/expr v0.18.0 // indirect` in `go.mod`, and `0.15.0` in `bazel mod graph` which is a transitive dep of `grpc-java` shown as following, so neither of them is a `direct dep`.
```
├───grpc-java@1.67.1
   ├───cel-spec@0.15.0
```

Buzz-Lightyear: I'd say that's desired behavior as it gives you the opportunity to explicitly add a `replace` directive for the version you'd like, if you so chose.

honnix: This check makes sense. I was just wondering whether this function `outdated_direct_dep_printer` was designed for this purpose.

PR #1967, opened 2024-10-25
go_deps: Support default naming convention override for Bazel modules

fmeum:
**What type of PR is this?**

Bug fix

go_deps

This unblocks using the `build_stackb_rules_proto` module with Gazelle and default settings, as it doesn't generate `go_default_library` targets.

Note that a manually applied `gazelle_override` still results in the Go module being preferred over the Bazel module, so this currently only works with default overrides.

Fixes #

**Other notes for review**


fmeum:
This has been verified in https://bazelbuild.slack.com/archives/CDBP88Z0D/p1729894012409729?thread_ts=1729455978.883849&cid=CDBP88Z0D, so I'll merge it. Please still take a look.

PR #1969, opened 2024-10-28
Issue #1965 - Fix label resolution for non-captured regex

### LINKED ISSUE #1965: Directive resolve_regexp returns invalid labels
`master`

`0.50.1`

`7.4.0`

Yes

MacOS Arm64

### What did you do?

Declare a resolve_regexp directive.
`# gazelle:resolve_regexp py foo\. //x/y/z`

Create a Python file with the following import:
from foo.bar import baz

`deps` should resolve to `//x/y/z` when running Gazelle

### What did you see instead?
`deps` resolves to //x/y/zbar when running Gazelle

[This pull request](https://github.com/bazel-contrib/bazel-gazelle/pull/1822) introduces a change that attempts to create a label from the import statement using regex.  This is problematic, because if the regex does not match on the entire import statement, it will only replace pieces of the import statement, resulting in an incorrect label that still passes the `label.Parse()` check and is then written to the BUILD file.

For example, the import statement passed to `FindRuleWithOverride()` and then `o.resolveRegexDep()` in the above example is the string `foo.bar`.  

The following is then called inside of `resolveRegexDep()`:
`o.ImpRegex.ReplaceAllString("foo.bar", "//x/y/z")`

[According to the docs](https://cs.opensource.google/go/go/+/refs/tags/go1.23.2:src/regexp/regexp.go;l=572):
> // ReplaceAllString returns a copy of src, replacing matches of the [Regexp]
> // with the replacement string repl.
> // Inside repl, $ signs are interpreted as in [Regexp.Expand].

Thus, we end up with the label `//x/y/zbar`.

Other regex passed into the same directive has other corner cases that can be hit, and result in various invalid labels.

Creating the label by running `ReplaceAllString()` appears to contain many corner cases where regex that has been successfully utilized in the past now results in invalid labels.

alandonham:
**What type of PR is this?**
Bug fix

resolve

This change resolves an issue with regex_replace where the default behavior attempts to perform a string replacement using regex, but results in an invalid label.

Fixes #1965

**Other notes for review**

### DIFF
--- a/resolve/config.go
+++ b/resolve/config.go
@@ -64,7 +64,7 @@ func (o regexpOverrideSpec) matches(imp ImportSpec, lang string) bool {
 }
 
 func (o regexpOverrideSpec) resolveRegexpDep(imp ImportSpec) label.Label {
-	// If "$"" is found in the dependency string, then backreferences exist and
+	// If "$" is found in the dependency string, then backreferences exist and
 	// ReplaceAllString() should be run on the string to substitute in the
 	// correct replacement strings to build the label.
 	if !strings.Contains(o.dep.String(), "$") {
--- a/resolve/config.go
+++ b/resolve/config.go
@@ -65,7 +65,7 @@ func (o regexpOverrideSpec) matches(imp ImportSpec, lang string) bool {
 
 func (o regexpOverrideSpec) resolveRegexpDep(imp ImportSpec) label.Label {
 	// If NumSubexp() returns 0, then there are no captures in the regex,
-	// therefor the ReplaceAllString() call above would likely generate
+	// therefore the ReplaceAllString() call above would likely generate
 	// an invalid label.
 	if o.ImpRegex.NumSubexp() == 0 {
 	    return o.dep
--- a/resolve/config.go
+++ b/resolve/config.go
@@ -64,10 +64,10 @@ func (o regexpOverrideSpec) matches(imp ImportSpec, lang string) bool {
 }
 
 func (o regexpOverrideSpec) resolveRegexpDep(imp ImportSpec) label.Label {
-	// If NumSubexp() returns 0, then there are no captures in the regex,
-	// therefore the ReplaceAllString() call below would likely generate
-	// an invalid label.
-	if o.ImpRegex.NumSubexp() == 0 {
+	// If "$"" is found in the dependency string, then backreferences exist and
+	// ReplaceAllString() should be run on the string to substitute in the
+	// correct replacement strings to build the label.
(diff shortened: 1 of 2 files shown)


resolve/config.go
```diff
@@ -64,7 +64,12 @@ func (o regexpOverrideSpec) matches(imp ImportSpec, lang string) bool {
 }
 
 func (o regexpOverrideSpec) resolveRegexpDep(imp ImportSpec) label.Label {
-
+	// If NumSubexp() returns 0, then there are no captures in the regex,
```
fmeum, line 67: Now that I thought about this more, I realized that adding a capturing group without the intention of capturing it is probably somewhat common (`(a|b)` is simply less verbose than `(?:a|b)`). 

Maybe looking for back references in the replacement string is a better signal? What do you think?

alandonham: Sorry for the delay, last week was very busy for me.  I should get back around to looking at this a bit today.

resolve/config.go
```diff
@@ -64,7 +64,12 @@ func (o regexpOverrideSpec) matches(imp ImportSpec, lang string) bool {
 }
 
 func (o regexpOverrideSpec) resolveRegexpDep(imp ImportSpec) label.Label {
-
+	// If "$" is found in the dependency string, then backreferences exist and
+	// ReplaceAllString() should be run on the string to substitute in the
+	// correct replacement strings to build the label.
+	if !strings.Contains(o.dep.String(), "$") {
+	    return o.dep
+	}
```
alandonham, line 72: This follows the [same logic](https://cs.opensource.google/go/go/+/refs/tags/go1.23.2:src/regexp/regexp.go;l=574) found in the `regex` Go stdlib to determine if backreferences exist.

I'm not entirely sure these are really backreferences, as the Go stdlib doesn't support backreferences, but it at least is indicating that the user expects regex to perform string replacement.  I think it is more accurately referred to as `Regexp.Expand` in this library.

resolve/config.go
alandonham: I looked into it a bit more and I agree with your statement.  Backreferences, or `regex.Expand` usage via "$" is a better signal.

resolve/config.go
fmeum: In theory `$` could be escaped, but since `$` is pretty unlikely to make it into a label, let alone into a replace pattern, I think that's okay for now.

fmeum (approved):

resolve/config.go
fmeum: Thanks, don't worry, we're all busy :⁠-⁠)

PR #1975, opened 2024-11-12
Revert to empty version and update repo metadata

fmeum:
**What type of PR is this?**

Bug fix

BCR

The new style also isn't supported by Publish to BCR and a fix is in flight: https://github.com/bazel-contrib/publish-to-bcr/pull/175

Fixes #

**Other notes for review**


fmeum:
FYI @tyler-french @linzhp

PR #1992, opened 2024-12-10
rule: unset attribute if srcAttr is nil when merge

### LINKED ISSUE #1989: Can't unset an attr with GlobValue during merge

0.37.0

0.49.0

7.2.1

Yes

macOS 15.1 aarch64

### What did you do?

I generated some empty rules to remove existing rules in BUILD files:

```go
func cleanUndefinedRules(res *language.GenerateResult, f *rule.File) {
	for _, fr := range f.Rules {
		if ... {
				continue
		}
		r := rule.NewRule(fr.Kind(), fr.Name())
		res.Empty = append(res.Empty, r)
	}
}
```

The kind definition has:

```go
		MergeableAttrs: map[string]bool{
			"srcs": true,
		}
```

and the current build file has:

```python
some_rule(
	name = "test",
    srcs = glob([...]),
)
```

the "test" target gets removed.

### What did you see instead?

```
gazelle: BUILD.bazel:124.12-124.29: could not merge expression
```

and the target is untouched after the run.

This is because when merging the attrs, if the dst is not a scalar, it can't be removed.

https://github.com/bazel-contrib/bazel-gazelle/blob/2898dda65bde76b45fd0f0cd1027a6bed62a3920/rule/merge.go#L105

ashi009:
**What type of PR is this?**

Bug fix

all

This change will allow merging a non-existing attribute to any kind dst attribute.

The current merge logic requires dst attribute to be a scaler to unset. Which will prevent gazelle to remove a target with attributes using `glob()`.

Fixes #1989

### DIFF
--- a/rule/merge.go
+++ b/rule/merge.go
@@ -102,31 +102,19 @@ func mergeAttrValues(srcAttr, dstAttr *attrValue) (bzl.Expr, error) {
 		return nil, nil
 	}
 	dst := dstAttr.expr.RHS
-	if srcAttr == nil && (dst == nil || isScalar(dst)) {
+	if srcAttr == nil {
 		return nil, nil
 	}
-	if srcAttr != nil && isScalar(srcAttr.expr.RHS) {
+	if isScalar(srcAttr.expr.RHS) {
 		return srcAttr.expr.RHS, nil
 	}
-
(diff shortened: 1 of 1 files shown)


ashi009:
breaks too many things with `# keep` comments.

PR #1993, opened 2024-12-10
rule: unset attribute if src is nil and no keep annotation in dst when merge

### LINKED ISSUE #1989: Can't unset an attr with GlobValue during merge

ashi009:
**What type of PR is this?**

Bug fix

all

This change will allow merging a non-existing attribute to any dst attribute not annotated with `# keep`.

The current merge logic requires the dst attribute to be a scaler to unset, which prevents Gazelle from removing an attribute using `glob()`.

Fixes #1989

### DIFF
--- a/rule/merge.go
+++ b/rule/merge.go
@@ -169,8 +169,6 @@ func shouldKeepRecursively(e bzl.Expr) bool {
 		if shouldKeepRecursively(e.X) || shouldKeepRecursively(e.Y) {
 			return true
 		}
-	default:
-		fmt.Printf("unexpected expression type %T\n", e)
 	}
 	return false
 }


rule/merge.go
```diff
@@ -138,6 +138,43 @@ func mergeAttrValues(srcAttr, dstAttr *attrValue) (bzl.Expr, error) {
...
+			if shouldKeepRecursively(arg) {
+				return true
+			}
+		}
+	case *bzl.BinaryExpr:
+		if shouldKeepRecursively(e.X) || shouldKeepRecursively(e.Y) {
+			return true
+		}
+	default:
+		fmt.Printf("unexpected expression type %T\n", e)
```
fmeum, line 173: Since we walk expressions recursively, wouldn't we always hit this eventually if there is no keep comment? Can we avoid recursing so deeply and e.g. only do it once?

ashi009: oops, this was for debugging. Let me get rid of it.

fmeum: Makes sense, thanks for the explanation!

fmeum (approved):
Could you add a test case with `glob` that demonstrates why this change is needed?

PR #1996, opened 2024-12-17
Automatically setting shard_count on go_test based on t.Parallel() calls

yushan26:
**What type of PR is this?**

>
> Bug fix
 Feature
> Documentation
> Other

> For example:
>
 language/go
> cmd/gazelle
> go_repository
> all

Bazel may not necessarily schedule tests cases under target in parallel and may assign one core to a target. This means that tests running in t.Parallel may run sequentially. To make sure that the test cases are truly parallelized on heavily loaded machines, you should instead set shard_count on the go_test target. We can automatically setting shard_count on go_test based on the number of t.Parallel() calls at the top level function. This matches what people expects under go test.
This PR counts the number of t.Parallel calls in each target of the top level function and set the shard_count to the number of calls to ensure the test target is truly parallelized

Fixes #

**Other notes for review**


language/go/fileinfo.go
```diff
@@ -307,6 +310,9 @@ func goFileInfo(path, srcdir string) fileInfo {
 					info.hasMainFunction = true
 					break
 				}
+				if info.isTest {
+					info.numParallel += countParallel(fdecl)
```
fmeum, line 314: Should we really be summing these up? My understanding has been that individual test functions in a file are not run in parallel, only the `t.Parallel` funcs in them. In that case taking the maximum would be more appropriate.

PR #2014, opened 2025-01-14
merger: handle ident load

chasestarr:
**What type of PR is this?**
Bug fix

merger

I found that used idents in assignment expression rhs position have their load statements removed. This PR adds those idents to used symbols collection in `FixLoads` fn.

### DIFF
--- a/merger/fix.go
+++ b/merger/fix.go
@@ -88,7 +88,7 @@ func FixLoads(f *rule.File, knownLoads []rule.LoadInfo) {
 			} else {
 				return
 			}
-    } else if ae, ok := x.(*bzl.AssignExpr); ok {
+		} else if ae, ok := x.(*bzl.AssignExpr); ok {
 			if id, ok := ae.RHS.(*bzl.Ident); ok {
 				idents = append(idents, id)
 			} else {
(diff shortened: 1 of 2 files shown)


merger/fix.go
```diff
@@ -86,10 +86,14 @@ func FixLoads(f *rule.File, knownLoads []rule.LoadInfo) {
 			if id, ok := d.X.(*bzl.Ident); ok {
 				idents = append(idents, id)
 			} else {
-				return
+				return nil
 			}
+		} else if _, ok := x.(*bzl.LoadStmt); ok {
+			return &bzl.StopTraversalError{}
+		} else if id, ok := x.(*bzl.Ident); ok {
+			idents = append(idents, id)
```
fmeum, line 94: This could be many idents, e.g. `bar` and `baz` in `foo.bar.baz` or even numeric literals. Could we limit this to the RHS of an assignment expression (or a dot expression in there) for now?

chasestarr: Thank you. I've updated the branch and PR description.

fmeum (approved):
Thanks, this looks safe to merge!

PR #2020, opened 2025-01-24
doc: fix the `rules_python` gazelle example

aignas:
This is a small fix to improve the docs.


aignas:
@fmeum, could you please take a look at this fix?

fmeum (approved):
Looks good, but you have to update the source of the docs too.

PR #2024, opened 2025-02-04
Fix a typo in the target name, and use relative target path

goretkin:
**What type of PR is this?**

>
> Bug fix
> Feature
> Documentation
> Other

> For example:
>
> language/go
> cmd/gazelle
> go_repository
> all

Fixes #

**Other notes for review**

### DIFF
--- a/extend.md
+++ b/extend.md
@@ -50,7 +50,7 @@ libraries through the `DEFAULT_LANGUAGES` list (you may want to use
 load("@bazel_gazelle//:def.bzl", "DEFAULT_LANGUAGES", "gazelle_binary")
 
 gazelle_binary(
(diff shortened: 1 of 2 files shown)


fmeum (commented):
The authoritative docs live under `docs`, this file is generated from them. Could you update the source of truth instead and run `bazel run //docs:update`?

ggoretkin-bdai:
> The authoritative docs live under `docs`, this file is generated from them. Could you update the source of truth instead and run `bazel run //docs:update`?

done

goretkin:
> The authoritative docs live under `docs`, this file is generated from them. Could you update the source of truth instead and run `bazel run //docs:update`?

Should be good now!

PR #2039, opened 2025-02-19
feat: add support for bazel8 REPO.bazel ignore_directories

### LINKED ISSUE #1994: support bazel 8 ignore_directories
Gazelle needs to support the new bazel 8 [ignore_directories](https://bazel.build/versions/8.0.0/rules/lib/globals/repo#ignore_directories) similar to `.bazelignore`

See:
* https://bazel.build/versions/8.0.0/rules/lib/globals/repo#ignore_directories
* https://github.com/bazelbuild/bazel/issues/7093
* https://github.com/bazelbuild/bazel/pull/24203

jbedard:
**What type of PR is this?**

Feature

all

Support the bazel8 [ignore_directories](https://github.com/bazel-contrib/bazel-gazelle/issues/1994)

Fixes #1994

**Other notes for review**

This is handled separately from `.bazelignore` because a) it only excludes directories and does not have to be checked on every file and b) it uses globs and is therefor a lot slower then `.bazelignore`

### DIFF
--- a/tests/bazelignore/REPO.bazel
+++ b/tests/bazelignore/REPO.bazel
@@ -1,0 +1,14 @@
+repo()
+
+# A single ignore_directories() with various comments and syntaxes
+ignore_directories([
+    "repo1",
+    # single line comment within array
+    "**/repo2",
+    "**/repo3/**", # trailing comment within array and doublestars
+    "sub*/repo4-*",
+])
+
+# Only a single ignore_directories() is supported so add a few more to ensure they are ignored
+ignore_directories(["*", "**", "**/repo1"])
+ignore_directories("very", "invalid", 42)
--- a/tests/bazelignore/repo3/sub2/file1
+++ b/tests/bazelignore/repo3/sub2/file1
@@ -1,0 +1,1 @@
+excluded because anything within **/repo3/** is excluded
--- a/walk/config.go
+++ b/walk/config.go
@@ -28,6 +28,7 @@ import (
 
 	"github.com/bazelbuild/bazel-gazelle/config"
 	"github.com/bazelbuild/bazel-gazelle/rule"
+	bzl "github.com/bazelbuild/buildtools/build"
 	"github.com/bmatcuk/doublestar/v4"
 
 	gzflag "github.com/bazelbuild/bazel-gazelle/flag"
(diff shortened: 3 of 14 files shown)


walk/config.go
```diff
@@ -166,12 +167,75 @@ func loadBazelIgnore(repoRoot string) (isIgnoredFunc, error) {
...
 		_, ok := excludes[p]
 		return ok
 	}
 
-	return isIgnored, nil
+	return isBazelIgnored, nil
+}
+
+func loadRepoDirectoryIgnore(repoRoot string) (isIgnoredFunc, error) {
+	ignorePath := path.Join(repoRoot, "REPO.bazel")
```
fmeum, line 179: Let's call this repoFilePath or similar

walk/config.go
```diff
@@ -166,12 +167,75 @@ func loadBazelIgnore(repoRoot string) (isIgnoredFunc, error) {
...
+
+	// Search for ignore_directories([...ignore strings...])
+	for _, expr := range ast.Stmt {
+		if call, isCall := expr.(*bzl.CallExpr); isCall {
+			if inv, isIdentCall := call.X.(*bzl.Ident); isIdentCall && inv.Name == "ignore_directories" {
+				if len(call.List) != 1 {
+					return nothingIgnored, fmt.Errorf("REPO.bazel ignore_directories() expects one argument")
+				}
+
+				list, isList := call.List[0].(*bzl.ListExpr)
```
fmeum, line 203: Users may be passing in a list comprehension. In that case, we probably shouldn't fail hard, but emit a warning.

jbedard: If it's not a list it just returns `nothingIgnored` and the returned error gets logged. That's the same as bazelignore.

walk/config.go
jbedard: Done. Thank you for pointing out my copy-paste errors 😅

tests/bazelignore/repo3/sub2/file1
```diff
@@ -0,0 +1 @@
+excluded because anything within **/repo4/** is excluded
```
fmeum, line 1:
```suggestion
excluded because anything within **/repo3/** is excluded
```

fmeum (approved):

tests/bazelignore/repo3/sub2/file1
jbedard: Thank you 👍

walk/walk.go
```diff
@@ -399,14 +404,14 @@ func buildTrie(c *config.Config, updateRels *UpdateFilter, isIgnored isIgnoredFu
...
-		return walkDir(c.RepoRoot, "", &eg, limitCh, updateRels, isIgnored, trie)
+		return walkDir(c.RepoRoot, "", &eg, limitCh, updateRels, isBazelIgnored, isRepoDirectoryIgnored, trie)
 	})
 
 	return trie, eg.Wait()
 }
 
 // walkDir recursively and concurrently descends into the 'rel' directory and builds a trie
-func walkDir(root, rel string, eg *errgroup.Group, limitCh chan struct{}, updateRels *UpdateFilter, isIgnored isIgnoredFunc, trie *pathTrie) error {
+func walkDir(root, rel string, eg *errgroup.Group, limitCh chan struct{}, updateRels *UpdateFilter, isBazelIgnored, isRepoDirectoryIgnored isIgnoredFunc, trie *pathTrie) error {
```
tyler-french, line 414: This function seems to take a lot of arguments. Can we make it a structure?

jbedard: Yeah, it probably is time to do that.

Can we do that in a followup PR though? Or I can do it first? Just to keep this actually feature PR minimal...

fmeum: Let's do it as a follow-up, the diff itself is clean enough.

tests/bazelignore/REPO.bazel
```diff
@@ -0,0 +1,13 @@
+repo()
+
+# A single ignore_directories() with various comments and syntaxes
+ignore_directories([
+    "repo1",
+    # single line comment within array
+    "**/repo2",
+    "**/repo3/**", # trailing comment within array and doublestars
```
fmeum, line 8: I didn't notice until now that we have no test for what I believe to be the most common usage: a pattern with a non-recursive `*` wildcard such as `sub/bazel-*`.

jbedard: Would `sub*/repo4-*` cover that? See latest push.

I thought things like `**/out` or `**/node_modules` would be the most common ones, maybe also basics like just `ci-scripts` etc.

fmeum: It does. You are right, those are likely even more common, but `bazel-*` was the original motivation for this feature.

fmeum (approved):
Thanks for the contribution, really happy to see this!

(Please still do the follow-up action)

jbedard:
Here's a followup, can discuss other ideas in the pr: https://github.com/bazel-contrib/bazel-gazelle/pull/2043

PR #2042, opened 2025-02-21
walk: add benchmark for Walk

jayconrod:
Somewhat related to #1891, since we've been talking about performance

**What type of PR is this?**

> Other

> walk

Adds a benchmark for the `Walk` function, since we've been talking speculatively about performance.

Also changes trie-building parallelism from 100 to 6. Lower parallelism is worse, higher is about the same.

Somewhat related to #1891

**Other notes for review**

Run with:

```
bazel test //walk:walk_test --test_arg=-test.bench=. --test_output=streamed --test_filter=none --cache_test_results=false
```

### DIFF
--- a/walk/walk.go
+++ b/walk/walk.go
@@ -394,7 +394,10 @@ func buildTrie(c *config.Config, updateRels *UpdateFilter, isIgnored isIgnoredFu
 	trie := &pathTrie{}
 
 	// A channel to limit the number of concurrent goroutines
-	limitCh := make(chan struct{}, 100)
+	// Value chosen by running BenchmarkWalk on a MacBook Pro M1.
+	// In most cases, walking a tree is memory or I/O bound, not CPU bound,
+	// so this should be lower than the number of cores.
+	limitCh := make(chan struct{}, 6)
 
 	// An error group to handle error propagation
 	eg := errgroup.Group{}
--- a/walk/walk_test.go
+++ b/walk/walk_test.go
@@ -16,8 +16,10 @@ limitations under the License.
 package walk
 
 import (
+	"bytes"
 	"flag"
 	"fmt"
+	"os"
 	"path"
 	"path/filepath"
 	"reflect"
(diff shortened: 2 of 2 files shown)


jayconrod:
cc @jbedard

jbedard (approved):
Thanks for the ping, and LGTM 👍

walk/walk_test.go
```diff
@@ -409,3 +411,76 @@ func (*testConfigurer) KnownDirectives() []string { return nil }
...
+//
+// There are 10 top-level directories. Each has 10 subdirectories. Each of
+// those has 10 subdirectories (so 1001 directories in total).
+//
+// Each directory has 10 files and a BUILD file with a filegroup that includes
+// those files (the content isn't really important, we just want to exercise
+// the parser a little bit.)
+//
+// This is somewhat unrealistic: the whole tree is likely to be in the kernel's
+// in memory in the kernel's file cache, so this doesn't measure I/O to disk.
```
fmeum, line 425:
```suggestion
// memory in the kernel's file cache, so this doesn't measure I/O to disk.
```

fmeum (approved):

walk/walk.go
```diff
@@ -394,7 +394,10 @@ func buildTrie(c *config.Config, updateRels *UpdateFilter, isIgnored isIgnoredFu
 	trie := &pathTrie{}
 
 	// A channel to limit the number of concurrent goroutines
-	limitCh := make(chan struct{}, 100)
+	// Value chosen by running BenchmarkWalk on a MacBook Pro M1.
+	// In most cases, walking a tree is memory or I/O bound, not CPU bound,
+	// so this should be lower than the number of cores.
+	limitCh := make(chan struct{}, 6)
```
tyler-french, line 400: While I do see similar performance on my machine, I wonder if we should make this configurable. 

For example, changing this from 100 to 6 now requires a release of gazelle just for users to pick up the change.

Although, if this is something we truly don't want to adjust in the future or let users change, then I support keeping the literal here.

jbedard: Does keeping this higher actually degrade performance at all? When it's not CPU bound does rounding up really cause any issues? But rounding up could allow more parallelism on other machines like CI?

jayconrod: > While I do see similar performance on my machine, I wonder if we should make this configurable.

Subjective opinion: I'd prefer it not be configurable. Fewer options means fewer decisions for the user to make, and I don't think this rises to the level of importance to justify a command line flag.

> Does keeping this higher actually degrade performance at all? When it's not CPU bound does rounding up really cause any issues? But rounding up could allow more parallelism on other machines like CI?

Empirically, not in this case. The performance with 6 is nearly identical to 100 on my machine. But in general, it depends.

I've seen cases in C++ and Java where adding more threads degrades performance a lot. The bottleneck is not CPU time but contention for cache, memory, or disk bandwidth. More threads compete with each other for the same resources and thrash.

For Go, the runtime limits the number of active OS threads to the number of cores. A goroutine remains bound to a thread when performing disk I/O. So here, when we increase the number of goroutines beyond the number of cores, we're allocating memory for the stack without actually increasing the number of threads that can run.

jbedard: > Subjective opinion: I'd prefer it not be configurable

That's my vote as well. I'd prefer giving the user less options if possible, at least until we have to.

> Empirically, not in this case. The performance with 6 is nearly identical to 100 on my machine. But in general, it depends.

If it doesn't degrade performance then I'd think we should still round up if we're not going to test on more machines. Wouldn't CI with more cores benefit from keeping the number higher? Especially when CI is more likely to be run across the repo and not on a subset of directories...

I'm just speculating though, 6 is probably fine most of the time 🤷

PR #2043, opened 2025-02-22
refactor: simplify walkDir signature

jbedard:
There's 2 commits here both simplifying it in different ways.

**What type of PR is this?**

Other

all

Cleanup regarding https://github.com/bazel-contrib/bazel-gazelle/pull/2039#discussion_r1964845284

### DIFF
--- a/walk/config.go
+++ b/walk/config.go
@@ -130,9 +130,39 @@ func (cr *Configurer) Configure(c *config.Config, rel string, f *rule.File) {
 	c.Exts[walkName] = wcCopy
 }
 
+type ignoreFilter struct {
+	isBazelIgnored         isIgnoredFunc
+	isRepoDirectoryIgnored isIgnoredFunc
+}
+
 type isIgnoredFunc = func(string) bool
 
 var nothingIgnored isIgnoredFunc = func(string) bool { return false }
+
+func newIgnoreFilter(repoRoot string) *ignoreFilter {
+	isBazelIgnored, err := loadBazelIgnore(repoRoot)
+	if err != nil {
+		log.Printf("error loading .bazelignore: %v", err)
+	}
+
+	isRepoDirectoryIgnored, err := loadRepoDirectoryIgnore(repoRoot)
+	if err != nil {
+		log.Printf("error loading REPO.bazel ignore_directories(): %v", err)
+	}
+
+	return &ignoreFilter{
+		isBazelIgnored:         isBazelIgnored,
+		isRepoDirectoryIgnored: isRepoDirectoryIgnored,
+	}
+}
+
+func (f *ignoreFilter) isDirectoryIgnored(p string) bool {
+	return f.isBazelIgnored(p) || f.isRepoDirectoryIgnored(p)
+}
+
+func (f *ignoreFilter) isFileIgnored(p string) bool {
+	return f.isBazelIgnored(p)
+}
 
 func loadBazelIgnore(repoRoot string) (isIgnoredFunc, error) {
 	ignorePath := path.Join(repoRoot, ".bazelignore")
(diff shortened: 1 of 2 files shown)


walk/walk.go
```diff
@@ -395,26 +386,21 @@ func newTrie(entry fs.DirEntry) *pathTrie {
 	}
 }
 
-func buildTrie(c *config.Config, updateRels *UpdateFilter, isBazelIgnored, isRepoDirectoryIgnored isIgnoredFunc) (*pathTrie, error) {
+func buildTrie(c *config.Config, updateRels *UpdateFilter, ignoreFilter *ignoreFilter) (*pathTrie, error) {
 	trie := &pathTrie{}
 
-	// A channel to limit the number of concurrent goroutines
-	limitCh := make(chan struct{}, 100)
```
jbedard, line 402: I'm not sure why this was a channel instead of using the error group limit? 🤔

fmeum (approved):
@tyler-french

walk/config.go
```diff
@@ -130,10 +130,40 @@ func (cr *Configurer) Configure(c *config.Config, rel string, f *rule.File) {
 	c.Exts[walkName] = wcCopy
 }
 
+type ignoreFilter struct {
```
jayconrod, line 133: (Moved this comment from #1928):

`ignoreFilter` seems too indirect to me. It's not possible from reading the code of its constructor or methods to understand what it does. I think `loadBazelIgnore` and `loadRepoDirectoryIgnore` should return the data needed to fully implement `isDirectoryIgnored` and `isFileIgnored` instead of returning closures.

jbedard: We could potentially merge the first commit for now and I can put this `ignoreFilter` thing in a separate PR. Either way both are required for https://github.com/bazel-contrib/bazel-gazelle/pull/1928

jayconrod: > That means the ignoreFilter methods need to understand each and every type of ignore including the implementation details? It would need to know that .bazelignore has no globs and has the map[string]interface{} implementation details?

Yes. Rather than splitting that code out into multiple places, it's clearer if you can read through the logic in one place. I don't think the abstraction here actually makes it easier to understand.

> So we'd basically be inlining the loadBazelIgnore and loadRepoDirectoryIgnore into a single place

Not exacty. `loadBazelIgnore` and `loadRepoDirectoryIgnore` could still be separate functions, but they'd return something concrete like `[]string` that would be stored as a field of `ignoreFilter`. The closures currently returned by those functions would be effectively inlined though, yes.

> We could potentially merge the first commit for now and I can put this ignoreFilter thing in a separate PR. Either way both are required for https://github.com/bazel-contrib/bazel-gazelle/pull/1928

SGTM. This is mostly a stylistic comment.

jbedard: I think we'd end up with:
```
type ignoreFilter struct {
    ignoreDirectoryGlobs []string
    ignorePaths map[string]interface{}
}
```

Then when we add `exclude` here we'll add:
```
   ignorePathsGlobs []string
```
(I use the word "ignore" here instead of "exclude" to allow expanding this beyond `# gazelle:exclude` in the future)

Then:
```
func (i *ignoreFilter) isDirectoryIgnored(p string) {
    if _, e := i.ignorePaths[p]; e { return true }
    return matchesAnyGlob(i.ignoreDirectoryGlobs, p) || matchesAnyGlob(i.ignorePathsGlobs, p)
}

func (i *ignoreFilter) isFileIgnored(p string) {
    if _, e := i.ignorePaths[p]; e { return true }
    return matchesAnyGlob(i.ignorePathsGlobs, p)
}
```

I'm fine with that if you'd prefer. That does remove an extra layer of indirection like you originally mentioned.

If my snippets above align with what you're thinking I'll add a commit to this PR to make the switch, or let me know if you'd prefer a new PR...

jbedard: See the fixup commit. It does reduce the total code as well as a couple function calls 👍

jayconrod: Thanks! That seems much clearer to me.

jayconrod (approved):

PR #2044, opened 2025-02-24
refactor: move fs/walk related configuration to walk/config

jbedard:
With the ongoing changes to fs-walking I think anything related to BUILD file location/generation should be part of the walk configuration.

I think this will be required or at least simplify the changes required for [parsing BUILD files during walk](https://github.com/bazel-contrib/bazel-gazelle/pull/1928) and [moving exclusion to walk](https://github.com/bazel-contrib/bazel-gazelle/pull/1928/files#r1962414111). Both of those require knowing the `ValidBuildFileNames` and `ReadBuildFilesDir` during the fs walk.

**What type of PR is this?**

Refactor

all

### DIFF
--- a/walk/config.go
+++ b/walk/config.go
@@ -24,6 +24,7 @@ import (
 	"log"
 	"os"
 	"path"
+	"path/filepath"
 	"strings"
 
 	"github.com/bazelbuild/bazel-gazelle/config"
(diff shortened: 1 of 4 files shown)


walk/config.go
```diff
@@ -73,18 +74,48 @@ func (wc *walkConfig) shouldFollow(p string) bool {
...
+	}
+	if wc.writeBuildFilesDir != "" {
+		if filepath.IsAbs(wc.writeBuildFilesDir) {
+			c.WriteBuildFilesDir = wc.writeBuildFilesDir
+		} else {
+			c.WriteBuildFilesDir = filepath.Join(c.WorkDir, wc.writeBuildFilesDir)
+		}
+	}
+
+	c.Exts[walkName] = &walkConfig{
```
jbedard, line 111: Note this assignment was moved from `RegisterFlags` to `CheckFlags`. Mainly just to make all these cli params more consistent, but imo this makes it clearer what comes from the CLI args (all put into `Configurer` fields) vs later getting appended from directives.

jayconrod: It probably doesn't matter since they're nearly called at the same time, but the doc for `RegisterFlags` indicates it's supposed to add to `c.Exts`.

Is there a reason to track excludes separately if they're set by flags or directives? Previously they were appended to the same slice.

walk/config_test.go
```diff
@@ -24,3 +30,53 @@ func TestCheckPathMatchPattern(t *testing.T) {
 		}
 	}
 }
+
+func TestConfigurerFlags(t *testing.T) {
+	dir, err := os.MkdirTemp(os.Getenv("TEST_TEMPDIR"), "config_test")
```
jayconrod, line 35: What does this directory do? I don't see it used below.

(Simplification: you can use `t.TempDir()` to create a directory. It automatically handles errors and cleanup.)

jayconrod (approved):
Thanks for splitting this out, this should make #1928 easier to review.

walk/config_test.go
jbedard: I simply copy+pasted the test from `config/config_test.go` and then deleted the parts unrelated to `ValidBuildFileNames`. I've never really looked into these tests and would rather not when just moving code.

walk/config.go
jbedard: I mainly did it just to align with how `config/config` was doing it. Now both here and in `config/config` the cli data is stored in fields and put into `c.*` and `c.Exts` at the same time 🤷 

I can change it if you have a preference.

PR #2053, opened 2025-03-14
fix: make language agnostic kinds generic

jbedard:
The generation of these kinds should not depend on the go language being enabled.

**What type of PR is this?**

Bug fix

language/go
cmd/gazelle
all

Allows any language to generate `filegroup`s without redeclaring the kinds.

### DIFF
--- a/language/go/kinds.go
+++ b/language/go/kinds.go
@@ -22,14 +22,6 @@ import (
 )
 
 var goKinds = map[string]rule.KindInfo{
-	"alias": {
-		NonEmptyAttrs:  map[string]bool{"actual": true},
-		MergeableAttrs: map[string]bool{"actual": true},
-	},
-	"filegroup": {
-		NonEmptyAttrs:  map[string]bool{"srcs": true},
-		MergeableAttrs: map[string]bool{"srcs": true},
-	},
 	"go_binary": {
 		MatchAny: true,
 		NonEmptyAttrs: map[string]bool{
--- a/v2/rule/kinds.go
+++ b/v2/rule/kinds.go
@@ -1,0 +1,29 @@
+/* Copyright 2026 The Bazel Authors. All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+   http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package rule
+
+// GenericKinds is the KindInfo for native.* bazel rules not owned by any
+// specific language extension.
+var GenericKinds = map[string]KindInfo{
+	"alias": {
+		NonEmptyAttrs:  map[string]bool{"actual": true},
+		MergeableAttrs: map[string]bool{"actual": true},
+	},
+	"filegroup": {
+		NonEmptyAttrs:  map[string]bool{"srcs": true},
+		MergeableAttrs: map[string]bool{"srcs": true},
+	},
+}
--- a/v2/rule/rule.go
+++ b/v2/rule/rule.go
@@ -1064,7 +1064,15 @@ func (r *Rule) InsertAt(f *File, index int) {
 // IsEmpty returns true when the rule contains none of the attributes in attrs
 // for its kind. attrs should contain attributes that make the rule buildable
 // like srcs or deps and not descriptive attributes like name or visibility.
+//
+// If info.NonEmptyAttrs is nil, IsEmpty falls back to metadata for Bazel
+// built-in kinds (e.g. "alias", "filegroup") so that language extensions
+// that generate these kinds without declaring them are still classified
+// correctly.
 func (r *Rule) IsEmpty(info KindInfo) bool {
+	if info.NonEmptyAttrs == nil {
+		info = GenericKinds[r.Kind()]
+	}
 	if info.NonEmptyAttrs == nil {
 		return false
 	}
--- a/internal/go_repository_tools_srcs.bzl
+++ b/internal/go_repository_tools_srcs.bzl
@@ -160,6 +160,7 @@ GO_REPOSITORY_TOOLS_SRCS = [
     Label("//v2/rule:BUILD.bazel"),
     Label("//v2/rule:directives.go"),
     Label("//v2/rule:expr.go"),
+    Label("//v2/rule:kinds.go"),
     Label("//v2/rule:merge.go"),
     Label("//v2/rule:platform.go"),
     Label("//v2/rule:platform_strings.go"),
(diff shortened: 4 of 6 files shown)


language/go/kinds.go
```diff
@@ -17,18 +17,11 @@ package golang
 
 import (
 	"fmt"
+
 	"github.com/bazelbuild/bazel-gazelle/rule"
 )
 
 var goKinds = map[string]rule.KindInfo{
-	"alias": {
```
jbedard, line 24: Issues:
1. a lot of the go tests were depending on this being in the go language
2. a lot of the proto tests were depending on this not being defined, yet in real scenarios there's a good chance it will be

Any opinions what the best way is to fix the tests? I've changed a few but others essentially copy+paste subsets of the `fix-update.go` walk/configure loop, so now we'd have to copy+paste a bit more?

jbedard: @jayconrod any opinions here?

jayconrod: I've been holding off on reviewing this because all the tests are red. I'd rather talk about approach in a separate issue. Feel free to open a new one. I've opened #2280 for one possible idea.

I do agree that `alias`, `filegroup`, and `package` should not be "owned" by any one extension. I'd actually like to move to a model where multiple extensions can generate rules with the same kinds, and Gazelle uses the `KindInfo` from whichever extension actually generated the rule in question.

(And yes, the Go tests are kind of problematic. I'd like to drop a lot of the unit tests since they're difficult to maintain, and instead cover the functionality with integration tests in `//tests` instead. But bandwidth is limited.)

jbedard: I've rebased and I think this PR is actually a reasonable change now. Although if it's not the direction you're thinking of going long term maybe it's not worth it.

I'm uncertain why protobuf tests are no longer being effected though? Previously I think dozens were failing because `filegroup`+`alias` were being changed by gazelle.

jayconrod: Apologies I haven't had time to review this yet. I'm getting a high volume of large PR review requests right now, and it will take me at least another day or two to get through the backlog.

jayconrod: So, as mentioned earlier and in #2280, I want to change to a model where multiple extensions can generate rules of of the same kind, and Gazelle uses the `KindInfo` from whichever extension generated the rule for purposes of resolving and merging. I think adding `rule.GenericKinds` takes us further away from that though.

I'm curious, what actually happens when multiple extensions return `KindInfo` for `alias` or `filegroup`? I think we might end up with the wrong resolver, but not sure how bad that is for these rules.

jbedard: I think it's just last-write wins atm, so it's kind of random? But the definitions of `alias` and `filegroup` are simple enough they are normally the same and it doesn't matter who wins. It's just annoying that these two native rules are defined by the go extension which users may or may not be using.

I do think #2280 is a better long term solution, I do worry about other side-effects of that though. For example, what if I write my own gazelle extension that generates a special `go_binary`, now I need to re-define what `go_binary` is instead of inheriting it from the go gazelle language? Will my re-defined `go_binary` still work with other go extension logic? That makes me wonder about all sorts of alternatives or variations of #2280 and make me assume it won't happen anytime soon 😅 

Also... for things as basic as `alias` and `filegroup`, even with #2280, should `native.*` maybe be defined in a common location so every extension doesn't have to redefine them? That makes me think this PR should still land...

v2/rule/kinds.go
```diff
@@ -0,0 +1,30 @@
+/* Copyright 2025 The Bazel Authors. All rights reserved.
```
jayconrod: 2026

v2/rule/kinds.go
```diff
@@ -0,0 +1,30 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package rule
+
+// GenericKinds is the KindInfo for Bazel built-in rules that are not owned by
+// any specific language extension. Rule.IsEmpty falls back to this map so that
+// languages that generate these kinds do not need to redeclare the metadata.
```
jayconrod: Let's not mention `Rule.IsEmpty` here specifically. If we're defining these here, it should be possible to use them for more than just that purpose.

v2/rule/rule.go
```diff
@@ -1064,7 +1064,15 @@ func (r *Rule) InsertAt(f *File, index int) {
...
 // for its kind. attrs should contain attributes that make the rule buildable
 // like srcs or deps and not descriptive attributes like name or visibility.
+//
+// If info.NonEmptyAttrs is nil, IsEmpty falls back to metadata for Bazel
+// built-in kinds (e.g. "alias", "filegroup") so that language extensions
+// that generate these kinds without declaring them are still classified
+// correctly.
 func (r *Rule) IsEmpty(info KindInfo) bool {
+	if info.NonEmptyAttrs == nil {
+		info = GenericKinds[r.Kind()]
```
jayconrod, line 1074: I don't think this should be hard-coded here. The caller should still be responsible for looking up the correct `KindInfo`. That should be in `v2/cmd/gazelle/update`.

language/go/kinds.go
jayconrod: > I do think #2280 is a better long term solution, I do worry about other side-effects of that though. For example, what if I write my own gazelle extension that generates a special go_binary, now I need to re-define what go_binary is instead of inheriting it from the go gazelle language? Will my re-defined go_binary still work with other go extension logic? That makes me wonder about all sorts of alternatives or variations of #2280 and make me assume it won't happen anytime soon 😅

That's kind of the point though? You'd be able to define a `go_binary` that works differently. Not inherited.

> Also... for things as basic as alias and filegroup, even with #2280, should native.* maybe be defined in a common location so every extension doesn't have to redefine them? That makes me think this PR should still land...

I'm kind of ambivalent on that. I feel like extensions should have to declare the rules they're going to return, but it's true the behavior for these is basically always going to be identical. So let's land it.

jbedard: > That's kind of the point though? You'd be able to define a go_binary that works differently. Not inherited.

What if I don't want it to work differently though? If I wanted `go_binary` to work differently then I would define a macro that wraps it and that wrapper would be independent and completely different.

This discussion goes down a rabbit hole though, which is why I fear #2280 won't be done anytime soon and I'd vote to land this PR just to get `native.*` out of golang for now 🤷

v2/rule/rule.go
jbedard: If this isn't here then a lot of golang tests fail because `filegroup` and `alias` targets are not longer considered "empty" and get left behind.

So we can...
1. duplicated `alias` + `filegroup` in `language/go` in addition to the new spot
2. change `goLang GenerateRules` to check `v2.rule.GenericKinds` (I was hoping to avoid that and keep `GenericKinds` as internal)
3. change `goLang` in some other way to not generate empty targets (instead of generating them and then checking if they are empty)?

Any ideas or preferences?

language/go/kinds.go
jayconrod: Then don't define a separate conflicting extension?

This is hypothetical, I agree this is too far down a rabbit hole to be productive.

v2/rule/rule.go
jayconrod: Hmm, the Go extension calls `IsEmpty` directly and fails if this isn't there.

Fine, let's land this as it is. I think the signature is kind of misleading though, so I may change this later with #2280 in v2.

jayconrod (approved):

PR #2056, opened 2025-03-21
feat(update-repos): allow users to profile gazelle update-repos

### LINKED ISSUE #2057: feat(cmd/update-repos): allow users to profile command
Similar to `fix-update`, running `gazelle update-repos` can be very slow, making it necessary to investigate if there are any bottlenecks in downloads or performance.

We should add the `--cpuprofile` and `--memprofile` flags to allow this command to output the profiles to a file.

tyler-french:
This PR enables CPU and Memory profiling capabilities in `gazelle update-repos`

Resolves #2057

Tested locally with `gazelle update-repos`

Screenshot 2025-03-21 at 12 54 31 PM

### DIFF
--- a/cmd/gazelle/update-repos.go
+++ b/cmd/gazelle/update-repos.go
@@ -20,6 +20,7 @@ import (
 	"errors"
 	"flag"
 	"fmt"
+	"log"
 	"os"
 	"path/filepath"
 	"sort"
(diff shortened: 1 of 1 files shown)


fmeum (approved):
Looks good, but I think we should still try to get rid of this together with WORKSPACE.

tyler-french:
> Looks good, but I think we should still try to get rid of this together with WORKSPACE.

Yeah, unfortunately, I think we'll need update-repos for the foreseeable future. cc @linzhp

PR #2065, opened 2025-04-13
Improve consistency by prefering bazel_dep over Go

### LINKED ISSUE #2060: Challenges migrating to bzlmod with go_deps vs. bazel_dep in non-monorepository environment
This is a bit of a challenge to write down. I have issues including and similar to #1640 and I am not entirely sure how to approach this. Either my mental model needs to be reconciled or there is some issue with gazelle.

### The stage

There are a few git repositories: Imagine a repository for libraries, tools to support CI and make deployments and then maybe the main application code. The tools and main application depend on the same libraries. The libraries use the same tools to support CI. We also have .go and .bzl files in our tools repositories. The dependencies across repos are circular, across the targets obviously not.

This is with:
bazel: 7.4.1
rules_go: 0.53.0
gazelle: 0.42.0

### The approach

Converting the first repository is easy. We have rules_go+gazelle as bazel_dep and then import the `go.mod` file in gazelle, have the `use_repo(go_deps, ...)` and everything is pleasant, upload this version to a bazel module registry. After one of the dependencies has been converted to bzlmod, I start converting the usage to `bazel_dep` and this is where the trouble starts. 

### What seems to happen

The tools require the libraries. The application require the libraries and the tools, the libraries require the tools for CI, the libraries the tools and I am not entirely sure that this happens however during resolving the dependencies this circle seems to go down. Something like:

`app->tool v8->lib v8->tool v7->lib v7->tool v6->lib v6->tool v5->lib v5...`

And then this generally ends in an error like:

```
ERROR: module extension "go_deps" from "@@gazelle~//:extensions.bzl" does not generate repository "com_github_zecke_tools", yet it is imported as "com_github_zecke_tools" in the usage at htttp://registry/modules/other_dep/v5/MODULE.bazel...`
```

v5 of `other_dep` would be the last version that didn't include "com_github_zecke_tools" through a bazel_dep and entirely relied on `go_deps` and the `go.mod` file. Fiddling with `repo_name` in `bazel_dep` and/or the `module` itself doesn't make a difference. In this case using `override_repos` doesn't work either. The only thing that does work is to use `inject_repo` in the root module and that is a challenge. It does feel "wrong" and one would need to do this in many repos.

What puzzles me is that `other_dep` has a direct dependency on `github.com/zecke/tools` via the go.mod file. How can it not "generate" that repo (especially as it is needed)? In the more general case that seems like a challenge to evolve dependencies. When would one bump the compat version?

I will try to create a public reproducer of this issue.

zecke:
Letting the bazel_dep and Go dependency participate equally in the version resolution is creating various inconsistencies. This can lead to to a root module having to either use `inject_repo` or `use_repo` and `override_repo`.

Resolve this by consistently preferring the `bazel_dep` over the Go dependency. Keep the existing code that warns (or can fail) if the dependencies are at different versions.

Example MODULE.bazel:
```
module(
    name = "bazel_dep_should_win",
)

bazel_dep(name = "rules_go", version = "0.53.0")
bazel_dep(name = "gazelle", version = "0.42.0")
bazel_dep(name = "circl", version = "1.3.8")

```

And the go.mod referring to a newer version.

Fixes #2060

**What type of PR is this?**

>
> Bug fix
> Feature
> Documentation
> Other

> For example:
>
> language/go
> cmd/gazelle
> go_repository
> all

Fixes #

**Other notes for review**

### DIFF
--- a/internal/bzlmod/go_deps.bzl
+++ b/internal/bzlmod/go_deps.bzl
@@ -563,9 +563,8 @@ def _go_deps_impl(module_ctx):
 
         bazel_dep_is_older = path in module_resolutions and bazel_dep.version < module_resolutions[path].version
 
-        # Version mismatches between the Go module and the bazel_dep can confuse Go tooling. If the bazel_dep version
-        # is lower, it won't be used, which can result in unexpected builds and should thus always be reported, even for
-        # indirect deps. Explicitly overridden modules are not reported as this requires manual action.
+        # Version mismatches between the Go module and the bazel_dep are problematic. For consistency always
+        # prefer the bazel_dep version and report any mismatch to the user.
         if (path in module_resolutions and
             bazel_dep.version != module_resolutions[path].version and
             bazel_dep.version != _HIGHEST_VERSION_SENTINEL and
@@ -616,10 +615,6 @@ Mismatch between versions requested for Go module {module}:
                 bazel_dep_version = bazel_dep_version,
                 go_module_version = go_module_version,
             ), *remediation)
-
-        # Only use the Bazel module if it is at least as high as the required Go module version.
-        if bazel_dep_is_older:
-            continue
 
         # TODO: We should update root_versions if the bazel_dep is a direct dependency of the root
         #   module. However, we currently don't have a way to determine that.
--- a/tests/bcr/go_mod/go.mod
+++ b/tests/bcr/go_mod/go.mod
@@ -13,7 +13,7 @@ require (
 	github.com/bazelbuild/rules_go v0.39.1
 	// NOTE: keep <4.7.0 to test the 'replace'
 	github.com/bmatcuk/doublestar/v4 v4.6.0
-	github.com/cloudflare/circl v1.3.7
+	github.com/cloudflare/circl v1.6.1
 	github.com/envoyproxy/protoc-gen-validate v1.0.1
 	github.com/fmeum/dep_on_gazelle v1.0.0
 	github.com/google/go-jsonnet v0.20.0
--- a/tests/bcr/go_mod/go.sum
+++ b/tests/bcr/go_mod/go.sum
@@ -17,6 +17,8 @@ github.com/chzyer/test v0.0.0-20180213035817-a1ea475d72b1/go.mod h1:Q3SI9o4m/ZMn
 github.com/client9/misspell v0.3.4/go.mod h1:qj6jICC3Q7zFZvVWo7KLAzC3yx5G7kyvSDkc90ppPyw=
 github.com/cloudflare/circl v1.3.7 h1:qlCDlTPz2n9fu58M0Nh1J/JzcFpfgkFHHX3O35r5vcU=
 github.com/cloudflare/circl v1.3.7/go.mod h1:sRTcRWXGLrKw6yIGJ+l7amYJFfAXbZG0kBSc8r4zxgA=
+github.com/cloudflare/circl v1.6.1 h1:zqIqSPIndyBh1bjLVVDHMPpVKqp8Su/V+6MeDzzQBQ0=
+github.com/cloudflare/circl v1.6.1/go.mod h1:uddAzsPgqdMAYatqJ0lsjX1oECcQLIlRpzZh3pJrofs=
 github.com/creack/pty v1.1.9/go.mod h1:oKZEueFk5CKHvIhNR5MUki03XCEU+Q6VDXinZuGJ33E=
 github.com/davecgh/go-spew v1.1.0 h1:ZDRjVQ15GmhC3fiQ8ni8+OwkZQO4DARzQgrnXU1Liz8=
 github.com/davecgh/go-spew v1.1.0/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
--- a/tests/bcr/go_work/pkg/go.mod
+++ b/tests/bcr/go_work/pkg/go.mod
@@ -9,7 +9,7 @@ require (
 	github.com/bazelbuild/rules_go v0.44.0
 	// NOTE: keep <4.7.0 to test the 'replace'
 	github.com/bmatcuk/doublestar/v4 v4.6.1
-	github.com/cloudflare/circl v1.3.7
+	github.com/cloudflare/circl v1.6.1
 	github.com/envoyproxy/protoc-gen-validate v1.0.4
 	github.com/fmeum/dep_on_gazelle v1.0.0
 	github.com/google/safetext v0.0.0-20240104143208-7a7d9b3d812f
--- a/tests/bcr/go_work/pkg/go.sum
+++ b/tests/bcr/go_work/pkg/go.sum
@@ -17,6 +17,8 @@ github.com/chzyer/test v0.0.0-20180213035817-a1ea475d72b1/go.mod h1:Q3SI9o4m/ZMn
 github.com/client9/misspell v0.3.4/go.mod h1:qj6jICC3Q7zFZvVWo7KLAzC3yx5G7kyvSDkc90ppPyw=
 github.com/cloudflare/circl v1.3.7 h1:qlCDlTPz2n9fu58M0Nh1J/JzcFpfgkFHHX3O35r5vcU=
 github.com/cloudflare/circl v1.3.7/go.mod h1:sRTcRWXGLrKw6yIGJ+l7amYJFfAXbZG0kBSc8r4zxgA=
+github.com/cloudflare/circl v1.6.1 h1:zqIqSPIndyBh1bjLVVDHMPpVKqp8Su/V+6MeDzzQBQ0=
+github.com/cloudflare/circl v1.6.1/go.mod h1:uddAzsPgqdMAYatqJ0lsjX1oECcQLIlRpzZh3pJrofs=
 github.com/davecgh/go-spew v1.1.0/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
 github.com/davecgh/go-spew v1.1.1 h1:vj9j/u1bqnvCEfJOwUhtlOARqs3+rkHYY13jYWTU97c=
 github.com/davecgh/go-spew v1.1.1/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=


zecke:
Any idea on how to test this? I think we can easily pin to an older version of circl and have the go.mod file have a newer entry. If things build without `use_repo(go_deps, "com_github_cloudflare_circl")` we should be fine.

fmeum:
That sounds decent. If you can't get that to work, I'm also fine with this PR as it is - it only deletes cases after all.

zecke:
Upgrading circl to v1.6.1 was enough. Without the patch I get the below build error showing that the go.mod dep was used instead of the bazel_dep (different reponame).

```
FAIL: //:gazelle_test (see /private/var/tmp/_bazel_ich/c0f651c55b5a6db13b7b6f55fed6bc48/execroot/_main/bazel-out/darwin_arm64-fastbuild/testlogs/gazelle_test/test.log)
INFO: From Testing //:gazelle_test:
==================== Test output for //:gazelle_test:
--- pkg/BUILD.bazel	1970-01-01 00:00:00.000000001 +0000
+++ pkg/BUILD.bazel	1970-01-01 00:00:00.000000001 +0000
@@ -52,9 +52,9 @@
     embed = [":pkg"],
     deps = [
         "//pkg/data",
-        "@circl//dh/x25519",
         "@com_github_bazelbuild_buildtools//labels:go_default_library",
         "@com_github_bmatcuk_doublestar_v4//:doublestar",
+        "@com_github_cloudflare_circl//dh/x25519",
         "@com_github_datadog_sketches_go//ddsketch",
         "@com_github_envoyproxy_protoc_gen_validate//validate",
         "@com_github_fmeum_dep_on_gazelle//:dep_on_gazelle",
```

fmeum (approved):

ash2k:
Hi! I'm struggling to update from gazelle 0.42.0 to 0.43.0 because of this change. If I revert it locally my build works. I may be doing something wrong, but so far I couldn't figure out how to make things work. Any help would be appreciated. Here is what I'm doing:

[Attempt 1](https://gitlab.com/gitlab-org/cluster-integration/gitlab-agent/-/merge_requests/2415): just bump gazelle. [I get the following error](https://gitlab.com/gitlab-org/cluster-integration/gitlab-agent/-/jobs/9844775955):

```
(03:42:48) ERROR: no such package '@@[unknown repo 'dev_cel_expr' requested from @@gazelle++go_deps+com_github_google_cel_go]//': The repository '@@[unknown repo 'dev_cel_expr' requested from @@gazelle++go_deps+com_github_google_cel_go]' could not be resolved: No repository visible as '@dev_cel_expr' from repository '@@gazelle++go_deps+com_github_google_cel_go'
(03:42:48) ERROR: /home/ci/.cache/bazel/_bazel_ci/9ca8268df963885560ea55034139bea5/external/gazelle++go_deps+com_github_google_cel_go/common/types/BUILD.bazel:8:11: no such package '@@[unknown repo 'dev_cel_expr' requested from @@gazelle++go_deps+com_github_google_cel_go]//': The repository '@@[unknown repo 'dev_cel_expr' requested from @@gazelle++go_deps+com_github_google_cel_go]' could not be resolved: No repository visible as '@dev_cel_expr' from repository '@@gazelle++go_deps+com_github_google_cel_go' and referenced by '@@gazelle++go_deps+com_github_google_cel_go//common/types:go_default_library'
(03:42:48) ERROR: Analysis of target '//internal/module/kubernetes_api/agent:agent_test' failed; build aborted: Analysis failed
```

[Attempt 2](https://gitlab.com/gitlab-org/cluster-integration/gitlab-agent/-/merge_requests/2416): bump gazelle and `cel-spec` to try to fix the warning. Same error:

```
(03:42:33) ERROR: no such package '@@[unknown repo 'dev_cel_expr' requested from @@gazelle++go_deps+com_github_google_cel_go]//': The repository '@@[unknown repo 'dev_cel_expr' requested from @@gazelle++go_deps+com_github_google_cel_go]' could not be resolved: No repository visible as '@dev_cel_expr' from repository '@@gazelle++go_deps+com_github_google_cel_go'
(03:42:33) ERROR: /home/ci/.cache/bazel/_bazel_ci/9ca8268df963885560ea55034139bea5/external/gazelle++go_deps+com_github_google_cel_go/cel/BUILD.bazel:7:11: no such package '@@[unknown repo 'dev_cel_expr' requested from @@gazelle++go_deps+com_github_google_cel_go]//': The repository '@@[unknown repo 'dev_cel_expr' requested from @@gazelle++go_deps+com_github_google_cel_go]' could not be resolved: No repository visible as '@dev_cel_expr' from repository '@@gazelle++go_deps+com_github_google_cel_go' and referenced by '@@gazelle++go_deps+com_github_google_cel_go//cel:go_default_library'
(03:42:33) ERROR: Analysis of target '//internal/module/kubernetes_api/agent:agent_test' failed; build aborted: Analysis failed
```

The warning with this change looks like this:

```
(03:42:00) DEBUG: /home/ci/.cache/bazel/_bazel_ci/9ca8268df963885560ea55034139bea5/external/gazelle+/internal/bzlmod/go_deps.bzl:607:40: 
Mismatch between versions requested for Go module cel.dev/expr:
  bazel_dep version (MODULE.bazel): 0.23.0 (as "cel-spec")
  Go module version (go.mod):       0.23.1
 
Either ensure that you have
  bazel_dep(module_name = "cel-spec", version = "0.23.1")
in your MODULE.bazel file or downgrade the Go module version via
  bazel run @rules_go//go -- get cel.dev/expr@v0.23.0
```

I don't think the version difference itself causes the problem, but probably the fact of the difference? I cannot use the exact same version since there is no `cel-spec` 0.23.1 in BCR.

p.s. there is no `module_name` attribute, its just `name`. The warning needs to be fixed.


What am I doing wrong and how do I fix it? 🤔

zecke:
Let me have a closer look tonight. Given it says "@@[unknown repo 'dev_cel_expr'.." have you tried to use the following in your attempt #2?

```
bazel_dep(name = "cel-spec", version = "0.23.0", repo_name = "dev_cel_expr")
```

ash2k:
Hm, I did not realize `cel-spec` and `cel.dev/expr` is the same thing! But it [still doesn't work](https://gitlab.com/gitlab-org/cluster-integration/gitlab-agent/-/jobs/9845163365#L148):

```
(04:51:39) ERROR: no such package '@@[unknown repo 'dev_cel_expr' requested from @@gazelle++go_deps+com_github_google_cel_go]//': The repository '@@[unknown repo 'dev_cel_expr' requested from @@gazelle++go_deps+com_github_google_cel_go]' could not be resolved: No repository visible as '@dev_cel_expr' from repository '@@gazelle++go_deps+com_github_google_cel_go'
(04:51:39) ERROR: /home/ci/.cache/bazel/_bazel_ci/9ca8268df963885560ea55034139bea5/external/gazelle++go_deps+com_github_google_cel_go/cel/BUILD.bazel:7:11: no such package '@@[unknown repo 'dev_cel_expr' requested from @@gazelle++go_deps+com_github_google_cel_go]//': The repository '@@[unknown repo 'dev_cel_expr' requested from @@gazelle++go_deps+com_github_google_cel_go]' could not be resolved: No repository visible as '@dev_cel_expr' from repository '@@gazelle++go_deps+com_github_google_cel_go' and referenced by '@@gazelle++go_deps+com_github_google_cel_go//cel:go_default_library'
(04:51:39) ERROR: Analysis of target '//internal/module/kubernetes_api/agent:agent_test' failed; build aborted: Analysis failed
(04:51:39) INFO: Elapsed time: 40.173s, Critical Path: 0.95s
```

zecke:
Something like this is the offender: https://github.com/google/cel-go/blob/master/MODULE.bazel#L82

We stop generating dev_cel_expr as a "go.mod" dependency and instead rely on the bazel_dep (somewhere in the dependency tree). As a workaround `inject_repo(go_deps, "dev_cel_expr")` is your friend. The fix will be to use the bazel dep consistently (or get rid of it).

(Let me have a closer look later tonight)

zecke:
> Something like this is the offender: https://github.com/google/cel-go/blob/master/MODULE.bazel#L82
> 
> We stop generating dev_cel_expr as a "go.mod" dependency and instead rely on the bazel_dep (somewhere in the dependency tree). As a workaround `inject_repo(go_deps, "dev_cel_expr")` is your friend. The fix will be to use the bazel dep consistently (or get rid of it).
> 
> (Let me have a closer look later tonight)

I think the above is accurate. The following workaround does seem to do the right thing (my build on macos fails with gRPC/protobuf/clang incompatibility issues). We should try to get upstream use the bazel-dep instead of a go_repository as dependency for dev_cel_expr.

```
@@ -31,6 +31,7 @@ go_sdk.download(version = "1.23.8")
 
+inject_repo(go_deps, "dev_cel_expr")
 
 # Update version imported via rules_proto_grpc_go
 go_deps.module(
```

zecke:
> I wonder if the "right" fix should be to have the Go module map into the BCR module automatically somehow? Right now this is not the case simply because the BCR module has a weird name (?). Would it work if BCR module was `dev_cel_expr` instead?

Gazelle sees both the Go module and the BCR module for the given importpath and will pick one. Before this change it picked the larger version number and this creates difficulties (see #2060) that cannot be automatically resolved (sometimes override_repo, sometimes inject_repo or it is just broken).

The repo_name is necessary as google/cel-go has picked that name for the dependency and has BUILD files that use that name (and no repo alias exists). 

 
> This seems like a fundamental issue in the bazel module system (I may be missing something!) - we can go from the BCR module name to the Go module name (by looking at the `go.mod`) but not the other way around.

`MODULE.bazel -> bazel_dep -> go_deps.from_file inside the MODULE.bazel`

 
> > We should try to get upstream use the bazel-dep instead of a go_repository as dependency for dev_cel_expr.
> 
> This might be problematic for them - there may not be a corresponding BCR version. At the moment the project is self-contained but that would add an "external" dependency they cannot control to it.
> 
> I wonder why this project even needs to be a BCR module. Probably so that people can consume the proto files from it...

Given that the cel-spec and cel-go are owned by Google. I am sure they can sprinkle a bit of automation on the cel-spec release process to publish the release to the BCR as well. I ended up creating https://github.com/bazelbuild/bazel-central-registry/pull/4500 to see where it is going.


I think the BCR version of cel-spec is useful for non Go languages. In the go only case, we could probably move https://github.com/google/cel-go/blob/845f2a8ec46a147297bb648edf41cfc2b68fb189/MODULE.bazel#L47-L59 to gazelle and be done.

PR #2066, opened 2025-04-14
Trim BCR versions before comparing them to Go versions

fmeum:
**What type of PR is this?**

Bug fix

`go_deps`

Otherwise a BCR version of 1.2.3.bcr.1 will always show a warning when the version in `go.mod` is 1.2.3.

**Other notes for review**


fmeum:
@shs96c

jayconrod:
Not sure if I understand. Do people publish Go modules on BCR with versions like `1.2.3.bcr.1`? I'd strongly recommend sticking to strict semver, but I'd like to understand why this happens.

fmeum:
> Not sure if I understand. Do people publish Go modules on BCR with versions like `1.2.3.bcr.1`? I'd strongly recommend sticking to strict semver, but I'd like to understand why this happens.

Yes, BCR specifically supports extended semver in case something needs to change that doesn't affect the upstream sources. For example, you would release `1.2.3.bcr.1` if upstream's release archive changes or you need to update a MODULE.bazel file for Bazel compatibility that upstream doesn't contain and you patched in.

jayconrod (approved):
I see, so this would be a case where the person creating the BCR release is different than the author of the original project and can't create a patch release?

fmeum:
Yes. In that (admittedly bad) situation, not messing with upstream versions is likely the best we can do.

PR #2067, opened 2025-04-14
[Feature] Enable support for go.work files in non-root modules

stefanpenner:
**What type of PR is this?**

Feature

language/go

As part of #1731, we disabled multiple go.work files due to only time
constraints and uncertainty about its implications. After thorough
consideration and nearly a year of running with this removed without
issues, I feel comfortable enabling it.

Reasoning: Since go.work files are essentially rollups of go.mod files,
and we already support multiple go.mod files, extending support to
go.work files is a natural and safe progression.

**Other notes for review**

### DIFF
--- a/internal/bzlmod/go_deps.bzl
+++ b/internal/bzlmod/go_deps.bzl
@@ -400,9 +400,6 @@ def _go_deps_impl(module_ctx):
             if from_file_tag.go_mod:
                 from_file_tags.append(from_file_tag)
             elif from_file_tag.go_work:
-                if module.is_root != True:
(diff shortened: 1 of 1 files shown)


fmeum (approved):
🎉

fmeum:
Could you change the title to emphasize that this allows go.work files in non-root modules, not multiple go.work files per Bazel module?

fmeum:
Also, please remove the mentions from the PR description. PR descriptions become commit messages, commits are frequently pushed to forks and each individual commit pings the person that's mentioned.

PR #2070, opened 2025-04-15
Generate multiple test rules

aaomidi:
**What type of PR is this?**

Bug Fix

language/go

The issue is as follows:

Imagine you have `package1` in Go. Then in that same directory, you create two tests. One of the tests is `package1_test` and the other is just in `package1`.

These are fundamentally different targets. The `package1` test should have an `embed` statement for the `go_library` that represents `package1`. While the `package1_test` test should import that `go_library` instead of embedding it.

Currently, gazelle generates a single `go_test` target, and that ends up behaving in very weird ways. For example, if you're writing a rule that wraps around a go_library, this construct causes the `go_linker` to get very weird about the `embed` bit.

Fixes #

**Other notes for review**

As it stands, this PR is very likely incorrectly implemented. It's really a hack around some of the restrictions around the code. I'm more than happy to make the refactors necessary to clean this up, or have a maintainer take this PR and bring it up to standard.

I'm also not aware of all the various flags and how the interactions are supposed to be handled there.

### DIFF
--- a/language/go/fileinfo.go
+++ b/language/go/fileinfo.go
@@ -222,7 +222,6 @@ func goFileInfo(path, srcdir string) fileInfo {
 
 	info.packageName = pf.Name.Name
 	if info.isTest && strings.HasSuffix(info.packageName, "_test") {
-		info.packageName = info.packageName[:len(info.packageName)-len("_test")]
 		info.isExternalTest = true
 	}
 
--- a/language/go/generate.go
+++ b/language/go/generate.go
@@ -376,6 +376,24 @@ func (gl *goLang) GenerateRules(args language.GenerateArgs) language.GenerateRes
 		}
 		rules = append(rules, g.generateBin(pkg, libName))
 		rules = append(rules, g.generateTests(pkg, libName)...)
+
+		// Generate test rules for test packages that are not in the main package.
+		for _, testPkg := range goPackageMap {
+			// Only handle _test packages that we haven't already handled.
+			if strings.HasSuffix(testPkg.name, "_test") && testPkg != pkg {
+				if testPkg.importPath == "" {
+					if err := testPkg.inferImportPath(c); err != nil {
+						log.Print(err)
+					}
+				}
+				tests := g.generateTests(testPkg, libName)
+				if len(tests) > 1 {
+					panic(fmt.Sprintf("multiple tests generated for %s", testPkg.name))
+				}
+				tests[0].SetAttr("name", fmt.Sprintf("%s_test", testPkg.name))
+				rules = append(rules, tests...)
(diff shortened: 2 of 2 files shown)


fmeum:
@jayconrod Could you lend your expertise here?

sluongng:
> Imagine you have package1 in Go. Then in that same directory, you create two tests. One of the tests is package1_test and the other is just in package1.
> 
> These are fundamentally different targets. 

This claim is false. The source of truth here should be how `go test` work because that's the official build tool maintained by the Go team.

If you have a setup like this

```
> tail -n +1 *.go
==> a_ext_test.go <==
package a_test

import "testing"

func TestExternal(t *testing.T) {
}

==> a_inner_test.go <==
package a

import "testing"

func TestHelloWorld(t *testing.T) {
	// t.Fatal("not implemented")
}
```

And run `go test -x`, you will see that there is only one `go tool link` invocation in the build log, which outputs only one binary (aka. 1 test target).

```
> go test -x ./... &>log.txt

> grep -c 'tool/darwin_arm64/link' log.txt
1
```

> The package1 test should have an embed statement for the go_library that represents package1. While the package1_test test should import that go_library instead of embedding it.
> 
> Currently, gazelle generates a single go_test target, and that ends up behaving in very weird ways. For example, if you're writing a rule that wraps around a go_library, this construct causes the go_linker to get very weird about the embed bit.

It would be easier to navigate this if you could be more specific about what you are expecting vs what you are observing. Perhaps with a couple of examples or a reproduce project for folks to test against.

As I mentioned on Slack, https://github.com/bazel-contrib/rules_go/blob/master/go/private/rules/test.bzl would be a great read to better understand the internals of `go test`. A quick recap: For each go_test target, we have 2 separate compile actions: one for the internal test package and one for the external test package. Then we issue a link action, which results in one final test executable as the output artifact. Bazel would then take the executable and run it in a TestRunner action.

I'm not against splitting the 2 compilation actions (internal+external packages) into separate go_library targets, but do keep in mind that it's a _VERY_ disruptive change. So it would be nice to understand the justification for doing that: what's not working? Are there alternative options that have yet to be explored?

jayconrod:
I don't have much to add beyond what @sluongng said.

An internal test package and external test package are *not* different targets. They are compiled separately (which rules_go handles correctly), but they are supposed to be linked together. The external test package may even import the internal test package (with the import string for the library package) to call functions that are defined only in the test files.

We cannot change this since it would break a lot of existing projects.

sluongng:
Separately, I raised https://github.com/bazel-contrib/rules_go/issues/4319, which hopefully would improve the readability of rules_go in the future.

PR #2073, opened 2025-04-18
walk: refactor and add Walk2 function

### LINKED ISSUE #1891: Proposal: Lazy indexing for faster subdirectory updates

jayconrod:
**What type of PR is this?**

> Feature

> walk

Walk2 is a replacement for Walk. Walk is now deprecated and implemented
in terms of Walk2. We avoid an incompatible change here since packages
outside Gazelle import and use walk.

Walk2 takes a callback function that accepts its arguments and returns
its results via structs, similar to language.Language.GenerateRules.
This allows us to add more arguments and results without incompatible
changes. This also lets the callback function return errors instead of
simply printing and maybe exiting.

This change will enable lazy indexing. The callback function will be
able to return a list of additional directories to visit.

This change includes a major refactoring of the directory caching and
parallel traversal functionality. Instead of storing data in a trie,
we now use a random access cache. This should let us read directories
more easily that weren't prepopulated. The cache population code still
pre-parses build files and respects excluded and ignored patterns.

Fixes #1891

**Other notes for review**

This is part 2/4. #2072 is the previous change.

### DIFF
--- a/cmd/gazelle/fix-update.go
+++ b/cmd/gazelle/fix-update.go
@@ -313,8 +313,16 @@ func runFixUpdate(wd string, cmd command, args []string) (err error) {
 		}
 	}()
 
-	var errorsFromWalk []error
-	walk.Walk(c, cexts, uc.dirs, uc.walkMode, func(dir, rel string, c *config.Config, update bool, f *rule.File, subdirs, regularFiles, genFiles []string) {
+	walkErr := walk.Walk2(c, cexts, uc.dirs, uc.walkMode, func(args walk.Walk2FuncArgs) walk.Walk2FuncResult {
+		dir := args.Dir
+		rel := args.Rel
+		c := args.Config
+		update := args.Update
+		f := args.File
+		subdirs := args.Subdirs
+		regularFiles := args.RegularFiles
+		genFiles := args.GenFiles
+
 		mrslv.AliasedKinds(rel, c.AliasMap)
 		// If this file is ignored or if Gazelle was not asked to update this
 		// directory, just index the build file and move on.
@@ -327,7 +335,7 @@ func runFixUpdate(wd string, cmd command, args []string) (err error) {
 					ruleIndex.AddRule(c, r, f)
 				}
 			}
-			return
+			return walk.Walk2FuncResult{}
 		}
 
 		// Fix any problems in the file.
@@ -360,7 +368,7 @@ func runFixUpdate(wd string, cmd command, args []string) (err error) {
 			imports = append(imports, res.Imports...)
 		}
 		if f == nil && len(gen) == 0 {
-			return
+			return walk.Walk2FuncResult{}
 		}
 
 		// Apply and record relevant kind mappings.
@@ -390,9 +398,10 @@ func runFixUpdate(wd string, cmd command, args []string) (err error) {
 			return nil, nil
 		}
 
+		var errs []error
 		for _, r := range allRules {
 			if replacementName, err := maybeRecordReplacement(r.Kind()); err != nil {
-				errorsFromWalk = append(errorsFromWalk, fmt.Errorf("looking up mapped kind: %w", err))
+				errs = append(errs, fmt.Errorf("looking up mapped kind: %w", err))
 			} else if replacementName != nil {
 				r.SetKind(*replacementName)
 			}
@@ -409,7 +418,7 @@ func runFixUpdate(wd string, cmd command, args []string) (err error) {
 						continue
 					}
 					if replacementName, err := maybeRecordReplacement(ident.Name); err != nil {
-						errorsFromWalk = append(errorsFromWalk, fmt.Errorf("looking up mapped kind: %w", err))
+						errs = append(errs, fmt.Errorf("looking up mapped kind: %w", err))
 					} else if replacementName != nil {
 						if err := r.UpdateArg(i, &build.Ident{Name: *replacementName}); err != nil {
 							log.Panicf("%s: %v", rel, err)
@@ -456,6 +465,10 @@ func runFixUpdate(wd string, cmd command, args []string) (err error) {
 				ruleIndex.AddRule(c, r, f)
 			}
 		}
+
+		return walk.Walk2FuncResult{
+			Err: errors.Join(errs...),
+		}
 	})
 
 	for _, lang := range languages {
@@ -464,17 +477,8 @@ func runFixUpdate(wd string, cmd command, args []string) (err error) {
 		}
 	}
 
-	if len(errorsFromWalk) == 1 {
-		return errorsFromWalk[0]
-	}
-
-	if len(errorsFromWalk) > 1 {
-		var additionalErrors []string
-		for _, err = range errorsFromWalk[1:] {
-			additionalErrors = append(additionalErrors, err.Error())
-		}
-
-		return fmt.Errorf("encountered multiple errors: %w, %v", errorsFromWalk[0], strings.Join(additionalErrors, ", "))
+	if walkErr != nil {
+		return walkErr
 	}
 
 	// Finish building the index for dependency resolution.
--- a/walk/cache.go
+++ b/walk/cache.go
@@ -1,0 +1,73 @@
+package walk
+
+import (
+	"fmt"
+	"sync"
+)
+
+// cache is an in-memory cache for file system information. Its purpose is to
+// speed up walking over large directory trees (commonly, the entire repo)
+// by parallelizing parts of the walk while still allowing random access
+// to parts of the directory tree that haven't been loaded yet.
+type cache struct {
+	entryMap sync.Map
+}
+
+type cacheEntry struct {
+	doneC chan struct{}
+	info  dirInfo
+	err   error
+}
+
+// get returns the result of calling the given function on the given key.
+//
+// If get has not yet been called with the key, it calls load and saves the
+// result.
+//
+// If get was called earlier with the key, it returns the saved result.
+//
+// get may be called by multiple threads concurrently. Later calls block
+// until the result from the first call is ready.
+func (c *cache) get(key string, load func(rel string) (dirInfo, error)) (dirInfo, error) {
+	// Optimistically load the entry. This is technically unnecessary, but it
+	// avoids allocating a new entry in the case where one already exists.
+	raw, ok := c.entryMap.Load(key)
+	if ok {
+		entry := raw.(*cacheEntry)
+		<-entry.doneC
+		return entry.info, entry.err
+	}
+
+	// Create a new entry. Another goroutine may do this concurrently, so if
+	// another entry is inserted first, wait on that one.
+	entry := &cacheEntry{doneC: make(chan struct{})}
+	raw, loaded := c.entryMap.LoadOrStore(key, entry)
+	if loaded {
+		entry = raw.(*cacheEntry)
+		<-entry.doneC
+		return entry.info, entry.err
+	}
+
+	// Read the directory contents.
+	defer close(entry.doneC)
+	entry.info, entry.err = load(key)
+	return entry.info, entry.err
+}
+
+// getLoaded returns the result of a previous call to get with the same key.
+// getLoaded panics if get was not called or has not returned yet.
+func (c *cache) getLoaded(rel string) (dirInfo, error) {
+	e, ok := c.entryMap.Load(rel)
+	if ok {
+		select {
+		case <-e.(*cacheEntry).doneC:
+		default:
+			ok = false
+		}
+	}
+	if !ok {
+		panic(fmt.Sprintf("getLoaded called for %q before it was loaded", rel))
+	}
+	ce := e.(*cacheEntry)
+	return ce.info, ce.err
+}
--- a/walk/config.go
+++ b/walk/config.go
@@ -69,6 +69,16 @@ func getWalkConfig(c *config.Config) *walkConfig {
 	return c.Exts[walkName].(*walkConfig)
 }
 
+func (wc *walkConfig) clone() *walkConfig {
+	wcCopy := *wc
+	// Trim cap of exclude and follow. We may append to these slices in multiple
+	// goroutines. Doing so should allocate a copy of the backing array.
+	// Other slices are either immutable or replaced when written.
+	wcCopy.excludes = wcCopy.excludes[:len(wcCopy.excludes):len(wcCopy.excludes)]
+	wcCopy.follow = wcCopy.follow[:len(wcCopy.follow):len(wcCopy.follow)]
+	return &wcCopy
+}
+
 func (wc *walkConfig) isExcludedDir(p string) bool {
 	return path.Base(p) == ".git" || wc.ignoreFilter.isDirectoryIgnored(p) || matchAnyGlob(wc.excludes, p)
 }
@@ -151,8 +161,7 @@ func (cr *Configurer) Configure(c *config.Config, rel string, f *rule.File) {
 }
 
 func configureForWalk(parent *walkConfig, rel string, f *rule.File) *walkConfig {
-	wc := &walkConfig{}
-	*wc = *parent
+	wc := parent.clone()
 	wc.ignore = false
 
 	if f != nil {
--- a/walk/config.go
+++ b/walk/config.go
@@ -52,20 +52,29 @@ const (
 // declared generated files, so we can't just stat.
 
 type walkConfig struct {
-	updateOnly bool
-	excludes   []string
-	ignore     bool
-	follow     []string
+	updateOnly          bool
+	ignoreFilter        *ignoreFilter
+	excludes            []string
+	ignore              bool
+	follow              []string
+	validBuildFileNames []string // to be copied to config.Config
 }
 
-const walkName = "_walk"
+const (
+	walkName       = "_walk"
+	walkNameCached = "_walkCached"
+)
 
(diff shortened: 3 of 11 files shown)


jayconrod:
As a bonus, `BenchmarkWalk` is about 40% faster on my machine with this implementation.

jayconrod:
Test failures seem like flakes in CI. Looks like I don't have the ability to hit the retry button in Buildkite, but I can poke with an empty commit when we're closer to merging.

cmd/gazelle/fix-update.go
```diff
@@ -456,6 +465,10 @@ func runFixUpdate(wd string, cmd command, args []string) (err error) {
 				ruleIndex.AddRule(c, r, f)
 			}
 		}
+
+		return walk.Walk2FuncResult{
+			Err: errors.Join(errs...),
```
fmeum, line 470: Just noting that this ties us to Go 1.20+, but we do already have test coverage for the minimum supported version and it appears to be at least that high.

walk/dirinfo.go
```diff
@@ -0,0 +1,157 @@
...
+			sem <- struct{}{} // acquire semaphore for child
+			wg.Add(1)
+			go func() {
+				defer wg.Done()
+				visit(subdirRel)
+			}()
+		}
+	}
+
+	// Call c.get for all directory prefixes. c.get always requires the parent to
```
fmeum, line 130: Comment refers to `c`

walk/cache.go
```diff
@@ -0,0 +1,73 @@
...
+
+// cache is an in-memory cache for file system information. Its purpose is to
+// speed up walking over large directory trees (commonly, the entire repo)
+// by parallelizing parts of the walk while still allowing random access
+// to parts of the directory tree that haven't been loaded yet.
+type cache struct {
+	entryMap sync.Map
+}
+
+type cacheEntry struct {
```
fmeum, line 16: Is memory usage of the cache a potential concern? Depending on the padding of Go structs, we could reduce the footprint by storing only one of info and err (via two different entry types).

fmeum (commented):
I've only reviewed half of the commit yet, but I already like what I'm seeing.

walk/config.go
```diff
@@ -119,22 +133,38 @@ func (*Configurer) KnownDirectives() []string {
...
-	*wcCopy = *wc
-	wcCopy.ignore = false
+	if c.Exts[walkNameCached] != nil {
+		// A normal Configurer implementation would process directives and set
+		// c.Exts[walkName] here. However, we've parallelized the tree walk and
+		// processed the configuration ahead of time in configureForWalk. So instead,
+		// the caller of this method (configure) sets c.Exts[walkNameCache] to the
+		// preprocessed configuration. We copy it to c.Exts[walkName] instead of
+		// re-processing directives.
+		c.Exts[walkName] = c.Exts[walkNameCached]
```
fmeum, line 143: Could we delete the `walkNameCached` key here to save memory?

fmeum (approved):

walk/config.go
```diff
@@ -119,22 +133,38 @@ func (*Configurer) KnownDirectives() []string {
...
+	} else {
+		// If c.Exts[walkNameCached] was not set, process directives normally.
```
jbedard, line 145: Can you explain when it would not be set? I would think it should have always been set during the fs-walk?

walk/walk.go
```diff
@@ -109,213 +110,379 @@ type WalkFunc func(dir, rel string, c *config.Config, update bool, f *rule.File,
...
+	update := !wc.ignore && shouldUpdate && !hasBuildFileError
+	if w.shouldCall(rel, updateParent) {
+		genFiles := findGenFiles(wc, info.file)
+		result := w.wf(Walk2FuncArgs{
+			Dir:          dir,
+			Rel:          rel,
+			Config:       c,
+			Update:       update,
+			File:         info.file,
+			Subdirs:      subdirs,
```
jbedard, line 457: The `subdirs`s was not modified based on `containedByParent` like `regularFiles` was

walk/dirinfo.go
```diff
@@ -0,0 +1,157 @@
...
+		}
+	}
+
+	return info, errors.Join(errs...)
+}
+
+// populateCache loads directory information in a parallel tree traversal.
+// This has no semantic effect but should speed up I/O.
+//
+// populateCache should only be called when recursion is enabled. It attempts
```
jbedard, line 93: "attempts"? Does it not always?

jbedard:
I think this accomplishes everything https://github.com/bazel-contrib/bazel-gazelle/pull/1928 was aiming for? Mainly making BUILD load+parse parse of the concurrent fs traversal, that allowing the `excludes` to be part of the fs-traversal as well.

walk/walk.go
```diff
@@ -350,105 +518,26 @@ func findGenFiles(wc *walkConfig, f *rule.File) []string {
...
-		if !wc.isExcluded(path.Join(f.Pkg, s)) {
+		if !wc.isExcludedFile(path.Join(f.Pkg, s)) {
 			genFiles = append(genFiles, s)
 		}
 	}
 	return genFiles
 }
 
-func shouldFollow(wc *walkConfig, dir, rel string, ent fs.DirEntry) bool {
+func resolveFileInfo(wc *walkConfig, dir, rel string, ent fs.DirEntry) fs.DirEntry {
```
jbedard, line 528: Was there a reason this was reverted? Was there an actual bug being fixed (and should I test+fix that before this PR is merged)?

jayconrod: Ah, I didn't intend to revert #2054. This was a rebasing mistake.

Looking at it closely though, I think it is better to go back to the original behavior of returning the resolved `fs.DirEntry`. I've renamed, documented, and added a test.

I believe the old code has a bug. It calls `os.ReadDir` to get a list of entries, then `IsDir()` to check whether something is a subdirectory, then `shouldFollow`. `IsDir` returns false on a symbolic link though, so this classifies a followable symbolic link to a directory as a regular file. We need to resolve the symbolic link first before calling `IsDir`.

(Technically we should call `filepath.EvalSymlinks` too before calling `os.Stat` in case there's a chain of symbolic links. I haven't changed that here though.)

walk/walk.go
jayconrod: Fixed, but please double check

walk/dirinfo.go
```diff
@@ -0,0 +1,157 @@
...
+	"sync"
+
+	"github.com/bazelbuild/bazel-gazelle/rule"
+)
+
+// dirInfo holds all the information about a directory that Walk2 needs.
+type dirInfo struct {
+	// entries holds the contents of the directory. Symbolic links are resolved
+	// or not depending on configuration. Excluded and ignored files are included.
+	entries []fs.DirEntry
```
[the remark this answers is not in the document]
jayconrod: Good catch, should save some memory.

walk/config.go
jayconrod: Updated comment to indicate this branch is only needed for tests. I didn't want to require that all unit tests go through `walk.Walk2` to call `populateCache`. But it makes this code a little awkward.

walk/config.go
jayconrod: Done, but note this only removes the map entry, an alias to the configuration. It won't save memory unless `c.Exts` is rehashed.

walk/dirinfo.go
jayconrod: Fixed

jayconrod:
There's a CI failure here that's kind of strange. I think it implies I've broken something with excludes, but I can't reproduce it locally. I'll try to track it down tomorrow.

walk/walk.go
```diff
@@ -109,213 +110,383 @@ type WalkFunc func(dir, rel string, c *config.Config, update bool, f *rule.File,
...
+	dir := filepath.Join(c.RepoRoot, rel)
+
+	// Load the build file.
+	info, err := w.cache.get(rel, w.loadDirInfo)
+	if err != nil {
+		w.errs = append(w.errs, err)
+	}
+	hasBuildFileError := err != nil
+
+	configure(w.cexts, w.knownDirectives, c, rel, info.file, info.config)
```
jbedard, line 398: I believe previously [`configure` was not called `if containedByParent`](https://github.com/bazel-contrib/bazel-gazelle/blob/a39914bf8d3d1b953ef79661df1b787b677b209a/walk/walk.go#L159-L162)

walk/walk.go
```diff
@@ -109,213 +110,383 @@ type WalkFunc func(dir, rel string, c *config.Config, update bool, f *rule.File,
...
+	wc := info.config
+	regularFiles := info.regularFiles
+	subdirs := info.subdirs
+
+	if wc.isExcludedDir(rel) {
+		return
+	}
+
+	containedByParent := info.file == nil && wc.updateOnly
+	w.visits[rel] = visitInfo{
```
jbedard, line 408: Shouldn't the `visitInfo` be be saved even if this directory was excluded? I think this should move above the `isExcludedDir(rel)` check?

walk/walk.go
```diff
@@ -109,213 +110,383 @@ type WalkFunc func(dir, rel string, c *config.Config, update bool, f *rule.File,
...
+		var collect func(string, string)
+		collect = func(rel, prefix string) {
+			vi := w.visits[rel]
+			if !vi.containedByParent {
+				return
+			}
+			for _, f := range vi.regularFiles {
+				regularFiles = append(regularFiles, path.Join(prefix, f))
+			}
+			for _, f := range vi.subdirs {
```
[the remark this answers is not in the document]
jayconrod: I'm not sure. That is the behavior at `master`, but it's not documented in `WalkFunc`. It looks like before #1921, we passed in all non-excluded subdirectories, regardless of whether they contain or will contain a build file. So was this an intended side effect?

I don't have a strong preference here either way. I have a slight preference for passing in the list of all subdirectories, since that's a simple contract to explain. Does this make a difference for extensions you work with?

The Go extension supports `testdata` directories that are usually just files (no `BUILD` file, so globbed) but are sometimes importable packages. It tracks whether `GenerateRules` generates stuff in each directory, but it would be simpler to use `Subdirs` as you suggest. If I remember correctly, that's actually why `GenerateRules` is called in post-order.

walk/walk.go
jayconrod: Yes, good catch.

walk/walk.go
jayconrod: Fixed

jayconrod:
I haven't yet been able to reproduce the CI failure. It's not always failing on the same job, so there's something non-deterministic here. It looks like the `# gazelle:exclude` directive is sometimes being ignored.

walk/walk.go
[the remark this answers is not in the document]
jbedard: Another way to think of it would be `Subdirs` are directories that `RegularFiles` does not traverse into, and #1921 maintained that.

I think I've never actually used `Subdirs` though so if you have a strong preference feel free to change it.

jbedard:
I was having a similar issue in https://github.com/bazel-contrib/bazel-gazelle/pull/1928.

I _think_ it had something to do with `# gazelle:exclude` excluding itself such as `# gazelle:exclude .`? Then in that scenario the current directory should not be traverse any further and none of the files should be recorded. Sorry I don't recall the details :/

walk/dirinfo.go
```diff
@@ -0,0 +1,151 @@
...
+		parentConfig = parentInfo.config
+	}
+
+	info.file, err = loadBuildFile(parentConfig, w.rootConfig.ReadBuildFilesDir, rel, dir, entries)
+	if err != nil {
+		errs = append(errs, err)
+	}
+
+	info.config = configureForWalk(parentConfig, rel, info.file)
+
```
jbedard, line 70: I think we need `info.config.isExcludedDir(rel)` here to check if the BUILD we just parsed excludes it's own directory 🤯

See https://github.com/bazel-contrib/bazel-gazelle/pull/1928/files#diff-6b93e57c1bff94796acb4cf6b14ec71e6f903b2895bef3e1ab0eb53bb7e9122eR480-R487

jayconrod:
> I think we need info.config.isExcludedDir(rel) here to check if the BUILD we just parsed excludes itself

Added the check you suggested, but there's a similar check in `walker.visit` already, so it shouldn't cause a change in behavior. `TestExcludeSelf` covers it too.

I added some debug prints to investigate the problem in CI. `language/proto/BUILD.bazel` contains `# gazelle:exclude testdata`, but `language/proto/testdata` was not added to excludes for `language/proto` in CI. It *was* added on my machine. So that's weird.

Running `//:gazelle_ci` with the race detector revealed the problem: multiple goroutines were appending to `wc.excludes`, so they could race and overwrite each other. Trimming the cap for `excludes` and `follow` when cloning `walkConfig` should fix this. CI is green now at least!

walk/dirinfo.go
jayconrod: Done

walk/config.go
```diff
@@ -69,6 +69,16 @@ func getWalkConfig(c *config.Config) *walkConfig {
 	return c.Exts[walkName].(*walkConfig)
 }
 
+func (wc *walkConfig) clone() *walkConfig {
+	wcCopy := *wc
+	// Trim cap of exclude and follow. We may append to these slices in multiple
```
fmeum, line 74: This is such a neat trick!

ptxmac:
This change broke gazelle for me. My test is creating a new `main.go` file in a empty directory and running gazelle. I used git bisect to discover that this is the first commit where running `gazelle` no longer created a new `BUILD.bazel` file

My setup is using bazel 8.2.1 and rules_go 0.55.1

gazelle aren't fond of logging much, so it's challenging to debug. I'll continue investigating it.

jayconrod:
@ptxmac Could you please open a new issue and tag me? Please include enough information to reproduce the issue.

ptxmac:
> @ptxmac Could you please open a new issue and tag me? Please include enough information to reproduce the issue.

after trying to add prints in the code, it is no longer reproducible... i suspect there's something odd happening related to the bazel cache :/

PR #2074, opened 2025-04-18
walk, language: support lazy indexing

jayconrod:
**What type of PR is this?**

> Feature

> walk
> language

A language extension may now return a list of additional directories
to index through GenerateResult.RelsToIndex. Gazelle visits these
directories and adds libraries to the dependency resolution index.
The build files in these directories are not updated unless they
would be otherwise.

The -index flag is still enabled by default, but in addition to
"all" (true), "none" (false), it can now be set to "lazy". In lazy mode,
Gazelle only indexes directories listed on the command line, their
subdirectories (if the -r flag is enabled), and directories returned
by extensions through RelsToIndex. This avoids the need to index
the entire repo on each run of Gazelle.

For #1891

**Other notes for review**

Part 3/4. #2073 is the previous PR.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -522,7 +522,7 @@ The following flags are accepted:
 | current repository. May be :value:`external`, :value:`static` or :value:`vendored`. See                    |
 | `Dependency resolution`_.                                                                                  |
 +-------------------------------------------------------------------+----------------------------------------+
-| :flag:`-index none|lazy|all`                                      | :value:`true`                          |
+| :flag:`-index none|lazy|all`                                      | :value:`all`                           |
 +-------------------------------------------------------------------+----------------------------------------+
 | Determines whether Gazelle should index the libraries in the current repository and whether it             |
 | should use the index to resolve dependencies.                                                              |
@@ -1201,14 +1201,16 @@ below to resolve dependencies:
       See `Avoiding conflicts with proto rules`_.
 
 4. If the import to be resolved is in the library index, the import will be resolved
-   to that library. If ``-index=true``, Gazelle builds an index of library rules in
-   the current repository before starting dependency resolution, and this is how
-   most dependencies are resolved.
+   to that library. If ``-index=all``, Gazelle builds an index of library rules in
+   the current repository before starting dependency resolution. This can take a
+   while, since Gazelle visits every directory in the repository. If
+   ``-index=lazy``, then language extensions may hint at specific directories
+   to visit, which can be much faster.
 
    a) For Go, the match is based on the ``importpath`` attribute.
    b) For proto, the match is based on the ``srcs`` attribute.
 
-5. If ``-index=false`` and a package is imported that has the current ``go_prefix``
+5. If ``-index=none`` and a package is imported that has the current ``go_prefix``
    as a prefix, Gazelle generates a label following a convention. For example, if
    the build file in ``//src`` set the prefix with
    ``# gazelle:prefix example.com/repo/foo``, and you import the library
--- a/README.rst
+++ b/README.rst
@@ -479,7 +479,7 @@ for details.
 
 Both commands accept a list of directories to process as positional arguments.
 If no directories are specified, Gazelle will process the current directory.
-Subdirectories will be processed recursively.
+Subdirectories will be processed recursively by default (unless ``-r=false``).
 
 The following flags are accepted:
 
@@ -522,11 +522,20 @@ The following flags are accepted:
 | current repository. May be :value:`external`, :value:`static` or :value:`vendored`. See                    |
 | `Dependency resolution`_.                                                                                  |
 +-------------------------------------------------------------------+----------------------------------------+
-| :flag:`-index true|false`                                         | :value:`true`                          |
+| :flag:`-index none|lazy|all`                                      | :value:`true`                          |
 +-------------------------------------------------------------------+----------------------------------------+
 | Determines whether Gazelle should index the libraries in the current repository and whether it             |
-| should use the index to resolve dependencies. If this is switched off, Gazelle would rely on               |
-| ``# gazelle:prefix`` directive or ``-go_prefix`` flag to resolve dependencies.                             |
+| should use the index to resolve dependencies.                                                              |
+|                                                                                                            |
+| If `none` or `false`, indexing is disabled, and Gazelle relies purely on conventions to translate          |
+| language-specific import strings into dependency labels.                                                   |
+|                                                                                                            |
+| If `lazy`, Gazelle indexes libraries in directories it visits explicitly. Language extensions may be       |
+| configured to index additional directories through directives like ``# gazelle:go_search``. This mode      |
+| is very fast when recursion is disabled with ``-r=false``.                                                 |
+|                                                                                                            |
+| If `all` or `true`, Gazelle indexes all directories in the repository, even when recursion is disabled.    |
+| This makes dependency resolution simple but can be slow for large repositories.                            |
 +-------------------------------------------------------------------+----------------------------------------+
 | :flag:`-go_grpc_compiler`                                         | ``@io_bazel_rules_go//proto:go_grpc``  |
 +-------------------------------------------------------------------+----------------------------------------+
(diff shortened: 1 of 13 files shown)


language/lang.go
```diff
@@ -189,4 +189,11 @@ type GenerateResult struct {
 	// correspond. These values are passed to Resolve after merge. The type
 	// is opaque since different languages may use different representations.
 	Imports []interface{}
+
+	// RelsToIndex is a list of additional directories to index for dependency
```
jbedard, line 193: Does `a/b` indicate `a/b/*` should also be indexed? So including the root directory causes a full indexing of the repo?

jbedard: Should `a/b/c` index `a/b` even if `a/b/c` does not exist (but its parent directories do)?

cmd/gazelle/fix-update.go
```diff
@@ -348,6 +350,7 @@ func runFixUpdate(wd string, cmd command, args []string) (err error) {
 		// Generate rules.
 		var empty, gen []*rule.Rule
 		var imports []interface{}
+		var relsToVisit []string
```
jbedard, line 353: Shouldn't this be a set to avoid duplicates?

walk/walk.go
```diff
@@ -210,6 +210,58 @@ func Walk2(c *config.Config, cexts []config.Configurer, dirs []string, mode Mode
...
 	}
 
+	// Visit additional directories that extensions requested for indexing.
+	for len(w.relsToVisit) > 0 {
+		// Don't simply range over relsToVisit. We may append more.
+		relToVisit := w.relsToVisit[0]
+		w.relsToVisit = w.relsToVisit[1:]
+
+		dirToVisit := filepath.Join(c.RepoRoot, relToVisit)
+		if fi, err := os.Stat(dirToVisit); err != nil || !fi.IsDir() {
```
jbedard, line 220: Similar to a comment above.. what if `a/b/c` does not exist but `a/b` does? Can't we avoid this stat and just `break` the `for` below once we reach a non-existing directory?

README.rst
```diff
@@ -522,11 +522,20 @@ The following flags are accepted:
 | current repository. May be :value:`external`, :value:`static` or :value:`vendored`. See                    |
 | `Dependency resolution`_.                                                                                  |
 +-------------------------------------------------------------------+----------------------------------------+
-| :flag:`-index true|false`                                         | :value:`true`                          |
+| :flag:`-index none|lazy|all`                                      | :value:`true`                          |
```
jbedard, line 525: I thought we could avoid this opt-in and gazelle would be lazy if all languages implement the API for it.

In what scenario would a user want to manually indicate `all` vs `lazy` if languages can indicate their support for it automatically?

walk/walk.go
```diff
@@ -210,6 +210,58 @@ func Walk2(c *config.Config, cexts []config.Configurer, dirs []string, mode Mode
...
+				c := parentCfg.Clone()
+				w.visit(c, rel, false)
+			}
+
+			if rel == relToVisit {
+				break
+			}
+		}
+
+		if _, ok := w.visits[relToVisit]; ok {
```
jbedard, line 256: Can't the for loop above invoke `visit` up to **and including** `relToVisit` instead of requiring this extra `if`+`visit` out here?

cmd/gazelle/fix-update.go
jbedard: Or maybe it doesn't matter because this just gets passed along and `relsToVisitSeen` handles it later?

README.rst
jayconrod: I'd like `-index=lazy` to be the default in the future, but it's a big change in default behavior, even if it's triggered by all language extensions opting in.

For example, let's say we implement `go_search` and `proto_search`, and those two plugins opt in so that the default build of Gazelle is opted in. If a user isn't aware of that change, they won't have added any `go_search` directives, so they may see dependency resolution stop working if they're using a non-standard layout.

I think it's best to implement the search directives in a few languages and give time for users to adopt them and provide feedback. Later on, we can look at the best way to make `-index=lazy` the default.

(I don't necessarily think extensions should opt in by the way. A lot of smaller, internal extensions may not even perform dependency resolution and shouldn't need to do anything.)

cmd/gazelle/fix-update.go
jayconrod: Yeah, `relsToVisitSeen` will handle it later.

I thought about this, but I expect duplicates to be uncommon here (multiple extensions want to index the same directory), so allocating a map to deduplicate might cost more memory than it saves.

walk/walk.go
jayconrod: Yes, removed! Actually it already did, so this was purely redundant.

language/lang.go
jayconrod: > Does a/b indicate a/b/* should also be indexed? So including the root directory causes a full indexing of the repo?

Good question. No, only `a/b` should be indexed, not subdirectories. I've updated the doc comment to be more explicit about this and disabled recursion in `walk.Walk2`.

> Should a/b/c index a/b even if a/b/c does not exist (but its parent directories do)?

I'm not sure. Initially I thought no, but your comment below about avoiding the stat made me think it's simpler to index parents. Changed to that.

walk/walk.go
jayconrod: Yes, converted this to a check on whether `loadDirInfo` returns `fs.ErrNotExist`. So there's no extra I/O, and the behavior is probably a bit simpler to explain.

jayconrod:
This is updated with `master` and should be ready for review now.

walk/walk_test.go
```diff
@@ -684,6 +684,57 @@ func TestSubdirsContained(t *testing.T) {
...
+		},
+	})
+	updateDir := filepath.Join(dir, "update")
+	err := Walk2(c, cexts, []string{updateDir}, UpdateDirsMode, func(args Walk2FuncArgs) Walk2FuncResult {
+		visitedRels = append(visitedRels, args.Rel)
+		if args.Update {
+			updatedRels = append(updatedRels, args.Rel)
+		}
+		res := Walk2FuncResult{}
+		if args.Rel == "update" {
```
fmeum, line 710: Could you extend this test to also return an additional dir to visit when `extra/a` is visited? Perhaps also include `update` to verify that we don't get into an endless loop.

fmeum (approved):

walk/walk_test.go
jayconrod: Yes, done. Also verified that parents of non-existent directories are configured (now fixed).

README.rst
jbedard, line 525: The default is `true` but it indicates the options are `none|lazy|all` 🤷

jayconrod: Fixed to `all`. Also updated the dependency resolution notes.

language/lang.go
```diff
@@ -189,4 +189,12 @@ type GenerateResult struct {
...
+	// resolution, expressed as slash-separated paths relative to the repository
+	// root, or "" for the root directory itself. If indexing is enabled,
+	// libraries in these directories are indexed before dependencies are
```
jbedard, line 196: *in these directories **and parent directories**?

walk/walk.go
```diff
@@ -217,6 +217,54 @@ func Walk2(c *config.Config, cexts []config.Configurer, dirs []string, mode Mode
...
+				c := parentCfg.Clone()
+				w.visit(c, rel, false)
+			}
+
+			if rel == relToVisit {
+				break
+			}
+		}
+
+		if c.Strict && len(w.errs) > 0 {
```
jbedard, line 264: Should this be checked after each `.visit` call? If not can we remove the check below the first `w.visit(c, "", false)` and just have this one (but moved out of the loop)?

language/lang.go
jayconrod: No: the parent directories are configured (the `.Configure` method is called on each extension), but the callback to `Walk2` is not called.

This, together with @fmeum's comment above made me think more about this, change the logic a bit, and add a test. We might first visit and configure a directory as a parent directory. Later, something else could add it to `RelsToVisit`. The callback should then be called, but we shouldn't reconfigure it.

walk/walk.go
jayconrod: This makes us break out of the loop early in strict mode if there's an error. `walker.visit` does the same thing and doesn't recurse into subdirectories after an error. I think it's important to be consistent there.

(I'm not sure I really like strict mode, but Gazelle doesn't really provide a good framework for error handling in extensions, so perhaps it's needed.)

jbedard: Doesn't that mean it should be invoked directly after `w.visit()` and not down here then?

jbedard:
I feel like this needs significantly more tests and experimenting with various languages. Should this be marked as EXPERIMENTAL so we can potentially still make changes to the API?

jayconrod:
Seems reasonable to mark it experimental for now. I added a comment in #2075.

walk/walk.go
jayconrod: Yes, follow-up in https://github.com/bazel-contrib/bazel-gazelle/pull/2077

PR #2077, opened 2025-04-25
walk: better error checking in strict mode for RelsToVisit

jayconrod:
**What type of PR is this?**

> Bug fix

> walk

When visiting directories in RelsToVisit, if strict mode is enabled, check for errors and return early after every directory component.

Strict mode implies that we return as soon as we find an error elsewhere, so let's do that here too.

For #1891

**Other notes for review**

Follow-up to [comment](https://github.com/bazel-contrib/bazel-gazelle/pull/2074#discussion_r2061002644).


jayconrod:
cc @jbedard, follow-up to your comment.

fmeum (approved):

PR #2081, opened 2025-04-29
Add `package_metadata` to `REPO.bazel`

Yannic:
**What type of PR is this?**
> Feature

> go_repository

This change adds the `package_metadata` from `@package_metadata` to replace the deprecated `package_info` from `@rules_license`.

### DIFF
--- a/internal/go_repository.bzl
+++ b/internal/go_repository.bzl
@@ -374,7 +374,12 @@ def _go_repository_impl(ctx):
         repo_file = ctx.path("REPO.bazel")
         if not repo_file.exists:
             ctx.file("REPO.bazel", """\
-repo(default_package_metadata = ["//:gazelle_generated_package_info"])
+repo(
+    default_package_metadata = [
+        "//:gazelle_generated_package_info",
+        "//:package_metadata",
+    ],
+)
 """)
 
             # Modify the top-level build file after patches have been applied as the
@@ -402,11 +407,20 @@ def _generate_package_info(*, importpath, version):
     # scheme:type/namespace/name@version?qualifiers#subpath
     purl = "pkg:golang/{namespace_and_name}@{version}".format(
         namespace_and_name = importpath,
-        version = version,
-    ) if version else None
+        version = version if version else "unknown",
+    )
 
     return """
+load("@package_metadata//rules:package_metadata.bzl", "package_metadata")
 load("@rules_license//rules:package_info.bzl", "package_info")
+
+package_metadata(
+    name = "package_metadata",
+    purl = {purl},
+    visibility = [
+        "//:__subpackages__",
+    ],
+)
 
 package_info(
     name = "gazelle_generated_package_info",
--- a/internal/go_repository.bzl
+++ b/internal/go_repository.bzl
@@ -405,10 +405,15 @@ def _generate_package_info(*, importpath, version):
     # See specification:
     # https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#golang
     # scheme:type/namespace/name@version?qualifiers#subpath
-    purl = "pkg:golang/{namespace_and_name}@{version}".format(
-        namespace_and_name = importpath,
-        version = version if version else "unknown",
-    )
+    if version:
+        purl = "pkg:golang/{namespace_and_name}@{version}".format(
+            namespace_and_name = importpath,
+            version = version,
+        )
+    else:
+        purl = "pkg:golang/{namespace_and_name}".format(
+            namespace_and_name = importpath,
+        )
 
     return """
 load("@package_metadata//rules:package_metadata.bzl", "package_metadata")
--- a/tests/bcr/go_mod/tests.bzl
+++ b/tests/bcr/go_mod/tests.bzl
@@ -1,4 +1,5 @@
 load("@bazel_features//:features.bzl", "bazel_features")
+load("@package_metadata//providers:package_metadata_info.bzl", "PackageMetadataInfo")
 load("@rules_license//rules:providers.bzl", "PackageInfo")
 load("@rules_testing//lib:analysis_test.bzl", "analysis_test", "test_suite")
 load("@rules_testing//lib:truth.bzl", "subjects")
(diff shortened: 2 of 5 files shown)


internal/go_repository.bzl
```diff
@@ -402,12 +407,21 @@ def _generate_package_info(*, importpath, version):
...
 
+package_metadata(
+    name = "package_metadata",
+    purl = {purl},
+    visibility = [
+        "//:__subpackages__",
+    ],
+)
+
 package_info(
```
fmeum, line 425: Let's get rid of this now that rules_license is deprecated.

Yannic: Are we sure that users currently don't depend on it?

Yannic:
CI failure doesn't immediately look related to me:

```

(17:12:24) ERROR: /private/var/tmp/_bazel_buildkite/e59377bde3e5c8f74ba33e9469092e03/external/io_bazel_stardoc/src/main/java/com/google/devtools/build/skydoc/renderer/BUILD:35:13 Building external/io_bazel_stardoc/src/main/java/com/google/devtools/build/skydoc/renderer/librenderer_lib.jar (4 source files) [for tool] failed: (Exit 1): cc_wrapper.sh failed: error executing CppCompile command (from target @@zlib//:zlib)
--
  | (cd /private/var/tmp/_bazel_buildkite/e59377bde3e5c8f74ba33e9469092e03/sandbox/darwin-sandbox/11/execroot/bazel_gazelle && \
  | exec env - \
  | PATH=/Users/buildkite/Library/Caches/bazelisk/downloads/sha256/45cca81a839d7495258b19ee8371c7b891f350586ef37b9940f7b531eb654cc8/bin:/opt/homebrew/bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/usr/bin:/bin:/usr/sbin:/sbin:/opt/homebrew/bin \
  | PWD=/proc/self/cwd \
  | external/local_config_cc/cc_wrapper.sh -U_FORTIFY_SOURCE -fstack-protector -Wall -Wthread-safety -Wself-assign -Wunused-but-set-parameter -Wno-free-nonheap-object -fcolor-diagnostics -fno-omit-frame-pointer -g0 -O2 '-D_FORTIFY_SOURCE=1' -DNDEBUG -ffunction-sections -fdata-sections -MD -MF bazel-out/darwin_arm64-opt-exec-ST-a828a81199fe/bin/external/zlib/_objs/zlib/zutil.d '-frandom-seed=bazel-out/darwin_arm64-opt-exec-ST-a828a81199fe/bin/external/zlib/_objs/zlib/zutil.o' -iquote external/zlib -iquote bazel-out/darwin_arm64-opt-exec-ST-a828a81199fe/bin/external/zlib -isystem external/zlib/zlib/include -isystem bazel-out/darwin_arm64-opt-exec-ST-a828a81199fe/bin/external/zlib/zlib/include '-mmacosx-version-min=15.4' -g0 -Wno-unused-variable -Wno-implicit-function-declaration -no-canonical-prefixes -Wno-builtin-macro-redefined '-D__DATE__="redacted"' '-D__TIMESTAMP__="redacted"' '-D__TIME__="redacted"' -c external/zlib/zutil.c -o bazel-out/darwin_arm64-opt-exec-ST-a828a81199fe/bin/external/zlib/_objs/zlib/zutil.o)
  | # Configuration: 99750b55bc0d4b9dbc994fcaccb3e8a4a8c09cb150573bb78b40a08a2cba00ba
  | # Execution platform: @@internal_platforms_do_not_use//host:host
  |  
  | Use --sandbox_debug to see verbose messages from the sandbox and retain the sandbox build root for debugging
```

internal/go_repository.bzl
Yannic: We do want to make `rules_license` a shim for similar providers and rules in package_metadata. The expectation is, however, that that's primarily for migration and we don't expect to update `rules_license` as the replacement evolves. So the expectation is that people that do migrate eventually (say, within 1-2 years, obviously assuming people do upgrade Bazel and dependencies. If you're stuck on Bazel 3 because you certified your toolchain, there's very little we can do).

What makes this somewhat complicated is that the APIs are quite different (on purpose to fix the limitations of rules_license), so users may not be able to get more than the most basic functionality out of rules_license (not that there's a lot today, just setting the expectation here).

It's really really unfortunate that `rules_license` tagged `v1.0.0` which gave people the impression it was ready to use when it really is not. I'll bring this up on Thursday in our meeting to see what we can do about it.

But cycling back to the original question, I think that we should keep both around for now.

Yannic: FWIW, there's currently 80 modules depending on `rules_license` according to registry.bazel.build. So a somewhat managable number to migrate

Yannic: let's do this in 2 stages. I'll follow-up removing the old package_info targets

internal/go_repository.bzl
```diff
@@ -402,12 +407,21 @@ def _generate_package_info(*, importpath, version):
     # scheme:type/namespace/name@version?qualifiers#subpath
     purl = "pkg:golang/{namespace_and_name}@{version}".format(
         namespace_and_name = importpath,
-        version = version,
-    ) if version else None
+        version = version if version else "unknown",
```
fmeum, line 410: Is this a well-known special string? Based on what the spec says, shouldn't we leave this empty and also drop the `@`?

tests/bcr/go_mod/tests.bzl
```diff
@@ -21,6 +22,11 @@ def _test_package_info_impl(env, target):
     # available.
     if not bazel_features.proto.starlark_proto_info:
         return
+    
+    # @package_metadata providers
+    env.expect.that_target(target).has_provider(PackageMetadataInfo)
```
fmeum, line 27: Could you add asserts on the contents of this provider?

internal/go_repository.bzl
Yannic: You're right, removed

fmeum (approved):

PR #2083, opened 2025-04-30
table fix

sluongng:
I accidentally broke the table in https://github.com/bazel-contrib/bazel-gazelle/pull/2071

So here is a fix to extend the existing table with a few additional columns.
After this, Github should render the file correctly.

Note: markdown would have been a bit easier with table syntax.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -483,169 +483,169 @@ Subdirectories will be processed recursively by default (unless ``-r=false``).
 
 The following flags are accepted:
 
-+-------------------------------------------------------------------+----------------------------------------+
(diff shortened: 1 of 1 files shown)


fmeum (approved):
@jayconrod What do you think of converting all the docs to Markdown to simplify future contributions?

jayconrod:
I'm fine with converting to Markdown. The one advantage that Restructured Text has is that it allows more legible tables in code with interior formatting, including paragraphs. I don't think Markdown allows things like lists in tables. So most tables will need to be raw HTML. It doesn't really feel like there's a good solution here.

PR #2085, opened 2025-05-02
proto: include go_proto in the default grpc compilers

sluongng:
Right now we flip between proto compilers vs grpc compilers based on
whether the proto file has a service definition or not.

It's possible for a proto file to contain both service definitions
and message definitions. So we would want to include go_proto in those
cases.


fmeum:
Could we instead do what you suggested in https://bazelbuild.slack.com/archives/CDBP88Z0D/p1746198402121529?thread_ts=1746189567.685639&cid=CDBP88Z0D and add the configured proto and/or grpc compilers depending on the number of service and/or message definitions?

sluongng:
I explored that briefly and went down a pretty deep rabbit hole and then backtracked. Can give that another try.

jayconrod (approved):
I'm curious what happens if a .proto file has only service definitions (no messages or enums) and we run `go_proto` on it? I expected that we'd run that compiler unconditionally. Does it do something bad? I assume it's safe if we can run it on an empty .proto file.

This is contrived but if the .proto file itself is generated, Gazelle may be aware of it but may act as if it has neither messages no services.

sluongng:
Good question, given https://github.com/bazel-contrib/rules_go/commit/077f15fe11b9da6aa0e3271db1260929f04fef87 (which was reverted shortly after merged), I think @linzhp might know about this better.

zecke:
> Good question, given [bazel-contrib/rules_go@077f15f](https://github.com/bazel-contrib/rules_go/commit/077f15fe11b9da6aa0e3271db1260929f04fef87) (which was reverted shortly after merged), I think @linzhp might know about this better.

Empirically with gazelle=0.45.0 and rules_go=0.56.1: An empty (no content, just proto syntax and options) will still generate the .pb.go file that at least includes the "FileDescriptor" definition.

PR #2095, opened 2025-05-16
language/go: add implicit go_search path when prefix is set

### LINKED ISSUE #2094: language/go: add implicit go_search path for a directory containing a go.mod or prefix directive
When lazy indexing is enabled, one expects to be able to find packages in one's own module without extra configuration.

jayconrod:
**What type of PR is this?**

> Bug fix

> language/go

See title and issue.

Fixes #2094

**Other notes for review**

Also skip a directory in RelsToVisit with any error, not just ErrNotExist. Before, the rest of this change caused //cmd/gazelle:gazelle_test TestMapKindEmbeddedResolver to fail because a .proto file gets added to RelsToIndex, and proto files are not directories. We should tolerate this kind of error.

### DIFF
--- a/walk/walk.go
+++ b/walk/walk.go
@@ -253,8 +253,11 @@ func Walk2(c *config.Config, cexts []config.Configurer, dirs []string, mode Mode
 					if getWalkConfig(parentCfg).isExcludedDir(rel) {
 						break
 					}
-					if _, err := w.cache.get(rel, w.loadDirInfo); errors.Is(err, fs.ErrNotExist) {
-						// Directory does not exist.
+					if _, err := w.cache.get(rel, w.loadDirInfo); err != nil {
+						// Error loading directory. Most commonly, this is because the
+						// directory doesn't exist, but it could actually be a file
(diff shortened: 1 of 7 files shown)


walk/walk.go
```diff
@@ -253,8 +253,11 @@ func Walk2(c *config.Config, cexts []config.Configurer, dirs []string, mode Mode
 					if getWalkConfig(parentCfg).isExcludedDir(rel) {
 						break
 					}
-					if _, err := w.cache.get(rel, w.loadDirInfo); errors.Is(err, fs.ErrNotExist) {
-						// Directory does not exist.
+					if _, err := w.cache.get(rel, w.loadDirInfo); err != nil {
+						// Error loading directory. Most commonly, this is because the
+						// directory doesn't exist, but it could actually be a file
+						// or we don't have permission, or some other I/O error.
+						// Skip it.
```
fmeum, line 260: Is skipping without any indication that we did so the right way to handle this? Maybe just skip non-existent dirs and unexpected files but log everything else?

fmeum (approved):

PR #2099, opened 2025-05-25
Avoid allocations in fileNameInfo

dzbarsky:
**What type of PR is this?**
Performance fix

language/go

Allocation from `strings.Split` shows up around 2-3% in pprof of Gazelle runs in my repo. We can parse these names without allocating.
While here, I removed the duplicate computation of `path.Ext(name)`, rearranged the initial computation of `ext` a bit to avoid computing a value that will be overwritten immediately, and used a switch to check the initial char instead of `strings.HasPrefix`, which should give the compiler the most freedom to optimize

The test coverage for this file is actually very good which gave me confidence that the new code is equivalent

Before:
```
goos: darwin
goarch: arm64
cpu: Apple M1 Max
BenchmarkFileNameInfo-10    	 3516370	       335.7 ns/op	     176 B/op	       4 allocs/op
```

After:
```
goos: darwin
goarch: arm64
cpu: Apple M1 Max
BenchmarkFileNameInfo-10    	 6399490	       178.9 ns/op	       0 B/op	       0 allocs/op
```

Fixes #

**Other notes for review**

### DIFF
--- a/language/go/fileinfo.go
+++ b/language/go/fileinfo.go
@@ -137,44 +137,59 @@ const (
 // a file. It does not read data from the file.
 func fileNameInfo(path_ string) fileInfo {
 	name := filepath.Base(path_)
-	var ext ext
-	switch path.Ext(name) {
-	case ".go":
-		ext = goExt
-	case ".c", ".cc", ".cpp", ".cxx", ".m", ".mm":
-		ext = cExt
-	case ".h", ".hh", ".hpp", ".hxx":
-		ext = hExt
-	case ".s":
-		ext = sExt
-	case ".S":
-		ext = csExt
-	case ".proto":
-		ext = protoExt
+	nameExt := path.Ext(name)
+
+	ext := unknownExt
+	switch name[0] {
+	case '.', '_':
 	default:
-		ext = unknownExt
-	}
-	if strings.HasPrefix(name, ".") || strings.HasPrefix(name, "_") {
-		ext = unknownExt
+		switch nameExt {
+		case ".go":
+			ext = goExt
+		case ".c", ".cc", ".cpp", ".cxx", ".m", ".mm":
+			ext = cExt
+		case ".h", ".hh", ".hpp", ".hxx":
+			ext = hExt
+		case ".s":
+			ext = sExt
+		case ".S":
+			ext = csExt
+		case ".proto":
+			ext = protoExt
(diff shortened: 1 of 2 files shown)


language/go/fileinfo.go
```diff
@@ -137,44 +137,59 @@ const (
...
-	case ".s":
-		ext = sExt
-	case ".S":
-		ext = csExt
-	case ".proto":
-		ext = protoExt
+	nameExt := path.Ext(name)
+
+	ext := unknownExt
+	switch name[0] {
```
fmeum, line 143: Could this crash if name is empty?

language/go/fileinfo.go
```diff
@@ -137,44 +137,59 @@ const (
...
+	toParse := name[:len(name)-len(nameExt)]
+	if strings.HasSuffix(toParse, "_test") {
 		isTest = ext == goExt
-		l = l[:len(l)-1]
+		toParse = toParse[:len(toParse)-len("_test")]
 	}
+
+	var segments [2]string
+	n := 0
+	for n < 2 {
```
fmeum, line 175: I could see this being more readable if we unroll it and give the intermediate values speaking names.

dzbarsky: I originally wrote it like this, but then went with the loop variant so the logic would more obviously match what was happening before. Were you thinking of a different spelling? Happy to go whichever way you prefer!

```
+       if i := strings.LastIndex(toParse, "_"); i != -1 {
+               goarch = toParse[i+1:]
+               toParse = toParse[:i]
+               if j := strings.LastIndex(toParse, "_"); j != -1 {
+                       goos = toParse[j+1:]
+               }
+
+               knownOS := rule.KnownOSSet[goos]
+               knownArch := rule.KnownArchSet[goarch]
+               if !knownOS || !knownArch {
+                       if rule.KnownOSSet[goarch] {
+                               goos = goarch
+                               goarch = ""
+                       } else {
+                               if !knownOS {
+                                       goos = ""
+                               }
+
+                               if !knownArch {
+                                       goarch = ""
+                               }
+                       }
+               }
        }
```

dzbarsky: FWIW the loop variant benchmarks 10% faster than the above on my M1 Max :)

fmeum (approved):

PR #2100, opened 2025-05-25
Speedup known os/arch checks

dzbarsky:
**What type of PR is this?**
Performance

language/go, although we could potentially reuse this for the `rule` package as well. I left it as-is since it has comments around these platforms being deprecated and needing to live in //language/go

The switch-case lets the compiler optimize more than the map lookup. This is a further 30-40% win on the benchmark added in https://github.com/bazel-contrib/bazel-gazelle/pull/2099. I also fixed the other callers in //language/go because it's also simpler

Fixes #

**Other notes for review**


fmeum (commented):
How much of a win is this for a real Gazelle execution? I'm asking since this adds a non-trivial amount of friction especially for new contributions in this area (say adding platforms).

dzbarsky:
> How much of a win is this for a real Gazelle execution? I'm asking since this adds a non-trivial amount of friction especially for new contributions in this area (say adding platforms).

It's around .7-1% for us. Is the friction related to having to manually update the checked-in file? I think having those checked-in is annoying; the comments suggest it's needed so code works with `go get` but I'm not sure if that's a constraint we care about anymore. Assuming we still do, here's some ideas to minimize the friction:
1) Have the checked-in file use the Set implementation instead of the generated switch case. It means everything will still work, no manual actions are necessary, but Gazelle will be a bit faster when built with bazel because it will use the optimized implementation
2) We can add some comments to the platforms list listing the commands to run to update the file in the repo.

As I typed this out, I think (1) feels like a great solution; the only drawback is it might make it a tad bit more confusing to navigate the code, but we can also leave a comment in the functions explaining that the optimized implementation is generated at build time. WDYT?

fmeum:
I think we do care as `go build` is used to bootstrap Gazelle for `go_repository`. 

Given that this shows a relevant improvement, I'm okay with the change. I think that the current approach is the closest match to idiomatic Go, so let's keep it as is. Thanks for sending this!

fmeum:
You probably need to update bazel_skylib to get around the failure.

dzbarsky:
> You probably need to update bazel_skylib to get around the failure.

That turned out to be non-trivial because the WORKSPACE version is pretty old and skylib split the bzl_library gazelle plugin into a separate workspace. Given bzlmod is the only future in medium term, I just applied the patch to green up CI.

dzbarsky:
ok @fmeum we are back to `go generate`, sorry about the confusion :)

rule/platform.go
```diff
@@ -47,6 +47,8 @@ func (p Platform) String() string {
 // will generate multi-platform build files using these tags. rules_go and
 // Bazel may not actually support all of these.
 //
+// If updating this list, please run `go generate ./...`
```
fmeum: Could you use `@rules_go//go` here so that users don't need a local Go SDK?

dzbarsky: done! (sorry I am used to having a bazel-env type thing so `go` is already aliased to the bazel-provided toolchain :) )

WORKSPACE
```diff
@@ -8,6 +8,8 @@ load(
 git_repository(
     name = "bazel_skylib",
     commit = "df3c9e2735f02a7fe8cd80db4db00fec8e13d25f",  # `master` as of 2021-08-19
```
fmeum: Sorry for missing this - can't we just update to latest here instead of adding a patch? This issue has been fixed for more than three years.

dzbarsky: It was really not obvious to me how to make it work with WORKSPACE - the problem is that skylib got split into multiple modules and the bzl_library gazelle plugin (which is used in gazelle itself) is now a submodule. I tried to follow the WORKSPACE boilerplate to bring it in but encountered some issues (don't remember exactly what they were since it's been a few weeks). Figured a 1-line patch was the lesser evil, but if you want to give it a shot maybe you'll have better luck?

dzbarsky: rebased on top of https://github.com/bazel-contrib/bazel-gazelle/pull/2116

fmeum (approved):

PR #2105, opened 2025-06-02
Replace illegal characters in underscores when deriving proto RuleName

### LINKED ISSUE #2103: Generated rule names should replace illegal charcters
When defining 2 proto files: `test.proto` and `test-extensions.proto` (under --proto=file) would generate following definitions

```bazel
proto_library(
    name = "test_proto",
    srcs = ["test.proto"],
    deps = ["@protobuf//:any_proto"],
)

proto_library(
    name = "extensions_proto",
    srcs = ["test-extensions.proto"],
    deps = [":test_proto"],
)
```

The `-` character is illegal for rule labels, however I would expect that the resulting rule would be named `test_extensions_proto` instead of `extensions_proto`. 

If we'd create 1 more file named `example-extensions.proto` it would lead to creating 2 `proto_library` rules with the same name. 

Additionally when using `//language/go` in the same setup it would create a single `go_proto_library` however under `--proto=file` there should be 2 rules defined: 

```bazel
go_proto_library(
    name = "src_go_proto",
    importpath = "src",
    protos = [
        ":extensions_proto",
        ":test_proto",
    ],
)
``` 
This problem does not exist if the proto file uses underscores in it's name. 

Probably the gazelle should replace all subsequent illegal characters with an underscore, eg. `foo--bar@baz.proto` should be replaced with `foo_bar_baz.proto`

WojciechMazur:
**What type of PR is this?**

Bug fix

language/proto

Also language/go when consuing outputs of language/proto 

It adjusts the rules of generating proto_library rule names to replace illegal characters with underscores instead of trimming the illegal prefix. 

Fixes #2103

**Other notes for review**

The https://bazel.build/concepts/labels#target-names allows more identifiers, we can also consider accepting all of these below. 
> Target names must be composed entirely of characters drawn from the set a–z, A–Z, 0–9, and the punctuation symbols `!%-@^_"#$&'()*-+,;<=>?[]{|}~/.`

### DIFF
--- a/internal/bzlmod/default_gazelle_overrides.bzl
+++ b/internal/bzlmod/default_gazelle_overrides.bzl
@@ -112,12 +112,18 @@ DEFAULT_DIRECTIVES_BY_PATH = {
     "k8s.io/api": [
         "gazelle:proto disable",
     ],
+    "k8s.io/apiserver": [
+        "gazelle:proto disable",
+    ],
     "k8s.io/apiextensions-apiserver": [
         "gazelle:proto disable",
     ],
     "k8s.io/apimachinery": [
         "gazelle:go_generate_proto false",
         "gazelle:proto_import_prefix k8s.io/apimachinery",
+    ],
+    "k8s.io/cri-api": [
+        "gazelle:proto disable",
     ],
     "k8s.io/kubelet": [
         "gazelle:proto disable",


WojciechMazur:
@jayconrod

jayconrod (approved):

PR #2106, opened 2025-06-03
perf: remove extraneous isExcluded(dir) checks

jbedard:
**What type of PR is this?**

Perf/cleanup

all

I'm uncertain if we just need more test cases or if these changes are safe. WDYT?

Ideally only the `isExcludedDir` calls within `loadDirInfo` would be needed. That should filter out all files+dirs loaded from the fs-walk.

Are there scenarios in the `RelsToVisit` logic where we need to check as well? Can that not be done by inspecting the `loadDirInfo` result instead of duplicating the exclusion logic? For example the call within `walk()` that I moved and within `Walk2`... I feel like we shouldn't need either of those...

**Other notes for review**

### DIFF
--- a/walk/dirinfo.go
+++ b/walk/dirinfo.go
@@ -109,13 +109,9 @@ func (w *walker) populateCache(rels []string) {
 		if err != nil {
 			return
 		}
-		wc := info.config
 
 		for _, subdir := range info.subdirs {
 			subdirRel := path.Join(rel, subdir)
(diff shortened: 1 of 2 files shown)


jayconrod (approved):
Looks good. The tests here caught a lot of my mistakes while I was working on this code, so if they're passing, I'm fairly confident the change is safe. :)

PR #2109, opened 2025-06-08
indexing all parents of visited directories

linzhp:
**What type of PR is this?**
Feature

walk

This is an alternative implementation of #2101. As suggested in the comments in #2101:

> Alternatively, what do you think we should index all parent directories here unconditionally? It wouldn't be that much extra work since we're already parsing them and applying configuration.

This also simplify the logic, so we no longer need to distinguish between "visited" and "called", because once a directory is "visited", it's also indexed. This also means all directories are "visited" at most once.

This fixes the issue of `foo/BUILD.bazel` file contains targets that provides imports like `foo.bar` which is common in Python, or generated Go packages like `foo/bar/baz`, which is common in Go code generated by Thrift.

**Other notes for review**


linzhp:
Fixed the tests. PTAL

jayconrod (approved):

PR #2110, opened 2025-06-08
Avoid duplicates in relToIndex

linzhp:
**What type of PR is this?**
Bug fix

language/go

This uses the existing `relsToIndexSeen` to avoid duplicates in `relsToIndex` and fix a bug in `addRelsToIndex()`

**Other notes for review**

### DIFF
--- a/language/go/generate.go
+++ b/language/go/generate.go
@@ -386,7 +386,10 @@ func (gl *goLang) GenerateRules(args language.GenerateArgs) language.GenerateRes
 			}
 		}
 	}
-	res.RelsToIndex = g.relsToIndex
+
+	for r := range g.relsToIndexSeen {
+		res.RelsToIndex = append(res.RelsToIndex, r)
+	}
 
 	if args.File != nil || len(res.Gen) > 0 {
 		gl.goPkgRels[args.Rel] = true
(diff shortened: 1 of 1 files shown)


language/go/generate.go
```diff
@@ -386,7 +386,10 @@ func (gl *goLang) GenerateRules(args language.GenerateArgs) language.GenerateRes
 			}
 		}
 	}
-	res.RelsToIndex = g.relsToIndex
+
+	for r := range g.relsToIndexSeen {
+		res.RelsToIndex = append(res.RelsToIndex, r)
+	}
```
jayconrod, line 392: Not strictly needed, but let's sort `res.RelsToIndex` before returning it for deterministic output.

language/go/generate.go
```diff
@@ -978,7 +980,9 @@ func (g *generator) addRelsToIndex(ps rule.PlatformStrings) {
 		for _, goSearch := range g.gc.goSearch {
 			if trimmed := pathtools.TrimPrefix(imp, goSearch.prefix); trimmed != goSearch.prefix {
```
jayconrod: Discussed on Slack. This should have been:

```
if trimmed := pathtools.TrimPrefix(imp, goSearch.prefix); goSearch.prefix == "" || trimmed != imp {
```

PR #1802, opened 2024-05-14
go_repository: add 'clean' build_file_generation

TvdW:
**What type of PR is this?**

> Feature

**What package or component does this PR mostly affect?**

> go_repository

**What does this PR do? Why is it needed?**

Before bzlmod, repositories were global, and BUILD files inside upstream repositories could load from the repositories defined at the root-level WORKSPACE file.

With the introduction of visibility rules, these result in errors that cannot trivially be worked around. One pattern that has emerged is ignoring existing build files by specifying `gazelle:build_file_name BUILD.bazel`, however this does not work for anything already using BUILD.bazel filenames.

This does affect real-world users: an example is https://github.com/google/go-jsonnet which has a BUILD.bazel file to `genrule` its AST. The module is pure-go, but the build step contains cpp code, and when imported under bzlmod-gazelle will not work because the cpp code is invoked. As a workaround, there is now a 'jsonnet' BCR module.

This patch introduces a `build_file_generation = clean` option, which removes existing build files before generating new ones. It allows loading the jsonnet code once again, and also makes it possible to load some of the complex protobuf repositories.

**Which issues(s) does this PR fix?**

Fixes #1773, #1769, #1535, https://github.com/bazelbuild/rules_go/issues/3685

**Other notes for review**

I have not made 'clean' the default, though I think that for most use cases it will be the correct behavior.

An example for fixing jsonnet:

```
go_deps.gazelle_override(
    build_file_generation = "clean",
    path = "github.com/google/go-jsonnet",
)
```

With some gymnastics this does also make it possible to import `googleapis` as if it were a regular Go dependency, without any switched_rules:
```
go_deps.module(
    path = "github.com/googleapis/googleapis",
    sum = "unused",
    version = "0.0.0",
)
go_deps.gazelle_override(
    build_file_generation = "clean",
    directives = [
        "gazelle:proto file",
    ],
    path = "github.com/googleapis/googleapis",
)
go_deps.archive_override(
    path = "github.com/googleapis/googleapis",
    sha256 = "7faad14f34810d1165a9873c140b378bc783a1448e46ebdd20a65ca66afa7bb1",
    strip_prefix = "googleapis-7002406181eb300da880701035a25157a5099abb",
    urls = ["https://github.com/googleapis/googleapis/archive/7002406181eb300da880701035a25157a5099abb.zip"],
)

# resolution
 "gazelle:resolve proto google/rpc/code.proto @com_github_googleapis_googleapis//google/rpc:code_proto",
[...]

PR #2111, opened 2025-06-12
language/go: allow clean as argument to -build_file_generation

oncilla:
**What type of PR is this?**

Bug fix

language/go

In #1802 the functionality was added, but the flag validation was not updated to allow specifying `clean` as an argument to `-build_file_generation` in `update-repos`.

With this change, we can now provide `-build_file_generation=clean` when running `update-repos`.

NaN (probably not worth creating one)

**Other notes for review**

None


fmeum (approved):
Thanks!

PR #2116, opened 2025-06-16
Update bazel-skylib in WORKSPACE setup

fmeum:
**What type of PR is this?**

Dep update

tests only

Use the same version as in `MODULE.bazel`, which requires updating past the split into `bazel_skylib` and `bazel_skylib_gazelle_plugin`.

Unblocks https://github.com/bazel-contrib/bazel-gazelle/pull/2100#discussion_r2135129237
**Other notes for review**


dzbarsky (approved):
thanks!

PR #2120, opened 2025-06-19
Test with Bazel 8

fmeum:
**What type of PR is this?**

CI

all

Fixes #

**Other notes for review**


fmeum:
@jayconrod This mostly works, but Stardoc now includes load statements that aren't the ones we would like to see since we use special internal entrypoints for doc generation. Do you see a way to restructure the docs that would give us the correct loads without having to set up a custom Stardoc template?

jayconrod:
I saw the internal load statements too when I tried to upgrade, but I didn't look deeply into it. I'll spend more time on it today.

Maybe it's as simple as moving the doc strings in the internal bzl files into a more appropriate place? It's not clear to me why it was structured like this.

jayconrod:
I think #2125 will fix the Stardoc issue. You'll still need to regenerate the (new) reference documentation, but the loads should be correct.

jayconrod (approved):

PR #2122, opened 2025-06-19
Add special handling for extracting explicit package name from rule name if being part of `go_package`

### LINKED ISSUE #2121: Incorrect rule name for proto when `go_package` name in proto file contains `;` in v0.44.0
rule names generated from `go_package` option like `bar;foo` are wrong after v0.44.0. For example, following proto file

```protobuf
syntax = "proto3";

option go_package = "bar;foo";

message Foo {}
```

will generate proto library with name `bar_foo_proto` and `bar_foo_go_proto` instead of `foo_proto` and `foo_go_proto`. This error is introduced by #2105.

a reproducible example [here](https://github.com/fardream/gazelleprotoname)

WojciechMazur:
**What type of PR is this?**

 Bug fix

langugage/proto

`go_package` with explicit package name is discouraged but allowed. Under default mode `go_package` has some special handling that extracts the package name, e.g. for `example.com/protos/foo;package_name` it uses `package_name` as base for RuleName. 
In cases of other modes there is no special handling. `Package.RuleName` containing `go_package` under `proto:package` mode was being directly passed to the `RuleName` method to resolve the rule name. 
Before my other change in #2105 the `RuleName` ignored all segments of the input name seperated by non letter or digit characters, which by pure luck also contained the `;` seperating import package from explicit package name 

The PR adds special case for extracting the explicit package name, as it was done previously in the `RuleName` before #2105 while still allowing to prevent duplicate rule names that could have originated after trimming the parts of name. 

Using explicit package name might under some circumstances lead to duplicate rule names, but this technique is already discouraged and users should be aware it can lead to problems.

Fixes #2121

**Other notes for review**

### DIFF
--- a/language/proto/generate.go
+++ b/language/proto/generate.go
@@ -102,6 +102,10 @@ func RuleName(names ...string) string {
 				'0' <= c && c <= '9' ||
 				c == '_')
 		}
+		// If name is explicit package name, e.g. `example.com/protos/foo;package_name` use package name instead of import path
+		if i := strings.LastIndexAny(name, `;`); i != -1 {
+			name = name[i+1:]
+		}
 		// If name is a path, take only the last segment
 		if i := strings.LastIndexAny(name, `/\\`); i != -1 {
 			name = name[i+1:]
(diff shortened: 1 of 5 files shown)


WojciechMazur:
cc @jayconrod

jayconrod (approved):

Issue #2127, opened 2025-06-23
Gazelle fails resolving generated packages in private repos


guw:
I have a repository where Gazelle fails to run cleanly. While `go mod tidy -e` works I am struggling with failing Gazelle runs.

Setup:
- private repo at `github.com/private/foo`
- Bazel workspace at `bar/` within that repo

`bar/MODULE.bazel`:

```Starlark
go_deps.config(
    go_env = {
        "GOPRIVATE": "github.com/private",
        "GOPROXY": "...",
        "GOWORK": "off",
    },
)
go_deps.from_file(go_mod = "//:go.mod")
```

Git configuration:

```
$ git config -l --show-origin | grep instead                              
file:/home/username/.gitconfig      url.git@github.com:.insteadof=https://github.com/
```

This might be important because Git auth is setup using SSH but not https.

```Shell
INFO: Running command line: bazel-bin/gazelle
gazelle: finding module path for import github.com/private/foo/bar/proto/whatever/v1: go: -d flag is deprecated. -d=true is a no-op
go: downloading github.com/private/foo/bar v0.0.0-...
github.com/private/foo@...: verifying module: github.com/private/foo@...: reading https://sum.golang.org/lookup/github.com/private/foo@...: 404 Not Found
        server response:
        not found: github.com/private/foo@...: invalid version: git ls-remote -q origin in /tmp/gopath/pkg/mod/cache/vcs/checksum: exit status 128:
                fatal: could not read Username for 'https://github.com': terminal prompts disabled
        Confirm the import path was entered correctly.
        If this is a private repository, see https://golang.org/doc/faq#git_https for additional information.
```

There seems to be a couple of issues at play here.

1. `github.com/private/foo` is a private repo
2. `github.com/private/foo/bar/proto/whatever/v1` is generated from poto file
3. there is an `empty.go` file at `bar/proto/whatever/v1` - it seems to be ignored by Gazelle but `go mod tidy` is happy
4. `GOPRIVATE` is defined in `bar/MODULE.bazel` but it seems to be ignored bz Gazelle
5. It should not attempt to hit `sum.golang.org` because `GONOSUMDB` is implied from `GOPRIVATE`. Perhaps another indicator that `GOPRIVATE` is not set?
6. There is a Git config override to use SSH instead of https for the Git URL, which seems to be ignored by the `git ls-remote` remote command executed by Gazelle

[...]

PR #2128, opened 2025-06-23
Fix missing go_env when running Gazelle

guw:
**What type of PR is this?**

Bug fix

cmd/gazelle

Work towards #2127

### DIFF
--- a/internal/bzlmod/go_deps.bzl
+++ b/internal/bzlmod/go_deps.bzl
@@ -259,7 +259,7 @@ def _go_repository_config_impl(ctx):
         ))
 
     ctx.file("WORKSPACE", "\n".join(repos))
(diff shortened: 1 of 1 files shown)


fmeum:
I can look into the docs issue.

Does this really fully fix the linked issue or is more work needed?

guw:
> Does this really fully fix the linked issue or is more work needed?

I suspect the Git config stuff is still broken. How does it work in Bazel and sandboxing? Should the Git config be made available somehow?

fmeum:
> I suspect the Git config stuff is still broken. How does it work in Bazel and sandboxing? Should the Git config be made available somehow?

`bazel run` should pick up all environment variables by default. Are you setting them via `--repo_env`?

PR #2140, opened 2025-07-09
Normilize windows paths

dmirell:
**What type of PR is this?**

Bug fix

cmd/gazelle

This patch normalizes Windows paths by replacing backslashes (\) with forward slashes (/). This modification ensures consistency in path formats across different platforms, which is particularly useful for maintaining compatibility and preventing potential errors in path handling. The change is applied within the go_repository_config_impl function in the internal/go_repository_config.bzl file.

This PR addresses the issue of inconsistent path formats on Windows systems by normalizing paths to use forward slashes (/).


fmeum:
Does this actually fix a bug? Could you provide more details on why this helps?

PR #2141, opened 2025-07-11
walk: expose cached GetDirInfo

### LINKED ISSUE #2132: resolver.Imports method needs RegularFiles list to evaluate glob

```
cc_library(
    name = "foo",
    hdrs = glob(["**/*.h"]),
)
```

```

type DirInfo struct {
  File *rule.File
}
```

jayconrod:
**What type of PR is this?**

> Feature

> walk

Adds `walk.GetDirInfo`, a function that exposes the internal I/O cache used by
`Walk` and `Walk2`. This function does no I/O on its own. It may only be called
during `Walk` or `Walk2` on directories that have been loaded.

This is meant to help `resolve.Resolver.Imports` implementations resolve globs.
In C++ for example, the returned imports are based on the `hdrs` attribute, which
for hand-written rules is likely to be a `glob`. Providing access to `RegularFiles` and
`GenFiles` lets us expand the `glob` without additional I/O while respecting `exclude`
and other directives.

Fixes #2132

**Other notes for review**


jayconrod:
cc @WojciechMazur

walk/cache.go
```diff
@@ -71,3 +71,42 @@ func (c *cache) getLoaded(rel string) (dirInfo, error) {
...
+)
+
+func maybeSetGlobalCache(c *cache) func() {
+	globalCacheMu.Lock()
+	defer globalCacheMu.Unlock()
+	if globalCache == nil {
+		globalCache = c
+		return func() { globalCache = nil }
+	} else {
+		return func() {}
```
fmeum: Could this case panic instead? This looks like it could hide lifecycle issues.

walk/cache.go
```diff
@@ -71,3 +71,42 @@ func (c *cache) getLoaded(rel string) (dirInfo, error) {
...
+// none was present. rel is a slash-separated path, relative to the repository
+// root directory or "" for the root directory itself.
+//
+// GetDirInfo does not directly perform any I/O. Instead, it provides access
+// to an internal cache to an internal cache that Walk and Walk2 use to speed
+// up directory operations. For this reason, GetDirInfo must not be called
+// on a directory Gazelle has not visited.
+//
+// In general, language extensions should prefer to use the RegularFiles,
+// Subdirs, and File fields of language.GenerateArgs. This function returns
```
fmeum, line 97: Could this be exposed on `language.GenerateArgs` instead of being a global?

walk/cache.go
jayconrod: Sure, done. I removed the mutex as well.

My intent here was to allow `Walk` and `Walk2` to be reentrant, but I'm probably overthinking it.

walk/cache.go
jayconrod: All the information returned by this function is already in `GenerateArgs`.

The problem is that `Resolver.Imports` gets called before `GenerateArgs` during indexing, so the information is not yet available. In C++, we wanted to expand `glob` expressions as part of `Imports` because it's common to have `hdrs = glob(...)`.

Ideally, I'd like to pass in these fields as arguments to `Imports`, but there's no way to do that without an incompatible change. If we do a v2 of the extension interfaces, that would be nice to have.

For now, adding this function seemed like a pragmatic option, though it does feel kludgy.

language/lang.go
```diff
@@ -161,7 +161,7 @@ type GenerateArgs struct {
 	// symbolic links to directories that Gazelle will follow.
 	// RegularFiles is a list of regular files including other symbolic
 	// links.
-	// GeneratedFiles is a list of generated files in the directory
+	// GenFiles is a list of generated files in the directory
 	// (usually these are mentioned as "out" or "outs" attributes in rules).
```
fmeum, line 165: Should we mention that these slices shouldn't be modified? The Java person in me wants to enforce immutability, but short of handing out a `Seq` I wouldn't see how.

fmeum (approved):

language/lang.go
jayconrod: Done. I've added some comments. Also set the cap of each of these slices to the len so that if the caller appends to the slice (most likely modification), they'll make a copy instead of writing to the same backing array.

I wish more languages had immutability built into the type system like Rust. Java doesn't have it either, and I think all extra code, copying, and allocation is far too costly.

walk/cache.go
```diff
@@ -71,3 +71,35 @@ func (c *cache) getLoaded(rel string) (dirInfo, error) {
...
+// root directory or "" for the root directory itself. The returned values
+// must not be modified.
+//
+// GetDirInfo does not directly perform any I/O. Instead, it provides access
+// to an internal cache to an internal cache that Walk and Walk2 use to speed
+// up directory operations. For this reason, GetDirInfo must not be called
+// on a directory Gazelle has not visited.
+//
+// In general, language extensions should prefer to use the RegularFiles,
+// Subdirs, and File fields of language.GenerateArgs. This function returns
```
jbedard, line 97: @jayconrod note that the fields in `DirInfo` are quite different then those in `GenerateArgs` when `generation_mode` is being used.

Was this intentional?

jbedard: FWIW I've always accessed this info via [a patch](https://github.com/aspect-build/aspect-cli/blob/dd3563abbc5b2f2844b80e789bd4e7049dfb1e23/patches/bazelbuild_bazel-gazelle_aspect-fs-direntry.patch) for what I think is the exact same scenario as yours: [expanding globs in `Imports()`](https://github.com/aspect-build/aspect-cli/blob/dd3563abbc5b2f2844b80e789bd4e7049dfb1e23/gazelle/js/resolve.go#L72).

That glob expansion needs to support globbing into subdirectories though...

jayconrod: Not intended specifically, but it's very difficult to support `generation_mode` outside of the normal tree walking process, so I think `GetDirInfo` is doing the right thing here.

Maybe it makes sense to add an `ExpandGlob` function that calls `GetDirInfo`, accounting for `generation_mode` and stopping at directories that have BUILD files? This is hard to predict though.

WojciechMazur: I've made some implementation of glob expansion based on this PR [in gazelle_cc ](https://github.com/EngFlow/gazelle_cc/pull/71/commits/afd3a4af89a09dae5c78ff945d19b0476d8db21e)  /
It does actually support globbing into subdirectories and is able to consume `**` syntax although. I wonder if we might want to push it to upstream gazelle instead? I might be useful, becouse on the user side we might not be able to handle cases when `GetDirInfo` would be invoked before actual tree walking, otherwise it's easy to trigger panic (typically it can happen if we have invalid, non-relative path)

I'm not yet sure how it would need to account for `generation_mode`, but so far the only limitation I've observed is fact the we cannot glob into directories that are contained in `.bazelignore` or using `# gazelle:exclude`, but that's expected.

jbedard: @WojciechMazur it looks like you're already recursing into `Subdirs` so I think that handles `generation_mode`, however I don't see (after a quick scan at least) anywhere that you're stopping that recursion when a BUILD is already present in the directory?

I'm also not sure if that `walk.GetDirInfo()` will always work... what if you recurse into a directory that hasn't been walked yet (it will `panic`)? Is there a race condition where the fs walk might be going on in parallel?

jbedard: I've have glob expansion working already using [a patch](https://github.com/aspect-build/aspect-cli/blob/dd3563abbc5b2f2844b80e789bd4e7049dfb1e23/patches/bazelbuild_bazel-gazelle_aspect-fs-direntry.patch) and invoking the [builtin glob method](https://github.com/aspect-build/aspect-cli/blob/dd3563abbc5b2f2844b80e789bd4e7049dfb1e23/gazelle/common/starlark/glob.go#L58) though.

Ideally I could drop that patch but this new `DirInfo` does not provide the info needed. Although now that I look at the recursion in https://github.com/EngFlow/gazelle_cc/pull/71/commits/afd3a4af89a09dae5c78ff945d19b0476d8db21e maybe it's possible?

Ideally we could put this logic somewhere we can share it though.

WojciechMazur: Yes, it can panic. For me it typically happend in case if I passed the absolute instead of relative path. I'd need to check how it would behave if there are no other languages enabled in the `gazelle_binary` 
I'm planning to update the logic with searching for `BUILD` files. I was previously not aware previously of the specific behaviour of `glob()` if the `BUILD` files are present.

jbedard: I think it can also panic due to race conditions though. IIUC if you invoke `GetDirInfo` faster then gazelle is walking those directories in the background...

I don't think `GetDirInfo` and a custom glob implementation in `gazelle_cc` is the ideal way of solving this. Maybe we can at least have a common implementation in gazelle instead of exposing utils like `GetDirInfo` exposing gazelle internals (in a way that is not ideal for our use case anyway)?

jayconrod: `GetDirInfo` should not panic as long as its contract is satisfied. That contract is rather complicated though, so if you see a panic, please open an issue and we can fix it if it's a bug or clarify the docs.

jbedard: Won't it panic if you try fetching a directory not loaded yet? Such as a subdirectory...? I'm speculating here so forgive me if I'm wrong :|

PR #2142, opened 2025-07-11
rule: add MatchGlobExpr for parsing glob expressions

### LINKED ISSUE #2133: rule: ability to parse glob expressions
In gazelle_cc, we're looking for a way to index `cc_library` targets with `glob` expressions like this:

```
cc_library(
    name = "foo",
    hdrs = glob(["**/*.h"]),
)
```

For now, we don't need to be able to generate or merge such expressions, but it would be nice to be able to extract them from the syntax tree without a custom pattern matcher.

There's already a [`rule.GlobValue`](https://pkg.go.dev/github.com/bazelbuild/bazel-gazelle/rule#GlobValue) type. It's used by the Go extension to generate `data = glob(["testdata/**"])`. It doesn't support merging yet (#1036).

A function like this would work for us:

```
// MatchGlob checks whether expr is a simple glob expression and returns a GlobValue and true
// if it is. MatchGlob returns the zero value and false if the expression is not a glob expression
// or if it can't be represented with GlobValue.
func MatchGlob(expr bzl.Expr) (GlobValue, bool)
```

jayconrod:
**What type of PR is this?**

> Feature

> rule

Adds `MatchGlobExpr`, a function for parsing a `GlobExpr` out of the syntax tree.
This will be used by the C++ extension's `resolve.Resolver.Imports` implementation.
It's common for hand-written `cc_library` targets to set the `hdrs` attribute to a `glob`
expression. We want to be able to expand that `glob` to be able to list the import
strings for the library.

Fixes #2133

**Other notes for review**

As a drive-by fix, `GlobValue.BzlExpr` now uses `Ident` instead of `LiteralExpr`, which
is for numeric constants.

### DIFF
--- a/rule/value.go
+++ b/rule/value.go
@@ -45,15 +45,62 @@ func (g GlobValue) BzlExpr() bzl.Expr {
 	if len(g.Excludes) > 0 {
 		excludesValue := ExprFromValue(g.Excludes)
 		globArgs = append(globArgs, &bzl.AssignExpr{
-			LHS: &bzl.LiteralExpr{Token: "exclude"},
+			LHS: &bzl.Ident{Name: "exclude"},
 			Op:  "=",
 			RHS: excludesValue,
 		})
 	}
 	return &bzl.CallExpr{
-		X:    &bzl.LiteralExpr{Token: "glob"},
+		X:    &bzl.Ident{Name: "glob"},
 		List: globArgs,
 	}
+}
+
+// MatchGlobExpr detects whether the given expression is a call to the glob
+// function. If it is, MatchGlobExpr returns the glob's patterns and excludes
+// (if they are literal strings) and true. If not, MatchGlobExpr returns false.
+func MatchGlobExpr(e bzl.Expr) (GlobValue, bool) {
+	call, ok := e.(*bzl.CallExpr)
+	if !ok {
+		return GlobValue{}, false
+	}
+	callee, ok := call.X.(*bzl.Ident)
+	if !ok || callee.Name != "glob" {
+		return GlobValue{}, false
+	}
+	var glob GlobValue
+	for i, arg := range call.List {
+		list, ok := arg.(*bzl.ListExpr)
+		if i == 0 && ok {
+			glob.Patterns = make([]string, 0, len(list.List))
+			for _, e := range list.List {
+				if str, ok := e.(*bzl.StringExpr); ok {
+					glob.Patterns = append(glob.Patterns, str.Value)
+				}
+			}
+			continue
+		}
+
+		kv, ok := arg.(*bzl.AssignExpr)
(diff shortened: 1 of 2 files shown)


jayconrod:
cc @WojciechMazur

rule/value_test.go
```diff
@@ -104,3 +104,66 @@ func TestExprFromValue(t *testing.T) {
 		})
 	}
 }
+
+func TestMatchGlobExpr(t *testing.T) {
```
fmeum: Could you add tests with `allow_empty` and `no_match_error`?

rule/value.go
```diff
@@ -45,17 +45,64 @@ func (g GlobValue) BzlExpr() bzl.Expr {
...
-		X:    &bzl.LiteralExpr{Token: "glob"},
+		X:    &bzl.Ident{Name: "glob"},
 		List: globArgs,
 	}
 }
 
+// MatchGlobExpr detects whether the given expression is a call to the glob
+// function. If it is, MatchGlobExpr returns the glob's patterns and excludes
+// (if they are literal strings) and true. If not, MatchGlobExpr returns false.
+func MatchGlobExpr(e bzl.Expr) (GlobValue, bool) {
```
fmeum: I first thought that `Match` refers to "evaluating" rather than "recognizing" a glob expression. How about `Parse`, `Get` or `As`?

jayconrod: I was thinking it's like pattern matching, but `Parse` works for me. Done.

rule/value_test.go
jayconrod: Done. The argument names shouldn't matter for this purpose, but now there's a test case with both `allow_empty` and `exclude_directories`. I also added a test case for invalid arguments.

fmeum (approved):

jbedard:
FYI I've always done similar for the `Imports()` method handling custom `ts_project(srcs)` or `js_library(srcs)` which often have globs. See [ExpandSrcs](https://github.com/aspect-build/aspect-cli/blob/dd3563abbc5b2f2844b80e789bd4e7049dfb1e23/gazelle/common/starlark/glob.go#L41).

Do you think it's worth putting a method similar to that alongside `ParseGlobExpr`? Maybe something such as `GlobFilter(GlobValue, sources []string) []string`?

jayconrod:
It's easy to compose `rule.ParseGlobExpr`, `walk.GetDirInfo`, and `doublestar.Match`, but it would also be easy to bring those together in a single function to support a common pattern. Where would such a function live? It doesn't really feel right having those packages depend on each other.

jbedard:
Even just `GlobValue.Match(path)` or `GlobValue.Filter(paths)` might be worth it? It's super simple but might allow languages to drop use of `doublestar` 🤷

PR #2145, opened 2025-07-17
Accept more forms of glob values when parsing `bzl.Expr`

WojciechMazur:
**What type of PR is this?**

 Bug fix

`rule` and possible integration in downstram projects

It adds handling for more forms of `glob` expressions: 
- accepts named `include` argument 
- accepts positional `exclude` arguments 

Both of these are allowed in Build files

No dedicated issue 
Addendum to #2142 

**Other notes for review**

### DIFF
--- a/rule/value.go
+++ b/rule/value.go
@@ -93,12 +93,12 @@ func ParseGlobExpr(e bzl.Expr) (GlobValue, bool) {
 			}
 			continue
 		}
-		allowPositionalArgs = false
 
 		kv, ok := arg.(*bzl.AssignExpr)
 		if !ok {
 			continue
 		}
+		allowPositionalArgs = false
 		key, ok := kv.LHS.(*bzl.Ident)
 		if !ok {
 			continue
--- a/rule/value.go
+++ b/rule/value.go
@@ -69,35 +69,49 @@ func ParseGlobExpr(e bzl.Expr) (GlobValue, bool) {
 		return GlobValue{}, false
 	}
 	var glob GlobValue
+	parseStringsList := func(list *bzl.ListExpr) []string {
+		parsed := make([]string, 0, len(list.List))
+		for _, e := range list.List {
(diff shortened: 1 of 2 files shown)


rule/value_test.go
```diff
@@ -153,7 +153,17 @@ func TestParseGlobExpr(t *testing.T) {
 		{
 			name: "invalid_args",
 			text: `glob(["a"], ["b"], exclude = ["x"], unknown = 1, *args, **kwargs)`,
-			want: GlobValue{Patterns: []string{"a"}, Excludes: []string{"x"}},
+			want: GlobValue{Patterns: []string{"a"}, Excludes: []string{"b"}},
```
WojciechMazur, line 156: Changes existing test case, however, this glob would not compile - when interpreting it would result in error as named `exclude` argument is already set using positional `["b"]` argument.

rule/value.go
```diff
@@ -69,35 +69,49 @@ func ParseGlobExpr(e bzl.Expr) (GlobValue, bool) {
...
+			case 0:
+				glob.Patterns = parseStringsList(list)
+			case 1:
+				glob.Excludes = parseStringsList(list)
+				// Last handled positional argument, no need to visit more
+				return glob, true
 			}
 			continue
 		}
+		allowPositionalArgs = false
```
jayconrod: I think this should move down to after we've matched the first `AssignExpr`. We might reach this line after seeing a positional argument that's not a `ListExpr`, for example:

```
include = ["*.txt"]
exclude = ["*.c"]
glob(include, exclude)
```

jayconrod (commented):
Looks good, thanks for adding to this. One small change needed.

rule/value.go
WojciechMazur: Good catch. We won't be able to read what's under the identifiers, but now we can still get content of the second positional argument. Added test case for that

jayconrod (approved):

PR #2149, opened 2025-07-23
language/go/generate: check len(segments) before adjusting srcdir

tzaman-stripe:
**What type of PR is this?**

Bug Fix

language/go

If len(segments) < 2, it will cause an index access error.

### DIFF
--- a/language/go/generate.go
+++ b/language/go/generate.go
@@ -126,12 +126,14 @@ func (gl *goLang) GenerateRules(args language.GenerateArgs) language.GenerateRes
 		// --experimental_sibling_repository_layout.
 		slashPath := filepath.ToSlash(c.RepoRoot)
 		segments := strings.Split(slashPath, "/")
-		repoName := segments[len(segments)-1]
-		previousSegment := segments[len(segments)-2]
-		if previousSegment == "external" {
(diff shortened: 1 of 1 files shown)


fmeum:
Generally happy to merge this, but I don't see how this case can arise. Could you share more details on the situation in which you see this crash?

tzaman-stripe:
honestly I'm not 100% sure about how our gazelle setup works but I know that RepoRoot=. in this case

fmeum:
The only source for repoPath in this code block is https://github.com/bazel-contrib/bazel-gazelle/blob/ffa4e789475efb24104bcc270fae91cd97e08747/internal/go_repository.bzl#L344 and I don't see how that could return `"."`. If we merge this, it might mask a bug somewhere else in your stack.

tzaman-stripe:
we essentially have our own version of go_repository that uses . I think. would they be equivalent?

fmeum:
Not sure whether the logic for cgo include paths would be correct in that case though. This will avoid the crash, but from the perspective of this repo, a crash is better than a build file that's silently generated in an incorrect way.

tzaman-stripe:
I don't know much about cgo (or go at all) but I feel like a better error could be raised than an indexing error, then?

PR #2150, opened 2025-07-24
Propagate to Gazelle extensions `proto_library` name achieved after merging with BUILD file

WojciechMazur:
**What type of PR is this?**

Bug fix

language/proto
Downstream gazelle extensions emitting `<lang>_proto_library` based on outputs of `//language/proto` including `//language/go`

Downstream extensions typically consume generated `proto_library` by emitting directly `<lang>_proto_library` with `deps` (or other language specific attribute) pointing to `proto_library` rule name obtained from `args.OtherGen`. 

However, if user has modified the name of `proto_library` manually `/language/proto` would still emit the generate name in emitted `proto_library` rules. Because of that all downstream languages would need to implement dedicated workarounds for this problem and search for actual `proto_library` in the files, typically using `Resolve` functionality.  

Example of this kind of dedicated support was initially drafted for `rules_cc` https://github.com/EngFlow/gazelle_cc/pull/82 but we'd like to fix it in upstream Gazelle so all extensions can benefit on it. 

No dedicated issue in this repository, see issue in downstream extensions https://github.com/EngFlow/gazelle_cc/issues/81

**Other notes for review**

`merger.Match` was used here to ensure emitted rule name would match it's name after the merge. `proto_library` is always matched by it's `srcs` attribute.  
`rule.SetName` overwrites the name assigned in `generateProto`


jayconrod (approved):
Approving: this works for proto. I have an idea for a more general but more complicated fix. However, this problem isn't very common (maybe just proto), and fixing it for proto is fine.

I think the most general way to fix this is in `runFixUpdate` and `merger.MergeFile`.

1. `runFixUpdate` could call `merger.MergeFile` after calling `language.GenerateRules` on each extension, rather than gathering all the generated rules into a list and merging them all at once.
2. `MergeFile` could modify its `genRules` list: if a generated rule is merged with an existing rule, the list would contain the merged existing rule afterward. This rule would have the correct final name and kind.
3. `runFixUpdate` would then pass the merged rules through `OtherGen` instead of the original rules.

This would be a lightly incompatible change depending on how we do it. Today we don't apply kind mapping to `OtherGen`, so if `proto_library` is replaced with some other kind, then other extensions wouldn't recognize the generated rules. So we might need to keep the original `OtherGen` and pass merged rules through a new `OtherGenMerged` list.

I'm talking myself out of this: it's complicated.

cc @fmeum

PR #2152, opened 2025-07-25
pathtools: add Prefixes helper function

### LINKED ISSUE #2132: resolver.Imports method needs RegularFiles list to evaluate glob

jayconrod:
**What type of PR is this?**

> Feature

> pathtools

In the fix for #2132, I'll need to iterate over the prefixes of a given
slash-separated path. Rather than duplicate the logic from Walk2,
let's extract it into pathtools.

Prefixes returns an iterator over prefixes to avoid needing to allocate
a slice. The code should be much simpler after we adopt Go 1.23 and can
use range-over-iterator loops.

Fixes #2132

**Other notes for review**

### DIFF
--- a/pathtools/path_test.go
+++ b/pathtools/path_test.go
@@ -15,7 +15,11 @@ limitations under the License.
 
 package pathtools
 
-import "testing"
+import (
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+)
 
 func TestHasPrefix(t *testing.T) {
 	for _, tc := range []struct {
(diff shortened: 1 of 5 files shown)


jbedard:
Go 1.23 is needed before we can use `iter.Seq`?

jbedard:
You could also use channels instead 🤷

I've avoided using iterators/yield in a few places where I've noticed they add a lot more overhead then channels.

jayconrod:
> Go 1.23 is needed before we can use iter.Seq?

Right, it's a new language feature. Older compilers don't support it.

> You could also use channels instead 🤷
> I've avoided using iterators/yield in a few places where I've noticed they add a lot more overhead then channels.

I'd be really surprised if that were true. Do you have an example with a benchmark?

Channels are synchronization primitives. Many operations on them involve flushing parts of the cache and context switching. They should not be used for iteration.

Iterators are not a zero cost abstraction, but they're implemented with stackless coroutines which are pretty cheap. I think they're usually a good choice when they help avoid allocating a slice or a map like this.

pathtools/path_test.go
```diff
@@ -110,3 +114,57 @@ func TestTrimPrefix(t *testing.T) {
...
+			want: []string{"", "aa", "aa/bb", "aa/bb/cc"},
+		},
+		{
+			name: "multiple_abs",
+			path: "/aa/bb/cc",
+			want: []string{"/", "/aa", "/aa/bb", "/aa/bb/cc"},
+		},
+		{
+			name: "unclean",
+			path: "a/../b//c/",
```
fmeum, line 155: Should we document this behavior on the function?

jbedard:
> I'd be really surprised if that were true. Do you have an example with a benchmark?

I can't reproduce it atm, so maybe I'm wrong or misremembering? Using sequences sure is a lot nicer when analyzing profiles like what I'm staring at right now though. Lets ignore my original comment for now...

jbedard (approved):

pathtools/path_test.go
jayconrod: Yes, done.

jayconrod:
> Should we document this behavior on the function?

I accidentally pushed this onto #2153 instead. About to merge that, so we'll end up in the same place.

PR #2153, opened 2025-07-25
walk: allow GetDirInfo to be called in directories not visited yet

### LINKED ISSUE #2132: resolver.Imports method needs RegularFiles list to evaluate glob

jayconrod:
**What type of PR is this?**

> Bug fix

> walk

GetDirInfo was intended to help expand glob expressions in
resolve.Resolver.Imports implementations. globs may cover subdirectories,
especially when gazelle:generate_mode is used, so GetDirInfo must
be able to read contents of subdirectories that may not have been
visited yet.

Previously, the contract of GetDirInfo did not allow visiting
directories that hadn't been visited yet, but it would still sometimes
succeed walker.populateCache won the race. This change fully
allows GetDirInfo to be called in directories that haven't been
visited yet.

Fixes #2132

**Other notes for review**

This PR is based on #2152. I tried to split them for easier review, but GitHub doesn't let me set the base to a branch in a forked repo, so I'm sorry if this makes things more difficult.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -11,7 +11,7 @@ bazel_dep(name = "protobuf", version = "3.19.6", repo_name = "com_google_protobu
 bazel_dep(name = "rules_cc", version = "0.0.17")
 bazel_dep(name = "rules_go", version = "0.53.0", repo_name = "io_bazel_rules_go")
 bazel_dep(name = "rules_license", version = "1.0.0")
-bazel_dep(name = "package_metadata", version = "0.0.2")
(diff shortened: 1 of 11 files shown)


jayconrod:
cc @WojciechMazur

jbedard (approved):
Thanks 👍

WojciechMazur:
Successfully tested out in https://github.com/EngFlow/gazelle_cc/pull/71

PR #2155, opened 2025-07-28
Upgrade `package_metadata` to 0.0.5 (was 0.0.2)

WojciechMazur:
**What type of PR is this?**

Other 

go_repository
downstream extensions

Downstream projects that depend on `gazelle` are adding the `package_metadata` to their builds, e.g. https://github.com/EngFlow/gazelle_cc/pull/72 
The `package_metadata` 0.0.3 has added `licenses` rules that are missing in 0.0.2. 

When adding `gazelle_cc` via WORKSPACE we typically want to include `gazelle` dependencies first before setting up extension specific dependenices, example: 
```
gazelle_dependencies()

load("@gazelle_cc//:deps.bzl", "gazelle_cc_dependencies")
gazelle_cc_dependencies()
```

Becouse older version of `package_metadata` would be requested in `gazelle_dependencies` the newer version present in `gazelle_cc_dependencies` would be ignored, making it hard to debug for non experienced users. 

By upgrading `package_metadata` to latest version we make dependency resolution easier.

No dedicated issue

**Other notes for review**

`package_metadata` should include the `license` attribute, but it can be done in the follow up PR

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -11,7 +11,7 @@ bazel_dep(name = "protobuf", version = "3.19.6", repo_name = "com_google_protobu
 bazel_dep(name = "rules_cc", version = "0.0.17")
 bazel_dep(name = "rules_go", version = "0.53.0", repo_name = "io_bazel_rules_go")
 bazel_dep(name = "rules_license", version = "1.0.0")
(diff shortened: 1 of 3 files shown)


WojciechMazur:
cc @Yannic

Yannic:
Thank you!

PR #2157, opened 2025-07-30
Extract either Module/Workspace RepoName in `CommonConfigurer`

WojciechMazur:
**What type of PR is this?**

Bug fix

cmd/gazelle 
config

`Config.RepoName` was never setup when using `MODULE.bazel`. 
With this change we now always try to extract repositoryName from either `module(name = ...)` in MODULE.bazel or `workspace(name = ...) in case of workspace files. 

The logic previously defined in `cmd/gazelle` was moved to `CommonConfigurer` as this is the place that should be responsible for setup of RepoName. 

No issue in this repo. I want to use have special logic in based on `Config.RepoName` in `gazelle_cc` 

**Other notes for review**

The hack (for mapping of google_apis) was preserved in the new location`. I don't know if it can be removed, or if the mapping should happen in the `cmd/gazelle` instead.

### DIFF
--- a/internal/module/module.go
+++ b/internal/module/module.go
@@ -41,6 +41,25 @@ func ExtractModuleToApparentNameMapping(repoRoot string) (func(string) string, e
 	}, nil
 }
 
+// ExtractModuleName collects name of the module from the MODULE.bazel file, if it exists.
+// Returns empty string if MODULE.bazel does not exist or does not define explicit name
+func ExtractModuleName(repoRoot string) (string, error) {
+	f, err := parseModuleSegment(repoRoot, "MODULE.bazel")
+	if err != nil {
+		// If there is no MODULE.bazel file, return an name by no error
+		if os.IsNotExist(err) {
+			return "", nil
+		}
+		return "", err
+	}
+	for _, dep := range f.Rules("") {
+		if dep.Kind() == "module" {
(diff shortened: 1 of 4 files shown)


WojciechMazur:
cc @jayconrod

internal/module/module.go
```diff
@@ -41,6 +41,25 @@ func ExtractModuleToApparentNameMapping(repoRoot string) (func(string) string, e
...
+	if err != nil {
+		// If there is no MODULE.bazel file, return an name by no error
+		if os.IsNotExist(err) {
+			return "", nil
+		}
+		return "", err
+	}
+	for _, dep := range f.Rules("") {
+		if dep.Kind() == "module" {
+			return dep.Name(), nil
```
fmeum, line 57: Depending on what this is meant to be used for, you may actually want the `repo_name` attribute (if set).

jayconrod: Hmm, I didn't know about the `repo_name` attribute. I suppose it would depend on the use case.

In this case, we're matching C++ include paths against a pre-generated index that includes headers from all the Bazel modules in the BCR. If we're running Gazelle in a module that's already in the BCR though, we want to resolve those locally, not with the index. The index has the module names, so module name is correct in this case.

PR #2161, opened 2025-08-12
perf: populate walk cache in background

jbedard:
Is there any reason to block waiting on this? We can most likely start `visit()`ing directories in parallel while the cache population is done async in the background...

Normally the cache population will always be a step (or many steps) ahead of `visit()` and this change should only make the 2 traversals overlap but introduce no change to who fills the cache or what order it is filled. The only scenario I can think of where `populateCache()` would no longer be the first to fill the cache would be if a language callback invokes `GetDirInfo`?

**What type of PR is this?**

Other

all

**Other notes for review**

### DIFF
--- a/walk/dirinfo.go
+++ b/walk/dirinfo.go
@@ -5,7 +5,6 @@ import (
 	"os"
 	"path"
 	"path/filepath"
-	"sync"
 
 	"github.com/bazelbuild/bazel-gazelle/rule"
 )
@@ -99,8 +98,8 @@ func (w *walker) loadDirInfo(rel string) (DirInfo, error) {
 // populateCache loads directory information in a parallel tree traversal.
 // This has no semantic effect but should speed up I/O.
 //
-// populateCache should only be called when recursion is enabled. It avoids
-// traversing excluded subdirectories.
+// populateCache will traverse only the directories that walker is configured
+// to visit. It will not traverse excluded directories.
 func (w *walker) populateCache() {
 	// sem is a semaphore.
 	//
@@ -111,7 +110,6 @@ func (w *walker) populateCache() {
 	// for each child. This prevents a deadlock that could occur for a deeply
 	// nested series of directories.
 	sem := make(chan struct{}, 6)
-	var wg sync.WaitGroup
 
 	var visit func(string)
 	visit = func(rel string) {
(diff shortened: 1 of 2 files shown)


walk/dirinfo.go
```diff
@@ -127,18 +125,12 @@ func (w *walker) populateCache() {
...
 			}
 		}
 	}
 
 	// Start the traversal at the root directory.
 	sem <- struct{}{}
-	visit("")
-
-	wg.Wait()
+	go visit("")
```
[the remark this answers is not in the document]
jbedard: We can revert all the changes in this file if preferred, the real change in this PR is in `walk.go` simply making `populateCache()` async. However that single invocation of `populateCache()` being async makes all the usage of `WaitGroup` here unused 🤷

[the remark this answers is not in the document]
jayconrod: I'd prefer to revert these changes. As a general principle, synchronous functions are much easier to understand. The caller can make the call asynchronous if they want, but making the function asynchronous removes that choice. With this change, there's no way to guarantee this function completes, and goroutines get leaked.

walk/walk.go
```diff
@@ -379,7 +379,8 @@ func newWalker(c *config.Config, cexts []config.Configurer, dirs []string, mode
 		relsToVisitSeen: make(map[string]struct{}),
 	}
 
-	w.populateCache()
+	// Asynchronously populate the walker cache in the background.
+	go w.populateCache()
```
jayconrod, line 383: I don't feel great about this change. I don't think functions should spawn goroutines that survive beyond the end of the function unless that's explicitly part of the contract and there's a way to track their progress or cancel them. That's a general principle, and in this case, it doesn't matter much because `populateCache` is supposed to be a transparent performance optimization.

That said, I only think we should do this if there's a measurable performance improvement. I ran `go test -bench=BenchmarkWalk` though, and I don't see a significant difference.

jayconrod (commented):
Sorry for the slow reply: was on vacation for the last two weeks.

walk/walk.go
jbedard: In one large repo I'm working in this improves performance ~10% when the other heavy operations (parsing source code) are cached while the full repo is still walked+indexed. I thought that was enough of a difference it was worth the extra complications 🤷

jayconrod: 10% is a good improvement! In that case, let's do it!

Issue #2164, opened 2025-08-12
Diff mode behavior inconsistent with other modes.


alandonham:
**What version of gazelle are you using?**
0.43.0

**What version of rules_go are you using?**
0.51.0

**What version of Bazel are you using?**
7.4.1

**Does this issue reproduce with the latest releases of all the above?**
Yes.

**What operating system and processor architecture are you using?**
MAC OS ARM64

**What did you do?**
Introduce a change to a BUILD.bazel file that has no trailing newline.

**What did you expect to see?**
Gazelle's `mode=fix` has the same exit code as `mode=diff`

**What did you see instead?**
Gazelle's `mode=diff` option can produce a different exit code than the other modes, due to stripping of newlines from the diff.

PR #2166, opened 2025-08-12
Adding unit test to ensure behavior of mode=diff

alandonham:
**What type of PR is this?**
Other

cmd/gazelle

Adding unit tests to verify `mode=diff` functionality with and without newlines present.

Updates #2165

**Other notes for review**
This is adding unit tests for a PR that was already approved.
#2164

### DIFF
--- a/cmd/gazelle/diff.go
+++ b/cmd/gazelle/diff.go
@@ -74,14 +74,24 @@ func diffFile(c *config.Config, f *rule.File) error {
 	} else {
 		diff.ToFile = outPath
 	}
+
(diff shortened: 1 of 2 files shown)


alandonham:
I meant to open #2164 as a draft, and it was merged before I was able to push the commit with the unit tests, so here they are.

fmeum:
> I meant to open #2164 as a draft, and it was merged before I was able to push the commit with the unit tests, so here they are.

The PR looked good as is and I wasn't sure whether it was easy to test. Glad to have the tests though :-)

PR #2169, opened 2025-08-14
support new ignore directive in go.mod

lbcjbb:
Go 1.25 introduces a new `ignore` directive to specify directories the go command should ignore.

**What type of PR is this?**

`Feature`

`bzlmod/go_mod`

This PR ignore the `ignore` directive in go.mod file, to avoid causing the parsing of the _go.mod_ file to fail when this new Go 1.25 directive is used.


lbcjbb:
> Could you send the same kind of PR to rules_go's go_sdk?

This function `version_from_go_mod()` was written to ignore all unknown _go.mod_ directives, and to only process `go` and `toolchain` directives.

PR #2172, opened 2025-08-15
go/proto: Do not assume the repo_name being "io_bazel_rules_go"

zecke:
Delay initializating the defaultProtoCompilers/defaultGrpcProtoCompilers until there is a valid module to apparent name mapping.

Generate the proto/grpc compiler label using the apparent name mapping and fallback to io_bazel_rules_go if none exist.

**What type of PR is this?**

> Bug fix

> language/go

Makes running gazelle with a proto _and_ `repo_name = "rules_go"` work.

Fixes #

**Other notes for review**

### DIFF
--- a/language/go/config.go
+++ b/language/go/config.go
@@ -154,10 +154,24 @@ const (
 	fileTestMode
 )
 
-var (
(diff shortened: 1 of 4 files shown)


zecke:
This has finally been fixed by #2271. It should make working with spiffe easier again! Thanks!

PR #2174, opened 2025-08-15
expr.go: return error in extractPlatformStringsExprs if expression type is unmatched

### LINKED ISSUE #2173: expr.go's extractPlatformStringsExprs falsely reports no error if the expression type is unmatched
https://github.com/bazel-contrib/bazel-gazelle/blob/3ca2882ceb34d9e515ba088d9b0a2e7523de5706/rule/expr.go#L221

If the expression type is unmatched, this will result in no error appearing.

I discovered this while merging the following rule (roughly speaking) into itself:
```bzl
first_list = []
second_list = []

scala_library(
    deps = first_list + second_list
)
```

first_list and second_list are not lists or calls. this function will return an empty platformStringsExprs, causing the deps array to become empty.

tzaman-stripe:
**What type of PR is this?**
Bug fix

rule

This will return an error if the expression type is unmatched.

Fixes #2173

**Other notes for review**

### DIFF
--- a/rule/BUILD.bazel
+++ b/rule/BUILD.bazel
@@ -26,6 +26,7 @@ go_test(
     name = "rule_test",
     srcs = [
         "directives_test.go",
+        "expr_test.go",
         "merge_test.go",
         "rule_test.go",
         "value_test.go",
@@ -46,6 +47,7 @@ filegroup(
         "directives.go",
         "directives_test.go",
         "expr.go",
+        "expr_test.go",
         "merge.go",
         "merge_test.go",
         "platform.go",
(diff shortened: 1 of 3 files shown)


tzaman-stripe:
need to think some more whether this is actually appropriate

tzaman-stripe:
okay the issue is that this is somewhat of a breaking change?

before, if you merged a regular list into an expression that did not match the expected platformStringExprsFormat (e.g. ["foo"] into an_ident), an_ident would be treated as empty then foo would merge successfully.

but if you merge another_ident into an_ident, then I believe it would remove the attribute entirely. that doesn't seem right.

let me know what you think of this.

Issue #2177, opened 2025-08-27
fetch_repo fails to go mod download for go_repository with local_path


xuxife:
### What version of gazelle are you using?

v0.45.0

### What version of rules_go are you using?

v0.53.0

### What version of Bazel are you using?

v7.3.1

### Does this issue reproduce with the latest releases of all the above?

Yes

### What operating system and processor architecture are you using?

macOS (darwin) arm64

### What did you do?

I create a minimal repro in https://github.com/xuxife/gazelle-repro.

### What did you expect to see?

Build successfully with `replace ( libmod => ../libmod )`

### What did you see instead?

```
❯ bazelisk build //:mainmod
INFO: Repository gazelle~~go_deps~com_github_xuxife_gazelle_repro_libmod instantiated at:
  <builtin>: in <toplevel>
Repository rule go_repository defined at:
  /private/var/tmp/_bazel_xuxife/2d301f0bfbd1f6e1a7f95c67a152cfd0/external/gazelle~/internal/go_repository.bzl:455:32: in <toplevel>
ERROR: An error occurred during the fetch of repository 'gazelle~~go_deps~com_github_xuxife_gazelle_repro_libmod':
   Traceback (most recent call last):
	File "/private/var/tmp/_bazel_xuxife/2d301f0bfbd1f6e1a7f95c67a152cfd0/external/gazelle~/internal/go_repository.bzl", line 297, column 13, in _go_repository_impl
		fail("%s: %s" % (ctx.name, result.stderr))
Error in fail: gazelle~~go_deps~com_github_xuxife_gazelle_repro_libmod: fetch_repo: error running '[/private/var/tmp/_bazel_xuxife/2d301f0bfbd1f6e1a7f95c67a152cfd0/external/rules_go~~go_sdk~mainmod__download_0/bin/go mod download -json -modcacherw]': exit status 1 in /private/var/tmp/_bazel_xuxife/2d301f0bfbd1f6e1a7f95c67a152cfd0/external/gazelle~~go_deps~com_github_xuxife_gazelle_repro_libmod. Output: go: github.com/xuxife/gazelle-repro/depmod@v0.0.0-00010101000000-000000000000 (replaced by ../depmod): reading ../depmod/go.mod: open /private/var/tmp/_bazel_xuxife/2d301f0bfbd1f6e1a7f95c67a152cfd0/external/depmod/go.mod: no such file or directory
[...]

PR #1776, opened 2024-04-05
Go deps replace with module path

stefanpenner:
**What type of PR is this?**
Feature

**What package or component does this PR mostly affect?**
language/go

**What does this PR do? Why is it needed?**

This builds on [the go.work support PR](https://github.com/bazelbuild/bazel-gazelle/pull/1731) and [the new rctx.watch_files support ](https://github.com/bazelbuild/bazel/commit/fffa0affebbacf1961a97ef7cd248be64487d480) to add support for [FilePath ReplaceDirective,](https://go.dev/ref/mod#go-mod-file-replace) in go.mod and go.work files when using bzlmod go_deps extension.

For example, now the following native go capability works in the bazel world:

```go
module github.com/bazelbuild/bazel-gazelle/tests/bcr/go_mod

go 1.21.5

require (
	example.org/hello v1.0.0
        // ... snip ...
)

// ... snip ...

replace example.org/hello => ../../../fixtures/hello
```

# Which issues(s) does this PR fix?

https://github.com/bazelbuild/bazel-gazelle/issues/1775

# Status

- [x] make it "work"
- [x] ensure relative and absolute pathing works
  - [x] /Users/user/path/to/mod
  - [x] ../../path/to/mod
- [x] test with nightly bazel + [rctx.watch_files](https://github.com/bazelbuild/bazel/commit/fffa0affebbacf1961a97ef7cd248be64487d480) and verify it fixes "rebuilds"
- [ ] expand bcr tests
  - [x] go.mod
  - [ ] go.work
- [ ] add go specific unit tests for new capabilities added to `cmd/fetch_repo`
- [ ] even less hack and even more quality
- [ ] fix tests that fail on the bazelbuild/bazel-gazelle CI

PR #2178, opened 2025-08-27
moduleFromPath don't need go mod download

xuxife:
**What type of PR is this?**

Bug fix

cmd/fetch_repo
go_repository

go_repository with local_path will fail if the `go.mod` having another replace clause for the repository.
Remove the unnecessary `go mod download` since the repository should already in local if `local_path` is set.

#2177 

Fixes #

**Other notes for review**

### DIFF
--- a/walk/walk.go
+++ b/walk/walk.go
@@ -379,7 +379,8 @@ func newWalker(c *config.Config, cexts []config.Configurer, dirs []string, mode 
 		relsToVisitSeen: make(map[string]struct{}),
 	}
 
-	w.populateCache()
+	// Asynchronously populate the walker cache in the background.
+	go w.populateCache()
 
 	return w, nil
 }


fmeum (commented):
@jayconrod I would like to get your opinion on this.

cmd/fetch_repo/path.go
```diff
@@ -15,23 +15,11 @@ limitations under the License.
 
 package main
 
-import (
-	"fmt"
-	"os/exec"
-)
-
 func moduleFromPath(from string, dest string) error {
```
jayconrod: @xuxife Could you inline this function and delete the file? If we only need to call `copyTree`, let's just do that at the call site.

jayconrod (commented):
I don't think I noticed when `go_repository` gained support for local module replacements. I think I'd advise using `local_path_override` instead, but it's probably useful in some cases.

In any case, it seems like we just need to copy the source code somewhere then run Gazelle. I can't think of why we'd need to run `go mod download`. The `-json` flag indicates we might have wanted to know something about the output, but we ignore it, so it must not be important. Seems fine to remove.

xuxife:
Thanks @jayconrod @fmeum for reviewing this.

Another issue we found during test is from `copyTree`.

```
ERROR: gazelle~~go_deps~<REPO_NAME>: fetch_repo: write /root/.aksbuilder/bazeloutput/a2fcf66cb2c332a524108539da4ddf37/external/gazelle~~go_deps~<REPO_NAME>/bazel-bin: copy_file_range: is a directory
```

When the go_repository (with local_path) having a symbolic link file (bazel-bin in our case),
`copyTree` fails to copy it around.

cmd/fetch_repo/path.go
xuxife: delete `path.go` and inline `copyTree` in `main.go`

jayconrod:
`copyTree` should not ignore symlinks. I think it's okay to either copy them as-is (preferred) or copy the underlying files or report an error. Ignoring them is not okay though.

A significant change to `copyTree` like this will require a test. It doesn't look like it's tested now at all, but it needs a test before a change. Could be a unit test in `cmd/fetch_repo` or preferably something in `internal/go_repository_test.go` that covers the `local_path` attribute of `go_repository`.

Side note: could you say more about how you're using `local_path`? I'm a little doubtful of the feature's usefulness, and I wonder if there's a simpler solution to the problem you're facing. You mentioned your target repository has a `bazel-bin` symbolic link, which implies that it's a Bazel workspace where you execute builds. It's probably better to have a single "main" Bazel workspace, but even if you have multiple, that implies that each workspace has its own `BUILD` files. Maybe you could use Bazel's `local_path_override` instead of `go_repository` with `local_path`?

xuxife:
Here is a minimal setup for our scenario (as I mentioned in issue #2177)
https://github.com/xuxife/gazelle-repro/blob/main/mainmod/MODULE.bazel#L47C1-L47C40

Basically, we have a monorepo with multiple go.mod historically, and these go modules have many **local replace** `replace (module => ../module)`.
We're onboarding our repo to Bazel and thanks to feature #1776 , gazelle adds support to **local replace**.
And now we have a `MODULE.bazel` for each `go.mod`

We didn't use `local_path_override` because the `go_repository` targets are generated from gazelle extension
```
go_deps.from_file(go_mod = "//:go.mod") // go_repository are generated from go.mod
```

jayconrod:
Thanks, I think I understand the structure.

If you want to fix the symlink issue and add a test for `copyTree`, that'd be fine for this PR.

I think I'd still advocate for checking in the generated build files and using `local_path_override`, but that's largely my personal taste. That maps closely to what the Go toolchain does. `go_repository` will need to be re-evaluated whenever a file changes in one of those modules, so it may make incremental builds much slower if those modules are large: it needs to copy all the source files and generate the build files again.

If you'll indulge my curiosity further, why multiple modules at all? When I worked on modules, we found most developers used multi-module repositories to avoid downloading dependencies from other projects, but both Go and Bazel fetch modules lazily, so it wasn't truly an issue. However, multi-module repositories can still be useful if the modules need to be released independently at different versions or if different teams cannot use the same versions of common dependencies.

xuxife:
Thanks.

I've update `copyTree` to cover symlink case:
- if the symlink is absolute, copy it as-is
- if the symlink is relative, will resolve it to abosolute path and create symlink with abs path

jayconrod:
> if the symlink is relative, will resolve it to abosolute path and create symlink with abs path

Why absolutize a relative symbolic link? This adds some risk of leaking information from the host machine into the build (home directory name and so on). It would be better to keep symlinks relative if we can.

xuxife:
Updated, copy symlink as-is

PR #2191, opened 2025-10-02
Add test coverage for external Go submodules

fmeum:
**What type of PR is this?**

Test

cmd/gazelle

Verify that Gazelle can resolve both a Go module and a submodule as deps of the same target.

Also add an exclude to the repo tools srcs generator that covers the new JetBrains Bazel plugin.

**Other notes for review**


sluongng (approved):
LGTM

jayconrod (approved):

PR #2194, opened 2025-10-03
Automatically deleting corrupted Go module cache

linzhp:
**What type of PR is this?**

Feature

go_repository

When the Go module sum mismatches, the most effective way is to delete the corrupted Go module, instead of deleting the whole Go module cache, which may contain many other unrelated modules. Because fetch_repo already know where the Go module is, it can automatically delete it and ask people to retry. 

**Other notes for review**

New experience:
```
$ export GO_REPOSITORY_USE_HOST_CACHE=1
# intentionally corrupt a module
$ rm $(go env GOMODCACHE)/github.com/mattn/go-sqlite3@v1.14.15/static_mock.go
$ bazel clean --expunge --async
$ bazel fetch --force @com_github_mattn_go_sqlite3//:go-sqlite3
...
ERROR: com_github_mattn_go_sqlite3: fetch_repo: resulting module with sum h1:NlqjV/4w2ZBKwamqKNXcb6I12S9NTFOi18B0VlNju6g=; expected sum h1:vfoHhTN1af61xCRSWzFIWzx2YskyMTwHLrExkBOjvxI=. Please try again. If the problem persists, please try clearing your host module cache with `go clean -modcache`
```

### DIFF
--- a/cmd/fetch_repo/module.go
+++ b/cmd/fetch_repo/module.go
@@ -66,9 +66,13 @@ func fetchModule(dest, importpath, version, sum string) error {
 	}
 
 	if repoSum != sum {
+		errMsg := fmt.Sprintf("resulting module with sum %s; expected sum %s.", repoSum, sum)
 		// The module cache is corrupt. Remove the downloaded directory.
-		os.RemoveAll(dl.Dir)
-		errMsg := fmt.Sprintf("resulting module with sum %s; expected sum %s. Please try again.", repoSum, sum)
+		if err := os.RemoveAll(dl.Dir); err != nil {
+			errMsg += fmt.Sprintf(" Additionally, failed to remove corrupt module cache directory %q: %v. Please remove it manaully and retry.", dl.Dir, err)
+		} else {
+			errMsg += " Please retry."
+		}
 		if goModCache := os.Getenv("GOMODCACHE"); goModCache != "" {
 			errMsg += fmt.Sprintf(" If the problem persists, please try clearing your host module cache with `go clean -modcache`")
 		} else {
--- a/cmd/fetch_repo/module.go
+++ b/cmd/fetch_repo/module.go
@@ -65,10 +65,15 @@ func fetchModule(dest, importpath, version, sum string) error {
 	}
 
 	if repoSum != sum {
+		// The module cache is corrupt. Remove the downloaded directory.
+		os.RemoveAll(dl.Dir)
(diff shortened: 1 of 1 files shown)


cmd/fetch_repo/module.go
```diff
@@ -65,10 +65,15 @@ func fetchModule(dest, importpath, version, sum string) error {
 	}
 
 	if repoSum != sum {
+		// The module cache is corrupt. Remove the downloaded directory.
+		os.RemoveAll(dl.Dir)
```
fmeum: This operation isn't atomic. Could this result in races with other concurrent runs? 

An alternative could be to move the directory into an unused location on the same file system and then delete it.

jayconrod: I'm not sure if racing will be an issue: I suspect it will work. But moving then deleting could introduce other problems on Windows. Deleting or renaming files that have recently been written or renamed is unreliable.

This doesn't need to be a fatal error, but it might be worth capturing any error returned and including that in the error message.

cmd/fetch_repo/module.go
```diff
@@ -65,10 +65,15 @@ func fetchModule(dest, importpath, version, sum string) error {
...
+		errMsg := fmt.Sprintf("resulting module with sum %s; expected sum %s. Please try again.", repoSum, sum)
 		if goModCache := os.Getenv("GOMODCACHE"); goModCache != "" {
-			return fmt.Errorf("resulting module with sum %s; expected sum %s, Please try clearing your module cache directory %q", repoSum, sum, goModCache)
+			errMsg += fmt.Sprintf(" If the problem persists, please try clearing your host module cache with `go clean -modcache`")
+		} else {
+			errMsg += fmt.Sprintf(" If the problem persists, please try clearing Bazel output directory with `bazel clean --expunge`")
 		}
-		return fmt.Errorf("resulting module with sum %s; expected sum %s", repoSum, sum)
+		return fmt.Errorf(errMsg)
```
jayconrod: Nit: use `errors.New` instead since this doesn't need formatting.

jayconrod (approved):

cmd/fetch_repo/module.go
linzhp: > Could this result in races with other concurrent runs?

It probably would, but since the Go module is already corrupted, the concurrent runs will fail anyways even without the race

fmeum (approved):

PR #2201, opened 2025-10-16
Replace deprecated rules_proto with protobuf rules

mmorel-35:
**What type of PR is this?**

Dependency management

language

As rules_proto is deprecated in favor of protobuf, this replaces the use rules_proto with the protobuf version

Fixes #

**Other notes for review**


fmeum:
While rules_proto is deprecated Protobuf is still working on feature parity, in particular regarding its toolchain story. Unless using rules_proto results in any concrete issues I would rather wait a bit longer before we migrate over. We don't want to migrate twice

PR #2204, opened 2025-10-19
perf: avoid use of reflection when generating bzl expressions

jbedard:
**What type of PR is this?**

Other

all

Perf. Reduce use of `reflect.*` in common cases.

### DIFF
--- a/rule/value.go
+++ b/rule/value.go
@@ -16,10 +16,10 @@ limitations under the License.
 package rule
 
 import (
-	"fmt"
 	"log"
 	"reflect"
 	"sort"
+	"strconv"
 
 	bzl "github.com/bazelbuild/buildtools/build"
 )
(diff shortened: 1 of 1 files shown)


rule/value.go
```diff
@@ -228,27 +229,46 @@ func ExprFromValue(val interface{}) bzl.Expr {
...
+	case []interface{}:
+		return interfaceSliceToExpr(v)
+	}
 
-	case reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64,
-		reflect.Uint, reflect.Uint8, reflect.Uint16, reflect.Uint32, reflect.Uint64:
+	// Fallback to reflection for less common types
+	rv := reflect.ValueOf(val)
+	switch rv.Kind() {
+	case reflect.Int8, reflect.Int16, reflect.Int32,
```
jbedard: All of these primitives could move really, I just did the ones that I (and magic AI robot) expected to be most common. WDYT?

jayconrod: Move everything that can be moved I think. You might need a second type switch without the assignment (`switch val.(type)` rather than `switch v := val.(type)`) so the code can be shared by these cases. But it's better than reflection.

rule/value.go
```diff
@@ -299,3 +319,25 @@ func (s byString) Less(i, j int) bool {
 func (s byString) Swap(i, j int) {
 	s[i], s[j] = s[j], s[i]
 }
+
+func stringSliceToExpr(strs []string) *bzl.ListExpr {
+	if len(strs) == 0 {
```
jayconrod: Nit: drop the `len(strs) == 0` case for both of these. The code looks like it should work without needing the special case.

jbedard:
I've been seeing `ExprFromValue` in both cpu and memory profiling. This change (and https://github.com/bazel-contrib/bazel-gazelle/pull/2205) reduces it a small amount and I no longer see the reflection showing at all.

If you think the code is better as a switch like this then it's worth the change imo, if the code was worse then it would be debatable.

rule/value.go
jbedard: I've moved all primitives. I'm not sure how to share code like previously though, when using the type-switch (with or without the assignment) it seems you can't fallthrough or share `case`s...

jayconrod (approved):

PR #2205, opened 2025-10-19
perf: dont use sprintf for simple string concat

jbedard:
**What type of PR is this?**

Other

all

Avoid use of `Sprintf` for frequent simple concatenation.


jbedard:
No, Go definitely can't optimize this itself, and it shows up when profiling sometimes. It's not significant, but the diff is simple enough to be worth it 🤷

PR #2209, opened 2025-10-22
Commit MODULE.bazel.lock file

### LINKED ISSUE #2208: Commit MODULE.bazel.lock files
This ensures Gazelle developers always see the same bits when downloading
files. We should see a security error if a module version changes upstream.

jayconrod:
**What type of PR is this?**

> Other

> all

This ensures Gazelle developers always see the same bits when downloading files. We should see a security error if a module version changes upstream.

Fixes #2208

**Other notes for review**

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -324,10 +324,12 @@ type indexFlag struct {
 }
 
 func (f indexFlag) String() string {
+	indexLibraries := f.indexLibraries != nil && *f.indexLibraries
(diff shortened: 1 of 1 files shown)


jayconrod:
I tried to add lock files in `tests/bcr`, but it looks like we test across a range of Bazel versions there, and one lock file won't satisfy them all.

PR #2212, opened 2025-10-24
fix: only return requested dirinfo on GetDirInfo error

jbedard:
**What type of PR is this?**

Bug fix

all

Ensure `GetDirInfo()` only returns `DirInfo` for the requested dir, or empty on error.

If there is an error loading a directory such as a syntax error in a BUILD file then `GetDirInfo` will detect that error and return without recursing further into the `GetDirInfo(rel)`. It will return the error but also the `DirInfo` for the directory containing the error.

If we are doing `GetDirInfo("a/b/c")` and `a/b/c/BUILD` had a syntax error, it is non-intuitive for `GetDirInfo("a/b/c")` to return the `DirInfo` for `a/b`.

**Other notes for review**

In my case I was recursing into subdirectories and when `GetDirInfo` returns the info of a parent directory it caused infinite recursion. This should probably by solved on my part by not recursing on error, but I think this change within gazelle makes return result on error clearer?

The alternative is `GetDirInfo` should return the directory the `DirInfo` is for. Maybe change the signature to `GetDirInfo(rel string) (string, DirInfo, error)` or adding `DirInfo.Pkg`.

### DIFF
--- a/walk/cache.go
+++ b/walk/cache.go
@@ -110,8 +110,10 @@ func GetDirInfo(rel string) (DirInfo, error) {
 	// configuration may exclude rel.
 	var prevCfg *walkConfig = nil
 	var di DirInfo
+	var d string
 	var err error
 	pathtools.Prefixes(rel)(func(prefix string) bool {
+		d = prefix
 		if prevCfg != nil && prevCfg.isExcludedDir(prefix) {
 			di = DirInfo{}
 			err = fmt.Errorf("directory %q is excluded", prefix)
@@ -121,5 +123,12 @@ func GetDirInfo(rel string) (DirInfo, error) {
 		prevCfg = di.config
 		return err == nil
 	})
+
+	if err != nil {
+		if d != rel {
+			di = DirInfo{}
+		}
+	}
+
 	return di, err
 }
--- a/walk/walk_test.go
+++ b/walk/walk_test.go
@@ -855,6 +855,38 @@ func TestGetDirInfoSubdir(t *testing.T) {
 	}
 }
 
+func TestGetDirInfoErrorOnParent(t *testing.T) {
+	dir, cleanup := testtools.CreateFiles(t, []testtools.FileSpec{
+		{
+			Path:    "BUILD.bazel",
+			Content: `filegroup(name = "foo". srcs = ["parent/child/file.txt"])`,
+		},
+		{
+			Path: "parent/child/file.txt",
+		},
+	})
+	defer cleanup()
+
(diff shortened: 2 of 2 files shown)


jbedard:
I've added a test.

I do wonder if the behaviour should be different though. When doing `GetDirInfo("a/b/c")` and there is a syntax error in `a/BUILD.bazel`... it could potentially still return the info about `a/b/c` despite an error in the parent BUILD. That would align more with what `Walk2` does...

jbedard:
Also note this code could all be simplified with go 1.23. I assume this bug wouldn't have existed if we could just use a plain loop:
```
	var prevCfg *walkConfig = nil
	var r DirInfo
	for prefix := range pathtools.Prefixes(rel) {
		if prevCfg != nil && prevCfg.isExcludedDir(prefix) {
			return DirInfo{}, fmt.Errorf("directory %q is excluded", prefix)
		}
		di, err := globalWalker.cache.get(prefix, globalWalker.loadDirInfo)
		if err != nil {
			return DirInfo{}, err
		}
		r = di
		prevCfg = di.config
	}
	return r, nil
```

Maybe upgrading the minimum go version can be added to the 2.0 list? Or can we now? 🤔

jayconrod:
> I do wonder if the behaviour should be different though. When doing GetDirInfo("a/b/c") and there is a syntax error in a/BUILD.bazel... it could potentially still return the info about a/b/c despite an error in the parent BUILD. That would align more with what Walk2 does...

Maybe, but it could end up doing something wrong or different because it's missing directives from that file. Syntax errors should be rare in any case.

> Also note this code could all be simplified with go 1.23. I assume this bug wouldn't have existed if we could just use a plain loop: ... Maybe upgrading the minimum go version can be added to the 2.0 list? Or can we now? 🤔

Semver doesn't require a major version release after upgrading a dependency by a minor version. So doesn't need to be in 2.0.

I think it's fine for 1.23 to be the minimum version at this point. It's already past its end-of-life by the Go team.

jbedard:
> Maybe, but it could end up doing something wrong or different because it's missing directives from that file. Syntax errors should be rare in any case.

Isn't that what `Walk2` does today though? It continues `Configure`ing directories despite there being errors parsing BUILDs. I'm not sure what is correct, but currently `Walk2` and this util seem inconsistent IIUC.

jbedard:
> I think it's fine for 1.23 to be the minimum version at this point. It's already past its end-of-life by the Go team.

I might try that then...

PR #2213, opened 2025-10-24
[bzlmod/go_deps] Clarify support for modules that differ only by case.

mkosiba:
**What type of PR is this?**

Bug fix / minor enhancement

bzlmod go_deps/go_repository

Go tooling allows go.mod files to contain entries that differ only by case. 
If that happens the `go_deps` extension will `fail(..)` with an error message describing the problem.

~In this PR we're adding support for the simple case, where the dependency is at the same version (as is the case [here](https://github.com/apognu/gocal/blob/master/go.mod#L6)).~

In this PR we're clarifying the error message so it doesn't suggest that this is desired/planned functionality.

Seems minor enough to not need an issue?

### DIFF
--- a/tests/bcr/go_mod/MODULE.bazel
+++ b/tests/bcr/go_mod/MODULE.bazel
@@ -16,7 +16,7 @@ local_path_override(
 
 bazel_dep(name = "bazel_features", version = "1.27.0")
 bazel_dep(name = "protobuf", version = "27.0", repo_name = "my_protobuf")
-bazel_dep(name = "rules_go", version = "0.58.3", repo_name = "my_rules_go")
+bazel_dep(name = "rules_go", version = "0.59.0", repo_name = "my_rules_go")
 bazel_dep(name = "rules_license", version = "1.0.0")
 bazel_dep(name = "package_metadata", version = "0.0.5")
 bazel_dep(name = "rules_proto", version = "6.0.0", repo_name = "my_rules_proto")
@@ -27,7 +27,7 @@ go_sdk.download(version = "1.24.2")
 
 # This bazel_dep provides the Go dependency github.com/cloudflare/circl, which requires custom
 # patches beyond what Gazelle can generate.
-bazel_dep(name = "circl", version = "1.3.7")
+bazel_dep(name = "circl", version = "1.3.8")
 
 go_deps = use_extension("@gazelle//:extensions.bzl", "go_deps")
 go_deps.config(
--- a/tests/bcr/go_mod/go.mod
+++ b/tests/bcr/go_mod/go.mod
@@ -10,10 +10,10 @@ require (
 	example.org/hello v1.0.0
 	github.com/DataDog/sketches-go v1.4.1
 	github.com/bazelbuild/buildtools v0.0.0-20251031164759-f48b23493530
-	github.com/bazelbuild/rules_go v0.58.3
+	github.com/bazelbuild/rules_go v0.59.0
 	// NOTE: keep <4.7.0 to test the 'replace'
 	github.com/bmatcuk/doublestar/v4 v4.6.0
-	github.com/cloudflare/circl v1.3.7
+	github.com/cloudflare/circl v1.3.8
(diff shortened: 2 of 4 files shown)


fmeum (commented):
~This looks pretty errorprone and not obviously safe. Could you share your use case for this? BCR modules for Go deps should already be rare.~ Nvm, I thought you referred to Bazel modules.

fmeum:
Repository names can contain uppercase letters: https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/cmdline/RepositoryName.java;l=55;drc=6728e381752f49171c1fa8419c37cae2875816ad

If we wanted to support this, we should probably just preserve the case if there would otherwise be a conflict.

cc @jayconrod

fmeum:
Is there a case in which you wouldn't just want to `replace` one casing with the other?

jayconrod:
I think it would be better not to support this. And I don't think the fix is correct because it treats two distinct modules as the same thing.

Around the time of the Go modules migration, there was a popular module that renamed itself from `github.com/Sirupsen/logrus` to `github.com/sirupsen/logrus`. Go treats these as two distinct modules. It's equivalent to releasing a new major version with a `/v2` suffix: both modules may exist in the same build. That was not the intent of the module authors nor anyone depending on the modules, and it caused a lot of bugs and conflicts.

It caused problems with Bazel too because both paths get converted to `com_github_sirupsen_logrus`. So when someone depended on both, they needed to manually manage their `WORKSPACE` files to say which one they wanted, or they needed to fix the case and stop Gazelle from fixing it back. There were bug fixes and workarounds in Gazelle for that, too.

If there were a naming conflict like this with another popular module, we could probably support a workaround for it. That would probably involve letting users override the repo name that `go_deps` uses for a particular module. I'd rather keep things simple and not support that if possible though: changing the repo name can break existing build files.

In this example, `github.com/ChannelMeter/iso8601duration` has no `go.mod` file, so there is no canonical module path. The last commit was 11 years ago. I don't think this justifies a workaround that adds complication.

mkosiba:
> Is there a case in which you wouldn't just want to replace one casing with the other?

@fmeum  Using `replace` was my first idea for a workaround, but doing that doesn't make the error go away.

@jayconrod I remember the logrus rename, fun times..

> In this example, github.com/ChannelMeter/iso8601duration has no go.mod file, so there is no canonical module path. The last commit was 11 years ago. I don't think this justifies a workaround that adds complication.

The don't think the example deserves a specific workaround in `bazel-gazelle`. That being said I'd love some sort of workaround for it. Do you think it'd be better to change the logic such that we don't error if there is a replace in place?

jayconrod:
`replace` changes the content of a module but doesn't change its existence.

I think this is really an edge case: a `go.mod` file determines the canonical path for a module, and there are very few modules without `go.mod` files. Why not find another module that does the same thing, or implement the functionality within your own module? It's basically a regular expression.

mkosiba:
> Why not find another module that does the same thing, or implement the functionality within your own module? It's basically a regular expression.

It's an indirect dependency. If it was a direct dependency it'd be much easier to fix. 

> replace changes the content of a module but doesn't change its existence.

That'd be OK, I think. Initially you were concerned that we'd be mapping two different Go modules to a single Bazzel repository. If we're saying that both `github.com/XXX` and `github.com/xxx` now have the same contents it's no longer incorrect to map both to `com_github_xxx`.

Do you see a path to supporting this case at all? If not then that's fine but maybe we should change the error message to say it's unlikely supported at all, instead of the "(yet)" which, to me at least, implies "if you want it - PRs welcome".

jayconrod:
If a widely used module changed its name in a way that caused a conflict like this, then it would be worthwhile to support a workaround. But I don't think it's worth it to support a small one without a go.mod file like this. It would be fine to drop the `"(yet)"`.

mkosiba:
> It would be fine to drop the "(yet)".

done.

jayconrod (approved):

jayconrod:
CI is red, but not due to this PR. Created #2234 to track that problem.

PR #2214, opened 2025-10-24
chore: upgrade go version to 1.23

jbedard:
Upgrade to go 1.23 and make a few changes using new APIs.

**What type of PR is this?**

Other

all

Upgrade golang to make use of modern language features

### DIFF
--- a/cmd/fetch_repo/module.go
+++ b/cmd/fetch_repo/module.go
@@ -247,7 +247,7 @@ func isBadNum(v string) bool {
 // The caller is assumed to have checked that semver.IsValid(v) is true.
 func isSemverPrefix(v string) bool {
 	dots := 0
-	for i := 0; i < len(v); i++ {
+	for i := range len(v) {
 		switch v[i] {
 		case '-', '+':
 			return false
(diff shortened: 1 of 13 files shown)


WORKSPACE
```diff
@@ -51,7 +52,7 @@ go_register_nogo(
 # TODO: Fix rules_go and set this back to 1.19.
 go_download_sdk(
     name = "go_compat_sdk",
-    version = "1.22.9",
+    version = "1.23.12",
```
fmeum, line 55: We definitely have users that lag behind the supported versions. Go the language is highly backwards compatible, but the Go tools are subject to frequent breaking changes (see our coverage story). If we gained something major I would be all in, but just for some cleanups here and there I would prefer to wait a bit longer.

fmeum: Cc @jayconrod

jbedard: IIUC @jayconrod seemed open to updating to at least 1.23: https://github.com/bazel-contrib/bazel-gazelle/pull/2212#issuecomment-3444571808

Otherwise maybe this should wait for "v2"?

jbedard:
I guess this was done in https://github.com/bazel-contrib/bazel-gazelle/pull/2274

PR #2215, opened 2025-10-26
Generate dictionary with go tool targets

mrdomino:
**What type of PR is this?**

> Feature

> go_repository

This adds Bazel support for [go1.24 tool directives][0]. A `go_tools.bzl` is emitted that contains a `GO_TOOLS` dictionary with binary names as keys and targets as values. This dictionary can be unioned into a [`bazel_env.bzl`][1] `tools` attribute to make a module’s tools accessible from that env.

Part of buildbuddy-io/bazel_env.bzl#68

**Other notes for review**

I consulted with Claude on how to build the `tool_targets` dict; I did not see any prior Starlark implementation of a Go package path to Bazel module transform.

[0]: https://tip.golang.org/doc/go1.24#tools
[1]: https://github.com/buildbuddy-io/bazel_env.bzl

### DIFF
--- a/internal/go_repository_config.bzl
+++ b/internal/go_repository_config.bzl
@@ -66,7 +66,7 @@ def _go_repository_config_impl(ctx):
     )
     ctx.file(
         "go_tools.bzl",
-        "TOOLS = {}",
+        "GO_TOOLS = {}",
         False,
     )
 


internal/bzlmod/go_deps.bzl
```diff
@@ -259,8 +259,12 @@ def _go_repository_config_impl(ctx):
         ))
 
     ctx.file("WORKSPACE", "\n".join(repos))
-    ctx.file("BUILD.bazel", "exports_files(['WORKSPACE', 'config.json', 'go_env.bzl'])")
+    ctx.file("BUILD.bazel", "exports_files(['WORKSPACE', 'config.json', 'go_env.bzl', 'go_tools.bzl'])")
     ctx.file("go_env.bzl", content = "GO_ENV = " + repr(ctx.attr.go_env))
+    ctx.file("go_tools.bzl", content = "TOOLS = {\n" + "\n".join([
```
fmeum: This can be cleaned up by replacing the manual dict printing with `repr`.

internal/bzlmod/go_deps.bzl
```diff
@@ -726,6 +733,32 @@ Mismatch between versions requested for Go module {module}:
...
+
+        # Start from the longest possible prefix and work down
+        for i in range(len(segments), 0, -1):
+            candidate = "/".join(segments[:i])
+            if candidate in importpath_to_repo:
+                module_path = candidate
+                break
+
+        if module_path:
+            tool_name = segments[-1]
```
fmeum, line 755: This is something I hadn't thought of before: how do we find a good name? The last segment could be `cmd`.

I see two options: 
* Provide a way for users to specify a custom name, perhaps by reading an end of line comment from `go.mod`.
* Export a dict of full tool paths to labels and let consumers implement their own heuristic (which can be as simple as `path.rpartition()[-1]` to match what we do here).

I have a slight preference for the second option. 

@jayconrod Based on your experience, do you see a good way to derive a (configurable) short name for a Go tool?

fmeum: From the go docs:

> Once this directive is added you can run the tool by passing the last non-major-version component of the tool’s import path to go tool

That makes me lean more towards the first option if we also add the logic required to do over major version components.

internal/bzlmod/go_deps.bzl
mrdomino: Ah, I'd thought I needed to do it manually to inject the Label calls, is that handled somehow?

internal/bzlmod/go_deps.bzl
mrdomino: Yeah, I wondered about that; it seems like for the moment tools just need to have globally unique names to be used in Go tool directives. I guess for versions, we would need logic to skip `v[0-9]+` as a last component?

mrdomino: Maybe `replace` is an escape hatch already for this sort of name conflict? (On a phone atm; am not exactly familiar with the guidance on replace.)

mrdomino:
> Could you add test coverage for this by loading the constant and asserting on its contents somewhere in `tests/bcr/go_mod`?

Yeah, I'll try to get to that today if I can.

internal/bzlmod/go_deps.bzl
fmeum: If the dict values are `Label`s, then `repr` should account for that. It is currently buggy for main repo labels, but those never come up here.

mrdomino: Hrm, I did not have luck with just applying `Label` to my labels and calling `repr`; the repo name was not recognized from gazelle’s context.

mrdomino:
@fmeum PTAL

internal/bzlmod/go_deps.bzl
jayconrod: I think it's best to use the same name that `go tool` would use. I think this would also be the same name `go install` would pick for a binary name. Hopefully authors know to avoid naming their tools `cmd`.

mrdomino:
IIUC it would be something like:

```go
load("@bazel_env.bzl", "bazel_env")
load("//:go_tools.bzl", GO_TOOLS = "TOOLS")

bazel_env(
    name = "bazel_env",
    tools = {
        "go": "@rules_go//go",
    } | GO_TOOLS,
)
```

mrdomino:
So then you’d run:
```sh
bazel run //:bazel_env
```
and afterwards, if you ran `gofumpt` you would get the `gofumpt` binary from the tool directive in your go.mod, as built by Bazel.

internal/bzlmod/go_deps.bzl
mrdomino: No wait, I think I get it; `Label` is just not needed explicitly here, right?

fmeum: They are needed since `use_repo` could be used to rename them on import in the module consuming the dictionary. My suggestion was wrong though, the `Label` calls have to be performed from within the generated file. Could you write out the string dict via `repr` and wrap it in `{k: Label(v) for k, v in ...}`? We don't have to worry about string escaping that way, but `Label` is still invoked from the correct point of view (the generated repo, not the Gazelle module repo).

mrdomino: Done.

internal/bzlmod/go_deps.bzl
```diff
@@ -726,6 +732,32 @@ Mismatch between versions requested for Go module {module}:
...
```
fmeum, line 755: Could you pick the last segment here that doesn't satisfy a condition such `len(s) >= 2 and s[0] == "v" and s[1:].isdigit()`? We would want to match the Go spec for tool names as closely as possible, but I think that we can ignore collisions for now.

mrdomino: Done, tried to just strip off the last `v\d+` path component.

fmeum (approved):

tests/bcr/go_mod/tests.bzl
```diff
@@ -1,4 +1,5 @@
 load("@bazel_features//:features.bzl", "bazel_features")
+load("@bazel_gazelle_go_repository_config//:go_tools.bzl", "TOOLS")
```
fmeum: Sorry, I missed that we don't want to make this repo public API. Could you create a new file, say `@rules_go//go:tools.bzl` that reexports this constant?

Could you also mention this feature in `bzlmod.md`?

mrdomino: All right; sounds like this needs work in rules_go as well as gazelle then? How should we handle backwards/forwards compatibility?

fmeum: Sorry, I'm switching between the repos too much. This can and should live in Gazelle, perhaps `@gazelle//:go_tools.bzl`?
@jayconrod any concerns about this location with respect to Gazelle 2.0

jayconrod: `@gazelle//:go_tools.bzl` seems reasonable. For 2.0, I want the Go extension to be less special. This makes it more special by adding a path outside `//language/go`, but the short name is worthwhile. I don't expect we'll move `go_deps` out of `extensions.bzl` either for example.

mrdomino: Done, re-exported `TOOLS` from `@gazelle//:go_tools.bzl`.

Looks like `bzlmod.md` is in `rules_go`, I can open up a doc PR there if this seems good.

fmeum (approved):

fmeum:
Thanks for going through all the iterations!

fmeum:
Whoops, one more. @mrdomino Could you run Gazelle or apply the diff printed by the failing test?

mrdomino:
D’oh, good catch. Done.

mrdomino:
Working on the test failures.

BUILD.bazel
```diff
@@ -136,3 +136,10 @@ bzl_library(
     visibility = ["//visibility:public"],
     deps = ["//internal/bzlmod:go_deps"],
 )
+
+bzl_library(
+    name = "go_tools",
+    srcs = ["go_tools.bzl"],
+    visibility = ["//visibility:public"],
+    deps = ["@bazel_gazelle_go_repository_config//:go_tools"],
```
fmeum:
```suggestion
    deps = ["@bazel_gazelle_go_repository_config//:go_tools.bzl"],
```

fmeum: May also need a `# keep`

mrdomino: Thanks—yeah, it also looked like maybe I needed to stub out that file for `WORKSPACE` builds so I did so.

mrdomino: > May also need a `# keep`

Seems like gazelle is okay with it after it has been populated.

mrdomino:
Hrm, looking at https://github.com/bazel-contrib/rules_go/pull/4496, I feel like the variable is probably better named `GO_TOOLS` so it doesn’t need to be renamed when imported. Thoughts?

fmeum:
> Hrm, looking at https://github.com/bazel-contrib/rules_go/pull/4496, I feel like the variable is probably better named `GO_TOOLS` so it doesn’t need to be renamed when imported. Thoughts?

Sounds reasonable!

mrdomino:
Done.

fmeum (approved):

PR #2217, opened 2025-10-28
language/go: handle cgo noescape and nocallback directives

### LINKED ISSUE #2216: cgo `noescape` and `nocallback` directives not handled
Gazelle fails to parse go files containg the cgo directives `#cgo noescape <func>` and `#cgo nocallback <func>` with the error message:
```
error reading go file: <path>: invalid #cgo line: #cgo noescape <func>
```

See https://pkg.go.dev/cmd/cgo#hdr-Optimizing_calls_of_C_code for a description of both directives.

torin-carey:
These directives currently cause gazelle to fail to process the given file. Update gazelle to use the same check as saveCgo does.

**What type of PR is this?**
Bug fix

language/go

This PR updates the `#cgo` directive parsing in `language/go` to handle these two new(ish) directives. Currently these directives will cause gazelle to bail on that file.

Fixes #2216

**Other notes for review**

### DIFF
--- a/language/go/fileinfo_go_test.go
+++ b/language/go/fileinfo_go_test.go
@@ -328,6 +328,28 @@ import ("C")
 				},
 			},
 		},
+		{
+			"noescape and nocallback directives",
+			`package foo
+
+/*
+#cgo noescape my_func
+#cgo nocallback my_func
+void my_func(void *param);
+
+// This directive won't be parsed if we failed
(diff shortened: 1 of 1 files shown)


torin-carey:
I wasn't sure which test directory you meant by `tests/core/cgo`, but I've added a test under `TestCgo` if that's okay?

PR #2220, opened 2025-10-29
CI: hopefully fix BCR tests

jayconrod:
**What type of PR is this?**

> Test

> all

The format for .bcr/presubmit.yml isn't well-documented. As far as I can tell, it's intepreted by buildkite/bazel-central-registry/bcr_presubmit.py from bazelbuild/bazel-central-registry. The "bcr_test_module" block is important, I shouldn't have deleted that. But it only accepts one module, so go_mod is it.

For #2208

**Other notes for review**

### DIFF
--- a/.bcr/presubmit.yml
+++ b/.bcr/presubmit.yml
@@ -1,32 +1,30 @@
-matrix:
-  platform:
-    - ubuntu2004
-    - macos
-    - windows
-  bazel: [6.*, 7.*, 8.*]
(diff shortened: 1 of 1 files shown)


jayconrod:
Sorry for all the noise on this topic. The BCR test script has some hardcoded assumptions about running on buildkite, and I can't easily test it directly. This is a relatively small change compared with the original, so I'm a bit more optimistic this time.

PR #2221, opened 2025-11-04
Fix BCR test module with Bazel 9

fmeum:
**What type of PR is this?**

Bug fix

BCR test

Work towards https://github.com/bazelbuild/bazel-central-registry/issues/6402

**Other notes for review**

### DIFF
--- a/tests/bcr/go_work/MODULE.bazel
+++ b/tests/bcr/go_work/MODULE.bazel
@@ -145,6 +145,13 @@ use_repo(
 # closure.
 go_tools = use_extension("@gazelle//:extensions.bzl", "go_deps", isolate = True)
 go_tools.from_file(go_mod = "//tools:go.mod")
+go_tools.gazelle_override(
+    build_file_generation = "clean",
+    directives = [
+        "gazelle:proto disable",
+    ],
+    path = "github.com/bazelbuild/buildtools",
+)
 use_repo(
     go_tools,
     buildtools = "com_github_bazelbuild_buildtools",


fmeum:
@jayconrod It looks like we need to drop Bazel 6 from the BCR test presubmit since it depends on a version of rules_jvm_external that has dropped support for 6 by using `use_repo_rule`. Do you see an ergonomic way to achieve that in the Bazel CI config?

jayconrod:
I'm not sure I understand. So the BCR presubmit depends on a version of rules_jvm_external that is no longer compatible with Bazel 6? Or rules_go? It sounds like the only thing we can do is drop Bazel 6 from the BCR test configuration.

You have that already for `.bcr/presubmit.yml`, but you'll also need to remove `6.x` from `.bazelci/presubmit.yml` for this PR to be green. We have no way of running the BCR tests ourselves, but `.bazelci/presubmit.yml` attempts to do the same thing.

fmeum:
The test module transitively depends on such a version now.

> but you'll also need to remove 6.x from .bazelci/presubmit.yml for this PR to be green

Ideally I would only do this for the BCR test, not for the rest of the CI jobs. But I don't know the syntax for using multiple test matrices - if that's supported at all?

jayconrod:
Multiple test matrices aren't supported. And actually the current matrix *only* applies to the `bcr_tests` task because that's the only place the parameters are expanded (with `${{ bazel }}` and so on). All other jobs use whatever's in `.bazelversion`.

.bazelci/presubmit.yml
```diff
@@ -4,7 +4,7 @@ matrix:
     - ubuntu2204
     - macos
     - windows
-  bazel: [6.*, 7.*, 8.*]
+  bazel: [7.*, 8.*]
```
jayconrod: Do you want to add `9.*` here too? Ideally we're testing the same thing in presubmit as in the BCR tests. The BCR tests only run post-release, so if we find an issue there, we need to create a new tag. 😭

jayconrod (approved):

PR #2223, opened 2025-11-06
Fix go tools at module roots

mrdomino:
**What type of PR is this?**

> Bug fix

> go_repository

This fixes a bug in bazel-contrib/bazel-gazelle#2215 wherein tools located at module roots would not be referenced correctly from the `GO_TOOLS` dictionary. The proposed fix is to always explicitly specify the target name, which is assumed to always be the last path component in the package path.

Part of buildbuddy-io/bazel_env.bzl#68

**Other notes for review**

Though the diff is in the end pretty small (I think), I leaned on Claude Code a fair bit to put this together, as you can see from mrdomino/bazel-gazelle#2. Mainly, I asked Claude to help me research how Gazelle generates the last path component in the binary targets it produces; from what I can tell, this is usually just the last path component, even if that component is a major version:

https://github.com/bazel-contrib/bazel-gazelle/blob/8ce02f76d6a7b248ac82ddb23136c1b6704183da/language/go/generate.go#L683-L686

https://github.com/bazel-contrib/bazel-gazelle/blob/8ce02f76d6a7b248ac82ddb23136c1b6704183da/language/go/package.go#L254-L257

https://github.com/bazel-contrib/bazel-gazelle/blob/8ce02f76d6a7b248ac82ddb23136c1b6704183da/pathtools/path.go#L55-L72

### DIFF
--- a/tests/bcr/go_mod/tests.bzl
+++ b/tests/bcr/go_mod/tests.bzl
@@ -74,13 +74,13 @@ _PackageInfoSubjectFactory = struct(
 
 def starlark_tests(name):
     if "staticcheck" not in GO_TOOLS:
-        fail("Expected 'staticcheck' in 'GO_TOOLS'")
+        fail("Expected 'staticcheck' in 'GO_TOOLS', got: {}".format(GO_TOOLS))
     if GO_TOOLS["staticcheck"] != Label("@co_honnef_go_tools//cmd/staticcheck"):
-        fail("Unexpected value for 'staticcheck': {} (GO_TOOLS={})".format(GO_TOOLS["staticcheck"], GO_TOOLS))
+        fail("Unexpected value for 'staticcheck': {}".format(GO_TOOLS["staticcheck"]))
     if "gofumpt" not in GO_TOOLS:
-        fail("Expected 'gofumpt' in 'GO_TOOLS'")
+        fail("Expected 'gofumpt' in 'GO_TOOLS', got: {}".format(GO_TOOLS))
     if GO_TOOLS["gofumpt"] != Label("@cc_mvdan_gofumpt//:gofumpt"):
-        fail("Unexpected value for 'gofumpt': {} (GO_TOOLS={})".format(GO_TOOLS["gofumpt"], GO_TOOLS))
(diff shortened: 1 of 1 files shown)


tests/bcr/go_mod/tests.bzl
```diff
@@ -75,8 +75,12 @@ _PackageInfoSubjectFactory = struct(
 def starlark_tests(name):
     if "staticcheck" not in GO_TOOLS:
         fail("Expected 'staticcheck' in 'GO_TOOLS'")
-    if GO_TOOLS["staticcheck"] != Label("@co_honnef_go_tools//cmd/staticcheck"):
+    if GO_TOOLS["staticcheck"] != Label("@co_honnef_go_tools//cmd/staticcheck:staticcheck"):
```
fmeum: You don't have to change this, the `Label` instances will be equal.

fmeum (commented):
Checks are failing. Could you let the `fail` calls print `GO_TOOLS` so that it's easier to understand why?

mrdomino:
Ah, I think Claude just edited `go.mod` instead of calling `go get -tool` for gofumpt.

Alas, after fixing that, I tried and failed to run `bazel run @my_rules_go//go -- mod tidy` just now:
```
go: finding module for package golang.org/x/tools/go/expect
go: finding module for package github.com/bazelbuild/bazel-gazelle/tests/bcr/proto/foo
go: found golang.org/x/tools/go/expect in golang.org/x/tools/go/expect v0.1.1-deprecated
go: finding module for package github.com/bazelbuild/bazel-gazelle/tests/bcr/proto/foo
go: github.com/bazelbuild/bazel-gazelle/tests/bcr/go_mod/proto tested by
	github.com/bazelbuild/bazel-gazelle/tests/bcr/go_mod/proto.test imports
	github.com/bazelbuild/bazel-gazelle/tests/bcr/proto/foo: module github.com/bazelbuild/bazel-gazelle/tests/bcr@latest found (v0.0.0-20240416214000-42b778147366), but does not contain package github.com/bazelbuild/bazel-gazelle/tests/bcr/proto/foo
```

fmeum:
`-e` should help it progress through that issue.

mrdomino:
Cool, thank you. That looks like it did the trick.

PR #2230, opened 2025-11-16
Add default Gazelle override for tflint-plugin-sdk

rorychatterton:
**What type of PR is this?**

Bug fix

bzlmod

Adds `gazelle:proto disable` directive for
`github.com/terraform-linters/tflint-plugin-sdk` to the default overrides.

The tflint-plugin-sdk repository includes pre-generated protobuf bindings.
When Gazelle attempts to regenerate these bindings, it produces incompatible
code that causes build failures with undefined type errors.

I have verified that disabling proto generation allows successful builds of
projects depending on tflint-plugin-sdk by using a MODULE.bazel override
like:

```python
go_deps.gazelle_override(
    directives = [
        "gazelle:proto disable",
    ],
    path = "github.com/terraform-linters/tflint-plugin-sdk",
)
```

N/A 

**Other notes for review**

Small fix similar to #2062 for hashicorp/go-plugin)

### DIFF
--- a/internal/bzlmod/go_deps.bzl
+++ b/internal/bzlmod/go_deps.bzl
@@ -683,7 +683,7 @@ Mismatch between versions requested for Go module {module}:
             root_module_direct_dev_deps.pop(_repo_name(path), None)
             continue
         if module.repo_name in repos_processed:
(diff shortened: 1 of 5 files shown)


fmeum (approved):

fmeum:
Thanks for sending the PR. We will have to fix unrelated issues with the latest Bazel version before we can merge it.

Issue #2228, opened 2025-11-13
[feature] remove keep comments if they are not required


peter-cardenas-ai:
after upgrading gazelle to support using capture groups when creating labels with the `resolve_regexp` directive, I found that I could remove a lot of unnecessary keep comments. However, I had to do this by a manual search and replace, when I hoped gazelle could do it for me. My ask would be to give the ability/configuration to remove keep comments if they are not needed to keep the target in deps/srcs.

jayconrod:
Could you post an example project where this would help, just with a couple build files?

I'm pretty uncomfortable though with the idea of Gazelle removing `# keep` comments though. They strongly signal user intent, and it's not really clear to me when it's safe to remove them.

jeffmace:
We occasionally go through a cleanup cycle for `#keep` comments. This involves removing the comments, running Gazelle diff and then adding comments back to where they are needed. It would be nice to support this type of behavior as a runtime flag or directive so it can be enabled when needed but doesn't change the default behavior.

```
gazelle --mode=diff --remove_noop_keep_comments

gazelle --mode=fix --remove_noop_keep_comments
```

fmeum:
I would have used this feature during numerous updates. I could even see it being enabled by default.

peter-cardenas-ai:
maybe if i could explain it in words: we have a few macros that create go library targets with well formatted names and importpaths. before yall added the support for using capture groups with resolve_regexp, we just resolved these imports to a bogus target but still required a keep comment. now with upgrading to use these capture groups, we can remove the keep comments because we can actually resolve to the dependency target name using the import, but going through to update the keep comments is a bit of a slog.


more generally though, if gazelle continues to add more features, like the resolution of generated files like from gomock, it would be nice to be able to keep up with and take advantage of these features without manual intervention

peter-cardenas-ai:
also on the point of signaling user intent, i find it hard to think of an example where an import can be resolved to a target name but we still want to keep the keep comment around to denote some reason

jayconrod:
Let's add this, since several people here are saying it would be useful.

For scoping: I think this should be constrained to `# keep` comments on specific values within rules generated/updated by Gazelle. We should not remove `# keep` comments on larger expressions like lists, attributes, or rules.

I like @jeffmace's suggestion for guarding this behavior with a flag. But it should also be fine to enable it by default within the `fix` command, not the `update` command. `fix` is already allowed to perform potentially destructive edits.

[...]

PR #2232, opened 2025-11-17
feat: remove unnecessary keep comments in lists

PeterCardenas:
**What type of PR is this?**

feature

cmd/gazelle

Removes keep comments in lists when the list item was going to be kept anyways. Allows for automatic cleanup of dependencies that do not have to be manually tracked.

Partially addresses #2228


jayconrod:
Commented in #2228. Let's discuss there before getting into PR details.

PeterCardenas:
@jayconrod i'm not super happy with the change but maybe it illustrates the files that need to be updated. also removed the ai generated tests for clarity.
some potential steps forward:
- create client interfaces for merging rules/files and convert functions that need the options to methods on the client so that they can access the remove_noop_keep_comments option
- pass the config struct down to mergerules so that all options can be seen
- just pass a bool

merger/merger.go
```diff
@@ -106,7 +110,7 @@ const UnstableInsertIndexKey = "_gazelle_insert_index"
 // If an attribute is marked with a "# keep" comment, it will not be merged.
 // If a rule is marked with a "# keep" comment, the whole rule will not
 // be modified.
-func MergeFile(oldFile *rule.File, emptyRules, genRules []*rule.Rule, phase Phase, kinds map[string]rule.KindInfo, aliasedKinds map[string]string) {
+func MergeFile(oldFile *rule.File, emptyRules, genRules []*rule.Rule, phase Phase, kinds map[string]rule.KindInfo, aliasedKinds map[string]string, opts *MergeFileOptions) {
```
jayconrod: Adding a parameter to a publicly exported function is an incompatible change, and unfortunately there are packages outside Gazelle that import and call this function.

I don't know that there's really a good option here. The least bad option might be to add a global `RemoveNoopKeepComments bool` flag to the `rule` package, then set that directly.

I'll make a note of this in https://github.com/bazel-contrib/bazel-gazelle/discussions/2207, since we can make incompatible changes for v2. I think it's a good idea to have an open-ended options struct like this.

rule/rule.go
```diff
@@ -1118,18 +1118,43 @@ func (r *Rule) sync() {
...
+		isKeep := isKeepComment(c)
+		if isKeep {
 			return true
 		}
 	}
 	return false
 }
 
+// RemoveKeep removes the keep comments from the expression.
+func RemoveKeep(e bzl.Expr) bzl.Comments {
```
jayconrod: Let's leave this unexported. I'd prefer not to expand the public interface of this package more than necessary.

PeterCardenas:
@jayconrod fixed according to your suggestions, would appreciate your preference on how to test this (integration/unit tests/none)

cmd/gazelle/fix-update.go
```diff
@@ -96,6 +97,7 @@ func (ucr *updateConfigurer) RegisterFlags(fs *flag.FlagSet, cmd string, c *conf
 	fs.StringVar(&ucr.memProfile, "memprofile", "", "write memory profile to `file`")
 	fs.Var(&gzflag.MultiFlag{Values: &ucr.knownImports}, "known_import", "import path for which external resolution is skipped (can specify multiple times)")
 	fs.StringVar(&ucr.repoConfigPath, "repo_config", "", "file where Gazelle should load repository configuration. Defaults to WORKSPACE.")
+	fs.BoolVar(&uc.removeNoopKeepComments, "remove_noop_keep_comments", false, "when set, gazelle will remove noop keep comments from BUILD files")
```
jayconrod, line 100: Could you document this in README.md as well?

jayconrod:
The current CI failure is unrelated. Should be fixed after merging `master`.

PeterCardenas:
@jayconrod updated readme + added integration tests. seems like i revealed that i'm somehow not removing the keep comments for visibility, but maybe can tackle that later and just have this merged?

PR #2237, opened 2025-11-18
CI: upgrade rules_go for BCR tests

### LINKED ISSUE #2234: CI error: 'struct' value has no field or method 'configure_features'
It looks like something was broken in Bazel 9.x upstream.

The error is emitted from rules_go, so consider this a tracking issue only. Either rules_go or Bazel will need to be fixed. This issue can be closed when CI is green again.

jayconrod:
**What type of PR is this?**

> Bug fix

tests/bcr

Bazel 9.x requires the cc_common module to be loaded explicitly. bazel-contrib/rules_go#4508 added the required load statement, and this fix was released in rules_go 0.59.0.

This change upgrades the tests/bcr/go_mod and go_work test modules to require rules_go 0.59.0 so they can be tested with Bazel 9.x. Gazelle itself doesn't require the fix, so this change doesn't touch MODULE.bazel, avoiding a forced upgrade.

Fixes #2234

**Other notes for review**

### DIFF
--- a/tests/bcr/go_mod/MODULE.bazel
+++ b/tests/bcr/go_mod/MODULE.bazel
@@ -27,7 +27,7 @@ go_sdk.download(version = "1.24.2")
 
 # This bazel_dep provides the Go dependency github.com/cloudflare/circl, which requires custom
 # patches beyond what Gazelle can generate.
-bazel_dep(name = "circl", version = "1.3.7")
+bazel_dep(name = "circl", version = "1.3.8")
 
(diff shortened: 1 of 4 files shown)


fmeum (approved):

fmeum:
Looks like we need to get a new version of circl into the BCR that has loads for rules_cc.

jayconrod:
Hmm, upgrading circl fixed the problem: the next point release has correct loads.

But the latest BCR release is pretty far behind the latest upstream release. That's a separate problem I suppose.

PR #2247, opened 2025-12-03
Add default_features directive

keith:
This sets the `package(features = ...)` attribute in all nested BUILD
files similar to `default_visibility`

Fixes https://github.com/bazel-contrib/bazel-gazelle/issues/2236


language/bazel/visibility/lang.go
```diff
@@ -64,7 +66,19 @@ func (*visibilityExtension) GenerateRules(args language.GenerateArgs) language.G
 	}
 
 	r := rule.NewRule("package", "")
-	r.SetAttr("default_visibility", cfg.visibilityTargets)
+	for _, er := range args.File.Rules {
+		if er.Kind() == "package" {
+			r = er
```
jayconrod: I'm curious why this was needed. Extensions should generate new rules and return them through `res.Gen`, relying on Gazelle to merge generated and existing rules. If an extension modifies an existing rule directly, it can break `# keep` directives or cause other unexpected behavior.

language/bazel/visibility/lang_test.go
```diff
@@ -286,3 +287,196 @@ func Test_MultipleDirectivesAcrossFilesSupercede(t *testing.T) {
 		t.Fatal("expected returned visibility to match '//src:__subpackages__'")
 	}
 }
+
+func Test_PackageFeatures_NewDirective(t *testing.T) {
```
jayconrod: I think it would be better to cover the new functionality with integration tests in `//tests`. Then we'll be able to see the change end to end.

language/bazel/visibility/lang.go
keith: ok thanks, let me try without this, might have also been because of https://github.com/bazel-contrib/bazel-gazelle/pull/2248

language/bazel/visibility/lang_test.go
keith: thanks, done

language/bazel/visibility/lang.go
```diff
@@ -65,9 +67,22 @@ func (*visibilityExtension) GenerateRules(args language.GenerateArgs) language.G
...
+	if len(cfg.visibilityTargets) > 0 {
+		r.SetAttr("default_visibility", cfg.visibilityTargets)
+	}
+	if len(cfg.features) > 0 {
+		r.SetAttr("features", cfg.features)
+	}
 
-	insertIndex := 0
+	// Start after the first statements if no rules exist
+	insertIndex := len(args.File.File.Stmt)
```
keith, line 90: partially unrelated change here as an oversight in https://github.com/bazel-contrib/bazel-gazelle/pull/2249 that i discovered in these new tests, if the file has no rules `0` isn't really nice because it put them above the gazelle comments.

i can split this into a later pr if you want

language/bazel/visibility/lang.go
keith: it does seem like this is still required, could it be something else i missed with how these should be merged?

language/bazel/visibility/lang.go
jayconrod: This is fine as it is.

language/bazel/visibility/lang.go
jayconrod: Some context: in `gazelle_cc`, we're fixing a bug caused by a broken `rule.Merger` implementation that unintentionally deleted `# keep` comments. Something like this can happen when overwriting the original parsed rules rather than generating new rules and letting the merger handle them.

Please add this test case:

```
-- tests/package_rule_keep_preserves_features/MODULE.bazel --
-- tests/package_rule_keep_preserves_features/BUILD.in --
# gazelle:default_features a

package(
    features = [
        "b",  # keep
        "c",
    ],
)
-- tests/package_rule_keep_preserves_features/BUILD.out --
# gazelle:default_features a

package(
    features = [
        "b",  # keep
        "a",
    ],
)
```

I'd guess that the reason to alias the parsed rule was to avoid overwriting or deleting a `package()` call when `# gazelle:default_visibility` or `# gazelle:default_features` was not set. If that's correct, then in that case, it's probably best not to generate `package()` at all: Gazelle will leave it alone. But if one directive is set, we should overwrite at least that argument.

I guess the question is if `default_features` is set, should we overwrite `default_visibility` (and vice versa)? That would be pretty surprising behavior, even if it is technically consistent. Perhaps instead, we can deep-copy the original rule instead of aliasing it.

keith: > Please add this test case:

thanks for this example, it was indeed broken and is now fixed i believe

> If that's correct, then in that case, it's probably best not to generate package() at all: Gazelle will leave it alone. 

yes if neither is set, we continue to do nothing:

```go
	if len(cfg.visibilityTargets) == 0 && len(cfg.features) == 0 {
		return res
	}
```

> But if one directive is set, we should overwrite at least that argument.

yep exactly what i was trying to achieve, i believe i have this correct now

> we can deep-copy the original rule instead of aliasing it.

doing this now by setting the attrs on the initial new rule, which seems to also be enough to work with the merger

jeffmace:
I've previously considered adding something like this but for `default_package_metadata`. Do you consider it better for each attribute to have a dedicated directive or have a generic directive for any attribute to the `package` function?

jayconrod:
> I've previously considered adding something like this but for default_package_metadata. Do you consider it better for each attribute to have a dedicated directive or have a generic directive for any attribute to the package function?

@jeffmace I think a separate directive for each (so `# gazelle:default_package_metadata`) is best. There aren't too many attributes, so supporting a few more shouldn't hurt.

jayconrod:
I'm seeing macOS failures in CI across multiple PRs: earlier OAuth2 failures, now errors checking out the repo. We'll need to wait for that to get straightened out upstream.

keith:
the repo checkout one is a CI flake, retry normally fixes

jayconrod:
So it did!

keith:
thanks!

PR #2249, opened 2025-12-03
Insert package() at the top of the file

keith:
This is a different approach to https://github.com/bazel-contrib/bazel-gazelle/pull/1458

### DIFF
--- a/language/bazel/visibility/BUILD.bazel
+++ b/language/bazel/visibility/BUILD.bazel
@@ -13,6 +13,7 @@ go_library(
         "//config",
         "//label",
         "//language",
+        "//merger",
         "//repo",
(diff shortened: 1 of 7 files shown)


keith:
ah i wrote the tests assuming https://github.com/bazel-contrib/bazel-gazelle/pull/2248 was fixed as well, i can change that if that's a problem, or i can rebase after that one is fixed

fmeum (approved):
Looks good to me but I would like @jayconrod to take a look as well.

jayconrod (approved):
Thanks for fixing this. I guess we should rename `merger.UnstableInsertIndexKey` in v2.

keith:
thanks!

Issue #2196, opened 2025-10-09
Clean up README, improve documentation


jayconrod:
- [x] Translate existing documentation from reStructuredText to Markdown for more familiar editing. (Unfortunately, this makes tables much harder, so we'll reformat flags, attributes, and directives.
  - [x] README.rst
  - [x] Design.rst
  - [ ] autogazelle.rst
- [ ] Split README into multiple documents, loosely following https://diataxis.fr/.
  - [ ] README should have a brief "quick start" guide to getting started with Gazelle with examples for Bzlmod and WORKSPACE. It should link to other documents and other Gazelle extensions.
  - [ ] Reference documents: separate pages for CLI flags, the `gazelle` rule, and directives.
  - [ ] Go- and protobuf-specific reference documents should be separated too.
- [ ] *Extending Gazelle* should be reframed as a tutorial.
- [ ] *Architecture of Gazelle* should be rewritten and reframed as an explanation: it predates the extension mechanism.

jayconrod:
Didn't mean to close this, not done.

PR #2251, opened 2025-12-05
docs: rewrite design explanation

jayconrod:
**What type of PR is this?**

> Documentation

> all

Replace Design.md with how-gazelle-works.md, a complete rewrite.

The old page was written before extensions, so was very obsolete.

The new page is intended to be an Explanation (https://diataxis.fr/) to help Gazelle developers and extension authors.

For #2196

**Other notes for review**

### DIFF
--- a/how-gazelle-works.md
+++ b/how-gazelle-works.md
@@ -2,6 +2,8 @@
 
 This page explains how Gazelle generates and updates `BUILD` files. It's intended to help Gazelle developers, extension authors, and users wishing to understand how (and why) Gazelle makes its changes.
 
+See [Configuration and command line reference](gazelle-reference.md) for details on specific directives and flags.
+
 ## Overview
 
 Gazelle updates `BUILD` files in the following stages. Each is described in detail below.
@@ -21,7 +23,7 @@ In each directory it visits, Gazelle parses the `BUILD` or `BUILD.bazel` file if
 
 Gazelle may or may not visit a directory based on directives and command line flags.
 
-- Gazelle always visits directories named as command line arguments. If no arguments are specified, Gazelle visits the repository root directory.
+- Gazelle always visits directories named with positional arguments on the command line. If no arguments are specified, Gazelle visits the repository root directory.
 - If recursion is enabled (with `-r=true`, enabled by default), Gazelle recursively visits subdirectories.
 - If eager indexing is enabled (with `-index=all`, enabled by default), Gazelle visits *all* directories.
 - If lazy indexing is enabled (with `-index=lazy`), Gazelle visits directories requested by language extensions in [`GenerateResult.RelsToIndex`](https://pkg.go.dev/github.com/bazelbuild/bazel-gazelle/language#GenerateResult.RelsToIndex). These directories are loaded lazily during the *Generate* stage.
@@ -32,21 +34,21 @@ The Load stage is implemented in [`walk.walker.populateCache`](https://github.co
 
 ### Generate
 
-As Gazelle visits each directory, it calls extension methods to apply configuration, fix deprecated usage, generate rules, and index libraries. Most of Gazelle's work happens during this stage.
+As Gazelle visits each directory, it calls extension methods to apply configuration, fix deprecated usage, generate rules, combine generated rules with existing rules, and index libraries. Most of Gazelle's work happens during this stage.
+
+Not all extension methods are called in each directory Gazelle visits. Methods that generate or modify rules (`Fix`, `GenerateRules`, and `MergeFile`) are only called in directories where Gazelle was asked to update the `BUILD` file (directories named with positional command line arguments and their subdirectories if the `-r` flag is enabled).
 
 1. [`Configure`](https://pkg.go.dev/github.com/bazelbuild/bazel-gazelle/config#Configurer.Configure) is called in each directory Gazelle visits. Each extension can read directives from the `BUILD` file (if there is one) to decide what to do. Most directives apply to the directory they appear in and to subdirectories.
     - `Configure` is called in each directory in pre-order. All other methods in this stage are called in post-order. Sibling directories are visited sequentially (not concurrently) in lexicographic order.
     - `Configure` is called in each directory Gazelle visits, even if Gazelle won't update the `BUILD` file.
-1. [`Fix`](https://pkg.go.dev/github.com/bazelbuild/bazel-gazelle/language#Language.Fix) is called in each directory that has an existing `BUILD` file. The purpose of this method is to fix deprecated rule usage, so extensions can make arbitrary transformations here.
+1. [`Fix`](https://pkg.go.dev/github.com/bazelbuild/bazel-gazelle/language#Language.Fix) is called in each directory that has an existing `BUILD` file. The purpose of this method is to fix deprecated rule usage, so extensions can make any necessary transformations here.
 1. [`GenerateRules`](https://pkg.go.dev/github.com/bazelbuild/bazel-gazelle/language#Language.GenerateRules) is called in each directory. This method returns rules that should be present in the `BUILD` file, and rules that should be removed. Unlike `Fix`, this method must not actually modify the rules parsed from the `BUILD` file.
 1. [`merger.MergeFile`](https://pkg.go.dev/github.com/bazelbuild/bazel-gazelle/merger#MergeFile) is called to combine the generated rules with existing rules.
     - `MergeFile` attempts to match each generated rule with an existing rule. If a rule is not matched, it's added to the end of the `BUILD` file. Usually the matching is based on the rule kind (a `go_library` named `client`), but it can be influenced by other heuristics. 
     - Each attribute is merged separately. An attribute can be *mergeable* or not. If an attribute is mergeable, it's expected to be managed by Gazelle, so the merger can overwrite existing values (except for values marked with `# keep`). If an attribute is not mergeable, Gazelle may set an initial value, but won't overwrite it later.
-    - `MergeFile` also merges empty rules returned by `GenerateRules` with existing rules. This can delete existing rules if they're not marked with `# keep`. For example, this allows Gazelle to delete a `go_test` rule after all the `_test.go` files were removed.
+    - `MergeFile` also merges rules from the empty list returned by `GenerateRules`'. This can delete existing rules if they're not marked with `# keep`. For example, this allows Gazelle to delete a `go_test` rule after all the `_test.go` files were removed.
     - Extensions don't directly participate in the merging process, though they can influence matching and merging by returning [`rule.KindInfo`](https://pkg.go.dev/github.com/bazelbuild/bazel-gazelle/rule#KindInfo) for each generated rule kind from the [`Kinds`](https://pkg.go.dev/github.com/bazelbuild/bazel-gazelle/language#Language.Kinds) method.
 1. [`Imports`](https://pkg.go.dev/github.com/bazelbuild/bazel-gazelle@v0.47.0/resolve#Resolver.Imports) is called on each rule after the merge to build an in-memory index for dependency resolution. This table maps import strings and language names to Bazel labels. `Imports` is not called if indexing is disabled with `-index=none`.
-
-The extension methods `Fix`, `GenerateRules`, and `MergeFile` are only called in directories where Gazelle was asked to update the `BUILD` file. `Configure` and `Imports` are called in all directories Gazelle visits.
 
 The Generate stage is implemented in the [callback function](https://github.com/bazel-contrib/bazel-gazelle/blob/028c500e9f911a73683b6ec390f3e59e8f31fccc/cmd/gazelle/fix-update.go#L322) passed to [`walk.Walk2`](https://pkg.go.dev/github.com/bazelbuild/bazel-gazelle/walk#Walk2) as part of the [`runFixUpdate`](https://github.com/bazel-contrib/bazel-gazelle/blob/028c500e9f911a73683b6ec390f3e59e8f31fccc/cmd/gazelle/fix-update.go#L263) function.
 
@@ -61,9 +63,7 @@ Gazelle resolves dependencies as a separate stage so it can use the index built 
 
 ### Write
 
-At this point, Gazelle has made all necessary changes to `BUILD` files in memory. It then formats these files with [`build.Format`](https://pkg.go.dev/github.com/bazelbuild/buildtools/build#Format) and writes them back to disk.
-
-The `-mode` flag modifies this slightly. `-mode=print` causes Gazelle to print the modified files. `-mode=diff` causes Gazelle to print the differences as a patch.
+At this point, Gazelle has made all necessary changes to `BUILD` files in memory. It then formats these files with [`build.Format`](https://pkg.go.dev/github.com/bazelbuild/buildtools/build#Format) and writes them back to disk or prints them, depending on the `-mode` flag (see [Flags](gazelle-reference.md#flags).
 
 ## Extensions
 
--- a/gazelle-reference.md
+++ b/gazelle-reference.md
@@ -52,7 +52,11 @@ If `all` or `true`, Gazelle indexes all directories in the repository, even when
 
 **Flag:** `-mode=fix|print|diff`<br>
 **Default:** `fix`<br>
-Method for emitting merged build files. In `fix` mode, Gazelle writes generated and merged files to disk. In `print` mode, it prints them to stdout. In `diff` mode, it prints a unified diff.
+Method for emitting merged build files.
+
+- In `fix` mode, Gazelle writes generated and merged files to disk.
+- In `print` mode, Gazelle prints updated files to stdout and does not write files to disk.
+- In `diff` mode, Gazelle prints a unified diff to stdout and does not write files to disk.
 
 **Flag:** `-r`<br>
 **Default:** `true`<br>


how-gazelle-works.md
```diff
@@ -0,0 +1,87 @@
...
+    - `Configure` is called in each directory in pre-order. All other methods in this stage are called in post-order. Sibling directories are visited sequentially (not concurrently) in lexicographic order.
+    - `Configure` is called in each directory Gazelle visits, even if Gazelle won't update the `BUILD` file.
+1. [`Fix`](https://pkg.go.dev/github.com/bazelbuild/bazel-gazelle/language#Language.Fix) is called in each directory that has an existing `BUILD` file. The purpose of this method is to fix deprecated rule usage, so extensions can make arbitrary transformations here.
+1. [`GenerateRules`](https://pkg.go.dev/github.com/bazelbuild/bazel-gazelle/language#Language.GenerateRules) is called in each directory. This method returns rules that should be present in the `BUILD` file, and rules that should be removed. Unlike `Fix`, this method must not actually modify the rules parse from the `BUILD` file.
```
fmeum:
```suggestion
1. [`GenerateRules`](https://pkg.go.dev/github.com/bazelbuild/bazel-gazelle/language#Language.GenerateRules) is called in each directory. This method returns rules that should be present in the `BUILD` file, and rules that should be removed. Unlike `Fix`, this method must not actually modify the rules parsed from the `BUILD` file.
```

jayconrod:
Hi @nikki-ef, could you review this if you have time? If not, no worries.

This is part of a larger Gazelle docs rewrite (#2196), but I found myself again referring a potential contributor to the old obsolete documentation, so I figured I'd better do this part soon.

fmeum (approved):

how-gazelle-works.md
```diff
@@ -0,0 +1,87 @@
...
+
+### Load
+
+When Gazelle starts, it begins traversing the directory tree. This process runs in parallel with later stages as an optimization.
+
+In each directory it visits, Gazelle parses the `BUILD` or `BUILD.bazel` file if present and makes a list of files and subdirectories, excluding those matched by `# gazelle:exclude` directives or `.bazelignore` files. This metadata is cached in memory so that later stages may access it quickly without requiring additional I/O.
+
+Gazelle may or may not visit a directory based on directives and command line flags.
+
+- Gazelle always visits directories named as command line arguments. If no arguments are specified, Gazelle visits the repository root directory.
```
nikki-ef:
```suggestion
- Gazelle always visits directories named as [command line arguments](/gazelle-reference.md). If no arguments are specified, Gazelle visits the repository root directory.
```
It'd be nice to make this resource easily findable, especially for folks who may be new to Gazelle and its repo structure.

how-gazelle-works.md
```diff
@@ -0,0 +1,87 @@
...
+- If indexing is disabled (with `-index=none`), Gazelle does not visit additional directories.
+- Gazelle visits parent directories within the repository in addition to other directories it visits. This is necessary to apply `# gazelle:exclude` directives, which may tell Gazelle to act as if a subdirectory does not exist.
+
+The Load stage is implemented in [`walk.walker.populateCache`](https://github.com/bazel-contrib/bazel-gazelle/blob/028c500e9f911a73683b6ec390f3e59e8f31fccc/walk/dirinfo.go#L104), which is called from [`walk.Walk2`](https://pkg.go.dev/github.com/bazelbuild/bazel-gazelle/walk#Walk2). No extension methods are called during this stage.
+
+### Generate
+
+As Gazelle visits each directory, it calls extension methods to apply configuration, fix deprecated usage, generate rules, and index libraries. Most of Gazelle's work happens during this stage.
+
+1. [`Configure`](https://pkg.go.dev/github.com/bazelbuild/bazel-gazelle/config#Configurer.Configure) is called in each directory Gazelle visits. Each extension can read directives from the `BUILD` file (if there is one) to decide what to do. Most directives apply to the directory the appear in and to subdirectories.
```
nikki-ef:
```suggestion
1. [`Configure`](https://pkg.go.dev/github.com/bazelbuild/bazel-gazelle/config#Configurer.Configure) is called in each directory Gazelle visits. Each extension can read directives from the `BUILD` file (if there is one) to decide what to do. Most directives apply to the directory they appear in and to subdirectories.
```

how-gazelle-works.md
```diff
@@ -0,0 +1,87 @@
...
+    - `Configure` is called in each directory in pre-order. All other methods in this stage are called in post-order. Sibling directories are visited sequentially (not concurrently) in lexicographic order.
+    - `Configure` is called in each directory Gazelle visits, even if Gazelle won't update the `BUILD` file.
+1. [`Fix`](https://pkg.go.dev/github.com/bazelbuild/bazel-gazelle/language#Language.Fix) is called in each directory that has an existing `BUILD` file. The purpose of this method is to fix deprecated rule usage, so extensions can make arbitrary transformations here.
```
nikki-ef:
```suggestion
1. [`Fix`](https://pkg.go.dev/github.com/bazelbuild/bazel-gazelle/language#Language.Fix) is called in each directory that has an existing `BUILD` file. The purpose of this method is to fix deprecated rule usage, so extensions can make necessary transformations here.
```

how-gazelle-works.md
nikki-ef: Would it be accurate to say this method "does not actually modify" instead of "must not actually modify" the rules parsed from the `BUILD` file?

how-gazelle-works.md
```diff
@@ -0,0 +1,87 @@
...
+    - Each attribute is merged separately. An attribute can be *mergeable* or not. If an attribute is mergeable, it's expected to be managed by Gazelle, so the merger can overwrite values that are already there (except for values marked with `# keep`). If an attribute is not mergeable, Gazelle may set an initial value, but won't overwrite it later.
+    - `MergeFile` also merges empty rules returned by `GenerateRules` with existing rules. This can delete existing rules if they're not marked with `# keep`. For example, this allows Gazelle to delete a `go_test` rule after all the `_test.go` files were removed.
+    - Extensions don't directly participate in the merging process, though they can influence matching and merging by returning [`rule.KindInfo`](https://pkg.go.dev/github.com/bazelbuild/bazel-gazelle/rule#KindInfo) for each generate rule kind from the [`Kinds`](https://pkg.go.dev/github.com/bazelbuild/bazel-gazelle/language#Language.Kinds) method.
```
nikki-ef:
```suggestion
    - Extensions don't directly participate in the merging process, though they can influence matching and merging by returning [`rule.KindInfo`](https://pkg.go.dev/github.com/bazelbuild/bazel-gazelle/rule#KindInfo) for each generated rule kind from the [`Kinds`](https://pkg.go.dev/github.com/bazelbuild/bazel-gazelle/language#Language.Kinds) method.
```

how-gazelle-works.md
```diff
@@ -0,0 +1,87 @@
...
+    - `Configure` is called in each directory Gazelle visits, even if Gazelle won't update the `BUILD` file.
+1. [`Fix`](https://pkg.go.dev/github.com/bazelbuild/bazel-gazelle/language#Language.Fix) is called in each directory that has an existing `BUILD` file. The purpose of this method is to fix deprecated rule usage, so extensions can make arbitrary transformations here.
+1. [`GenerateRules`](https://pkg.go.dev/github.com/bazelbuild/bazel-gazelle/language#Language.GenerateRules) is called in each directory. This method returns rules that should be present in the `BUILD` file, and rules that should be removed. Unlike `Fix`, this method must not actually modify the rules parse from the `BUILD` file.
+1. [`merger.MergeFile`](https://pkg.go.dev/github.com/bazelbuild/bazel-gazelle/merger#MergeFile) is called to combine the generated rules with existing rules.
+    - `MergeFile` attempts to match each rule with an existing rule. If a rule is not matched, it's added to the end of the file. Usually the matching is based on the rule kind (a `go_library` named `client`), but it can be influenced by other heuristics.
```
nikki-ef: @copilot 
```suggestion
    - `MergeFile` attempts to match each generated rule with an existing rule. If a rule is not matched, it's added to the end of the `BUILD` file. Usually the matching is based on the rule kind (a `go_library` named `client`), but it can be influenced by other heuristics. 
```

how-gazelle-works.md
```diff
@@ -0,0 +1,87 @@
...
+1. [`GenerateRules`](https://pkg.go.dev/github.com/bazelbuild/bazel-gazelle/language#Language.GenerateRules) is called in each directory. This method returns rules that should be present in the `BUILD` file, and rules that should be removed. Unlike `Fix`, this method must not actually modify the rules parse from the `BUILD` file.
+1. [`merger.MergeFile`](https://pkg.go.dev/github.com/bazelbuild/bazel-gazelle/merger#MergeFile) is called to combine the generated rules with existing rules.
+    - `MergeFile` attempts to match each rule with an existing rule. If a rule is not matched, it's added to the end of the file. Usually the matching is based on the rule kind (a `go_library` named `client`), but it can be influenced by other heuristics. 
+    - Each attribute is merged separately. An attribute can be *mergeable* or not. If an attribute is mergeable, it's expected to be managed by Gazelle, so the merger can overwrite values that are already there (except for values marked with `# keep`). If an attribute is not mergeable, Gazelle may set an initial value, but won't overwrite it later.
```
nikki-ef:
```suggestion
    - Each attribute is merged separately. An attribute can be *mergeable* or not. If an attribute is mergeable, it's expected to be managed by Gazelle, so the merger can overwrite existing values (except for values marked with `# keep`). If an attribute is not mergeable, Gazelle may set an initial value, but won't overwrite it later.
```

how-gazelle-works.md
```diff
@@ -0,0 +1,87 @@
...
+- If lazy indexing is enabled (with `-index=lazy`), Gazelle visits directories requested by language extensions in [`GenerateResult.RelsToIndex`](https://pkg.go.dev/github.com/bazelbuild/bazel-gazelle/language#GenerateResult.RelsToIndex). These directories are loaded lazily during the *Generate* stage.
+- If indexing is disabled (with `-index=none`), Gazelle does not visit additional directories.
+- Gazelle visits parent directories within the repository in addition to other directories it visits. This is necessary to apply `# gazelle:exclude` directives, which may tell Gazelle to act as if a subdirectory does not exist.
+
+The Load stage is implemented in [`walk.walker.populateCache`](https://github.com/bazel-contrib/bazel-gazelle/blob/028c500e9f911a73683b6ec390f3e59e8f31fccc/walk/dirinfo.go#L104), which is called from [`walk.Walk2`](https://pkg.go.dev/github.com/bazelbuild/bazel-gazelle/walk#Walk2). No extension methods are called during this stage.
+
+### Generate
+
+As Gazelle visits each directory, it calls extension methods to apply configuration, fix deprecated usage, generate rules, and index libraries. Most of Gazelle's work happens during this stage.
```
nikki-ef:
```suggestion
As Gazelle visits each directory, it sequentially calls extension methods to apply configuration, fix deprecated usage, generate rules, combine generated rules with existing rules, and index libraries. Most of Gazelle's work happens during this stage.
```

how-gazelle-works.md
```diff
@@ -0,0 +1,87 @@
...
+1. [`merger.MergeFile`](https://pkg.go.dev/github.com/bazelbuild/bazel-gazelle/merger#MergeFile) is called to combine the generated rules with existing rules.
+    - `MergeFile` attempts to match each rule with an existing rule. If a rule is not matched, it's added to the end of the file. Usually the matching is based on the rule kind (a `go_library` named `client`), but it can be influenced by other heuristics. 
+    - Each attribute is merged separately. An attribute can be *mergeable* or not. If an attribute is mergeable, it's expected to be managed by Gazelle, so the merger can overwrite values that are already there (except for values marked with `# keep`). If an attribute is not mergeable, Gazelle may set an initial value, but won't overwrite it later.
+    - `MergeFile` also merges empty rules returned by `GenerateRules` with existing rules. This can delete existing rules if they're not marked with `# keep`. For example, this allows Gazelle to delete a `go_test` rule after all the `_test.go` files were removed.
```
nikki-ef:
```suggestion
    - `MergeFile` also removes rules that `GenerateRules` flagged for removal. This can delete existing rules if they're not marked with `# keep`. For example, this allows Gazelle to delete a `go_test` rule after all the `_test.go` files were removed. 
```
Suggestion for clarity, if it retains technical accuracy.

how-gazelle-works.md
```diff
@@ -0,0 +1,87 @@
...
+    - Extensions don't directly participate in the merging process, though they can influence matching and merging by returning [`rule.KindInfo`](https://pkg.go.dev/github.com/bazelbuild/bazel-gazelle/rule#KindInfo) for each generate rule kind from the [`Kinds`](https://pkg.go.dev/github.com/bazelbuild/bazel-gazelle/language#Language.Kinds) method.
+1. [`Imports`](https://pkg.go.dev/github.com/bazelbuild/bazel-gazelle@v0.47.0/resolve#Resolver.Imports) is called on each rule after the merge to build an in-memory index for dependency resolution. This table maps import strings and language names to Bazel labels. `Imports` is not called if indexing is disabled with `-index=none`.
+
+The extension methods `Fix`, `GenerateRules`, and `MergeFile` are only called in directories where Gazelle was asked to update the `BUILD` file. `Configure` and `Imports` are called in all directories Gazelle visits.
```
nikki-ef:
```suggestion
Note that not all extension methods may be called on each Gazelle run:

- The extension methods `Fix`, `GenerateRules`, and `MergeFile` are only called in directories where Gazelle was asked to update the `BUILD` file. 
- `Configure` and `Imports` are called in all directories Gazelle visits.
```

Suggestion for phrasing + document flow: Regardless of this phrasing update, I think we'd benefit from moving this paragraph higher up in this section, right before the numbered list begins. On my first read, I thought all of these methods were always called until I saw these conditions.

how-gazelle-works.md
```diff
@@ -0,0 +1,87 @@
...
+
+The Generate stage is implemented in the [callback function](https://github.com/bazel-contrib/bazel-gazelle/blob/028c500e9f911a73683b6ec390f3e59e8f31fccc/cmd/gazelle/fix-update.go#L322) passed to [`walk.Walk2`](https://pkg.go.dev/github.com/bazelbuild/bazel-gazelle/walk#Walk2) as part of the [`runFixUpdate`](https://github.com/bazel-contrib/bazel-gazelle/blob/028c500e9f911a73683b6ec390f3e59e8f31fccc/cmd/gazelle/fix-update.go#L263) function.
+
+### Resolve
+
+Gazelle resolves dependencies as a separate stage so it can use the index built by calling `Imports` during the previous stage.
```
nikki-ef:
```suggestion
Gazelle resolves dependencies as a separate stage so it can use the index built by calling `Imports` during the Generate stage.
```

how-gazelle-works.md
```diff
@@ -0,0 +1,87 @@
+# How Gazelle Works
+
+This page explains how Gazelle generates and updates `BUILD` files. It's intended to help Gazelle developers, extension authors, and users wishing to understand how (and why) Gazelle makes its changes.
+
+## Overview
+
+Gazelle updates build files in the following stages. Each is described in detail below.
```
nikki-ef:
```suggestion
Gazelle updates `BUILD` files in the following stages. Each is described in detail below.
```

how-gazelle-works.md
```diff
@@ -0,0 +1,87 @@
...
+
+1. [`Resolve`](https://pkg.go.dev/github.com/bazelbuild/bazel-gazelle/resolve#Resolver.Resolve) is called on each generated rule.
+    - The extension that generated the rule should resolve dependencies, often using the index, then set `deps` and any related attributes.
+    - To avoid redundant I/O, an extension may return information about import strings found in source files through [`GenerateResult.Imports`](https://pkg.go.dev/github.com/bazelbuild/bazel-gazelle/language#GenerateResult.Imports) when returning from [`GenerateRules`](https://pkg.go.dev/github.com/bazelbuild/bazel-gazelle/language#Language.GenerateRules). This value is opaque to Gazelle. It's passed back to `Resolve`.
+1. [`merger.MergeFile`](https://pkg.go.dev/github.com/bazelbuild/bazel-gazelle/merger#MergeFile) is called again to merge changes with existing rules in `BUILD` files.
+
+### Write
+
+At this point, Gazelle has made all necessary changes to `BUILD` files in memory. It then formats files with [`build.Format`](https://pkg.go.dev/github.com/bazelbuild/buildtools/build#Format) and writes files back to disk.
```
nikki-ef:
```suggestion
At this point, Gazelle has made all necessary changes to `BUILD` files in memory. It then formats these files with [`build.Format`](https://pkg.go.dev/github.com/bazelbuild/buildtools/build#Format) and writes them back to disk.
```

how-gazelle-works.md
```diff
@@ -0,0 +1,87 @@
...
+
+The `-mode` flag modifies this slightly. `-mode=print` causes Gazelle to print the modified files. `-mode=diff` causes Gazelle to print the differences as a patch.
```
nikki-ef: If `-mode=print` or `-mode=diff`, does the printing as modified files or patch happen _in addition to_ saving to disk or instead of? I couldn't tell from the [reference doc](https://github.com/bazel-contrib/bazel-gazelle/blob/master/gazelle-reference.md) either.

how-gazelle-works.md
```diff
@@ -0,0 +1,87 @@
...
+
+Gazelle provides a language-agnostic framework for generating rules and updating `BUILD` files. All the language-specific functionality is implemented in *extensions*. An extension is an implementation of the [`language.Language`](https://pkg.go.dev/github.com/bazelbuild/bazel-gazelle/language#Language) interface, which requires `GenerateRules` and a few other methods. `language.Language` also embeds [`config.Configurer`](https://pkg.go.dev/github.com/bazelbuild/bazel-gazelle@v0.47.0/config#Configurer) and [`resolve.Resolver`](https://pkg.go.dev/github.com/bazelbuild/bazel-gazelle/resolve#Resolver).
+
+Gazelle can't dynamically load extensions at run-time: the Gazelle binary must be built with all extensions the user might need. The [`gazelle_binary`](https://github.com/bazel-contrib/bazel-gazelle/blob/master/reference.md#gazelle_binary) rule makes this easy: the user lists packages built with `go_library` that contain extensions. Each package must contain a `New()` function that returns a value implementing the extension interface. `gazelle_binary` generates a source file that calls each of those `New()` functions, then compiles that together with the rest of Gazelle.
```
nikki-ef:
```suggestion
Gazelle can't dynamically load extensions at run-time: the Gazelle binary must be built with all extensions the user might need. The [`gazelle_binary`](https://github.com/bazel-contrib/bazel-gazelle/blob/master/reference.md#gazelle_binary) rule makes this easy: the user lists packages built with `go_library` that contain necessary extensions. Each package must contain a `New()` function that returns a value implementing the extension interface. `gazelle_binary` generates a source file that calls each of those `New()` functions, then compiles that together with the rest of Gazelle.
```

nikki-ef (approved):
All suggestions are non-blocking, most of them are for slight clarity improvements.

how-gazelle-works.md
jayconrod: I think "must not" is better: it is *possible* for an extension to modify the original rules, but it's likely to result in a subtle bug. Extensions aren't supposed to do that, but there's no technical measure preventing it.

how-gazelle-works.md
jayconrod: Instead of. I've updated the reference doc for clarity, linked to it, and removed some text here. The `-mode` flag isn't too important for understanding how Gazelle works. I only mentioned it because I don't want to imply that changes are always written to disk.

how-gazelle-works.md
jayconrod: That sounds good. I moved this up, but I dropped mention of `Configure` and `Imports`. Rephrased a little bit for clarity.

how-gazelle-works.md
jayconrod: Added a slight clarification to the original. I realize I actually need to describe in detail how this works. I'll add a new section in a follow-up PR.

how-gazelle-works.md
jayconrod: Changed to "any necessary".

how-gazelle-works.md
jayconrod: Rephrased this slightly, but added the link at the top rather than here.

how-gazelle-works.md
nikki-ef: I see. Well, since it'll likely result in a bug I think that serves as a sufficient alert for any changes that were made.

PR #2255, opened 2025-12-16
docs: added explanation on how the merger works

jayconrod:
**What type of PR is this?**

> Documentation

> all

- How extensions generate rules
- How the merger combines existing and generated rules
- Examples
- keep comments and parser subtleties

For #2196

**Other notes for review**

### DIFF
--- a/language/go/build_constraints.go
+++ b/language/go/build_constraints.go
@@ -280,7 +280,7 @@ func matchAuto(tokens []string) (*buildTags, error) {
 // build constraints for a file and will instead defer to the compiler at compile
 // time.
 func isDefaultIgnoredTag(tag string) bool {
-	if tag == "cgo" || tag == "purego" || tag == "race" || tag == "msan" || strings.HasPrefix(tag, "goexperiments.") {
+	if tag == "cgo" || tag == "purego" || tag == "race" || tag == "msan" || strings.HasPrefix(tag, "goexperiment.") {
 		return true
 	}
 	if len(tag) < 5 || !strings.HasPrefix(tag, "go") {
--- a/language/go/fileinfo_go_test.go
+++ b/language/go/fileinfo_go_test.go
@@ -506,12 +506,12 @@ func TestGoExperimentsTagsIgnored(t *testing.T) {
 		desc, content string
 	}{
 		{
-			desc:    "goexperiments tag satisfied",
-			content: "//go:build goexperiments.range\n\npackage foo",
+			desc:    "goexperiment tag satisfied",
+			content: "//go:build goexperiment.range\n\npackage foo",
 		},
 		{
 			desc:    "negated goexperiments tag satisfied",
-			content: "//go:build !goexperiments.range\n\npackage foo",
+			content: "//go:build !goexperiment.range\n\npackage foo",
 		},
 	} {
 		t.Run(tc.desc, func(t *testing.T) {
--- a/walk/walk.go
+++ b/walk/walk.go
@@ -19,6 +19,7 @@ package walk
 
 import (
 	"errors"
+	"fmt"
 	"io/fs"
 	"log"
 	"os"
@@ -229,6 +230,12 @@ func Walk2(c *config.Config, cexts []config.Configurer, dirs []string, mode Mode
 
 		// Make sure to visit prefixes of relToVisit as well so we apply
 		// configuration directives.
+		if path.IsAbs(relToVisit) {
+			panic(fmt.Sprintf("relToVisit must not be absolute: %q", relToVisit))
+		}
+		if relToVisit != "" && relToVisit != path.Clean(relToVisit) {
+			panic(fmt.Sprintf("relToVisit is not clean: %q", relToVisit))
+		}
 		pathtools.Prefixes(relToVisit)(func(rel string) bool {
 			if _, ok := w.visits[rel]; !ok {
 				// Never visited this directory.


jayconrod:
@nikki-ef I'd appreciate if you could take a look at this one too. Not in a hurry though.

how-gazelle-works.md
```diff
@@ -85,3 +85,182 @@ gazelle_binary(
     ],
 )
 ```
+
+## Manipulating the syntax tree
+
+Gazelle uses the [`build` package](https://pkg.go.dev/github.com/bazelbuild/buildtools/build) (from buildifier and buildozer) to parse, edit, and format `BUILD` files. This package provides a low-level interface to the syntax tree. For more convenient editing and merging, Gazelle provides its own [`merger`](https://pkg.go.dev/github.com/bazelbuild/bazel-gazelle/merger) and [`rule`](https://pkg.go.dev/github.com/bazelbuild/bazel-gazelle/rule) packages.
+
+The `rule` package lets extensions create, update, and delete "rules" (which become call expressions in the underlying syntax tree) and read or write their attributes using simple values rather than syntax tree nodes. For example, an extension can create a new rule as follows:
```
fmeum, line 108: Should we add a paragraph about how this nomenclature differs from Bazel end user documentation (but aligns with Bazel's Java code)? This would be a new target of the `go_binary` rule, not a new rule.

how-gazelle-works.md
```diff
@@ -85,3 +85,182 @@ gazelle_binary(
...
+
+### Merging changes to the syntax tree
+
+`BUILD` files often contain a mix of human-written and machine-generated rules and attributes. Updating the machine-generated parts while preserving the human-written portion is a delicate process, so extensions do not directly modify the syntax tree. 
+
+Instead, each extension's [`GenerateRules`](https://pkg.go.dev/github.com/bazelbuild/bazel-gazelle/language#Language.GenerateRules) method creates and returns two lists: a `Gen` list of rules the `BUILD` file should contain, and an `Empty` list of rules that should be deleted from the `BUILD` file, if they're present. This is often enough information to regenerate the `BUILD` file from scratch.
+
+After calling `GenerateRules`, Gazelle calls [`merger.MergeFile`](https://pkg.go.dev/github.com/bazelbuild/bazel-gazelle/merger#MergeFile) to merge both lists with the rules that are already present. `MergeFile` is language-neutral, but its behavior is controlled by the map returned by each extension's [`Kinds`](https://pkg.go.dev/github.com/bazelbuild/bazel-gazelle/language#Language.Kinds) method.
```
nikki-ef:
```suggestion
After calling `GenerateRules`, Gazelle calls [`merger.MergeFile`](https://pkg.go.dev/github.com/bazelbuild/bazel-gazelle/merger#MergeFile) to merge the `Gen` and `Empty` lists with the rules that are already present in the `BUILD` file. `MergeFile` is language-neutral, but the way it handles rule attributes is controlled by the map returned by each extension's [`Kinds`](https://pkg.go.dev/github.com/bazelbuild/bazel-gazelle/language#Language.Kinds) method. 
```
Polish - makes the paragraph stand alone for someone who may land on it while scanning quickly.

how-gazelle-works.md
```diff
@@ -85,3 +85,182 @@ gazelle_binary(
...
+
+After calling `GenerateRules`, Gazelle calls [`merger.MergeFile`](https://pkg.go.dev/github.com/bazelbuild/bazel-gazelle/merger#MergeFile) to merge both lists with the rules that are already present. `MergeFile` is language-neutral, but its behavior is controlled by the map returned by each extension's [`Kinds`](https://pkg.go.dev/github.com/bazelbuild/bazel-gazelle/language#Language.Kinds) method. 
+
+`MergeFile` processes each rule as follows:
+
+1. `MergeFile` attempts to match the rule with an existing rule. An existing rule matches if:
+    - It has the same kind and name (a `go_binary` with `name = "server"`).
+    - One of its *matchable attributes* (determined by the `Kinds` map) has the same value (a `go_library` with `importpath = "example.com/hello/server"`).
+    - If the rule kind's `MatchAny` flag is set in the `Kinds` map, then any rule of that kind can match. This is useful when only one rule is expected per directory.
+1. If no match is found, then `MergeFile` either adds the rule if it was from the `Gen` list or ignores the rule if it was from the `Empty` list.
```
nikki-ef:
```suggestion
1. If `MergeFile` doesn't find a match, then it either adds the rule if it was from the `Gen` list or ignores the rule if it was from the `Empty` list.
```
Nit: make active voice

how-gazelle-works.md
```diff
@@ -85,3 +85,182 @@ gazelle_binary(
...
+1. If a match was found, `MergeFile` calls [`rule.MergeRules`](https://pkg.go.dev/github.com/bazelbuild/bazel-gazelle/rule#MergeRules) to combine the rules.
```
nikki-ef:
```suggestion
1. If `MergeFile` finds a match, it calls [`rule.MergeRules`](https://pkg.go.dev/github.com/bazelbuild/bazel-gazelle/rule#MergeRules) to combine the rules.
```
Nit: make active voice

how-gazelle-works.md
jayconrod: Yes, certainly. Added a new section at the top, explaining why we use *rule* and not *target*.

I thought about fixing this as part of the v2 proposal, but I think it would be too painful of a change for extensions.

PR #2259, opened 2025-12-19
Add default Gazelle override for go.chromium.org/luci

erock2112:
**What type of PR is this?**
Bug fix

internal/bzlmod/default_gazelle_overrides.bzl

Turns off Gazelle generation of go_proto_library targets for go.chromium.org/luci.  The repository includes the generated files, so these targets are unnecessary and they produce errors.

N/A

**Other notes for review**

### DIFF
--- a/internal/bzlmod/default_gazelle_overrides.bzl
+++ b/internal/bzlmod/default_gazelle_overrides.bzl
@@ -16,6 +16,7 @@ visibility("private")
 
 DEFAULT_BUILD_FILE_GENERATION_BY_PATH = {
     "cel.dev/expr": "on",
+    "github.com/bazelbuild/remote-apis-sdks": "on",
     "github.com/cncf/xds/go": "on",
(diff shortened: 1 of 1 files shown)


jayconrod:
Do we still want `proto_library` targets? If not, I'd suggest `gazelle:proto disable` instead.

erock2112:
> Do we still want `proto_library` targets? If not, I'd suggest `gazelle:proto disable` instead.

We don't currently depend on them, but they don't seem to hurt anything so I figured it was reasonable to generate them.

jayconrod:
Let's use `gazelle:proto disable` then. Other overrides in this list do that, so let's stay consistent.

(Proto rule generation in external repos is difficult to use and rarely works. We're planning to remove it in Gazelle v2).

erock2112:
Sounds good. Done.

jayconrod (approved):

PR #2262, opened 2026-01-02
Add check to see if testdata is empty when determining when to add data

### LINKED ISSUE #2233: Gazelle adds data source even for empty testdata directory
Recently ran into an issue where we removed a testdata directory in favor of a different testing approach. In the aftermath of this change we had a couple PRs that attempted to add `data = glob(["testdata/**"]),` to our BUILD.bazel. This was due to the presence of an untracked empty testdata directory after rebase.

I can't seem to think of a scenario in which we would want an empty testdata directory to be added as a data src. I believe the change could be narrowed down to adding a check for an empty subdir in the following code.

https://github.com/bazel-contrib/bazel-gazelle/blob/b13c1573c378bb3dd9b9de1397db89d0392f8e72/language/go/generate.go#L113-L115

sohumk-stripe:
**What type of PR is this?**

Bug fix

language/go

Adds logic to check that a testdata subdirectory is not empty before adding it as a data source.

Fixes #2233

### DIFF
--- a/cmd/gazelle/integration_test.go
+++ b/cmd/gazelle/integration_test.go
@@ -5114,3 +5114,42 @@ go_library(
 		},
 	})
 }
+
+// TestEmptyTestdataNoData checks that an empty testdata subdirectory does not
+// result in a data attribute being added to go_test rules.
+func TestEmptyTestdataNoData(t *testing.T) {
+	dir, cleanup := testtools.CreateFiles(t, []testtools.FileSpec{
+		{Path: "WORKSPACE"},
+		{
+			Path: "example_test.go",
+			Content: `
+package example
+
+import "testing"
+
+func TestExample(t *testing.T) {}
+`,
+		},
+		{Path: "testdata/"},
+	})
+	defer cleanup()
+
+	if err := runGazelle(dir, []string{"-go_prefix", "example.com/foo"}); err != nil {
+		t.Fatal(err)
+	}
+
+	buildContent, err := os.ReadFile(filepath.Join(dir, "BUILD.bazel"))
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Verify that the BUILD file doesn't contain a data attribute
+	if strings.Contains(string(buildContent), "data = ") {
+		t.Errorf("BUILD file should not contain data attribute for empty testdata\n%s", string(buildContent))
+	}
+
+	// Verify that a go_test rule was generated
+	if !strings.Contains(string(buildContent), "go_test(") {
+		t.Errorf("BUILD file should contain a go_test rule\n%s", string(buildContent))
+	}
+}


language/go/generate.go
```diff
@@ -107,12 +108,21 @@ func (gl *goLang) GenerateRules(args language.GenerateArgs) language.GenerateRes
...
 	var hasTestdata bool
 	for _, sub := range args.Subdirs {
 		if sub == "testdata" {
 			_, ok := gl.goPkgRels[path.Join(args.Rel, "testdata")]
 			hasTestdata = !ok
+
+			// Check if testdata directory is empty
+			if hasTestdata {
+				testdataPath := filepath.Join(args.Dir, "testdata")
+				entries, err := os.ReadDir(testdataPath)
```
jayconrod: Use [`walk.GetDirInfo`](https://pkg.go.dev/github.com/bazelbuild/bazel-gazelle/walk#GetDirInfo) to read directory contents rather than direct I/O. The `walk` package already caches this in memory.

language/go/generate.go
```diff
@@ -107,12 +108,21 @@ func (gl *goLang) GenerateRules(args language.GenerateArgs) language.GenerateRes
...
-	// not contain a buildable package.
+	// not contain a buildable package and is not empty.
 	var hasTestdata bool
 	for _, sub := range args.Subdirs {
 		if sub == "testdata" {
 			_, ok := gl.goPkgRels[path.Join(args.Rel, "testdata")]
 			hasTestdata = !ok
+
+			// Check if testdata directory is empty
+			if hasTestdata {
```
jayconrod: Nit: maybe just check `!ok` here instead of assigning `hasTestdata` above? It's a little confusing that `hasTestdata` is assigned more than once with different meaning.

language/go/generate.go
sohumk-stripe: We're not in a walk here, so I don't believe this would work?

language/go/generate.go
sohumk-stripe: Done, and also added a test.

language/go/generate.go
jayconrod: There's a hack in place to make it work through global state. (We'll improve that for v2, but don't worry about it for now.)

sohumk-stripe: Can you share an example of that workaround?

jayconrod: It's in https://github.com/bazel-contrib/bazel-gazelle/blob/master/walk/cache.go#L104, `globalWalker` is set by `Walk2`.

I realize that may not be active in the unit test, but `GenerateRules` is normally only called by `Walk2`.

Did you try adding a test in `//tests` instead? I imagine that might run into a problem with `git` not creating empty directories. `cmd/gazelle/integration_test.go` might be another place we could test this.

sohumk-stripe: Okay, I see what you mean. My one concern here is whether we're certain `GenerateRules` is only used within a walk, as it would break if not.

Will look at adding a test in `//tests` and/or `cmd/gazelle/integration_test.go`

sohumk-stripe: I updated the unit test to use Walk and changed the `os.ReadDir` to `walk.GetDirInfo`.

Still working on adding a test in `//tests` or `cmd/gazelle/integration_test.go`

sohumk-stripe: Added an integration test, couldn't add anything to `//tests` that could be checked in because of `git` not tracking directories.

jayconrod (approved):
Thanks, this looks good.

I appreciate your patience with `Walk` and with the tests. Reducing I/O should speed things up. I'm working on Gazelle v2, and fixing `walk.GetDirInfo` will be part of that (among many other things).

PR #2263, opened 2026-01-09
Fix the lockfile and let it update locally

fmeum:
**What type of PR is this?**

Other

developer experience

**Other notes for review**

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -20,8 +20,9 @@ common:ci --sandbox_debug
 common:ci --spawn_strategy=standalone
 common:ci --genrule_strategy=standalone
 common:ci --test_strategy=standalone
+common:ci --lockfile_mode=error
 
-common --lockfile_mode=error
+common --lockfile_mode=update
 common --test_output=errors
 
 common:incompatible --incompatible_load_proto_rules_from_bzl
--- a/MODULE.bazel.lock
+++ b/MODULE.bazel.lock
@@ -165,6 +165,123 @@
         "recordedRepoMappingEntries": []
       }
     },
+    "@@pybind11_bazel+//:python_configure.bzl%extension": {
+      "general": {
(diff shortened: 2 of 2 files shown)


fmeum:
The lockfile updated incrementally and not every build touches all the extensions. This update was the result of `bazel mod deps`, which forcefully evaluates all extensions, including stuff we don't transitively build but have a transitive `bazel_dep` on.

The way the lockfile is used by Bazel means that you can't get unexpected changes in `update` mode as extensions would also rerun when their declared inputs change. But if you do e.g. update a `bazel_dep`, you would currently first run into an error and then have to retry the command, which just seems unnecessary to me. The way we would catch unnecessary or unexpected lockfile changes would be during code review.

jayconrod:
> The lockfile updated incrementally and not every build touches all the extensions. This update was the result of bazel mod deps, which forcefully evaluates all extensions, including stuff we don't transitively build but have a transitive bazel_dep on.

Thanks, that makes sense. I suppose there's no build-time cost to having that in the lock file.

> The way the lockfile is used by Bazel means that you can't get unexpected changes in update mode as extensions would also rerun when their declared inputs change. But if you do e.g. update a bazel_dep, you would currently first run into an error and then have to retry the command, which just seems unnecessary to me. The way we would catch unnecessary or unexpected lockfile changes would be during code review.

The main issue I've run into is that a build can fill in missing parts of the lock file, and builds can run unexpectedly in the background (e.g., triggered by an LSP server). For example, I'll switch to a branch to look at something, then try to switch back but can't due to uncommitted changes. I'd like to force the lock file to be complete.

The experience you describe is frustrating too though. I suppose I'm bringing in some Go bias here. Go originally worked the same way, but we switched to `-mod=readonly` being the default (equivalent to `--lockfile_mode=errors`) because there were so many unexpected changes. But for Go, I think it's more realistic to expect that most changes are done via `go get` and `go mod tidy` and not manually edited into `go.mod`.

In any case, I'm happy as long as CI continues to run with `--lockfile_mode=errors`.

Issue #2269, opened 2026-01-21
[language/go]: regression in resolving package-relative dependency labels


grrtrr:
### Problem description

We are testing `v0.46.0`.

In several places I observed how _package-relative labels_ (short form, eg. `":host_config_proto_go"`) were resolved to _absolute labels_ (e.g. `"//tools/system_command_service/host_config:host_config_proto_go"`).

This then causes issues with `MergeFile`, which would _replace the relative path with the absolute path_.

I can confirm that this is a problem in Go. Will update this ticket if I see similar cases in other places.

In the case of Golang, this is not a critical bug, just annoying; since it needlessly converts relative labels to absolute ones.

### Root-causing the problem

I traced this back to `FindRuleWithOverride`. We are using a `resolve_regexp`:
```bazel
# gazelle:resolve_regexp go corp.tech/av/gen/(.*)/(.*)_pb //${1}:${2}_proto_go
```
The corresponding `label.Label` has an empty `Repo` (no `@`). Now, in #2157, the `config.RepoName` is now always filled in as part of `CheckFlags`.
As a result, the `from` label now has a non-empty `Repo`:
```go
// language/go/resolve.go
func (gl *goLang) Resolve(c *config.Config, ix *resolve.RuleIndex, rc *repo.RemoteCache, r *rule.Rule, importsRaw interface{}, from label.Label) {
  // ...
    deps, errs := imports.Map(func(imp string) (string, error) {
        l, err := resolve(c, ix, rc, imp, from)
        if err == errSkipImport {
            return "", nil
        } else if err != nil {
            return "", err
        }
        for _, embed := range gl.Embeds(r, from) {
            if embed.Equal(l) {
                return "", nil
            }
        }
        l = l.Rel(from.Repo, from.Pkg)   // <== No longer relative since l.Repo != from.Repo
        return l.String(), nil
    })
```

When the Go `Resolve` function is called, the `ResolveGo` calls `FindRuleWithOverride`, which resolves the `resolve_regexp` import with an empty `l.Repo`.
As a result, the expression `l.Rel(from.Repo, from.Pkg)` no longer applies, i.e. it returns the _absolute path_ of `l`, even though `l` is relative to `config.RepoName`.

### Possible fix

As a preliminary fix, the following addresses the problem with overrides that are relative to `config.RepoName`:
```diff
--- a/resolve/config.go
+++ b/resolve/config.go
[...]

PR #2270, opened 2026-01-22
resolve: use a fallback RepoName for overrides

grrtrr:
_PR type_:  bug fix
_Scope_: resolve overrides (all languages)

### Why this PR?

As discussed in #2269, when an override does not set an explicit repository name, use the configured one instead.

This avoids problems in resolving package-local labels.

### Description of changes

Substitute the configured `RepoName` when resolving overrides (`resolve/resolve_regexp`), and for `go` targets and the resolved `Label.Repo` is empty.


grrtrr:
@jayconrod - I have also added the fallback for regular overrides (in addition to the snippet in #2269, which only addressed `o.resolveRegexpDep()`).

jayconrod:
Thanks! Could you please add a regression test in `//tests` to verify the fix? Something similar to your original example would be great.

grrtrr:
Sure, will see to it that I get this done before the end of next week.

language/go/resolve.go
```diff
@@ -89,6 +89,9 @@ func (gl *goLang) Resolve(c *config.Config, ix *resolve.RuleIndex, rc *repo.Remo
 				return "", nil
 			}
 		}
+		if l.Repo == "" {
+			l.Repo = c.RepoName
+		}
```
grrtrr, line 94: I have added this fix (mentioned in #2269) as well.
When running the tests, this statement alone resolves the problem.

(Can be verified by taking out the 2 `FindRuleWithOverride` changes).

There are several cases where the above fix comes into play, e.g. `bazel` rules that generate code and wrap a `go_library`.

resolve/config.go
```diff
@@ -33,12 +33,18 @@ import (
 func FindRuleWithOverride(c *config.Config, imp ImportSpec, lang string) (label.Label, bool) {
 	rc := getResolveConfig(c)
 	if dep, ok := rc.findOverride(imp, lang); ok {
+		if dep.Repo == "" {
+			dep.Repo = c.RepoName
+		}
```
grrtrr, line 38: This is tested by the `gazelle:resolve` case.

resolve/config.go
```diff
@@ -33,12 +33,18 @@ import (
...
 		return dep, true
 	}
 	for i := len(rc.regexpOverrides) - 1; i >= 0; i-- {
 		o := rc.regexpOverrides[i]
 		if o.matches(imp, lang) {
 			dep := o.resolveRegexpDep(imp)
+			if dep.Repo == "" {
+				dep.Repo = c.RepoName
+			}
```
grrtrr, line 47: This is tested by the `gazelle:resolve_regexp` case.

grrtrr:
This is the output of the test without code changes:
```console
INFO: Analyzed target //tests:go_resolve_with_empty_repo_name (0 packages loaded, 0 targets configured).
FAIL: //tests:go_resolve_with_empty_repo_name (Exit 1) (see /home/grenker/.cache/bazel/_bazel_grenker/d3b620f996ac5e418dce3977fc7e6c38/execroot/_main/bazel-out/k8-fastbuild/testlogs/tests/go_resolve_with_empty_repo_name/test.log)
INFO: From Testing //tests:go_resolve_with_empty_repo_name:
==================== Test output for //tests:go_resolve_with_empty_repo_name:
--- FAIL: TestFullGeneration (0.01s)
    --- FAIL: TestFullGeneration/go_resolve_with_empty_repo_name (0.01s)
        files.go:192: go_resolve_with_empty_repo_name/BUILD.bazel diff (-want,+got):
              (
                """
                ... // 20 identical lines
                    deps = [
            -           ":host_config_proto_go",
            +           "//go_resolve_with_empty_repo_name:host_config_proto_go",
                    ],
                )
                ... // 21 identical lines
                """
              )
        files.go:321: 
            Run UPDATE_SNAPSHOTS=true bazel run //tests:go_resolve_with_empty_repo_name to update BUILD.out and expected{Stdout,Stderr,ExitCode}.txt files.
FAIL
```

PR #2071, opened 2025-04-16
go_grpc_compiler: default to go_grpc_v2

sluongng:
Set the default grpc compiler to go_grpc_v2 because go_grpc is considered deprecated.

https://github.com/bazel-contrib/rules_go/blob/8251e45ee86fd17f3fedb34caa2d06a40089f22e/proto/BUILD.bazel#L34

**What type of PR is this?**

Feature

**What package or component does this PR mostly affect?**

language/go

**What does this PR do? Why is it needed?**

**Which issues(s) does this PR fix?**

Fixes #

**Other notes for review**

PR #2271, opened 2026-01-23
Add proto compiler for grpc libraries

mering:
**What type of PR is this?**

Bug fix

language/go

The go_grpc compiler generated both proto and grpc bindings. The go_grpc_v2 compiler only generates the grpc bindings but requires explicit use of go_proto for the proto bindings.

This change in behavior was missed in #2071.

**Other notes for review**


mering:
@jayconrod @fmeum Is there a release of Gazelle including this PR planned soon? We are eagerly waiting to start using this fix.

Issue #2272, opened 2026-01-25
Gazelle v2 implementation plan


jayconrod:
This is a tracking issue for tasks that need to be done for v2. [Gazelle 2.0 proposal](https://github.com/bazel-contrib/bazel-gazelle/discussions/2207) is the place to discuss what's actually included in v2.

## Beta release (Bazelcon 2026)

- [x] Add go.mod, go.work file for development
- [x] Move fix-update logic to v2/cmd/gazelle/update, to be shared by v1 and v2
- [x] Introduce v2/cmd/gazelle binary
- [x] Let `gazelle_binary` build either v1 or v2; make integration tests cover both
- [x] Move tests in `cmd/gazelle/integration_test.go` to `//tests` so they cover both versions
- [x] Make `go_repository_tools` build v2
- [x] Add version awareness, `-version` flag
- [x] Move and forward basic packages: `flag`, `label`, `merger`, `pathtools`, `rule`, `testtools`
- [x] Copy and forward interface packages, but add new interfaces: `config`, `resolve`, `language`
- [x] Add compatibility wrappers, and convert `update`, `testtools`, etc to use v2 interfaces
- [ ] Rewrite `go_deps` and `go_repository` to resolve packages in multi-module workspaces
    - [ ] #2344 
    - [ ] #2413 
- [x] Migrate miscellaneous `Configurer` implementations to new interfaces
- [x] `gazelle_binary`: support a mix of v1 and v2 extensions
- [x] Migrate language extensions to use new interfaces
    - [x] Write agent skill to aid migration
    - [x] Test languages
    - [x] Visibility
    - [x] Proto
    - [x] Go
- [ ] Refactoring
    - [x] `walk.Walk2`: rename to `Walk`, discard old `Walk`.
    - [x] `rule.KindInfo`: use `LoadedFrom` instead of `ApparentLoads`
    - [x] `visibility`: promote to be an internal extension
    - [ ] #2279 
    - [x] #2280
    - [x] #2421 
- [x] `-strict` mode (also #499): finish the run, collect errors, print them, exit non-zero on failure.
- [x] Enable lazy indexing by default: use `-r=false -index=lazy` when given at least one positional argument
- [ ] Add a -fix flag, replacing the `fix` subcommand
- [x] Make the `GenerateResult.Imports` field optional
- [ ] Clean up remaining `TODO(v2)` comments
[...]

PR #2274, opened 2026-01-25
Upgrade Go to 1.24.12, rules_go to 0.59.0, and sync deps to WORKSPACE

jayconrod:
**What type of PR is this?**

> Other

> all

- Go is still a pretty old version, but a year newer than what we had before. This gives us fancy range loops.
- rules_go is now the latest version, needed to resolve some nogo issues in WORKSPACE mode, though I'm not sure if it was actually a rules_go problem or upgrading the ancient version of x/tools we were on.
- Synced dependencies from go.mod / MODULE.bazel to WORKSPACE. These had completely diverged.
- Removed some apparently unused compatibility definitions referencing multiple versions of Go.
- Disabled Stardoc targets in WORKSPACE mode. They no longer build due to some hidden dependency on rules_cc. Don't care.

For #2272

**Other notes for review**

### DIFF
--- a/internal/bzlmod/default_gazelle_overrides.bzl
+++ b/internal/bzlmod/default_gazelle_overrides.bzl
@@ -38,6 +38,9 @@ DEFAULT_DIRECTIVES_BY_PATH = {
     "github.com/authzed/spicedb": [
         "gazelle:proto disable",
     ],
+    "github.com/bufbuild/protocompile": [
+        "gazelle:proto disable",
+    ],
     "github.com/census-instrumentation/opencensus-proto": [
         "gazelle:proto disable",
     ],


cmd/fetch_repo/BUILD.bazel
```diff
@@ -25,13 +25,6 @@ go_binary(
     visibility = ["//visibility:public"],
 )
 
-# Verify that fetch_repo builds with Go 1.22.
```
fmeum, line 28: The idea behind this target is that it provides assurance that Gazelle builds with the lowest SDK we aim to support. Could we keep this target, but bump the referenced SDK to 1.24.12? We could then bump the overall Go version to the latest Go SDK version to verify that we aren't broken by any incompatible changes to Go?

jayconrod: My problem with `go_cross_binary` is that it causes us to download and build a separate copy of the standard library, which makes the build considerably slower. Actually I'm seeing four `GoStdlib` actions since I think it's building for both exec and target.

Would it be all right if we consolidated on supporting a single Go version? I'm not too worried about either forward or backward compatibility being tested here, but if we need to test another version, I think it would be an end-to-end test, not run by `bazel test //...`, but a separate command (similar to the BCR tests).

PR #2276, opened 2026-01-25
v2: extract fix-update logic to v2/cmd/gazelle/update

jayconrod:
**What type of PR is this?**

> Other

> v2/cmd/gazelle/update

This new package now implements Gazelle's core fix/update functionality. cmd/gazelle and v2/cmd/gazelle will both we wrappers around this, though the update-repos command will remain in v1, since we're dropping it in v2.

Also modifed go_repository_tools to work in module mode, since it would actually require deeper changes to continue working in GOPATH mode.

For #2272

**Other notes for review**

Also copied `internal/wspace`. It's an internal package, so it can't be moved and forwarded unfortunately.

### DIFF
--- a/internal/go_repository_tools.bzl
+++ b/internal/go_repository_tools.bzl
@@ -62,7 +62,7 @@ def _go_repository_tools_impl(ctx):
         env["GOPROXY"] = ctx.os.environ["GOPROXY"]
 
     bin_dir = str(ctx.path("bin"))
-    gazelle_dir = str(ctx.path(Label("//:WORKSPACE")).dirname)
+    gazelle_dir = str(ctx.path(Label("//:MODULE.bazel")).dirname)
 
     # Make sure the list of source is up to date.
     # We don't want to run the script, then resolve each source file it returns.


internal/go_repository_tools.bzl
```diff
@@ -68,6 +61,9 @@ def _go_repository_tools_impl(ctx):
     if "GOPROXY" in ctx.os.environ:
         env["GOPROXY"] = ctx.os.environ["GOPROXY"]
 
+    bin_dir = str(ctx.path("bin"))
+    gazelle_dir = str(ctx.path(Label("//:WORKSPACE")).dirname)
```
fmeum: Let's go with `MODULE.bazel` here as we might delete WORKSPACE in the future.

v2/internal/wspace/finder.go
```diff
@@ -0,0 +1,75 @@
+/* Copyright 2016 The Bazel Authors. All rights reserved.
```
fmeum, line 1: Can we already delete `wspace` package in the non-v2 Gazelle module? We can duplicate this temporarily if necessary, but deleting it now would simplify the git diff.

jayconrod: Unfortunately not. It's used in a few packages beyond `cmd/gazelle` including `config` and `cmd/move_labels`. I've added a refactoring task to the list in #2272, but I'm planning to leave most of the refactoring to the end, after the new interfaces are in place.

internal/go_repository_tools.bzl
jayconrod: Done, looking forward to it!

PR #2277, opened 2026-01-25
v2: add v2/cmd/gazelle

jayconrod:
**What type of PR is this?**

> Other

> v2/cmd/gazelle

The new entry point for v2. It just wraps update.Run. Doesn't need to do much more than that.

For #2272

**Other notes for review**

### DIFF
--- a/v2/cmd/gazelle/main.go
+++ b/v2/cmd/gazelle/main.go
@@ -31,6 +31,11 @@ func main() {
 	log.SetPrefix("gazelle: ")
 	log.SetFlags(0) // don't print timestamps
 
+	// TODO(#2279): interpret arguments as paths relative to either
+	// BUILD_WORKSPACE_DIRECTORY or BUILD_WORKING_DIRECTORY, depending on whether
+	// the paths were specified as arguments to the gazelle macro or passed as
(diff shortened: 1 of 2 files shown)


v2/cmd/gazelle/main.go
```diff
@@ -0,0 +1,60 @@
...
+
+	"github.com/bazel-contrib/bazel-gazelle/v2/cmd/gazelle/update"
+)
+
+func main() {
+	log.SetPrefix("gazelle: ")
+	log.SetFlags(0) // don't print timestamps
+
+	var wd string
+	if wsDir := os.Getenv("BUILD_WORKSPACE_DIRECTORY"); wsDir != "" {
```
fmeum, line 40: Is this the correct variable compared to `BUILD_WORKING_DIRECTORY`? The `else` branch uses the current wd rather than searching up to a go.mod/MODULE.bazel file.

jayconrod: It's not really correct, but copied from v1. We should fix it for v2, though arguably it's a bug in v1, and we should fix it there too. I opened #2279, added a TODO comment here, and added to the list in #2272.

PR #2281, opened 2026-02-01
gazelle_binary: support either v1 or v2

jayconrod:
**What type of PR is this?**

> Other

> gazelle_binary

gazelle_binary accepts a "version" attribute, used to select between either the ./cmd/gazelle or ./v2/cmd/gazelle main package. The default is determined by Gazelle's module version.

//tests now has separate gazelle_binary targets and test targets for each version so that we don't regress either v1 or v2.

For #2272

**Other notes for review**


internal/gazelle_binary.bzl
```diff
@@ -126,3 +150,27 @@ def format_import(importpath):
...
+
+    Returns 1 if the major version is 0 or 1 or when building in WORKSPACE mode.
+    Very little difference in behavior is expected between these versions.
+
+    Returns 2 if the major version is 2 or unset. When the major version is not
+    set, it's assumed to be a development version.
+    """
+    if not GAZELLE_IS_BAZEL_MODULE:
+        return 1
+    if not GAZELLE_MODULE_VERSION:
```
fmeum, line 165: Do we want dev builds to always use v2?

fmeum (approved):

internal/gazelle_binary.bzl
jayconrod: I think so yes. For now, this only affects whether `gazelle_binary` uses `cmd/gazelle` or `v2/cmd/gazelle` by default, and that can be overridden if needed. Defaulting to v2 in module mode when no module version is found I think should only affect Gazelle developers.

PR #2284, opened 2026-02-11
Switch to "protobuf" module when generating proto_library rules

### LINKED ISSUE #2283: Module "rules_proto" is deprecated
We can see the information at:

- https://github.com/bazelbuild/rules_proto
- https://registry.bazel.build/modules/rules_proto

Thus, we should change the implementation of:

```go
func (*protoLang) ApparentLoads(moduleToApparentName func(string) string) []rule.LoadInfo
```

so we generate:

```starlark
load("@protobuf//bazel:proto_library.bzl", "proto_library")
```

instead of:

```starlark
load("@rules_proto//proto:defs.bzl", "proto_library")
```

mateusz-olczyk:
**What type of PR is this?**

> Other

> language/proto

Because `"rules_proto"` is a deprecated module, `proto_library` rule should be fetched from the `"protobuf"` module. This change updates `ApparentLoads()` method for `proto` extension and updates the test scenarios accordingly. Moreover, `rules_proto` is not mentioned in `README.md` anymore; now it refers to `protobuf`.

Fixes #2283 

**Other notes for review**

I wasn't sure if `"rules_proto"` should also be removed from `MODULE.bazel`. It looks like `bazel test //...` passes after removing both `"rules_proto"` and `"protobuf"` from `MODULE.bazel`. But I see those may be required by the submodules: `tests/bcr/go_mod` and `tests/bcr/go_work`.

### DIFF
--- a/gazelle-reference.md
+++ b/gazelle-reference.md
@@ -203,6 +203,8 @@ Comma-separated list of visibility specifications. This directive adds the visib
 # gazelle:default_visibility //foo:__subpackages__,//src:__subpackages__
 ```
 
+You must include the extension `@gazelle//language/bazel/visibility` to use this directive.
+
 ### `WORKSPACE` directives
 
 Gazelle also reads directives from the WORKSPACE file. They may be used to discover custom repository names and known prefixes. The `fix` and `update` commands use these directives for dependency resolution. `update-repos` uses them to learn about repository rules defined in alternate locations.


fmeum:
+1 for doing this, even if we may have to migrate again at some point in the future if protobuf is split up.

language/proto/kinds.go
```diff
@@ -41,13 +41,13 @@ func (pl *protoLang) Loads() []rule.LoadInfo {
 }
 
 func (*protoLang) ApparentLoads(moduleToApparentName func(string) string) []rule.LoadInfo {
-	rulesProto := moduleToApparentName("rules_proto")
+	rulesProto := moduleToApparentName("protobuf")
 	if rulesProto == "" {
-		rulesProto = "rules_proto"
+		rulesProto = "com_google_protobuf"
```
jayconrod: Is there a reason to use `com_google_protobuf`? The name of the module is `protobuf`, so it seems better to prefer that.

README.md
```diff
@@ -31,7 +31,7 @@ Gazelle can generate Bazel BUILD files for many languages:
...
 * **JavaScript / TypeScript:** Aspect provides [JavaScript and TypeScript Support](https://github.com/aspect-build/aspect-gazelle/tree/main/language/js). BenchSci's [rules_nodejs_gazelle](https://github.com/benchsci/rules_nodejs_gazelle) supports generating `ts_project`, `js_library`, `jest_test`, and `web_asset` rules, and is able to support module bundlers like Webpack and Next.js.
 * **Kotlin:** Aspect Build provides some [Kotlin Support](https://github.com/aspect-build/aspect-gazelle/tree/main/language/kotlin). Still under development, please check the README for currently available features.
-* **Protocol Buffers:** Support for the `proto_library` rule, as well as `go_proto_library` is in this repository, see below. Other language-specific proto rules are not supported here. [stackb/rules_proto](https://github.com/stackb/rules_proto) is a good resource for these rules.
+* **Protocol Buffers:** Support for the `proto_library` rule, as well as `go_proto_library` is in this repository, see below. Other language-specific proto rules are not supported here. [protocolbuffers/protobuf](https://github.com/protocolbuffers/protobuf/tree/main/bazel) is a good resource for these rules.
```
jayconrod:
```suggestion
* **Protocol Buffers:** Support for the `proto_library` rule. Language-specific rules like `go_proto_library` are supported by other language extensions.
```

language/proto/kinds.go
jayconrod, line 1: One missing piece here: if the build file loads `@rules_proto//proto:defs.bzl` AND the current module depends on `protobuf` (or some renaming of it), then we should implement `Fix` to update the name. That might be as simple as deleting the old `load` statement; if it is restored automatically, then all is well. If not, we'd have to modify the statement there.

jayconrod:
> I wasn't sure if "rules_proto" should also be removed from MODULE.bazel. It looks like bazel test //... passes after removing both "rules_proto" and "protobuf" from MODULE.bazel. But I see those may be required by the submodules: tests/bcr/go_mod and tests/bcr/go_work.

Please update `tests/bcr/go_mod/MODULE.bazel` and `tests/bcr/go_work/MODULE.bazel` to depend on `protobuf` rather than `rules_proto`. These are end-to-end tests that run Gazelle in a fake repo and verify that it generates real targets that can be built. There are .proto files here.

Let's drop the `rules_proto` dependency from the main `MODULE.bazel` completely. Gazelle itself doesn't have .proto files or `proto_library` targets, so we don't need it. (I think... perhaps CI will prove me wrong.)

language/proto/kinds.go
mateusz-olczyk: I saw that pattern once in Gazelle; I mean, defaulting to the workspace name when there's no appropriate definition found in `MODULE.bazel` (or rather `MODULE.bazel` does not exist at all).

See the implementation in `language/go/kinds.go`:

```go
 func (*goLang) ApparentLoads(moduleToApparentName func(string) string) []rule.LoadInfo {
	return apparentLoads(moduleToApparentName)
}

func apparentLoads(moduleToApparentName func(string) string) []rule.LoadInfo {
	rulesGo := moduleToApparentName("rules_go")
	if rulesGo == "" {
		rulesGo = "io_bazel_rules_go"
	}
	gazelle := moduleToApparentName("gazelle")
	if gazelle == "" {
		gazelle = "bazel_gazelle"
	}

	# ...
}
```

README.md
mateusz-olczyk: Yes, this definition is more precise. Acknowledged.

language/proto/kinds.go
mateusz-olczyk: Good catch!

> That might be as simple as deleting the old load statement

That would be true, when implementing `Fix()` only for the `proto_library` symbols. But I decided to fix all the symbols from [@rules_proto](https://github.com/bazelbuild/rules_proto/blob/main/proto/defs.bzl), so the implementation is more comprehensive.

mateusz-olczyk:
> > I wasn't sure if "rules_proto" should also be removed from MODULE.bazel. It looks like bazel test //... passes after removing both "rules_proto" and "protobuf" from MODULE.bazel. But I see those may be required by the submodules: tests/bcr/go_mod and tests/bcr/go_work.
> 
> Please update `tests/bcr/go_mod/MODULE.bazel` and `tests/bcr/go_work/MODULE.bazel` to depend on `protobuf` rather than `rules_proto`. These are end-to-end tests that run Gazelle in a fake repo and verify that it generates real targets that can be built. There are .proto files here.
> 
> Let's drop the `rules_proto` dependency from the main `MODULE.bazel` completely. Gazelle itself doesn't have .proto files or `proto_library` targets, so we don't need it. (I think... perhaps CI will prove me wrong.)

Done 👍

language/proto/kinds.go
jayconrod: Ah, that's right. I suppose we need to keep supporting `WORKSPACE` dependents for a while longer, though I hope we can drop that at some point soon.

language/proto/fix.go
```diff
@@ -16,9 +16,96 @@ limitations under the License.
...
+		if l.Name() == deprecatedFile {
+			deprecatedLoads = append(deprecatedLoads, l)
+		}
+	}
+
+	if len(deprecatedLoads) == 0 {
+		return
+	}
+
+	if !c.ShouldFix {
```
jayconrod: I think we can skip this check, since we know at this point that there's a dependency on `protobuf`, so the `load` statements should work. `ShouldFix` is meant for potentially destructive cases, like when we're renaming rules or combining multiple rules into one.

language/proto/fix_test.go
jayconrod, line 1: It's fine to keep this unit test, but for the future, just adding or updating a case in `//tests` is good enough. We don't always need unit test coverage when we have good integration tests. For Gazelle, the unit tests are often more complicated.

language/proto/fix.go
mateusz-olczyk: Done.

language/proto/fix_test.go
mateusz-olczyk: Sure. I saw the existing `language/go/fix_test.go` and was inspired by it.

PR #2286, opened 2026-02-11
Add sorting of load statements in BUILD files

aaron-skydio:
**What type of PR is this?**

Feature

cmd/gazelle

Previously, gazelle was emitting unsorted load statements which conflicts with buildifier's enforced sorted load order (the out-of-order-load rule), causing buildifier to fail if BUILD files were not explicitly formatted after running gazelle.  Add sort_loads.patch that introduces a SortLoads() method on rule.File and calls it at the end of merger.FixLoads() so gazelle produces loads in sorted order natively.  We match buildifier's definition of the sort order.

This is part of
https://github.com/bazel-contrib/bazel-gazelle/issues/19, although it doesn't address all of the possible features mentioned there, just ordering within blocks of loads.

**Other notes for review**

Should this be gated behind some configuration?  I'm not sure if there's any reason you'd need to disable this

### DIFF
--- a/rule/rule.go
+++ b/rule/rule.go
@@ -406,6 +406,59 @@ func (f *File) Format() []byte {
 	return bzl.Format(f.File)
 }
 
+// SortLoads sorts contiguous chunks of load statements in the file by name,
+// matching buildifier's out-of-order-load behavior. Load statements separated
+// by non-load statements (comments, docstrings, rules) are sorted independently.
+// This method calls Sync internally.
+func (f *File) SortLoads() {
+	f.Sync()
+
+	if len(f.Loads) <= 1 {
+		return
+	}
+
+	// Ensure f.Loads is ordered by statement index. After Sync, recently
+	// inserted loads may be appended out of AST order, which would cause
+	// loadsByName.Swap to produce incorrect results.
+	sort.SliceStable(f.Loads, func(i, j int) bool {
+		return f.Loads[i].index < f.Loads[j].index
+	})
+
+	// Group loads into contiguous chunks based on their position in the
+	// statement list. Two loads are in the same chunk if there are no
+	// non-load statements between them.
+	type chunk struct {
+		start, end int // indices into f.Loads
+	}
+	var chunks []chunk
+	chunkStart := 0
+	for i := 1; i < len(f.Loads); i++ {
+		prevIdx := f.Loads[i-1].index
+		currIdx := f.Loads[i].index
+		// Check if there are any non-load statements between these two loads
+		contiguous := true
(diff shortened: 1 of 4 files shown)


rule/rule.go
```diff
@@ -406,6 +406,59 @@ func (f *File) Format() []byte {
 	return bzl.Format(f.File)
 }
 
+// SortLoads sorts contiguous chunks of load statements in the file by name,
+// matching buildifier's out-of-order-load behavior. Load statements separated
```
fmeum, line 410: Is there a way to reuse buildifier's sorting here? It would be great if we could guarantee that Gazelle and buildifier won't fight about this when the versions align.

aaron-skydio: I think we'd want to call compareLoadLabels and comparePaths from there, which aren't currently exported, we could make them exported though with a PR to the buildtools repo and a version bump here if it's important?  I don't think we can just reuse Rewriter from there because of the other bookkeeping here

Not sure if the buildifier sort order is something that might change

We can also add a test that the results match buildifier's without exporting those functions?

Not sure which of those is preferred

jayconrod: Perhaps we should call `github.com/bazelbuild/buildtools/build.Rewrite`? That would apply all the other buildifier rewrites, so it would be a bigger change, but everything in the list looks fine. I'd certainly prefer not to duplicate the code from buildifier.

Not sure how complicated this will be. Maybe we can call from `File.Sync`.

rule/rule_test.go
```diff
@@ -509,6 +509,225 @@ def repos():
 	}
 }
 
+func TestSortLoads(t *testing.T) {
```
jayconrod, line 512: I'd prefer to have an integration test for this in `//tests` than a unit test. The place where we implement this doesn't really matter.

aaron-skydio:
Thanks for the review @jayconrod - I looked into the `Rewrite` thing, and it looks like on the current version of gazelle loads are actually already sorted - `Format` in `rule/rule.go` calls `bzl.Format`, which calls `Rewrite`, which does sort loads.

This came in from https://github.com/bazelbuild/buildtools/pull/1244, which was pulled into gazelle in version 0.36 [here](https://github.com/aaron-skydio/bazel-gazelle/commit/1b331b6d9cb2d981580a8d3e2dfc551185790737#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6), I ported this change forward to master from an older version of gazelle where the vendored version of buildtools didn't do load sorting and didn't notice it was already fixed on master (really should have checked that)

Sorry for the noise, I'll close this

jayconrod:
Thanks, that's good news!

PR #2287, opened 2026-02-13
(docs) Note that an extension is required for default_visibility

### LINKED ISSUE #2285: default_visibility directive not recognized
I realize that this [directive does not work well](https://github.com/bazel-contrib/bazel-gazelle/issues/1666), but it also appears to be unrecognized. Running gazelle-0.47.0:
```
gazelle: /Users/djmitche/go/src/foo/bar/BUILD.bazel: unknown directive: gazelle:default_visibility
```
That file contains
```
#gazelle:default_visibility //bar:default_visibility
```

djmitche:
**What type of PR is this?**

>
> Bug fix
> Feature
Documentation
> Other

> For example:
>
> language/go
> cmd/gazelle
> go_repository
all

Adds a note that an extension is required in order to use the default_visibility directive. I was unaware of this and that led to #2285. I do not see the extension referenced in any other documentation in the repo.

Fixes #2285.

**Other notes for review**

This file only references language extensions, and from a user perspective my thinking was, "I have the Go extension installed, so what more do I need?"

If it would be better to add some kind of `**Required Extension:** to each directive, I can do that, especially with help to figure out which extension each one requires.

### DIFF
--- a/gazelle-reference.md
+++ b/gazelle-reference.md
@@ -203,6 +203,8 @@ Comma-separated list of visibility specifications. This directive adds the visib
 # gazelle:default_visibility //foo:__subpackages__,//src:__subpackages__
 ```
 
+You must include the extension `@gazelle//language/bazel/visibility` to use this directive.
+
 ### `WORKSPACE` directives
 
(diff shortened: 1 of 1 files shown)


jayconrod:
Thanks for fixing this.

There are only a couple directives that work this way, and I think `default_visibility` is the only one that would be commonly used. I hope to have this squashed into v2 before too long.

PR #2289, opened 2026-02-13
Add ability to show versions

### LINKED ISSUE #2288: `gazelle version` command
In #2285 I spent a good amount of time trying to convince myself I was using Gazelle 0.47.0 and not an older version. Bazel is a bit opaque to non-experts, so I wasn't confident that just changing the version in MODULE.bazel would do the trick. And in fact I needed to run `bazel run @rules_go//go -- mod tidy` to update some other files, so perhaps I was running the old version for a while? Unclear.

It'd be really helpful to get the ground-truth on this by just running `bazel run //:gazelle -- version` or something similar.

Adding a new command is pretty trivial, but it's much less clear how the binary should know its version, so I'd like some guidance there:
 * Put the version in a `.go` file somewhere and remember to update that on release
 * Use `git describe` in whatever automation builds releases to get the version information and embed that with `-X` or something similar
 * Use [`debug.ReadBuildInfo`](https://pkg.go.dev/runtime%2Fdebug#ReadBuildInfo), but this will probably not return any data for common deployment scenarios.

I can work on the Go part of this (adding the command) given some guidance on the second question. I don't know much about BCR or how Gazelle is released in general.

(also, I don't know why the issue template is not populating in the GitHub UI, but it does not seem relevant for this issue)

djmitche:
**What type of PR is this?**

>
> Bug fix
Feature
> Documentation
> Other

> For example:
>
> language/go
cmd/gazelle
> go_repository
> all

This adds -V, -version, and --version, giving the current Gazelle version.

Fixes #2288.

**Other notes for review**

Thanks for the advice in the issue -- I would have tried a lot of much more difficult things first!

### DIFF
--- a/cmd/gazelle/BUILD.bazel
+++ b/cmd/gazelle/BUILD.bazel
@@ -17,6 +17,7 @@ go_library(
     importpath = "github.com/bazelbuild/bazel-gazelle/cmd/gazelle",
     tags = ["manual"],
     visibility = ["//visibility:public"],
+    x_defs = {"main.GazelleVersion": module_version() or "dev"},
     deps = [
         "//config",
         "//internal/wspace",
--- a/cmd/gazelle/main.go
+++ b/cmd/gazelle/main.go
@@ -91,6 +91,9 @@ func run(wd string, args []string) error {
 	if len(args) == 1 && (args[0] == "-h" || args[0] == "-help" || args[0] == "--help") {
 		return help()
 	}
+	if len(args) == 1 && (args[0] == "-version" || args[0] == "--version") {
+		return version()
+	}
 	if len(args) == 0 {
 		return update.Run(ctx, languages, wd, args)
 	}
@@ -104,6 +107,14 @@ func run(wd string, args []string) error {
 		// update.Run knows what to do with it.
 		return update.Run(ctx, languages, wd, args)
 	}
+}
+
+// This is substituted at build time using x_defs.
+var GazelleVersion = "unknown"
+
+func version() error {
+	fmt.Fprintf(os.Stderr, "gazelle %s\n", GazelleVersion)
+	return flag.ErrHelp
 }
 
 func help() error {


cmd/gazelle/BUILD.bazel
```diff
@@ -17,6 +17,7 @@ go_library(
     importpath = "github.com/bazelbuild/bazel-gazelle/cmd/gazelle",
     tags = ["manual"],
     visibility = ["//visibility:public"],
+    x_defs = {"main.GazelleVersion": "0.47.0"},
```
fmeum:
```suggestion
    x_defs = {"main.GazelleVersion": native.module_version() or "dev"},
```
something like this should work

djmitche: Oops, I missed that! AIUI `native.` is unnecessary and discouraged in `BUILD` files, so I've left it out.

cmd/gazelle/main.go
```diff
@@ -91,6 +91,9 @@ func run(wd string, args []string) error {
 	if len(args) == 1 && (args[0] == "-h" || args[0] == "-help" || args[0] == "--help") {
 		return help()
 	}
+	if len(args) == 1 && (args[0] == "-V" || args[0] == "-version" || args[0] == "--version") {
```
fmeum: Is capital V common for this? Lowercase v should probably be reserved for a verbose mode. Perhaps we should only allow the long form?

djmitche: My thinking was `-v` is usually verbose. I can omit the short form entirely though.

jayconrod:
This needs to be supported in v2 as well, otherwise it looks like we're deliberately removing a feature.

I'd ask if we can hold off on this for a bit though, until I can give it some more thought. Gazelle will need to change behavior based on whether it's v1 or v2, and that needs to be cleanly communicated to extensions. The `cmd/gazelle` package is not the right place for this information, since it's only v1 and not available to extensions.

djmitche:
I'm happy to hold off. Hopefully this sort of thing isn't too difficult to add for v2, though!

djmitche:
Looks like https://github.com/bazel-contrib/bazel-gazelle/pull/2307 supersedes this.

PR #2290, opened 2026-02-13
Update pull-request template for better results

djmitche:
**What type of PR is this?**

Other

all

Update the pull request template, as #2288 shows that removing the `> ` prefix from one of the lines does not change the rendered output.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -11,22 +11,20 @@
 -->
 
 **What type of PR is this?**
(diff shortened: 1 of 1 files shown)


jayconrod:
I don't really see a reason to change this. The rendered markdown looks worse in preview.

djmitche:
I won't push the question, but based on the appearance of #2288 where I did "uncomment" one item in each question, is the template really achieving its purpose right now?

jayconrod:
The key instruction there is to remove the other lines.

PR #2294, opened 2026-02-22
tests: convert most remaining integration tests

jayconrod:
**What type of PR is this?**

> Other

> tests

Convert tests that weren't converted earlier by #2293 and remove the corresponding old integration tests.

The remaining integration tests can't be converted because they aren't compatible with the directory test framework. Most of them cover `update-repos`, which isn't supported there and won't be supported in v2 anyway. A few tests have excessively long names or subtests, and runfiles couldn't be accessed on Windows.

For #2272

**Other notes for review**


jayconrod:
Sorry this is huge. Should be the last one of these conversions.

PR #2295, opened 2026-02-24
Implement Rule.AttrBool() method

mateusz-olczyk:
**What type of PR is this?**

> Feature

> rule

`Rule` already has methods like `AttrString` or `AttrStrings`. Boolean values are another common attribute type in Bazel rules. As a developer of a `gazelle` extension, I'd appreciate having this as a part of the `Rule` interface instead of having to write side helper functions.

I have not created an issue for such a simple task.

**Other notes for review**

### DIFF
--- a/rule/rule_test.go
+++ b/rule/rule_test.go
@@ -778,7 +778,7 @@ func TestAttributeBool(t *testing.T) {
 		},
 	} {
 		t.Run(tc.desc, func(t *testing.T) {
-			f, err := LoadData(filepath.Join("old", "BUILD.bazel"), "", []byte(tc.src))
+			f, err := LoadData("BUILD.bazel", "", []byte(tc.src))
 			if err != nil {
 				t.Fatal(err)
 			}


rule/rule.go
```diff
@@ -946,6 +946,28 @@ func (r *Rule) AttrStrings(key string) []string {
 	return strs
 }
 
+// AttrBool returns the value of the rule attribute key as a bool. If the
+// attribute is absent or not a boolean literal (True/False), it returns
+// defaultValue.
+func (r *Rule) AttrBool(key string, defaultValue bool) bool {
```
jayconrod: I agree we should have this method, but I don't think the `defaultValue` parameter is useful. The correct signature is ambiguous to me though. There are a few options:

1. `AttrPool(key string) bool`, return `true` if the attribute is `True`, otherwise `false` in all other cases. (Seems best to me since it's consistent with `AttrString` and `AttrStrings`.
1. `AttrBool(key string) bool`, return `true` if the attribute exists and is truthy, like `bool(x)` in Python.
1. `AttrBool(key string) bool`, panic if the attribute is not a `bool` (probably bad idea).
1. `AttrPool(key string) (bool, bool)`, return a second value indicating whether the attribute existed and was a `bool`. (more precise but complicated)

mateusz-olczyk: - 3 & 4 are not as usage-friendly as `AttrString` or `AttrStrings`.
- 2 is problematic in terms of consistency, because one can put a variable name or a function call as the RHS, which, of course, won't be evaluated by the simple `bazelbuild/buildtools/build` parser. So it's impossible to ideally mimic Python's `bool(expr)` call.
- Because of the arguments above, option No. 1 seems to be the best choice.

PR #2298, opened 2026-03-02
cmd/fetch_repo: fix dropped error

### LINKED ISSUE #2297: Dropped Error Variable
There is a dropped error variable in `cmd/fetch_repo`

alrs:
**What type of PR is this?**

Bug fix

/cmd/fetch_repo/

This fixes an err variable that was dropped and unhandled.

Fixes #2297.


tyler-french (approved):
In the future, it seems like it might be cleaner to just `return err` at the end of each case

PR #2301, opened 2026-03-11
Fix MergeDict() logic for empty lists

### LINKED ISSUE #2300: BUG: Empty lists inside "select" expressions are dropped during merging
# Description

When Gazelle merges rules, attributes that use `select()` with **explicit empty lists** in some branches can have those empty lists dropped in the merged result. Dropping them changes the semantics of the build.

In Bazel, an explicit empty list in a `select` branch (e.g. `"@platforms//os:linux": []`) means "for this configuration, emit no values." When the merger drops that branch, `"//conditions:default"` takes over for that platform — so a user who explicitly wanted no deps on that platform instead gets the default branch's value. Semantics are changed and build behavior can break.

# Expected behavior

After merging, any branch that was explicitly given an empty list in the source or destination should still appear in the merged `select` with an empty list, for example:

```starlark
select({
    "@platforms//os:linux": [],
    "//conditions:default": ["//lib"], # use this when NOT on linux
})
```

# Reference

From the [official Bazel documentation on configurable attributes](https://bazel.build/configure/attributes) and [build encyclopedia (select)](https://bazel.build/reference/be/functions#select):

> The built-in condition `//conditions:default` automatically matches **when nothing else does**.

So when a branch with an explicit empty list is dropped, that condition no longer applies and `//conditions:default` takes over for that configuration — the opposite of what the user expressed.

mateusz-olczyk:
**What type of PR is this?**

> Bug fix

> rule

`MergeDict()` won't drop explicit empty lists. It's needed to properly merge `select` expressions without breaking the semantics.

Fixes #2300

**Other notes for review**


mateusz-olczyk:
> I'm curious about the broader context of the change.

Is the description of #2300 not sufficient for you?

> I'm not authorized to retry tests, so try pushing an empty commit to re-run it.

👍👍👍

jayconrod:
> Is the description of https://github.com/bazel-contrib/bazel-gazelle/issues/2300 not sufficient for you?

In Go, we don't have a way to say that a dependency should be available for the default configuration, but not for a specific OS. There's no exclusion mechanism. Normally, we'd have a generic list and an OS-specific dictionary like this:

```
deps = ["//example/generic"] + select({
    "@platforms//os:linux": ["//example/linux"],
}),
```

If the generic library actually is not compatible with Linux, then we'd just have the OS-specific dictionary, with explicit entries for every OS.

```
deps = select({
    "@platforms//os:linux": ["//example/linux"],
    "@platforms//os:macos": ["//example/generic"],
    "@platforms//os:windows": ["//example/generic"],
})
```

So this is different than what I'm used to. It does still seem valuable and correct, but I'd also like to see it in context.

mateusz-olczyk:
I got your point. In C++ you can write something like:

```c++
#ifndef __linux__
#include "linux_incompatible.h"
#endif
```

So it should be resolved as:

```starlark
deps = select({
    "@platforms//os:linux": [],
    "//conditions:default": ["//linux_incompatible"],
})
```

I'm not sure if this makes any practical sense. Preprocessor conditions like this usually cover **all branches** with an appropriate platform-specific header. But at least theoretically, you can do something like this, and gazelle should not break it.

jayconrod:
In Go, you can express the same thing with a constraint like `//go:build !linux`.

But it doesn't really matter how the constraint expressed in the language. Gazelle's Go extension evaluates all the constraints and checks what dependencies are needed for each platform, then lists those as a concatenation of four things: a platform-neutral list, an OS-specific `select`, an arch-specific `select`, and an OS-and-arch-specific `select`. Technically we could put everything in the last one, but it saves some redundancy to split it out like this.

What you suggest could potentially eliminate more redundancy. But I think it's still possible to express platform-specific dependencies with `rule.PlatformStrings`, if we clean that up and make it less Go-specific.

PR #2303, opened 2026-03-12
bzlmod: allow non-default go.work file names

### LINKED ISSUE #2302: go_deps.from_file(go_work=...) rejects non-default go.work file names

0.47.0 (also present on master at baca6fb)

0.59.0

8.4.2

Yes.

linux/amd64 (not platform-specific; this is a Starlark name check)

### What did you do?

```starlark
go_deps.from_file(go_work = "//:go.bazel.work")
```

The workspace file loaded and its sum file (`go.bazel.work.sum`) used for
checksums, matching the `go` command's behavior.

### What did you see instead?

```
Error in fail: go_deps.from_file requires a 'go.work' file, not 'go.bazel.work'
```

---

The `go` command itself accepts any file name via `GOWORK` and writes
checksums to `<name>.sum`:

```console
$ GOWORK=$PWD/go.bazel.work go mod download
$ ls *.sum
go.bazel.work.sum
```

`sums_from_go_work` in `internal/bzlmod/go_mod.bzl` has two issues:
1. `_check_go_work_name` hard-fails on anything other than `go.work`
2. The sum file name is hardcoded as `"go.work.sum"` rather than derived
   from the work file name

Note that `go_work_from_label` (the other entry point, called on the same
label from `go_deps.bzl`) has no name check at all, so this is already
inconsistent.

**Use case:** a non-default name lets a repository maintain a Bazel-only
workspace that the `go` toolchain won't auto-discover. The `go` command
walks parent directories looking specifically for `go.work`, so a file
named `go.bazel.work` is invisible to native `go build` invocations. This
lets Bazel and native Go builds coexist during a gradual migration.

I have a fix ready and will open a PR.

findleyr:
The go command accepts any file as a workspace file via the GOWORK environment variable, and writes checksums to a sibling file with a .sum suffix (e.g. GOWORK=go.bazel.work produces go.bazel.work.sum).

go_deps.from_file(go_work = ...) currently rejects any name other than go.work, and hardcodes go.work.sum as the sum file name. This is stricter than the go command itself. Notably, go_work_from_label already accepts any name; only the sum-file path checks.

Remove the name check and derive the sum file name from the work file name.

A non-default name is useful when a repository wants a Bazel-only workspace that the go toolchain won't auto-discover (the toolchain walks up directories looking specifically for go.work), so native go builds remain unaffected by the Bazel workspace configuration.

**What type of PR is this?**

 Bug fix

go_repository

Fixes #2302

**Other notes for review**


findleyr:
CC @jayconrod @achew22

PR #2305, opened 2026-03-15
bzlmod: fix _noop arity for check_direct_dependencies = "off"

### LINKED ISSUE #2304: `check_direct_dependencies = "off"` fails with `_noop() accepts no more than 1 positional argument`

0.47.0 (also present on master at 40c2c73)

0.59.0

8.4.2

Yes — `_noop(_)` at go_deps.bzl:325 is unchanged on master.

linux/amd64 (but this is a pure Starlark bug, platform-independent)

### What did you do?

Set `go_deps.config(check_direct_dependencies = "off")` in a workspace where MVS resolves a higher version than a root `go.mod` requires (or where a `bazel_dep` version differs from the corresponding `go.mod` version).

Silent module rresolution.

### What did you see instead?

```
  ERROR: .../go_deps.bzl:325:5: Traceback (most recent call last):
        File ".../go_deps.bzl", line 660, column 36, in _go_deps_imp
                outdated_direct_dep_printe
        File ".../go_deps.bzl", line 325, column 5, in _noo
                def _noop(_
  Error: _noop() accepts no more than 1 positional argument but got 5
```

`outdated_direct_dep_printer` defaults to `print` (variadic). Both call sites (lines 636 and 660) pass multiple positional args. `"off"` swaps in `_noop(_)`, which takes exactly one.

Fix: `def _noop(*_):`

findleyr:
outdated_direct_dep_printer defaults to print (variadic). Both call sites (go_deps.bzl:636, :660) pass multiple positional arguments. When check_direct_dependencies = "off" swaps in _noop, the single-argument signature fails with:

  Error: _noop() accepts no more than 1 positional argument but got 5

tests/bcr/go_work already has a circl bazel_dep (1.3.8) vs pkg/go.mod (1.6.1) version skew that fires the :636 path; setting check_direct_dependencies = "off" there makes bazel build //... fail without this fix.

Fixes #2304

### DIFF
--- a/internal/bzlmod/non_module_deps.bzl
+++ b/internal/bzlmod/non_module_deps.bzl
@@ -40,6 +40,12 @@ load(
 visibility("//")
 
 def _non_module_deps_impl(module_ctx):
+    gazelle_version = ""
+    for module in module_ctx.modules:
+        if module.name == "gazelle":
+            gazelle_version = module.version
+            break
+
     go_repository_cache(
         name = "bazel_gazelle_go_repository_cache",
         # Label.repo_name is always a canonical name, so use a canonical label.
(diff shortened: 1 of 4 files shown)


findleyr:
CC @jayconrod @achew22

findleyr:
@jayconrod @fmeum do you want me to drop the test change for this before merging?

jayconrod:
Let's keep the test change. Lacking a better place to put a regression test for now, some coverage is better than none.

fmeum:
I don't see a better way either. I started building a unit testing framework for module extensions at some point, but that turned out to be surprisingly tricky. I might revisit this at some point though.

PR #2307, opened 2026-03-15
Add -version flag to print the current version

### LINKED ISSUE #2288: `gazelle version` command

jayconrod:
**What type of PR is this?**

> Feature

> v2/cmd/gazelle/update

When set, Gazelle prints its version to stdout, then exits without doing anything else.

This flag is recognized by the fix/update commands, so usage can be 'gazelle -version' or 'gazelle update -version' but not 'gazelle update-repos -version'. It feels awkward for v1, but I didn't want to add a new 'version' subcommand because v2 doesn't have subcommands.

For #2272
Fixes #2288

**Other notes for review**

### DIFF
--- a/internal/bzlmod/non_module_deps.bzl
+++ b/internal/bzlmod/non_module_deps.bzl
@@ -40,6 +40,12 @@ load(
 visibility("//")
 
 def _non_module_deps_impl(module_ctx):
+    gazelle_version = ""
+    for module in module_ctx.modules:
+        if module.name == "gazelle":
+            gazelle_version = module.version
+            break
+
     go_repository_cache(
         name = "bazel_gazelle_go_repository_cache",
         # Label.repo_name is always a canonical name, so use a canonical label.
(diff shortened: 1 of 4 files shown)


jayconrod:
I couldn't figure out a good way to write a test for this. Hopefully it's small enough and low enough risk that we can verify by inspection. Let me know if you have ideas though.

I tried building a test module that depends on Gazelle with `local_path_override`, but no version is set when that override is used. The version from `bazel_dep` doesn't show up anywhere.

I could write a test that passes `-version` and verifies *something* is printed. But that's not very useful. The expected version and mode would be unknown to the test.

Issue #2308, opened 2026-03-16
Add gazelle:include to re-add paths after an earlier gazelle:exclude


nickcoai:
gazelle:exclude is currently one-way: once a path or subtree is excluded, there is no directive to selectively add part of it back.

This makes it hard to express common cases like:
- exclude a large generated or vendored subtree, but keep one checked-in source file
- exclude a directory by default, but re-enable one subdirectory
- keep a narrow allowlist inside an otherwise excluded tree

A new directive, gazelle:include, would solve this by letting users re-add a path matched after an earlier gazelle:exclude.

### Proposed semantics

Add support for:
`gazelle:include pattern`
with ordered behavior shared with gazelle:exclude:

- directives are evaluated in source order
- later matching directives win
- exclude removes a matching path from consideration
- include re-adds a matching path, even if an earlier exclude matched it

For directories, include should also reopen traversal as needed to reach matching descendants under an excluded parent.

### Example:
```
# gazelle:exclude blocked
# gazelle:include blocked/sub/keep.go
# gazelle:include blocked/allowed
```

repo:
```
blocked/sub/keep.go
blocked/sub/drop.go
blocked/allowed/file.go
blocked/other/file.go
```

Expected behavior:
```
blocked/sub/keep.go is included
blocked/allowed/... is included
blocked/sub/drop.go stays excluded
blocked/other/... stays excluded
```

nickcoai:
Draft implementation of the proposal: https://github.com/bazel-contrib/bazel-gazelle/pull/2309

jayconrod:
I'm kind of skeptical of this feature. I've worked with tools that support both inclusion and exclusion like this, and I often ended up with an incredible mess. Someone would give me a Perforce client for example with 30+ lines, and it would be really difficult to tell whether a directory was going to be synced or not. I'd generally prefer a single flat list of exclusions, even if it's significantly longer.

Could you give a more detailed scenario where this feature would be useful? Your example common cases are pretty similar, and not generally something I've run into with Go, Java, or protobuf. Maybe it's a problem for other languages though and I'm not understanding the pain point here, so I'd like to hear more.

A possible simpler alternative solution to consider: either `gazelle:include` or `gazelle:exclude` are allowed, but not both. When `gazelle:include` is used, everything that doesn't match is excluded.

nickcoai:
[...]

PR #2309, opened 2026-03-16
Add gazelle:include to counter exclude as an option for sub-inclusion

nickcoai:
Feature implementation of https://github.com/bazel-contrib/bazel-gazelle/issues/2308 

**What type of PR is this?**

Feature

cmd/gazelle

gazelle:exclude is currently one-way: once a path or subtree is excluded, there is no directive to selectively add part of it back.

https://github.com/bazel-contrib/bazel-gazelle/issues/2308


jayconrod:
Holding on review for design discussion in #2308.

From a quick look, I'm concerned about time complexity. `shouldTraverseDir` is recursive in the case where a directory is excluded but contains included subdirectories. It's called for every directory in the tree (more than once since it's also called on subdirectories), so this makes directory traversal O(n^2). That's going to be too inefficient for large repos, and I don't think that's essential to solve the problem.

nickcoai:
Yeah very happy to iterate on this, figured it's important to settle the design issue first.

PR #2311, opened 2026-03-22
Move and forward basic packages to v2

jayconrod:
**What type of PR is this?**

> Other

- flag
- label
- merger
- pathtools
- rule
- testools

These packages won't require substantial changes in v2, so we don't need two separate sets of definitions. v2 packages should eventually not import anything from v1, so v2 is the source of truth. Consequently, the v1 packages are now just aliases for v2.

For #2272

**Other notes for review**

This PR was generated with a small program, https://github.com/jayconrod/bazel-gazelle/blob/v2/tools/forward/forward.go. It's throw-away code, so I'm not planning to check it in. No manual changes other than re-running Gazelle.


fmeum:
Could we use https://go.dev/blog/inliner to simplify this migration for users (and possibly us)? The deprecation comments are useful, but don't seem to be particularly machine readable.

jayconrod:
Yes, good idea! Done.

fmeum (approved):

tyler-french (approved):

PR #2312, opened 2026-03-24
Bugfix: embed resolution skipping subdirectory having other go files.

### LINKED ISSUE #1392: embedsrcs does not work for subdirs that contains .go files

v0.28.0

not relevant

not relevant

yes

not relevant

### What did you do?

main.go
```go
package main

import _ "embed"

//go:embed subdir/*
```
subdir/whaterver.go

subdir/whatever.go is in embedsrcs for main.go

### What did you see instead?

Error: `pattern subdir/*: matched no files` and no embedsrcs

wonderyl:
**What type of PR is this?**

Bug fix

language/go

It's a common pattern to use a dummy.go in a folder that contains the artifacts for embedding. E.g.
https://github.com/kcl-lang/lib/tree/main/go/lib/linux-amd64 Previously, Gazelle will skip the folder because it will generate a separate BUILD.bazel for the sub-folder. Which prevents //go:embed to embed any files in the sub-folders.
This PR fixes it.

Fixes #1392

**Other notes for review**


jayconrod:
Generally, a Bazel target can't depend on files in a different Bazel package unless those files are named in the other build file and are made visible. `exports_files` is generally the way to do that.

How does this PR get around that problem? It looks like it allows direct reference from `embedsrcs` but doesn't add `exports_files`.

wonderyl:
> Generally, a Bazel target can't depend on files in a different Bazel package unless those files are named in the other build file and are made visible. `exports_files` is generally the way to do that.
> 
> How does this PR get around that problem? It looks like it allows direct reference from `embedsrcs` but doesn't add `exports_files`.

Thanks, @jayconrod, for the feedback. 

My understanding is, `go:embed` only takes relative path strings, it doesn't recognize package boundry.

Based on https://github.com/bazel-contrib/rules_go/blob/master/docs/go/core/rules.md#go_library-embedsrcs
> All files must be in the same logical directory or a subdirectory as source files. 

It further restricts the embedded artifacts to be within the same directory or subdirectory. 

This rule implies the embedding go file already has visibility of the embedded files.

On the other hand, it seems to me `exports_files` is mostly used to change visibility of files, [source](https://github.com/bazel-contrib/bazel-docs/blob/8c78f341e6298114e5f54ad4c947da1465bbd416/concepts/visibility.mdx#source-file-target-visibility-source-file-target-visibility).
In this case, since the embeded files are already visible, there is really no need to use `exports_files` explictly.

jayconrod:
You're right that `go:embed` in plain Go (without Bazel) does not care about package boundaries. On module boundaries matter there: you can't embed a file in another module. But that doesn't matter for this purpose.

Bazel is quite picky about package boundaries (and here I mean Bazel packages, defined by BUILD files including subdirectories down to the next BUILD file).

For example, let's say we want to embed `data/message.txt`.

- If `data` does not have a `BUILD` file, then `embedsrcs = ["data/message.txt"]` works because the file is part of the same Bazel package, even though it's in a subdirectory.
- If `data` has a `BUILD` file and the file is exported with `exports_files`, then `embedsrcs = ["//data:message.txt"]` works instead. You must use an absolute label, not a relative path.
- If `data` has a `BUILD` file and the file is not exported but is mentioned somewhere, then it is [implicitly exported](https://bazel.build/concepts/visibility#source-file-target-visibility). So `embedsrcs = ["//data:message.txt"]` might work unless `--incompatible_no_implicit_file_export` is enabled (and we should assume it might be enabled).
- If `data` has a `BUILD`  file and the file is neither exported nor mentioned, it cannot be embedded.

The last case is what I'm concerned about. This would be a problem in the example you linked because `dummy.go` doesn't embed anything. Gazelle would generate a `go_library` in that directory and would not reference any data files. 

This is a somewhat hard problem to solve Gazelle because it's hard to predict, when operating in one directory, whether Gazelle will write BUILD files in other directories. In this case, you'd probably want Gazelle to write `exports_files` in the directory with embeddable files, but there's no clean way to do that in `GenerateRules`. 

Here's the concrete example I tested.

```
-- MODULE.bazel --
bazel_dep(name = "rules_go", version = "0.60.0")
bazel_dep(name = "gazelle", version = "0.48.0")
local_path_override(
    module_name = "gazelle",
    path = "../bazel-gazelle",
)

go_deps = use_extension("@gazelle//extensions.md", "go_deps")
go_deps.from_file(go_mod = "go.mod")
-- go.mod --
module example.com/m

go 1.25.0
-- BUILD --
load("@gazelle//:def.bzl", "gazelle")
load("@rules_go//go:def.bzl", "go_binary", "go_library")

# gazelle:prefix example.com/m
gazelle(name = "gazelle")

go_library(
    name = "m_lib",
    srcs = ["embed.go"],
    embedsrcs = ["data/message.txt"],
    importpath = "example.com/m",
)

go_binary(
    name = "m",
    embed = [":m_lib"],
)
-- embed.go --
package main

import (
	_ "embed"
	"fmt"
)

//go:embed data/message.txt
var message string

func main() {
	fmt.Printf("message: %s\n", message)
}
-- data/BUILD.bazel --
load("@rules_go//go:def.bzl", "go_library")

go_library(
    name = "data",
    srcs = ["dummy.go"],
    importpath = "example.com/m/data",
)
-- data/dummy.go --
package data
-- data/message.txt --
hello world
```

wonderyl:
@jayconrod thanks for the comments. 
I have made a few new commits. Now it adds `exports_files()` in sub-packages and use the bazel labels to reference them.
PTAL

jayconrod (commented):
I see how this can work, but I'm not comfortable with adding `language.GenerateResult.ExternalGen`. That's a major change to the language interface.

Gazelle is often invoked in specific directories. It *may read* from other directories (and it *must read* from parent directories), but it *must not* write to directories, other than where it's invoked. `ExternalGen` breaks that rule.

`ExternalGen` also ignores local configuration. For example, if the `go` extension is enabled in the directory with the sources containing the `//go:embed` directive but disabled in the directory with the embedded files, we shouldn't write `export_files` there. `gazelle:exclude` directives in the child directory should also influence this.

I think a new design is needed here. `Configure` is called in pre-order. So we could move package parsing (the stuff in `package.go`) into that step from `GenerateRules`. We'd need to save the parsed package info so that it's available later in `GenerateRules`. This may increase memory usage.

When we get to `GenerateRules`, we'd know which files in the current directory were matched by `//go:embed` directives in parent directories, and we could generate `exports_files` declarations appropriately.

wonderyl:
@jayconrod PTAL, I made a new commit based on your feedback.

wonderyl:
@jayconrod, I noticed some ci checks are failing, but I don't know how to reproduce them locally. Could you give me some pointers, please.

config/config.go
```diff
@@ -124,6 +124,9 @@ type Config struct {
 	// to the apparent name (repo_name) specified in the MODULE.bazel file. It
 	// returns the empty string if the module is not found.
 	ModuleToApparentName func(string) string
+
+	// DirInfo holds directory content information for the current directory.
+	DirInfo DirInfo
```
jayconrod, line 129: Please don't modify anything outside of `language/go` for this change, other than tests. This information is already available through `walk.GetDirInfo`.

language/go/testdata/embedsrcs/deep/BUILD.want
jayconrod, line 1: I'd prefer if you could add new integration tests as subdirectories of `//tests` rather than adding new things here. The structure's pretty similar, but it provides better, more realistic coverage instead of just focusing on `GenerateRules`.

language/go/lang.go
```diff
@@ -62,10 +62,18 @@ type goLang struct {
 	// Go code. If the value is false, it means the directory does not contain
 	// buildable Go code, but it has a subdir which does.
 	goPkgRels map[string]bool
+
+	// cer encapsulates the pre-resolved embed sources and their
+	// Bazel labels. Populated during Configure (pre-order), consumed during
+	// GenerateRules (post-order).
+	cer *cachedEmbedResolver
```
jayconrod, line 69: Needs a more descriptive name, expand to  `cachedEmbedResolver` or `embedResolver`. In places where this is used, it's not obvious what it is.

jayconrod (commented):
Sorry, this is a large PR, and I won't have time for a full review for at least a couple days. Only a couple structural comments for now.

It would help if you can resolve the test failures. You can reproduce these by changing to the `tests/bcr/go_mod` directory and running `bazel run //:gazelle` then `bazel build //...`. These failures mean that an external dependency with build files generated by `go_repository` was broken. I'd guess that a new BUILD file was generated but labels are not correctly referring to files within its package.

language/go/embed.go
```diff
@@ -199,6 +174,130 @@ func (er *embedResolver) resolve(embed fileEmbed) (list []string, err error) {
...
+		// need exports_files for its own embed targets.
+		if shallowestEmbedPackage == rel {
+			continue
+		}
+		fileRelToPackage := strings.TrimPrefix(embedRel, prefix)
+
+		// Delete the file from resolvedEmbeds to indicate that we've claimed it.
+		delete(r.resolvedEmbeds, embedRel)
+		exportFiles = append(exportFiles, fileRelToPackage)
+		r.embedSrcLabels[embedRel] = "//" + rel + ":" + fileRelToPackage
```
jayconrod, line 296: Generally, use the `label` package to construct and parse labels, don't concatenate strings for this.

language/go/fileinfo.go
```diff
@@ -41,6 +41,7 @@ import (
 type fileInfo struct {
 	path string
 	name string
+	rel string
```
jayconrod, line 44: Why is the new field needed? This requires extra memory and updating a lot of call sites.

language/go/embed.go
```diff
@@ -199,6 +174,130 @@ func (er *embedResolver) resolve(embed fileEmbed) (list []string, err error) {
...
+func (r *cachedEmbedResolver) addEmbedSrc(rel, fileRel, embedRel string) {
+	// Only record the shallowest originator of an embed source. So that it knows when to stop exporting embedded files.
+	// A embeded file is exported, if it's not exprted by a sub-package and there's parent package that embeds it.
+	if _, exists := r.resolvedEmbeds[embedRel]; !exists {
+		r.resolvedEmbeds[embedRel] = rel
+	}
+	r.relToEmbedSrcs[fileRel] = append(r.relToEmbedSrcs[fileRel], embedRel)
+}
+
+// discoverCrossPkgEmbeds reads Go files in the directory, resolves //go:embed
```
jayconrod, line 240: Documented name doesn't match method name.

language/go/embed.go
```diff
@@ -199,6 +174,130 @@ func (er *embedResolver) resolve(embed fileEmbed) (list []string, err error) {
...
+// patterns, and stores repo-root-relative paths for files in subdirectories
+// into resolvedEmbeds. This is called during Configure (pre-order), so
+// when child directories' GenerateRules runs (post-order), they can check
+// whether they should generate exports_files rules.
+func (r *cachedEmbedResolver) resolveDir(c *config.Config, rel string, di config.DirInfo) {
```
jayconrod, line 245: So this is called by `Configure` in every visited directory, and it does a fair amount of I/O, reading each .go file. The result of that I/O is not cached. This may also consider files that aren't used in rules (e.g., wrong package name, excluded by build tags).

(I'll have some thoughts overall, just pointing out the issue here).

jayconrod (commented):
Reviewed a bit further. I think this is still pretty far from what I can merge. My review bandwidth is pretty limited. We might need to shelve this one for a while.

I'd like the change to be simpler. Rather than caching all the embed info, I wonder if `Configure` can cache only the `go:embed` patterns that appear in source files. Then when `GenerateRules` is called in a directory, it can check if any of the files in that directory match embed patterns from parent directories, and then generate the corresponding `exports_files`.

Unfortunately, this is necessarily still a complicated change. I'd like to minimize redundant work between `Configure` and `GenerateRules` and to keep them consistent. A substantial amount of code needs to be extracted from `GenerateRules` into a new method and called from `Configure`. Basically anything that reads file metadata. I think we can skip the filtering and package grouping, and consider embeds from all .go files, even if we don't end up generating rules for them.

With that in mind, I'd also like this change to be as small and simple as possible. It should not modify other packages outside of `language/go`.

wonderyl:
> Reviewed a bit further. I think this is still pretty far from what I can merge. My review bandwidth is pretty limited. We might need to shelve this one for a while.
> 
> I'd like the change to be simpler. Rather than caching all the embed info, I wonder if `Configure` can cache only the `go:embed` patterns that appear in source files. Then when `GenerateRules` is called in a directory, it can check if any of the files in that directory match embed patterns from parent directories, and then generate the corresponding `exports_files`.
> 
> Unfortunately, this is necessarily still a complicated change. I'd like to minimize redundant work between `Configure` and `GenerateRules` and to keep them consistent. A substantial amount of code needs to be extracted from `GenerateRules` into a new method and called from `Configure`. Basically anything that reads file metadata. I think we can skip the filtering and package grouping, and consider embeds from all .go files, even if we don't end up generating rules for them.
> 
> With that in mind, I'd also like this change to be as small and simple as possible. It should not modify other packages outside of `language/go`.

Thanks for the review. I completely understand that the PR is pretty complex. It took me a few rounds of refactorings to wrap my head around.

There are two places where `go:embed` directives need to expand to a list of files. The first one is in when generating `embedsrcs`, in the parent folder, the second one is when generating `embeds_files`, in the child folder. Your proposal of remembering the `go:embed` directives would mean each directive will be resolved at least twice, more times when a directive embeds files from more than one folder. On the contrary, my PR only resolves each directive once in `Configure` and stores the results so that `GenerateRules` doesn't need to resolve it again. In addition, this approach won't generate addition disk I/O, compared to the original implementation. The original implementation calls `EmbedResolver.resolve` once in `GenerateRules`, while my PR calls `EmbedResolver.resolve` once in `Configure`.

Unless the concern is memory consumption, then the "remembering go:embed" approach would essentially trade disk I/O and CPU for less memory consumption.

I don't really mind which way I implement. I care more about fixing the bug. Please let me know which way you prefer.

jayconrod:
I'm sorry I haven't gotten back to this yet. It's on my list, but I need to give it some thought, and it will take time.

(Apologies, it's a busy time right now. I'm preparing for a conference, and my team has sent me 17 large PRs to review today, so bandwidth is limited.)

language/go/embed.go
```diff
@@ -199,6 +174,130 @@ func (er *embedResolver) resolve(embed fileEmbed) (list []string, err error) {
...
+	// child packages remove entries they claim via exports_files.
+	// Key: repo-root-relative path (e.g., "embedsrcs/m_go/m.go")
+	// Value: rel of the directory that originated the embed
+	resolvedEmbeds map[string]string
+	// relToEmbedSrcs maps a Go file's repo-root-relative path to the
+	// repo-root-relative paths of its resolved embed sources.
+	relToEmbedSrcs map[string][]string
+	// embedSrcLabels maps repo-root-relative embed source paths to Bazel
+	// labels for cross-package access.
+	embedSrcLabels map[string]string
```
jayconrod: Let's use `label.Label` for everything that's a label. I think that's values in `relToEmbedSrcs`, `embedSrcLabels`, and a few of the method returns.

language/go/embed.go
```diff
@@ -199,6 +174,130 @@ func (er *embedResolver) resolve(embed fileEmbed) (list []string, err error) {
...
+			result = append(result, relToDir)
+		}
+	}
+	return result
+}
+
+// addEmbedSrc records that a resolved embed source (embedRel, repo-root-relative)
+// originated from directory rel and is referenced by the Go file at fileRel.
+// The shallowest originator is preserved in resolvedEmbeds.
+func (r *cachedEmbedResolver) addEmbedSrc(rel, fileRel, embedRel string) {
```
jayconrod: I'm a little confused by the type signature. `rel` is the directory above `fileRel`? Could we use `path.Dir` to get it instead of passing it as a parameter?

language/go/embed.go
```diff
@@ -199,6 +174,130 @@ func (er *embedResolver) resolve(embed fileEmbed) (list []string, err error) {
...
+	// Only record the shallowest originator of an embed source. So that it knows when to stop exporting embedded files.
+	// A embeded file is exported, if it's not exprted by a sub-package and there's parent package that embeds it.
+	if _, exists := r.resolvedEmbeds[embedRel]; !exists {
+		r.resolvedEmbeds[embedRel] = rel
```
jayconrod: How do we know it's the shallowest? I think it's implied that this is called by `Configure` in top-down order, but let's make that explicit in comment or code.

language/go/embed.go
```diff
@@ -199,6 +174,130 @@ func (er *embedResolver) resolve(embed fileEmbed) (list []string, err error) {
...
+		}
+	}
+	return result
+}
+
+// addEmbedSrc records that a resolved embed source (embedRel, repo-root-relative)
+// originated from directory rel and is referenced by the Go file at fileRel.
+// The shallowest originator is preserved in resolvedEmbeds.
+func (r *cachedEmbedResolver) addEmbedSrc(rel, fileRel, embedRel string) {
+	// Only record the shallowest originator of an embed source. So that it knows when to stop exporting embedded files.
```
jayconrod: What does "when to stop exporting" mean? Is it for visibility? I think it's actually fine if we just use public visibility, the default.

EDIT: I see in `claimExportFiles` this is used to determine whether the embedding is from the same directory, in which case, `export_files` isn't needed. But it doesn't look like the path is used other than that. Maybe we could just omit entries for embedding within the same directory.

language/go/embed.go
```diff
@@ -199,6 +174,130 @@ func (er *embedResolver) resolve(embed fileEmbed) (list []string, err error) {
...
+// (rel) that were originated by a different (ancestor) package. It removes
+// claimed entries from resolvedEmbeds, records their Bazel labels in
+// embedSrcLabels, and returns the package-relative file paths to export.
+func (r *cachedEmbedResolver) claimExportFiles(rel string) []string {
+	prefix := rel + "/"
+	if rel == "" {
+		prefix = ""
+	}
+	var exportFiles []string
+	for embedRel, shallowestEmbedPackage := range r.resolvedEmbeds {
```
jayconrod: This could potentially include a large number of files that aren't relevant to this directory. Could we index embedded files by directory so we don't need to iterate over all files checking the prefix?

language/go/embed.go
```diff
@@ -199,6 +174,130 @@ func (er *embedResolver) resolve(embed fileEmbed) (list []string, err error) {
...
+		}
+		// Skip files that belong to the embedder's own package — it doesn't
+		// need exports_files for its own embed targets.
+		if shallowestEmbedPackage == rel {
+			continue
+		}
+		fileRelToPackage := strings.TrimPrefix(embedRel, prefix)
+
+		// Delete the file from resolvedEmbeds to indicate that we've claimed it.
+		delete(r.resolvedEmbeds, embedRel)
```
jayconrod: I'm not sure I understand what it means to claim a file. I think it should be possible for a file to be embedded from more than one parent directory ("a/a.go" and "a/b/b.go" might both embed "a/b/c/c.txt").

language/go/embed.go
```diff
@@ -199,6 +174,130 @@ func (er *embedResolver) resolve(embed fileEmbed) (list []string, err error) {
...
+
+	er := newEmbedResolver(dir, di.Subdirs, di.RegularFiles, di.GenFiles)
+
+	// Parse Go files for embed directives.
+	for _, name := range di.RegularFiles {
+		if !strings.HasSuffix(name, ".go") {
+			continue
+		}
+
+		info := goFileInfo(filepath.Join(dir, name), rel, path.Join(rel, name))
```
jayconrod: This is the extra I/O I'm concerned about. This opens and parses the file. We need to do that again later in `GenerateRules` to find the package name, imports, and build constraints.

It's not a huge deal: the file will be in cache memory anyway, so it's mostly the system calls that are expensive. But let's add a TODO with an issue number here so we remember to fix it later.

jayconrod (commented):
Thanks for your patience. I've read through this again more carefully, and I think I understand it better.

> There are two places where go:embed directives need to expand to a list of files. The first one is in when generating embedsrcs, in the parent folder, the second one is when generating embeds_files, in the child folder. Your proposal of remembering the go:embed directives would mean each directive will be resolved at least twice, more times when a directive embeds files from more than one folder. On the contrary, my PR only resolves each directive once in Configure and stores the results so that GenerateRules doesn't need to resolve it again. In addition, this approach won't generate addition disk I/O, compared to the original implementation. The original implementation calls EmbedResolver.resolve once in GenerateRules, while my PR calls EmbedResolver.resolve once in Configure.
> 
> Unless the concern is memory consumption, then the "remembering go:embed" approach would essentially trade disk I/O and CPU for less memory consumption.

I was thinking that we'd record `go:embed` directives in `Configure`, then in `GenerateRules` for each file, check each file against each `go:embed` directive in any parent directory. If it matches in the current directory, `embedsrcs`. If it matches in any parent directory, `export_files`. So I think each file would be matched against each embed once. There's probably some optimization available there, but CPU time is relatively cheap compared with memory and especially disk.

That said, I think your approach here works fine. I have a few comments on the details, but I think we can merge once those are fixed.

language/go/embed.go
wonderyl: Consider this case. The folder's structure is 
```
a
|- foo.go
|- b
   |- bar.go
   |- c
      |- d
         |- data.txt
```
### Case 1
Only `bar.go` embeds `data.txt`.
Since both c and d are not go packages, data.txt is not exported. In the package of `b`, even though `bar.go` embeds `data.txt`, it references `data.txt` using relative path, `c/d/data.txt` rather than label. 
### Case 2
The file structure is the same, but both `foo.go` and `bar.go` embeds `data.txt`.
In this case, package b still reference `data.txt` using relative path. But since `foo.go` also embeds it and it's not exported yet, package b needs to export `data.txt`. 

The path is used to distinguish the 2 cases.
In case 1, the shallowest embedding source is `a/b`, then b doesn't need to export `data.txt`.
In case 2, the shallowest embedding source is `a`, then b needs to export `data.txt`.

language/go/embed.go
wonderyl: This extra `goFileInfo` is inevitable even in the alternative solution you proposed. You proposed to remember the embeds during Configure, this will require this extra `goFileInfo` to get the embeds as well.

jayconrod: The result can be cached to avoid that.

Don't worry about that for this PR though, it's big enough already. Let's just leave a TODO.

language/go/embed.go
wonderyl: One file is only exported once. Claiming a file means finding the right package to export it, then delete it from the cache to prevent it from exported again.
e.g. a file is in `a/b/c/data.txt` all a, b, c may export it, depend on which one is a go package. e.g. if b folder has a go file, while c doesn't data.txt will be exported in b package.

The `more than one files embedding the same file` case is handled by `shallowestEmbedPackage`

language/go/testdata/embedsrcs/deep/BUILD.want
wonderyl: done

config/config.go
wonderyl: done

language/go/lang.go
wonderyl: done

language/go/embed.go
wonderyl: done

language/go/embed.go
wonderyl: done

language/go/fileinfo.go
wonderyl: removed

language/go/embed.go
wonderyl: done

language/go/embed.go
wonderyl: done

language/go/embed.go
wonderyl: done

language/go/embed.go
wonderyl: Used a Radix instead.

jayconrod:
I apologize in advance, it will take some time to review this. I'm traveling to a conference this week.

wonderyl:
> I apologize in advance, it will take some time to review this. I'm traveling to a conference this week.

Hi Jay, I really appreciate the heads up, take your time. It took me a while to complete the update as well. 

The PR touches 100 files now, however most of them are in the /tests folder. The main logic is pretty much the same as before.

Hope you have a good trip.

PR #2314, opened 2026-03-26
Add a `directive_file` directive

shs96c:

When managing a large monorepo (or possibly a small, chaotic repo) it's possible for the number of directives in build files (especially the top-level build file) to become unmanageable, obscuring the actual content of the build file in a mess of directives.

This adds a new `directive_file`, which is treated as if the directives in it are inline with the build file. Recursion is forbidden (so you can't use the `directive_file` directive in a directive file) to simplify processing, both by machine and by people.

**What type of PR is this?**

Feature

`rule` and `walk`

You've disabled issues in the GitHub repo....


walk/dirinfo.go
```diff
@@ -96,6 +107,59 @@ func (w *walker) loadDirInfo(rel string) (DirInfo, error) {
...
+// directives parsed from that file. Directives from the external file are
+// inserted at the position of the directive_file entry, preserving ordering so
+// that later inline directives can override earlier external ones.
+//
+// Paths in directive_file values are resolved relative to the directory
+// containing the BUILD file (i.e., the package directory). The loaded
+// directives behave as if they were written inline in that BUILD file.
+// Directive files may not themselves contain directive_file entries (no
+// recursion); any such entries are reported as errors.
+func expandDirectiveFiles(f *rule.File, repoRoot string) []error {
```
jayconrod: Return `error`, not `[]error`: use `errors.Join(...)` to return a list of errors as a single `error`.

walk/dirinfo.go
```diff
@@ -96,6 +107,59 @@ func (w *walker) loadDirInfo(rel string) (DirInfo, error) {
...
+	var errs []error
+	for _, d := range f.Directives {
+		if d.Key != "directive_file" {
+			expanded = append(expanded, d)
+			continue
+		}
+		filePath := filepath.Join(pkgDir, filepath.FromSlash(d.Value))
+		loaded, err := rule.ParseDirectivesFromFile(filePath)
+		if err != nil {
+			log.Printf("%s: %v", f.Path, err)
```
jayconrod: Remove the `log.Print` calls from here: let the caller decide what to do with the error. I think `Walk2` or whatever calls it will log these already.

(Gazelle has a lot of `log.Print` calls as an unfortunate consequence of not having good error handling built into the extension API. I'm trying to fix that in v2. Ideally the library code shouldn't log anything.)

jayconrod (commented):
Thanks for sending this. I'd like to get this into a new release tomorrow.

We've needed something like this for a while. I've been holding my breath for more sophisticated support for offline indexing, or the ability to use a separate indexing service, but that hasn't materialized yet. Better to have something simple for now.

jayconrod (approved):

Issue #2315, opened 2026-03-26
Spurious `No sum for github.com/bazel-contrib/bazel-gazelle/v2@2.0.0-1 found` warnings


rdesgroppes:
### What version of gazelle are you using?

0.48.0

### What version of rules_go are you using?

0.60.0

### What version of Bazel are you using?

9.0.1

### Does this issue reproduce with the latest releases of all the above?

Yes.

### What operating system and processor architecture are you using?

Linux x86_64 (also reproducible on macOS and Windows)

### What did you do?

Used `gazelle` as a non-root Bazel module (via `bazel_dep`) in various projects, then ran any Bazel command that triggers the `go_deps` extension.

### What did you expect to see?

No warnings.

### What did you see instead?

In any fresh Bazel context:
```
DEBUG: /path/to/external/gazelle+/internal/bzlmod/go_deps.bzl:819:18: No sum for github.com/bazel-contrib/bazel-gazelle/v2@2.0.0-1 found, run bazel run @rules_go//go -- mod tidy to generate it
```

Running `bazel run @rules_go//go -- mod tidy` from the workspace/repo root does not help.

jayconrod:
I've seen this warning developing Gazelle, which seemed fine, but didn't realize it was visible outside Gazelle too.

#2317 should fix it.

Ahajha:
I'm still seeing this after updating to 0.49.0, it's even noisier now
```
Mismatch between versions requested for Go module github.com/bazel-contrib/bazel-gazelle/v2:

  bazel_dep version (MODULE.bazel): 0.49.0 (as "gazelle")
  Go module version (go.mod):       2.0.0-1
 
Either ensure that you have

  bazel_dep(name = "gazelle", version = "2.0.0-1")

in your MODULE.bazel file or downgrade the Go module version via

  bazel run @rules_go//go -- get github.com/bazel-contrib/bazel-gazelle/v2@v0.49.0
```
Similarly, running the suggested command does nothing.

jayconrod:
Sorry for the churn. Hopefully #2325 should fix this. I'll tag a patch release when that lands.

Ahajha:
Looks fixed now, thanks @jayconrod!

jayconrod:
Thanks for confirming! Sorry again for the disruption.

Issue #2242, opened 2025-11-26
handling local replace statements with go.mods in a chained dependency


SaurabhDRao:
I noticed an issue with the replace statement resolution when the go.mod are in a chained dependency format

```a/go.mod -> b/go.mod -> c/go.mod```

with the contents being

**a/go.mod**
```
replace b => ../b
```

**b/go.mod**
```
replace c => ../c
```

**c/go.mod**
No replace statements, fully self contained.

Now when building `b/go.mod` with bazel, it is able to resolve `c` and build.
However, when building `a/go.mod`, its able to resolve `b/go.mod`, but its unable to resolve `c/go.mod` correctly. Its trying to resolve `c` from bazel's output base instead of from the actual code workspace directory.

Sample error:
```
exit status 1 in /home/myuser/.cache/bazel/_bazel_myuser/2992cb5291dc9ca5aac988b11f7aeea3/external/gazelle++go_deps+b.
Output: go: c@v0.0.0-00010101000000-000000000000 (replaced by ../c): reading ../c/go.mod:
open /home/myuser/.cache/bazel/_bazel_myuser/2992cb5291dc9ca5aac988b11f7aeea3/c/go.mod: no such file or directory
```

SaurabhDRao:
i dont think there is a reason to resolve the replace statement in the `b/go.mod` at all since if `c/go.mod` is needed to build `a/go.mod`, `go` asks to explicitly declare it in `a/go.mod`.
since its not declared, it can be ignored

jayconrod:
Could you provide an example project (as repo, zip file, or txtar archive)?

Which of these is the main module, if any? `replace` directives outside the main module should be ignored.

SaurabhDRao:
I had trouble replicating this with a minimal set of code. However after checking, I see that this issue seems to have been fixed in `v0.46.0`.

I had been using `v0.45.0` in my code base where I had seen this issue and I was also able to replicate it with `v0.45.0`.

This seems to be the fix: https://github.com/bazel-contrib/bazel-gazelle/pull/2178

So this issues seems to be good to close.

Issue #1775, opened 2024-04-05
go.mod FilePath ReplaceDirective is missing when using the go-deps bzlmod extension


stefanpenner:
### What version of gazelle are you using?

<= 0.36.0

### What version of rules_go are you using?

<= 0.46.0

### What version of Bazel are you using?

<= 7.1.0

### Does this issue reproduce with the latest releases of all the above?

yes, it's a feature gap

### What operating system and processor architecture are you using?

All

### What did you do?

Attempted to use go's FilePath ReplaceDirective in a go bazel project whos dependencies are managed by the go_deps bzlmod extension.

```go
module github.com/bazelbuild/bazel-gazelle/tests/bcr/go_mod

go 1.21.5

require (
	example.org/hello v1.0.0
        // ... snip ...
)

// ... snip ...

replace example.org/hello => ../../../fixtures/hello
```

### What did you expect to see?

The ability to develop locally on the replaced example.org/hello go.mod residing at `../../../fixtures/hello`.

### What did you see instead?
The Feature does not exist, so it simply fails saying the feature is not support

stefanpenner:
FYI: I've also been working on this [here](https://github.com/bazelbuild/bazel-gazelle/pull/1776)

PR #2316, opened 2026-03-26
fix(go_deps): honor local-path replaces from non-root Bazel modules

rdesgroppes:
**What type of PR is this?**

Bug fix

go_repository

`go_deps` intentionally ignores `replace` directives from non-root Bazel modules as a supply-chain safeguard: a transitive dependency should not redirect a Go module to an arbitrary remote. Local-path replaces (`=> ./subdir`) are however safe - they redirect within their module itself - but were caught by the same guard.

A visible consequence is that `gazelle` users would repeatedly face the following warning message:
```
DEBUG: /path/to/external/gazelle+/internal/bzlmod/go_deps.bzl:819:18:
  No sum for github.com/bazel-contrib/bazel-gazelle/v2@2.0.0-1 found,
  run bazel run @rules_go//go -- mod tidy to generate it
```
... because gazelle's own `go.mod` carries `replace github.com/bazel-contrib/bazel-gazelle/v2 => ./v2` for its `v2` sub-module, which was silently discarded.

The proposed fix consists in admitting `go.mod` entries with **local** path to the `replace_map` even for non-root, non-isolated modules.

The changes comes with a new regression test that exercises the code path by depending on `@com_github_bazel_contrib_bazel_gazelle_v2`, which resolves correctly only when the local-path replace is honored.

#2315

**Other notes for review**

There might be a simpler solution to only silence the warning, it's just nothing came to mind.

Somewhat related:
- #1449
- #1775
- #2242

### DIFF
--- a/internal/bzlmod/go_deps.bzl
+++ b/internal/bzlmod/go_deps.bzl
@@ -572,6 +572,7 @@ def _go_deps_impl(module_ctx):
             new_version = semver.to_comparable(replace.version)
             module_resolutions[path] = with_replaced_or_new_fields(
                 module_resolutions[path],
+                local_path = replace.local_path,
                 replace = replace.to_path,
                 version = new_version,
                 raw_version = replace.version,


jayconrod:
Go ignores `replace` statements in modules other than the main module. I don't think we should allow these in general, certainly not by default.

Gazelle is doing something weird here: we have a single Bazel module but multiple Go modules. I wouldn't recommend that in general, but we're trying to develop v1 and v2 at the same time from the same code, keeping them compatible. They need to refer to each other directly, not through a `go_repository`.

This may not be a completely isolated case, so I'm adding a flag to allow it in #2317.

rdesgroppes:
Closing as per above review comments.

PR #2328, opened 2026-04-09
go_deps: allow inheriting host env vars

sluongng:
This adds `go_env_inherit` to both `gazelle_dependencies` and
`go_deps.config`, so callers can forward selected host environment
variables into dependency fetching without checking literal values into
source control.

That is especially useful for BuildBuddy users whose workflows provide
repository authentication or other fetch-time settings through runtime
environment variables. With this change, those workflows can keep the
values in the runner environment and tell Gazelle which variable names
to inherit.

Bzlmod example:

```bzl

go_deps.config(
    go_env = {
        "GOPRIVATE": "github.com/my-org/*",
    },
    go_env_inherit = [
        "REPO_USER",
        "REPO_TOKEN",
    ],
)
```

### DIFF
--- a/internal/common.bzl
+++ b/internal/common.bzl
@@ -51,3 +51,30 @@ def getenv(repo_ctx, name, default = ""):
     if hasattr(repo_ctx, "getenv"):
         return repo_ctx.getenv(name, default)
     return repo_ctx.os.environ.get(name, default)
+
+def resolve_env(ctx, direct = {}, inherit = [], reserved = []):
+    """Builds an environment map from explicit values and host env passthroughs."""
+    env = {}
+    reserved_keys = {key: True for key in reserved}
+    inherited_keys = {}
+    for key, value in direct.items():
+        if key in reserved_keys:
+            fail("{} cannot be set in go_env".format(key))
+        env[key] = value
+
+    for key in inherit:
+        if key in inherited_keys:
+            continue
+        inherited_keys[key] = True
+        if key in reserved_keys:
+            fail("{} cannot be set in go_env_inherit".format(key))
+        if key in env:
+            fail("{} cannot be set in both go_env and go_env_inherit".format(key))
+
+        # Use getenv when available so repository rules invalidate if the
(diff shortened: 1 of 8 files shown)


sluongng:
cc: @fmeum @tyler-french

internal/common.bzl
```diff
@@ -31,6 +31,18 @@ def env_execute(ctx, arguments, environment = {}, **kwargs):
     arguments = env_args + arguments
     return ctx.execute(arguments, **kwargs)
 
+_ABSENT_ENV_SENTINEL = "__GAZELLE_ABSENT_ENV__7e5475b0c2e343fe__"
+
+def valid_env_variable_name(name):
```
fmeum: Do we need this? If we didn't validate before, I don't think we need to do it now. Weird characters in env var names are supported by most OSes and Windows has a few.

internal/common.bzl
```diff
@@ -51,3 +63,34 @@ def getenv(repo_ctx, name, default = ""):
...
+        if not valid_env_variable_name(key):
+            fail("Invalid environmental variable name: '{}'".format(key))
+        if key in reserved_keys:
+            fail("{} cannot be set in go_env_inherit".format(key))
+        if key in env:
+            fail("{} cannot be set in both go_env and go_env_inherit".format(key))
+
+        # Use getenv when available so repository rules invalidate if the
+        # underlying host environment changes.
+        value = getenv(ctx, key, _ABSENT_ENV_SENTINEL)
```
fmeum: Could you use `None` instead of a sentinel?

internal/common.bzl
sluongng: Ah i think the clanker vibed this for the sentinel value.

sluongng: removed

internal/common.bzl
sluongng: Oh i thought default needs to be a string. None makes a lot more sense.

jayconrod (approved):

PR #2331, opened 2026-04-20
go.mod: drop unused indirect deps incl. grpc and x/net (clears CVE scanner findings)

rkk-ant:
## What

Removes 9 stale `// indirect` entries from `go.mod` that aren't imported by any gazelle package (v1 or v2): `google.golang.org/grpc`, `grpc/cmd/protoc-gen-go-grpc`, `golang.org/x/net`, `x/text`, `x/tools`, `gogo/protobuf`, `golang/mock`, `genproto`, `genproto/googleapis/rpc`. Prunes `go.sum`, `deps.bzl`, `vendor/modules.txt`, `go.work.sum`, and `MODULE.bazel.lock` to match.

## Why

Downstream supply-chain scanners (govulncheck, OSV, Socket.dev) read `go.mod` literally and flag v0.50.0 for:

- `google.golang.org/grpc@v1.67.3` — CVE-2026-33186, GHSA-p77j-4mvh-x3m3, GO-2026-4762
- `golang.org/x/net@v0.35.0` — GHSA-qxp5-gwg8-xv66, GHSA-vvgc-356p-c3xw, GO-2025-3503/3595, GO-2026-4440/4441

`go mod tidy -e` and a grep across all `.go` files confirm none of the dropped modules are imported — they're indirect-block entries that accumulated because plain `go mod tidy` can't complete past the `tests/` fixture imports. Removing them clears the scanner findings without behavior change and shrinks the dep closure back to roughly what v0.47.0 declared.

## How regenerated

```sh
go mod tidy -e
go work vendor -e
bazel mod deps --lockfile_mode=update
```

`deps.bzl` was pruned by hand to match what `gazelle update-repos -prune` would emit (8 `go_repository` blocks removed). Happy to re-run via the canonical target if there's a preferred incantation under the current go.work layout.

## Tested

- `go build ./...` and `go vet ./...` pass in workspace mode (excluding `tests/` fixtures)
- `internal/go_repository_tools_srcs.bzl` bootstrap check passes
- `bazel mod deps --lockfile_mode=update` clean, no warnings

Relying on CI for the full `bazel build //...`.


jayconrod:
I generally don't like to accept PRs changing dependencies because they're hard to verify without repeating the same work. Gazelle in particular has very hard dependencies to update. It usually takes me hours, which I can't spare now.

Checking out this PR, `go mod tidy -e` does still make changes (moving `go-difflib` to indirect), and removing v2 from go.work.sum may cause problems for downstream projects (we just had a release fixing that).

Unless there's an actual vulnerability that needs to be fixed, I'd prefer to leave dependencies alone until we're further along with v2 and can drop WORKSPACE support. It's too much work just to fix spurious warnings.

PR #2333, opened 2026-04-22
Allow placement of `go.work` in a nested (non-root) package

### LINKED ISSUE #2332: Allow non-root `go.work` file location
This is a feature request for a use-case of a multi-language monorepo that among other things utilizes Golang.

The idea is that if you have the following repo structure:
```
* .git
* go
    - foo
        - go.mod
    - bar
        - go.mod
    - go.work
* java
     - baz
     - <...>
* MODULE.bazel
```

It becomes quite logical to move `go.work` one level up to a parent `go/` directory instead of keeping everything at root.

In fact, the "fix" is very simple and doesn't require any additional configuration.

kberezin-nshl:
**What type of PR is this?**

> Feature

> bzlmod, go.work support

Allows an arbitrary location of `go.work` within a Bazel module

Fixes #2332

**Other notes for review**

Disclosure: I used an AI agent to make this change. I reviewed it myself and then manually ran full build and test suite.

I also tested it in our (private) monorepo by using `archive_override` and verified that it works with the `go.work` relocated into a subfolder.

### DIFF
--- a/internal/bzlmod/go_mod.bzl
+++ b/internal/bzlmod/go_mod.bzl
@@ -26,7 +26,7 @@ def _validate_go_version(path, state, tokens, line_no):
     if len(tokens) > 2:
         fail("{}:{}: unexpected token '{}' after '{}'".format(path, line_no, tokens[2], tokens[1]))
 
-def use_spec_to_label(repo_name, use_directive):
+def use_spec_to_label(go_work_label, use_directive):
     if use_directive.startswith("../") or "/../" in use_directive or use_directive.endswith("/.."):
         fail("go.work use directive: '{}' contains '..' which is not currently supported.".format(use_directive))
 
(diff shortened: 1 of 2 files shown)


kberezin-nshl:
@jayconrod may I have your feedback on this please? Thanks!

internal/bzlmod/go_mod.bzl
```diff
@@ -26,7 +26,7 @@ def _validate_go_version(path, state, tokens, line_no):
     if len(tokens) > 2:
         fail("{}:{}: unexpected token '{}' after '{}'".format(path, line_no, tokens[2], tokens[1]))
 
-def use_spec_to_label(repo_name, use_directive):
+def use_spec_to_label(repo_name, use_directive, package = ""):
```
jayconrod: This signature feels awkward. Instead of adding a `package` parameter, let's change `repo_name` to `go_work_label`. We can get both `.repo_name` and `.package` out of that.

jayconrod (commented):
Looks good. One small comment.

internal/bzlmod/go_mod.bzl
kberezin-nshl: Thanks for the feedback, I've just addressed that!

jayconrod (approved):

jayconrod:
Could you update this to `master` so we can merge it? I don't have permission to update your PR branch.

kberezin-nshl:
@jayconrod I've just done that!

PR #2335, opened 2026-04-22
Respect caller GOAUTH when building content

### LINKED ISSUE #2334: Gazelle build does not respect GOAUTH
When building content, gazelle will respect the user's GOPROXY (ref https://github.com/bazel-contrib/bazel-gazelle/blob/a472876ab09f8d86010222e77ea9bdcc247d03f8/internal/go_repository.bzl#L231) but will not respect the user's GOAUTH. For environments where the configured GOPROXY requires authentication to work, we must configure both. The GOAUTH environment variable instructs the go toolchain on how to get authentication tokens that it should use when accessing GOPROXY.

yabberyabber:
**What type of PR is this?**
Bug fix

 go_repository

Pass the user's GOAUTH env var (alongside GOPROXY) when invoking go. If both GOAUTH and GOPROXY are set, the go toolchain will use GOAUTH to generate access tokens that will be passed to GOPROXY. [See here for documentation](https://pkg.go.dev/cmd/go@master#hdr-GOAUTH_environment_variable) on the GOAUTH environment variable.

GOAUTH was introduced in go1.24 which explains why it wasn't in the original list of env vars. Now that the go toolchain supports it, gazelle should support it as well.

Fixes #2334

**Other notes for review**

### DIFF
--- a/internal/go_repository.bzl
+++ b/internal/go_repository.bzl
@@ -225,6 +225,7 @@ def _go_repository_impl(ctx):
         # not go out to the network at all. This means *the build*
         # goes out to the network. We tolerate this for downloading
         # archives, but finding module roots is a bit much.
+        "GOAUTH",
         "GONOPROXY",
         "GONOSUMDB",
         "GOPRIVATE",


fmeum (approved):

yabberyabber:
Thanks for the quick review, @fmeum 

What does the release process look like for bazel-gazelle? Are releases created on a schedule?

I'm happy to use `archive_override` on my end until the next release is cut off; I'm just curious when I should be checking back to see if/when I can remove that override and get back on an officially released version.

fmeum:
No, we tend to just cut them when we think it's worth it. You can text these repo's releases to be notified.

PR #2339, opened 2026-04-28
fix(go): do not set cgo when import C appears only in excluded files

jgautier-dd:
**What type of PR is this?**
Bug fix

language/go

Gazelle could set cgo = True on a go_library when the only import "C" lived in a //go:build ignore (or otherwise constraint-excluded) file. go build would not compile that file, so the library was not actually CGO, but Gazelle still flipped cgo and could pull in neighboring .c / .S sources as if the package used CGO.

**Fix**
getPlatformStringsAddFunction now returns (add, included bool) in one constraint pass: included means the file contributes on some known rules_go OS/arch/platform.

goTarget.addFile sets cgo with:

t.cgo = t.cgo || (info.isCgo && included)

so import "C" in a file that is excluded everywhere (e.g. //go:build ignore) no longer forces cgo.


jgautier-dd:
The CI error I think is something flakey not related to the actual change

```

(20:31:56) ERROR: /private/var/tmp/_bazel_buildkite/8cacacd4f537b4411198655663c2f8f1/external/bazel_tools/tools/build_defs/repo/http.bzl:137:45: An error occurred during the fetch of repository 'rules_jvm_external+':
--
Traceback (most recent call last):
File "/private/var/tmp/_bazel_buildkite/8cacacd4f537b4411198655663c2f8f1/external/bazel_tools/tools/build_defs/repo/http.bzl", line 137, column 45, in _http_archive_impl
download_info = ctx.download_and_extract(
Error in download_and_extract: java.io.IOException: Error downloading [https://github.com/bazelbuild/rules_jvm_external/releases/download/6.3/rules_jvm_external-6.3.tar.gz] to /private/var/tmp/_bazel_buildkite/8cacacd4f537b4411198655663c2f8f1/external/rules_jvm_external+/temp13826203162849904283/rules_jvm_external-6.3.tar.gz: GET returned 502 Bad Gateway or Proxy Error
(20:31:56) ERROR: no such package '@@rules_jvm_external++maven+stardoc_maven//': no such package '@@rules_jvm_external+//': java.io.IOException: Error downloading [https://github.com/bazelbuild/rules_jvm_external/releases/download/6.3/rules_jvm_external-6.3.tar.gz] to /private/var/tmp/_bazel_buildkite/8cacacd4f537b4411198655663c2f8f1/external/rules_jvm_external+/temp13826203162849904283/rules_jvm_external-6.3.tar.gz: GET returned 502 Bad Gateway or Proxy Error
(20:31:56) ERROR: /private/var/tmp/_bazel_buildkite/8cacacd4f537b4411198655663c2f8f1/external/stardoc+/src/main/java/com/google/devtools/build/stardoc/renderer/BUILD:29:13: no such package '@@rules_jvm_external++maven+stardoc_maven//': no such package '@@rules_jvm_external+//': java.io.IOException: Error downloading [https://github.com/bazelbuild/rules_jvm_external/releases/download/6.3/rules_jvm_external-6.3.tar.gz] to /private/var/tmp/_bazel_buildkite/8cacacd4f537b4411198655663c2f8f1/external/rules_jvm_external+/temp13826203162849904283/rules_jvm_external-6.3.tar.gz: GET returned 502 Bad Gateway or Proxy Error and referenced by '@@stardoc+//src/main/java/com/google/devtools/build/stardoc/renderer:renderer_lib'
(20:31:56) ERROR: Analysis of target '//docs:reference_docgen' failed; build aborted: Analysis failed
(20:31:56) INFO: Elapsed time: 18.291s, Critical Path: 0.02s
(20:31:56) INFO: 50 processes: 50 internal.
(20:31:56) ERROR: Build did NOT complete successfully
(20:31:56) FAILED:
Fetching repository @@+go_deps+com_github_bazelbuild_buildtools; starting 12s
Fetching repository @@rules_cc++cc_configure_extension+local_config_cc; starting 12s
Fetching repository @@+go_deps+com_github_pmezard_go_difflib; starting 12s
Fetching repository @@+go_deps+org_golang_x_tools_go_vcs; starting 12s
Fetching repository @@+go_deps+org_golang_x_mod; starting 12s
Fetching repository @@+go_deps+com_github_google_go_cmp; starting 12s
Fetching repository @@+go_deps+com_github_bmatcuk_doublestar_v4; starting 12s
Fetching repository @@+go_deps+org_golang_x_sync; starting 12s ... (12 fetches)
bazel build failed with exit code 1
🚨 Error: The command exited with status 1


```

jayconrod:
I'm not sure about this. It feels like a partial solution, and I wonder if it would be simpler to work around it and properly exclude or delete the ignored files.

You're right that if all of the cgo sources are excluded, the library should not be built with cgo.

But the complication is that if all the sources are excluded *for a specific platform*, cgo should be disabled on that platform but still enabled on other platforms. So the `cgo` attribute might need a `select` on it. Same for `cdeps`.

rules_go does some filtering of build constraints already, so it's usually easier to figure out things at compile time, though it doesn't look like it works in this case. It decides whether cgo should be enabled before applying build constraints, not after. And it's still technically better to figure things out in BUILD files so we don't unnecessarily require a C toolchain or build C dependencies that go unused.

Could you say more about when this was a problem for you? A proper fix for this is complicated, so if it was just `import "C"` in a file with `//go:build ignore`, I'd be inclined to say just work around it with `# gazelle:exclude`.

jgautier-dd:
I have a related change here: https://github.com/bazel-contrib/rules_go/pull/4603. Basically what I want is that only if a cgo is actually used in a binary (and `static` is set) should it use the external linker. The PR in rules_go fixes this by analyzing what is actually included at build time so would fix this case for the //go:build ignore and some of the other edge cases you mentioned. However there is still a case where a library you pull in might have use cgo but with a pure go fallback, in those cases I think it would be reasonable for a user to have to specify an override. In those cases would `# gazelle:exclude` work, or does it make sense to have a new directive like `# gazlle:cgo false` that you could use to selectively disable cgo?

jayconrod:
> However there is still a case where a library you pull in might have use cgo but with a pure go fallback, in those cases I think it would be reasonable for a user to have to specify an override.

The `cgo` attribute only applies to the library where it's specified. A library with `cgo = False` (the default) may depend on a library with `cgo = True`, and vice versa. It only indicates whether any source files may need cgo preprocessing. I think it's best to override this by excluding the cgo files (or correctly noticing that they're excluded already by build constraints).

If the all libraries you're interested in have non-cgo alternative code, you may be interested in the `pure` attribute on `go_binary` and the flag of the same name. You can still build libraries with `cgo = True` in that mode, but files with `import "C"` will be excluded, and the `cgo` build tag will be false.

PR #2351, opened 2026-06-01
feat: don't require `# gazelle:` prefix in directive_file

rafikk:

This is a follow-up to #2314.

With the introduction of `directive_file`, directives in an external file still require the `# gazelle:` prefix. This harms readability as comments preceding directives are not visually distinguished from the directives themselves. With this PR, directives in external files can now be declared without the `# gazelle:` prefix, e.g.:

```
# Generate a single proto_library per file
proto file
```

**What type of PR is this?**

Feature

`rule`

N/A

**Other notes for review**

### DIFF
--- a/v2/rule/directives.go
+++ b/v2/rule/directives.go
@@ -76,11 +76,12 @@ func parseDirectives(stmt []bzl.Expr) []Directive {
 }
 
 var directiveRe = regexp.MustCompile(`^#\s*gazelle:(\w+)\s*(.*?)\s*$`)
+var fileDirectiveRe = regexp.MustCompile(`^(?:#\s*gazelle:)?(\w+)\s*(.*?)\s*$`)
 
 // ParseDirectivesFromFile reads a file and extracts Gazelle directives from it.
-// Each line is matched against the same pattern used for BUILD file comments
-// (# gazelle:key value). Blank lines and comment lines that don't match
-// the directive pattern are ignored.
+// Each line is matched against a permissive version of the pattern used
+// for BUILD file comments ("# gazelle:key value" or "key value"). Blank lines
+// and comment lines that don't match the directive pattern are ignored.
 func ParseDirectivesFromFile(filePath string) ([]Directive, error) {
 	f, err := os.Open(filePath)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


v2/rule/directives.go
```diff
@@ -76,6 +76,7 @@ func parseDirectives(stmt []bzl.Expr) []Directive {
 }
 
 var directiveRe = regexp.MustCompile(`^#\s*gazelle:(\w+)\s*(.*?)\s*$`)
+var fileDirectiveRe = regexp.MustCompile(`^#?\s*gazelle:(\w+)\s*(.*?)\s*$`)
```
rafikk: Review note: We don't actually need to introduce a new regex for this and could update the existing `directiveRe` in place since `directiveRe` is only ever searched over parsed `Comments`. However, I thought this was safer and clearer, but let me know if you'd prefer to keep a single regex.

jayconrod: Yes, let's keep it separate.

jayconrod: Let's drop the `gazelle:` prefix too. Every uncommented non-blank line in a directive file should be a directive, and I can't think of a good reason to require the `gazelle:` prefix. It's needed in BUILD files since there may be other tools like buildifier with their own directives, but no other tools should be reading Gazelle directive files.

jayconrod (commented):
Please also update gazelle-reference.md.

v2/rule/directives.go
rafikk: Done. PTAL. I wasn't sure if you wanted to only support the new format, but I've supported all three variants (`# gazelle:key value`, `gazelle:key value` and `key value`) for flexibility. Let me know if you'd like to me change that.

jayconrod: Let's not support `gazelle:key value`, only the other two.

Fewer formats is better. In hindsight, it would have been better to support `key value` and not `# gazelle:key value`, but too late to change that now.

rafikk: Done. PTAL.

Note: I wanted to add a test to ensure that `gazelle:key value` is ignored, but the regex currently captures this as `rule.Directive{Key:"gazelle", Value:":resolve go example.com/foo //third_party:foo"}}` which is wrong. I tried to update the regex from `^(?:#\s*gazelle:)?(\w+)\s*(.*?)\s*$` to `^(?:#\s*gazelle:)?(\w+)\s+(.*?)\s*$` (`\s*` to `\s+`) but then `mixed directive types` fails because `# gazelle:ignore` is not parsed.

Let me know if this should be fixed and if you have a suggestion for how to go about it. I don't want to overcomplicate the regex. One option would be to update `fileDirectiveRe` to only parse the new format and update `ParseDirectivesFromFile` to match against both `directiveRe` and `fileDirectiveRe`.

jayconrod: Let's make it an error in `ParseDirectivesFromFile` if a matched line starts with `gazelle:`. That should result in a clearer behavior than if we simply skipped those lines by not matching the regex.

rafikk: That's a good solution. Done. Let me know if you'd like me to reword the error message or make any other changes.

jayconrod (approved):
Looks good, thanks!

PR #2354, opened 2026-06-03
gazelle_binary needs cc toolchain

dzbarsky:
**What type of PR is this?**
Bug fix

rules_go has become stricter in requiring rules to declare their cc toolchain (the previous ways predated toolchain resolution and could lead to incorrect toolchains being selected for multiplatform builds)

https://github.com/bazel-contrib/rules_go/issues/4626

**Other notes for review**

### DIFF
--- a/internal/gazelle_binary.bzl
+++ b/internal/gazelle_binary.bzl
@@ -131,7 +131,10 @@ proto extension stores metadata in hidden attributes of generated
         ),
     },
     "executable": True,
-    "toolchains": ["@io_bazel_rules_go//go:toolchain"],
+    "toolchains": [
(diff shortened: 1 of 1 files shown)


fmeum (approved):
We could expose the constants added in https://github.com/bazel-contrib/rules_go/commit/8286eae655d83412d11169ea707592cf0c0979b3#diff-f6aaf6101bc7030a9b8d09d47a1af5918ef80ca113d867cf8b781d8d4d9d5722 publicly for use by third-party rulesets. But future LTS versions of Bazel will no longer require the fragments, so we would need to clean up the interface anyway.

Which is to say that this looks good. :⁠-⁠)

PR #2356, opened 2026-06-04
Add publish-to-bcr workflow

jayconrod:
**What type of PR is this?**

> Other

n/a

After this, I think the release process should just be publishing a tag. The release_ruleset workflow will create the source archive and attach it to a new GitHub release.

n/a

**Other notes for review**

Copied from bazel_features with light modification

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -1,0 +1,36 @@
+# Publish new releases to Bazel Central Registry.
+name: Publish to BCR
+on:
+  # Run the publish workflow after a successful release
+  # Will be triggered from the release.yaml workflow
+  workflow_call:
+    inputs:
+      tag_name:
+        required: true
+        type: string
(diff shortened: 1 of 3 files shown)


dzbarsky (approved):

fmeum (approved):

kormide:
I suggest upgrading to `v1.2.0` of the workflow so that you can enable immutable GitHub releases. See https://github.com/bazel-contrib/publish-to-bcr#immutable-releases.

jayconrod:
I see there's a v1.3.0. If we're upgrading, should we use that?

kormide:
> I see there's a v1.3.0. If we're upgrading, should we use that?

Yup. While you're making changes, you can also remove `.bcr/config.yml`. The `fixedReleaser` property was used by the GH app and isn't needed for the reusable workflow.

PR #2359, opened 2026-06-08
Upgrade publish-to-bcr workflow

jayconrod:
**What type of PR is this?**

> Other

release process

Title says it all.

Also delete .bcr/config.yml: was only used by the app

n/a

**Other notes for review**

Follow-up to #2356, based on @kormide's recommendation.

### DIFF
--- a/.bcr/config.yml
+++ b/.bcr/config.yml
@@ -1,3 +1,0 @@
-fixedReleaser:
-  login: fmeum
-  email: fabian@meumertzhe.im
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -21,7 +21,7 @@ on:
         type: string
 jobs:
   publish:
-    uses: bazel-contrib/publish-to-bcr/.github/workflows/publish.yaml@v1.1.0
+    uses: bazel-contrib/publish-to-bcr/.github/workflows/publish.yaml@v1.3.0
     with:
       tag_name: ${{ inputs.tag_name }}
       # GitHub repository which is a fork of the upstream where the Pull Request will be opened.


jayconrod:
I think we would need to change the workflows a bit to actually enable immutable releases though. Currently, `release.yaml` is triggered by pushing a tag, and `publish.yaml` is triggered afterward. But publishing the tag effectively sets the release in stone as far as Go is concerned, so it doesn't really make sense to create a draft release at that point.

kormide (approved):

kormide:
> I think we would need to change the workflows a bit to actually enable immutable releases though. Currently, `release.yaml` is triggered by pushing a tag, and `publish.yaml` is triggered afterward. But publishing the tag effectively sets the release in stone as far as Go is concerned, so it doesn't really make sense to create a draft release at that point.

I haven't worked much in Go so I don't know the nuances of releasing, but it would only remain a draft for a short period. If the release is created as a draft, then the publish and finalizing of the release should be a matter of seconds. If it's not acceptable for the release to remain as a draft if the publish job failed, you could always have the finalization step run unconditionally.

In any case, immutable releases aren't necessary and you could look into in the future if you wanted to.

jayconrod:
I'll go ahead and merge this. Seems like a good idea in any case, though it's worth thinking about our release process a bit more.

> I haven't worked much in Go so I don't know the nuances of releasing, but it would only remain a draft for a short period. If the release is created as a draft, then the publish and finalizing of the release should be a matter of seconds. If it's not acceptable for the release to remain as a draft if the publish job failed, you could always have the finalization step run unconditionally.
>
> In any case, immutable releases aren't necessary and you could look into in the future if you wanted to.

I very much want immutable releases. Go requires them. Go's module system is based on Git tags. Once a tag is created and fetched by proxy.golang.org, it's added to a global transparency log at sum.golang.org. If the tag is changed later, a user will either get the old content from proxy.golang.org, or they'll get a security error when fetching the new content because the hash of the new content doesn't match sum.golang.org.

So I basically want to create the tag as late as possible in the process, after all tests pass and approvals are granted. This is somewhat incompatible with BCR processes: we need to create a tag first, then PR the BCR release, which can still fail.

PR #2362, opened 2026-06-10
Fix empty rule deletion for aliased kinds

### LINKED ISSUE #2363: `alias_kind` rules not deleted when empty rule is emitted
When a language extension emits an empty rule (via `result.Empty`) to delete a stale target, and the existing BUILD file uses an aliased kind (via `# gazelle:alias_kind`), the rule is not deleted.

In `v2/merger/merger.go`
```go
  if oldRule.IsEmpty(kinds[oldRule.Kind()]) {
      oldRule.Delete()
  }
```

`oldRule.Kind()` returns the aliased kind (e.g., `my_py_library`), which has no entry in the kinds map. `kinds["my_py_library"]` returns a zero-value KindInfo with nil NonEmptyAttrs. IsEmpty returns false when NonEmptyAttrs is nil , so the rule is never deleted.

Discovered this issue while working on https://github.com/taowang487/rules_python/pull/1 . The added unit test is a reproducible project.

taowang487:
**What type of PR is this?**
Bug fix

merger

Fixes empty rule deletion for aliased kinds (`# gazelle:alias_kind`).

When a language extension emits an empty rule to delete a stale target, `MergeFile` uses `kinds[oldRule.Kind()]` to look up `KindInfo` for the `IsEmpty` check. For aliased kinds (e.g., `my_py_library` aliased to `py_library`), the aliased name has no entry in the `kinds` map, returning a zero-value `KindInfo` with `nil` `NonEmptyAttrs`. Since `IsEmpty` returns `false` when `NonEmptyAttrs` is nil, the rule is never deleted — it gets stripped to a skeleton but remains in the BUILD file.

The fix resolves through `aliasedKinds` before looking up `KindInfo`.  `map_kind` is not affected — `update.go` rewrites empty rule kinds before  they reach the merger, so the kind lookup already works correctly

Fixes #2363

**Other notes for review**


jayconrod (approved):
Thanks for fixing!

PR #2365, opened 2026-07-01
go_deps: allow overriding a module's Bazel repo name

### LINKED ISSUE #2364: go_deps: allow overriding the generated Bazel repo name to disambiguate colliding modules
Hi, I encounter this issue when using gazelle. @fmeum @jayconrod  Could you review this proposal?

---

Two distinct modules whose import paths differ only by / vs _ (or only by case) mangle to the same Bazel repo name, so go_deps hard-fails. E.g. example.com/foo/bar/baz and example.com/foo_bar_baz both become com_example_foo_bar_baz

When both are pulled in transitively, "use only one in your go.mod" isn't always actionable.

Prior art: #1949, #1954 (kept the hard fail), #2213. An earlier version of the message ended with "isn't supported in Gazelle (yet)".

Proposal
Let the root module override the generated repo name for a colliding module via the existing go_deps.gazelle_override:

```starlark
go_deps.gazelle_override(
    path = "example.com/foo_bar_baz",
    repo_name = "com_example_foo_bar_baz_alt",
)
```

When set, use repo_name instead of the mangled default. The collision check still runs on the resulting names. Purely additive; root-module-only, like other gazelle_override attributes. A dedicated tag works too if preferred. Happy to send a PR.

H5-O5:
Two distinct Go modules whose import paths differ only by `/` vs `_` (or only by case) mangle to the same Bazel repo name, so go_deps hard-fails. When both are pulled in transitively, dropping one from the `go.mod` is not always actionable.

Add an optional `repo_name` attribute to the `gazelle_override` tag. When set, it is used as the Bazel repo name for that module instead of the mangled default, letting the root module break such collisions. The change is purely additive and root-module-only, like other gazelle_override attributes; the collision check still runs on the resulting names, and its error message now points at this attribute. `repo_name` is intentionally kept out of `_GAZELLE_ATTRS` so it cannot be set via `gazelle_default_attributes`, since a repo name is necessarily per-module.

**What type of PR is this?**

Feature

none. This is a new feature.

See above.

Fixes #2364

**Other notes for review**

### DIFF
--- a/extensions.md
+++ b/extensions.md
@@ -15,7 +15,7 @@ go_deps.gazelle_override(<a href="#go_deps.gazelle_override-build_extra_args">bu
 go_deps.gazelle_default_attributes(<a href="#go_deps.gazelle_default_attributes-build_extra_args">build_extra_args</a>, <a href="#go_deps.gazelle_default_attributes-build_file_generation">build_file_generation</a>, <a href="#go_deps.gazelle_default_attributes-directives">directives</a>)
 go_deps.module(<a href="#go_deps.module-build_file_proto_mode">build_file_proto_mode</a>, <a href="#go_deps.module-build_naming_convention">build_naming_convention</a>, <a href="#go_deps.module-indirect">indirect</a>, <a href="#go_deps.module-local_path">local_path</a>, <a href="#go_deps.module-path">path</a>, <a href="#go_deps.module-sum">sum</a>,
                <a href="#go_deps.module-version">version</a>)
-go_deps.module_override(<a href="#go_deps.module_override-patch_cmds">patch_cmds</a>, <a href="#go_deps.module_override-patch_strip">patch_strip</a>, <a href="#go_deps.module_override-patches">patches</a>, <a href="#go_deps.module_override-path">path</a>)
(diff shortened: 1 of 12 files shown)


H5-O5:
Failure looks irrelevant. Can someone re-run the failed tests?

fmeum:
Bazel CI is broken in multiple ways. I'll review this now, but can only merge when the infra issues have been fixed.

extensions.md
```diff
@@ -88,6 +88,7 @@ Override Gazelle's behavior on a given Go module defined by other tags in this e
...
 | <a id="go_deps.gazelle_override-path"></a>path |  The Go module path for the repository to be overridden.<br><br>This module path must be defined by other tags in this extension within this Bazel module.   | String | required |  |
+| <a id="go_deps.gazelle_override-repo_name"></a>repo_name |  The Bazel repository name to use for this Go module.<br><br>By default, Gazelle derives the repository name from the module's import path. Two distinct modules whose import paths differ only by "/" vs "_" (or only by case) derive the same default name, which Gazelle rejects with an error. When both are pulled in transitively, dropping one from the go.mod is not always possible. Setting this attribute lets the root module assign a distinct repository name to one of the colliding modules so they can coexist. The collision check still runs on the resulting names.   | String | optional |  `""`  |
```
fmeum: Logically this would fit better on a `go_deps.module_override` as it's relevant even if you don't run Gazelle on a module.

fmeum (commented):
Looks good to me, but could you exercise the new functionality in one of the `tests/bcr` modules with a real dep pair that runs into this issue?

H5-O5:
@fmeum Hi, I updated this using AI (force pushed). Could you have a look again?

fmeum (approved):

PR #2366, opened 2026-07-02
fetch_repo: verify module sum using "go mod download" JSON, not the extracted tree

malt3:
fetch_repo re-hashed the extracted module directory with dirhash.HashDir and compared the result against the go.sum entry. This duplicates the verification go itself does, but derives the hash from the filesystem rather than the authoritative source, so it fails whenever the filesystem cannot faithfully round-trip the zip's file names.

github.com/vbatts/go-mtree v0.7.0 ships a testdata file whose name ends in a trailing space (testdata/dirwithbrokenlink/...). On Windows the trailing space is stripped during extraction, so HashDir walks a different set of names than the zip contains and computes a different hash. This surfaces as a spurious sum mismatch on Windows only, even though the downloaded module is correct:

    fetch_repo: resulting module with sum
    h1:0eFy2k8hOMfWGZ0EXIWhDdNMeCo0+DSQWB23GP6xdWg=; expected sum
    h1:ytmOc3MTRidZiBi9VBCyZ2BHe4fZS47L5v7BVXDWW4E=. Please retry.

The h1: hash in go.sum is defined as dirhash.HashZip over the module zip, which is exactly what the go toolchain computes and verifies before extraction. Trust the value reported by Go instead: `go mod download -json` already reports the sum it calculated, so we can use that value. This means we perform the same check standard Go tooling performs, which is immune to filesystem round-trip quirks (trailing spaces, case-folding, unicode normalization, symlink handling, ...).

The verification also moves ahead of copyTree, so a mismatch is caught before anything is written to the destination.

**What type of PR is this?**

Bug fix

go_repository

Allows users to depend on vbatts/go-mtree on Windows

No issue created

**Other notes for review**

Happy to add a test, but really unsure how to write one that's not overly complex.


malt3:
Note: we could also just use `dl.Sum` directly (and trust the validation performed by Go). Let me know what you prefer.

fmeum:
> Note: we could also just use `dl.Sum` directly (and trust the validation performed by Go). Let me know what you prefer.

+1 for doing that. @jayconrod

malt3:
@fmeum I updated the PR to use `dl.Sum`.

malt3:
An interesting observation I made is that the `HashDir` check was added deliberately [here](https://github.com/bazel-contrib/bazel-gazelle/pull/1755) while adding support for the host module cache via `GO_REPOSITORY_USE_HOST_MODCACHE`.
Technically, my PR makes it so the host mod cache is now trusted without extra validation.

Hypothetically, we could first check if the `GOMODCACHE` env var is set and only calculate `dirhash.HashDir` then. Technically, this means that modules like `github.com/vbatts/go-mtree` would always fail on Windows with a host mod cache set.

I don't think there's a single best option.

jayconrod:
I think perhaps what we can do here is delegate the sum verification to `go mod download` by writing a temporary `go.sum` file before we run the command. Then we don't actually need to calculate or verify the sum within `fetch_repo`. We already write a `go.mod` file there. Both files should be removed before the `copyTree` call.

Experimentally, if I write a `go.sum` file with the wrong sum, then `go mod download -json <module>` fails correctly with a security error. I don't think it downloads the archive if it's already in the cache, but I'm not sure whether it recalculates it from the zip or if the sum is saved somewhere.

In any case, I think it's fine to trust the contents of the local module cache, whether it's managed by Bazel or the user. The main threat we're protecting against is an upstream repo being compromised; if the user's own system is compromised, not much we can do to help.

jayconrod:
Right. This is good and solves the problem, but I think telling `go mod download` the expected sum is a slight improvement.

cmd/fetch_repo/BUILD.bazel
```diff
@@ -13,10 +13,7 @@ go_library(
     ],
     importpath = "github.com/bazelbuild/bazel-gazelle/cmd/fetch_repo",
     visibility = ["//visibility:private"],
-    deps = [
-        "@org_golang_x_mod//sumdb/dirhash",
```
fmeum, line 17: Could you also delete the vendored dirhash.go?

fmeum (approved):

malt3:
I'll update the PR with your suggestion. Thanks for the input!

jayconrod (approved):
Thanks! LGTM.

PR #2368, opened 2026-07-07
Fix alias_kind directive leaking across directories

alex-torok:
Config.Clone() deep-copies KindMap but never copied AliasMap, so every directory's config shared a single AliasMap by reference. A "# gazelle:alias_kind" directive set in one package therefore mutated the map seen by every other directory in the repo. For most people this likely isn't an issue, but could lead to unexpected behavior.

Fix Clone() to copy AliasMap the same way it copies KindMap, making alias_kind properly directory-scoped like map_kind. Add a regression test that reproduces the cross-directory leak and asserts an unrelated sibling package is left untouched.

**What type of PR is this?**

Bug fix

v2/config

Fixes a bug with alias_kind not being appropriately scoped to the package that defines the directive.

N/A


cmd/gazelle/integration_test.go
```diff
@@ -406,6 +406,71 @@ my_custom_macro(
...
 }
 
+// TestAliasKindIsDirectoryScoped verifies that a "# gazelle:alias_kind"
+// directive set in one package does not leak into unrelated packages
+// elsewhere in the repository. The repo root declares an alias_kind of its
+// own so that its AliasMap is allocated and inherited by both "leaf" and
+// "other"; "leaf" then adds a second alias_kind, and "other" is an unrelated
+// sibling package that never opts in and must be unaffected by "leaf"'s
+// directive.
+func TestAliasKindIsDirectoryScoped(t *testing.T) {
```
jayconrod: Could you move this to `//tests`? New integration tests should go there (if they're compatible with the framework) so that we can test against both v1 and v2.

jayconrod (commented):
Nice find, thanks for fixing this.

jayconrod (approved):
Looks good, thanks!

Issue #2371, opened 2026-07-13
Feature request: a directive to set gc_goopts on generated Go rules


H5-O5:
### Problem

Gazelle has no way to set `gc_goopts` on the Go rules it generates. In particular, there's no supported way to apply Go compiler flags to a **specific third-party package**: `go_deps` external repos are generated entirely by Gazelle, and `--@rules_go//go/config:gc_goopts` only affects the current module, so it never reaches them.

### Proposal

Add a `# gazelle:go_gc_goopts <flags>` directive that sets the `gc_goopts` attribute on generated rules. It rides the existing `directives` channel, so it also covers external deps:

```python
go_deps.gazelle_default_attributes(directives = ["gazelle:go_gc_goopts <flags>"])
```

It only sets an attribute and never changes the rule kind, so it composes with the rest of Gazelle.

### Design question

A "set **any** attribute" directive would be more general, but the value's type is the problem: a directive is just text, while the emitted attribute's Bazel type depends on the value's Go type (`gc_goopts` → string list, `cgo` → bool, `deps` → label list, some attrs → `select(...)`). Going fully generic needs either type annotations in the syntax or an "everything is a string list" assumption, plus dynamic handling in the merge/idempotency logic.

A middle ground: limit it to **string-list compiler attributes** (`gc_goopts`, `gopts`, …), where the type is fixed. Would maintainers prefer **(a)** `gc_goopts`-only, **(b)** a string-list compiler-attr family, or **(c)** fully generic? Happy to implement whichever you'd accept.

### Contribution

I have a working `gc_goopts`-only implementation (directive parsing + emission, `MergeableAttrs` for idempotency, tests) and would be happy to upstream it.

*Drafted with Claude Code.*

jayconrod:
Adding a `go_gc_goopts` directive seems reasonable to me.

Let's also support similar Go attributes, `clinkopts`, `copts`, `cppopts`, `cxxopts`, and `gc_linkopts`. `clinkopts`, `copts`, and `cppopts` should only be set when `cgo = True`. `clinkopts` and `gc_linkopts` should only be set on `go_binary` and `go_test` rules.

I don't think it makes sense to generalize this outside of Go. Attributes are almost always valid for certain kinds of rules in certain situations, and they often need special handling (e.g., merging with cgo directives). Let's leave that complexity up to the extensions.

H5-O5:
[...]

PR #2372, opened 2026-07-14
feat(go): add directives to set Go/C compiler and linker flags on generated rules

H5-O5:
Implements #2371.

Adds a family of Go directives that set string-list compiler/linker attributes on generated rules. The main motivation is applying flags to external dependencies generated by `go_deps`, which have no hand-written build files to edit — the directives flow through the existing `directives` channel of `go_deps.gazelle_default_attributes` / `gazelle_override`.

| Directive | Attribute | Applied to | Gate |
|---|---|---|---|
| `# gazelle:go_gc_goopts` | `gc_goopts` | `go_library` / `go_binary` / `go_test` with sources | — |
| `# gazelle:go_gc_linkopts` | `gc_linkopts` | `go_binary`, `go_test` | — |
| `# gazelle:go_copts` | `copts` | cgo targets (`go_library`) | `cgo = True` |
| `# gazelle:go_cppopts` | `cppopts` | cgo targets (`go_library`) | `cgo = True` |
| `# gazelle:go_cxxopts` | `cxxopts` | cgo targets (`go_library`) | `cgo = True` |
| `# gazelle:go_clinkopts` | `clinkopts` | cgo targets (`go_library`) | `cgo = True` |

**Notes on design (per discussion in the issue):**
- The C flags (`copts`/`cppopts`/`cxxopts`/`clinkopts`) are **merged** with the flags Gazelle derives from `#cgo` comments, not overwritten.
- `clinkopts` is set on `go_library`, matching where the source-derived `clinkopts` already go. Gazelle never generates a cgo `go_binary`/`go_test` (cgo isn't allowed in `_test.go`, and a cgo `main` becomes a cgo `go_library` embedded by a plain `go_binary`), so in practice all C flags land on `go_library`.
- `gc_goopts`/`gc_linkopts` are set directly (not derived from source); `gc_goopts` needs sources to compile, `gc_linkopts` applies to the linking rules.
- Values may be separated by whitespace and/or commas (`-N -l` is equivalent to `-N,-l`); an empty value resets the directive.

**Tests**
- Unit tests for directive parsing, reset, and inheritance (`config_test.go`).
- A generate golden (`testdata/compiler_flags`) exercising emission, the cgo gate, the library-vs-linking gate, and merging with `#cgo`-derived flags.
- `go test ./language/go/` passes; new/changed files are `gofmt`-clean.

**Docs**
- `language/go/reference.md` documents all six directives.

*Drafted with Claude Code.*


language/go/config.go
```diff
@@ -739,6 +789,21 @@ func splitValue(value string) []string {
 	return values
 }
 
+// splitCompilerFlags parses the value of a compiler/linker flag directive
+// (go_gc_goopts, go_copts, ...) into a list of flags. Flags may be separated by
+// commas and/or whitespace, so "# gazelle:go_gc_goopts -N -l" and
+// "# gazelle:go_gc_goopts -N,-l" both yield ["-N", "-l"]. An empty value resets
+// the directive, returning nil.
+func splitCompilerFlags(value string) []string {
```
jayconrod: Let's not split on commas: some flags need to include commas, especially when passing flags to the external linker, like `-Wl,-rpath,$ORIGIN/libs`.

Let's use `splitQuoted` to split the fields. That also will allow flags with spaces.

language/go/config.go
```diff
@@ -591,6 +623,24 @@ Update io_bazel_rules_go to a newer version in your WORKSPACE file.`
 					continue
 				}
 
+			case "go_gc_goopts":
+				gc.gcGoopts = splitCompilerFlags(d.Value)
```
jayconrod: For these new directives, let's follow similar semantics to `go_search`: instead of replacing the list of flags, a directive appends to it. Multiple directives with the same key may appear in a BUILD file: that lets the user specify long flags more readably, one per line. A directive with an empty value clears the current list.

jayconrod (approved):
Thanks, looks good!

PR #2375, opened 2026-07-21
create workflow to draft release

tyler-french:
Let's try setting up a simple workflow


fmeum:
If possible, we should follow the examples of other rules-contrib modules (all derived from the rules-template setup), e.g., https://github.com/bazel-contrib/bazel_features/blob/main/.github/workflows/release.yml. That would give us "push to tag" or alternatively a standard workflow that accepts the name of a tag.

PR #2383, opened 2026-07-28
Fix cgo race detection

keith:
This is required for rules_go 0.62.0+

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -65,6 +65,14 @@ tasks:
     - "..."
     - "-//internal:bazel_test"
     - "-//cmd/gazelle:gazelle_test"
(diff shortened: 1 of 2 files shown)


fmeum (approved):

keith:
I think the new test would pass as is w/o the fix until we updated to rules_go 0.62.0

hunshcn:
Thanks for the fix! Could we get a release with it (0.52.3 or 0.53.0)? This is the last piece needed to make rules_go 0.63.0 usable for repos that enable race mode globally: 0.63.0 restored the mechanism in #4672, and this PR makes gazelle's own `std_package_list` use it. Until a gazelle release includes both, we're pinned to rules_go 0.61.1 (see bazel-contrib/rules_go#4662).

PR #2384, opened 2026-07-28
Test fix without CI job

keith:


keith:
testing https://github.com/bazel-contrib/bazel-gazelle/pull/2383 without the CI changes to make sure it doesn't break anything that already exists

Issue #2344, opened 2026-05-04
go_deps underresolves transitive dep versions in a multi-module workspace


findleyr:
**What version of gazelle are you using?**

0.50.0

**What version of Bazel are you using?**

8.5.0

**Description**

When go_deps.from_file(go_work = ...) is used with a multi-module workspace, gazelle can resolve a lower version of a transitive dependency than Go's own MVS, because it reads only the workspace go.mod files' require lines and does not follow the one-level indirection that Go 1.21+ module graph pruning relies on.

With go 1.21 or later in a go.mod, go mod tidy omits // indirect lines that are already implied by a listed require's own go.mod (https://go.dev/ref/mod#graph-pruning). Go's resolver recovers the omitted requirement by reading that one transitive go.mod. Gazelle's go_deps does not, so if the only workspace module that records the dep has it at a lower version, gazelle resolves that lower version.

**Minimal repro**

```
go.work          (use ./mod_a ./mod_b)
mod_a/go.mod     require golang.org/x/text v0.35.0   # x/text@v0.35.0 requires x/mod@v0.33.0
                 # x/mod is PRUNED — implied by x/text's go.mod
mod_b/go.mod     require golang.org/x/mod v0.31.0
```

Both modules are tidied (GOWORK=off go mod tidy in each). Then:

```sh
$ go list -m golang.org/x/mod
golang.org/x/mod v0.33.0

$ bazel mod show_repo @org_golang_x_mod | grep version
  version = "v0.31.0",
```

Full repro in the associated PR.

**Why it matters**

Bazel-built binaries link a different (older) version of the dep than go build would. In our case this surfaced as golang.org/x/mod v0.31.0 vs v0.33.0 across a workspace where every module's go.mod was individually tidy.

**Related**

- #1091 tracks graph pruning generally; this is a specific consequence for go work files.
- #1585 is the WORKSPACE-era analogue (update-repos).

fmeum:
Wow, this is tricky to solve as it would require downloading all deps' go.mod files. Is it true that this can't happen without `go.work` and that using `x/mod` directly in your example would prevent this?

I hope that @jayconrod has an idea how to address this.

findleyr:
> Is it true that this can't happen without go.work

I suppose this also applies to untidy `go.mod` files, but the `go.work` case is where everything is correct and it still breaks.

> Is it true that this can't happen without go.work and that using x/mod directly in your example would prevent this

[...]

PR #2388, opened 2026-08-05
go_deps: add a unit testing framework

jayconrod:
**What type of PR is this?**

> Test

> go_deps

In preparation to rework go_deps to call 'go list -m' instead of simulating MVS internally. We should have some tests that verify we're getting the same versions before and after. Writing unit tests for this is difficult, but the tests are MUCH faster than tests/bcr/go_mod and go_work, which have gotten unwieldy.

In each test case, we represent a mock workspace in Starlark with modules, tags, and files, accessed through a mock module_ctx. After running go_deps_impl, we check that expected repos were declared with expected versions.

For #2344

**Other notes for review**

I wish there were `bazel coverage` for Starlark.


tests/bzlmod/go_deps_test.bzl
```diff
@@ -1,5 +1,135 @@
...
+        path = lambda v: _mock_module_ctx_path(case, v),
+        read = lambda filename: _mock_module_ctx_read(case, filename),
+        is_dev_dependency = lambda tag: False,
+        declare_repo = lambda rule, *, name, **kwargs: _mock_module_ctx_declare_repo(repos, rule, name = name, **kwargs),
+        _repos = repos,
+    )
+
+def _mock_module_ctx_path(case, v):
+    repo_name = case.modules[0].name
+    if type(v) == "path":
```
fmeum: Can this ever happen? The mocks can only create `struct`s, not actual `path` objects. We should probably return those unchanged instead?

tests/bzlmod/go_deps_test.bzl
```diff
@@ -1,5 +1,135 @@
...
+def _go_deps_test_impl(ctx):
+    env = unittest.begin(ctx)
+    errors = []
+    for case in _GO_DEPS_TEST_CASES:
+        module_ctx = _mock_module_ctx(case)
+        go_deps_impl(module_ctx)
+
+        missing_repo = False
+        for want_repo in case.want:
+            if want_repo.name not in module_ctx._repos:
```
fmeum, line 34: These assertions may benefit from the existing collection assertions in `rules_testing`, especially when it comes to comparing repo attributes.

fmeum (approved):
I couldn't resist looking into this. This is actually very similar to what I wanted to build when working on go_deps originally, but never did since I wanted to build a generic version and that was too much work at the time. Thanks for making this, it will over a much better experience!

internal/bzlmod/go_deps.bzl
```diff
@@ -397,7 +413,8 @@ _SHARED_REPOS = [
     "google.golang.org/protobuf",
 ]
 
-def _go_deps_impl(module_ctx):
+def go_deps_impl(module_ctx):
+    module_ctx = _wrap_module_ctx_for_testability(module_ctx)
```
fmeum, line 417: An alternative could be to pass in `go_repository = go_repository` style arguments, but this does seem less hacky.

jayconrod: Oh something like:

```
def go_deps_impl(module_ctx, go_repository = go_repository):
```

And then `go_repository` would be a wrapper that captures the declaration? I didn't think of that, it would work. I have some slight preference for the current approach: no additional boilerplate needed per repo rule.

tests/bzlmod/go_deps_test.bzl
jayconrod: Hmm, haven't used rules_testing before. How's this?

I'll admit some distaste for fluent APIs in general. The code can be easier to read but is often abstract, and I have trouble writing it. I suppose writing code is a lot easier than it used to be though!

tests/bzlmod/go_deps_test.bzl
jayconrod: Right, good catch. Changed to pass through something that looks like `_mock_path`.

jayconrod (commented):
Thanks for reviewing, sorry this is huge!

Certainly hard to make a generic version of this. Probably a lot more would need to be done to support repository rules too.

I like how this turned out, but the test cases are unavoidably still quite large and difficult to write. Agents make that process a lot easier though: I wrote empty, module, and bcr_go_mod by hand, but mvs was entirely generated. I'll add more while working on the `go list -m` change.

tests/bzlmod/go_deps_test.bzl
fmeum: I find the most important benefit of using rules_testing to be the diff logic for failed dict/collection asserts. The fluent style is more of a question of individual preference, I guess (although Starlark types will make it much more useful). This looks good!

fmeum (approved):

PR #2389, opened 2026-08-06
Use ephemeral GOMODCACHE for module-mode fetches

### LINKED ISSUE #2292: Make bazel_gazelle_go_repository_cache friendly to remote caching
With the introduction of the remote repo contents cache, content produced by repositories can we cached remotely. When using remote execution and BwoB this makes it possible to run builds without downloading repository contents to the host. Only reproducible repositories benefit from this, and it would therefore be very useful is the go repo cache could somehow be made reproducible.

blampe:
**What type of PR is this?**

Bug fix

go_repository

fetch_repo consumes the downloaded module immediately and nothing downstream reads GOMODCACHE, so persisting it in the shared cache repo only grows unbounded (many GB on projects with large SDK dependencies). Scope it to a per-invocation tempdir under the go_repository and delete it after fetch_repo returns.

The GOCACHE at <cache_dir>/gocache is unchanged; it's populated by go_repository_tools and benefits from persistence.

Host-cache opt-ins are unchanged: they surface as GOMODCACHE already set in env, which skips the ephemeral override.

Fixes #2292.

**Other notes for review**


fmeum (approved):
@jayconrod Could you confirm this is safe if you get the chance?

PR #2392, opened 2026-08-10
go_deps: compute environment outside of go_repository_cache

jayconrod:
**What type of PR is this?**

> Refactor

> go_deps

This lays the groundwork for being able to run the go command from go_deps. go_deps allows the user to set Go's environment via go_deps.config.go_env and go_env_inherit. There may be multiple isolated instances of go_deps, so these settings should only apply to the repos declared within each go_deps.

Previously in module mode, the environment was stored in @bazel_gazele_go_repository_cache (declared from @non_module_deps, shared by all go_deps). go_repository loaded from here, ignoring config.go_env settings on isolated go_deps instances. This didn't really seem consistent, and I didn't want to extend the ambiguity by using that configuration when invoking the go command from go_deps.

Now instead, go_repository_config (declared by go_deps) writes a go.env file, and go_repository loads that.

For #2344

**Other notes for review**


internal/bzlmod/go_deps.bzl
```diff
@@ -613,6 +615,15 @@ def go_deps_impl(module_ctx):
     _fail_on_unmatched_overrides(gazelle_overrides.keys(), module_resolutions, "gazelle_overrides")
     _fail_on_unmatched_overrides(module_overrides.keys(), module_resolutions, "module_overrides")
 
+    # Compute the environment based on the config tag and available go_sdks.
+    go_env = compute_env(
+        module_ctx,
+        # Label.repo_name is always a canonical name, so use a canonical label.
```
fmeum, line 621: A canonical label would need to start with `@@`, not just `@`. It's possible that the comment needs to be updated or that consumers confusingly expect a single `@`.

internal/env.bzl
```diff
@@ -0,0 +1,290 @@
...
+        if value != None:
+            env[key] = value
+
+    return env
+
+def getenv(ctx, name, default = ""):
+    """
+    Gets an environment variable.
+
+    Uses ctx.getenv if it exists (after Bazel 7.1.0) to also invalidate
```
fmeum: I think that we can assume the highest minor version of Bazel 7 at this point and drop this helper.

internal/bzlmod/go_deps.bzl
jayconrod: Fixed. This was copy-and-pasted, but it was confusing because we were also adding `@` in `compute_env` (previously `go_repository_cache`). So now `compute_env` does not add that `@` but verifies that it's there.

internal/env.bzl
jayconrod: Sounds good, replaced `getenv` with direct calls to `ctx.getenv`.

jayconrod (commented):
Thanks for reviewing! The dependencies between these module extensions and repo rules have kept my head spinning for months. I think this is only marginally more consistent, but hopefully it will get better once we drop WORKSPACE mode.

fmeum (approved):

PR #2401, opened 2026-08-16
perf(language/go): parse build tags without Scanner

blico-v2:
**What type of PR is this?**

Other

language/go

Parse legacy `// +build` constraints by iterating directly over the source
bytes instead of using `bufio.Scanner`.

The new implementation follows the approach used by Go's `go/build` package:

- Iterate over lines using `bytes.IndexByte`.
- Trim lines without converting them to strings.
- Cheaply reject lines that cannot be `+build` constraints.
- Convert only candidate constraint lines to strings for
  `constraint.IsPlusBuild` and `constraint.Parse`.

The previous scanner token-size limit and `bufio.ErrTooLong` behavior are
preserved.

An allocation profile from a large ARM64 repository attributed 3.42 GB of
flat allocation to `bufio.Scanner.Scan`.

A local microbenchmark, removed before submission, measured:

| Case | Before | After | Allocations before | Allocations after |
|---|---:|---:|---:|---:|
| No build tags | 45.4 µs/op | 37.6 µs/op | 118 allocs/op | 16 allocs/op |
| `+build` constraint | 19.8 µs/op | 17.2 µs/op | 21 allocs/op | 19 allocs/op |
| 1,000-line header | 294.6 µs/op | 258.4 µs/op | 1,026 allocs/op | 24 allocs/op |

Tests cover CRLF input, optional spacing, similar non-constraint prefixes,
and both sides of the existing scanner token-size boundary.

N/A


jayconrod (approved):
Nice improvement, thanks for including benchmark results. Feel free to add benchmarks to the test, too!

PR #2402, opened 2026-08-17
perf(merger): reuse load indexes across BUILD files

blico-v2:
## Summary

- add a reusable `merger.LoadFixer` that indexes known load files and symbols once
- reuse the fixer across BUILD files in the update loop
- retain `FixLoads` as a compatibility wrapper and preserve per-file handling when `map_kind` changes the known loads

In a large monorepo profile, load fixing decreased from 10.01s to about 5.7s cumulative CPU time and from 4.00 GB to about 460 MB cumulative allocation. These are function-level measurements; this change does not claim an equivalent end-to-end wall-time improvement.

## Testing

- `go test ./v2/merger ./v2/cmd/gazelle/update ./v2/cmd/gazelle`
- `go test -race ./v2/merger ./v2/cmd/gazelle/update`
- `go vet ./v2/merger ./v2/cmd/gazelle/update`


v2/merger/fix.go
```diff
@@ -23,21 +23,56 @@ import (
...
+	}
+	return fixer
+}
+
 // FixLoads removes loads of unused go rules and adds loads of newly used rules.
 // This should be called after FixFile and MergeFile, since symbols
 // may be introduced that aren't loaded.
 //
 // This function calls File.Sync before processing loads.
 func FixLoads(f *rule.File, knownLoads []rule.LoadInfo) {
```
jayconrod: Since `FixLoads` is made obsolete by this change, let's delete it from v2, since we're not locked in yet.

v1 `merger.FixLoads` calls this, so please call `v2.NewLoadFixer(knownLoads).Fix(f)` from there instead.

jayconrod (commented):
Thanks looks like a nice improvement.

jayconrod (approved):

PR #2407, opened 2026-08-19
fix(go_repository): add package metadata regardless of build file generation

### LINKED ISSUE #2406: Package metadata is missing for source repositories shipping their own BUILD files

0.52.2

0.63.0

9.2.0

Yes.

Linux/macOS/Windows on x86_64/aarch64.

### What did you do?

Leverage bazel-contrib/rules_go#4595 released with `rules_go` 0.63.0.

`Deps` entries of `runtime/debug.ReadBuildInfo()` for all Gazelle-generated `go_deps`.

### What did you see instead?

`Deps` entries are present when Gazelle generates the BUILD file in full, but package metadata is missing at all when the source repository ships one.

Proposed fix:
- #2407.

rdesgroppes:
**What type of PR is this?**

Bug fix

go_repository

bazel-contrib/rules_go#4595 builds the `Deps` entries of `runtime/debug.ReadBuildInfo()` from the `package_metadata` that `go_repository` attaches to each module via `REPO.bazel`.
A module without that label is therefore missing from `Deps`.

`go_repository` only writes a module's `REPO.bazel` and `package_metadata` label inside the `if generate:` block, so a module whose repository ships its own `BUILD` file gets no label at all, even though the label describes the module rather than its build rules.
Forcing `build_file_generation = "on"` would attach the label too, but at the cost of regenerating build files a repository ships on purpose.

@fmeum [pointed out](https://github.com/bazel-contrib/bazel-gazelle/pull/2407#discussion_r3814829409) that appending the generated targets onto whatever build file already contained assumes that content is safe to extend, e.g. free of a target name collision with what we generate.
That risk was never specific to a repository-provided file either: appending onto Gazelle's own freshly generated content made the same assumption, just with better odds of holding.

This revised change therefore consists in writing the generated targets to a dedicated `_gazelle_generated` package instead, so `go_repository` never has to reason about what a build file already contains.
A leading underscore is enough to make collisions practically impossible: Go's own tooling ignores any directory starting with `_`, so a real module is exceedingly unlikely to populate one.
`REPO.bazel`'s `default_package_metadata` now points there, and the generated `package_info` target drops its `gazelle_generated_` prefix, redundant once the package name says as much.

Nothing about this write depends on `build_file_name` any longer, so attach the metadata unconditionally rather than gating it behind `generate`.

A new analysis test checks that against the existing `proto_compat` fixture, which ships its own `BUILD.bazel` file and previously carried no package metadata.
The existing test is renamed for consistency.

Prior to the fix:
```
cd tests/bcr/go_mod
bazel test //:test_package_info_for_repo_provided_build
...
Error: <merged target @@gazelle++go_deps+org_example_proto_compat//:proto_compat>
  (rule 'go_library') doesn't contain declared provider 'PackageInfo'
```

Fixes #2406.

**Other notes for review**

A few examples of Go repositories shipping their own `BUILD`/`BUILD.bazel` that would gain `package_metadata` this way:
- https://github.com/cel-expr/cel-spec
- https://github.com/chrusty/protoc-gen-jsonschema
- https://github.com/envoyproxy/protoc-gen-validate
- https://github.com/go-resty/resty
- https://github.com/google/flatbuffers
- https://github.com/grpc-ecosystem/grpc-gateway
- https://github.com/iceber/iouring-go
- etc.

Prior art:
- external deps: bazel-contrib/rules_go#4595,
- package path: bazel-contrib/rules_go#4690,
- build settings: bazel-contrib/rules_go#4691.

### DIFF
--- a/internal/go_repository.bzl
+++ b/internal/go_repository.bzl
@@ -417,32 +417,27 @@ def _go_repository_impl(ctx):
     # Apply patches if necessary.
     patch(ctx)
 
-    if generate:
-        # Do not override a REPO.bazel patched in by users. This also provides a
-        # way for users to opt out of Gazelle-generated package_info.
-        repo_file = ctx.path("REPO.bazel")
-        if not repo_file.exists:
-            ctx.file("REPO.bazel", """\
+    # Do not override a REPO.bazel patched in by users. This also provides a
+    # way for users to opt out of Gazelle-generated package_info.
+    repo_file = ctx.path("REPO.bazel")
+    if not repo_file.exists:
+        ctx.file("REPO.bazel", """\
 repo(
     default_package_metadata = [
-        "//:gazelle_generated_package_info",
-        "//:package_metadata",
+        "//_gazelle_generated:package_info",
+        "//_gazelle_generated:package_metadata",
     ],
 )
 """)
-
-            # Modify the top-level build file after patches have been applied as the
-            # patches may otherwise conflict with our generated content.
-            build_file = ctx.path(build_file_name)
-            if build_file.exists:
-                build_file_content = ctx.read(build_file)
-            else:
-                build_file_content = ""
-            build_file_content += _generate_package_info(
+        # Write the generated targets to a dedicated package to avoid name collisions
+        # with targets already defined in the top-level build file.
+        ctx.file(
+            "_gazelle_generated/BUILD.bazel",
+            _generate_package_info(
                 importpath = ctx.attr.importpath,
                 version = ctx.attr.version,
-            )
-            ctx.file(build_file_name, build_file_content)
+            ),
+        )
 
     if reproducible and hasattr(ctx, "repo_metadata"):
         return ctx.repo_metadata(reproducible = True)
@@ -482,7 +477,7 @@ package_metadata(
 )
 
 package_info(
-    name = "gazelle_generated_package_info",
+    name = "package_info",
     package_name = {package_name},
     package_url = {package_url},
     package_version = {package_version},
--- a/tests/bcr/go_mod/tests.bzl
+++ b/tests/bcr/go_mod/tests.bzl
@@ -5,10 +5,10 @@ load("@rules_license//rules:providers.bzl", "PackageInfo")
 load("@rules_testing//lib:analysis_test.bzl", "analysis_test", "test_suite")
 load("@rules_testing//lib:truth.bzl", "subjects")
 
-def _test_package_info(name):
+def _test_package_info_for_gazelle_generated_build(name):
     analysis_test(
         name = name,
-        impl = _test_package_info_impl,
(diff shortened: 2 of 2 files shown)


internal/go_repository.bzl
```diff
@@ -417,7 +416,10 @@ def _go_repository_impl(ctx):
     # Apply patches if necessary.
     patch(ctx)
 
-    if generate:
+    # Attach package metadata regardless of whether the build file is generated here
```
fmeum: This comment isn't really a justification for why we can do this. Instead, we would need to convince ourselves that what we do to the BUILD file (appending new lines) is safe no matter what it contained (not assuming an adversarial BUILD file, of course). For example, what if the target names collide?

What I could see us doing is adding a BUILD file under a unique `_bazel_gazelle_generated_package_metadata` package.

@jayconrod @sluongng What do you think?

rdesgroppes: Working on it.

jayconrod: That does sound safer.

aiuto: Weighing in from supply_chain.
Our policy is that the BUILD file from the upstream repo is the truth. We should not be modifying it.

If they don't include metadata, that is their deficiency.

rdesgroppes: > Working on it.

fmeum: @aiuto This is the case of a "package manager" (Go modules) that doesn't have Bazel-aware upstreams. I don't think we can get anywhere without injecting metadata information for that.

sluongng: > If they don't include metadata, that is their deficiency.

I don't think it's a fair characterization. These are relatively newer features and we can't really blame the OSS maintainers for keeping up with latest upstream.

A better way to think about this would be to empower whoever is setting this up for their repo to be able to fix this without much hassle. Automatically or semi-automatically.

For this particular instance, I think the idea of using a hard-coded package that Gazelle generated would be a good short-term solution. Longer term, I would prefer if this information could be added directly to REPO.bazel to support a simpler setup (1 repo, 1 version), while the existing attributes remain as an option for a more complicated setup (1 repo, multiple package versions).

rdesgroppes: I revised the PR to address the collision concern: generated targets now live in a dedicated package instead of being appended to `build_file_name`, so `go_repository` no longer has to reason about what a build file already contains, whether that file is Gazelle's own or the repository's.
This also means the repository's own `BUILD`/`BUILD.bazel` is never modified, which speaks to @aiuto's concern directly: the upstream file stays untouched either way.

This PR stays scoped to the short-term fix, since the `REPO.bazel`-native direction is a larger design question independent of the collision concern it started from.

Would appreciate another look when you have a chance, thanks in advance!

fmeum (approved):

PR #2409, opened 2026-08-24
feat(go_repository): expose go.sum checksum on the golang PURL

### LINKED ISSUE #2408: go_repository's package_metadata PURL carries no checksum

0.53.0

0.63.0

9.2.0

Yes.

Linux/macOS/Windows on x86_64/aarch64.

### What did you do?

A plain `go build` on a module that depends on a `go_repository`-managed package already reports that dependency's checksum through `runtime/debug.ReadBuildInfo()`, with no extra work on the user's part.
I verified this directly, building a scratch module against `github.com/fmeum/dep_on_gazelle@v1.0.0` with vanilla `go build` (not Bazel) and printing its build info:
```
ok: true
Main.Path: vanilla_buildinfo_check
Main.Version: (devel)
Dep: Path=github.com/fmeum/dep_on_gazelle Version=v1.0.0 Sum=h1:7gEtQ2CoD77tYca+1iUnKjIBUZ4mX7mZwjdWp3uuN/E=
```

`Sum` there is exactly the `h1:` line from that module's `go.sum` entry, populated automatically because `cmd/go` already has it on hand when it resolves the module.

A Bazel-built binary of the same dependency gives nothing in that field.
bazel-contrib/rules_go#4595 builds the `Deps` entries of `runtime/debug.ReadBuildInfo()` for Bazel-built Go binaries, reading them from the `pkg:golang/...` PURL that `go_repository` writes into each module's `package_metadata`.
That PURL currently carries only the module path and version, so the `Module.Sum` field rules_go populates from it is always empty, even though `go_repository`'s own `sum` attribute already has the exact same `go.sum` hash vanilla Go reports, and already verifies the downloaded module against it.

So the discrepancy is: vanilla `go build` reports a dependency's checksum in build info for free, a Bazel build of the same dependency, via `go_repository`, currently cannot report it at all, because the one piece of metadata rules_go reads (the PURL) never carries it.

The same checksum information vanilla Go already reports for this dependency: a Bazel-built binary's `runtime/debug.ReadBuildInfo()` reporting `Sum=h1:7gEtQ2CoD77tYca+1iUnKjIBUZ4mX7mZwjdWp3uuN/E=` for `github.com/fmeum/dep_on_gazelle`, the same value shown above, instead of an empty one.

### What did you see instead?

No checksum is available anywhere in `package_metadata`'s PURL for a Go module fetched by `go_repository`, so no build-info or supply-chain consumer reading that PURL can recover it, even though `go_repository` already has and verifies against that exact checksum internally, and even though a plain `go build` of the same dependency reports it with no extra effort.

rdesgroppes:
**What type of PR is this?**

Feature

go_repository

`go_repository`'s `sum` attribute already carries a module's `go.sum` `h1:` hash, and the bzlmod `go_deps` extension already sets it for every fetched module.
Add it to the generated `pkg:golang/...` PURL as a `checksum` qualifier, the PURL spec's own extension point for exactly this, so build-info consumers reading that PURL can recover it.

bazel-contrib/rules_go#4700 builds the `Deps` entries of `runtime`/`debug.ReadBuildInfo()` from this same PURL, and reads the `checksum` qualifier back out.

A `go.sum` `h1:` hash only ever contains the base64 alphabet plus the literal `h1:` prefix, so only `+`, `/`, and `=` need percent-escaping to be a valid PURL qualifier value.

Fixes #2408.

**Other notes for review**

This started out building the PURL with `@package_metadata//purl:purl.bzl`'s `purl.builder()`, instead of hand-formatting it, so escaping would be handled by that library instead of reimplemented here.

That required bumping `package_metadata` past `0.0.5`, and `go_repository.bzl` loading `purl.bzl` at its own top level broke WORKSPACE mode: `deps.bzl` needs to load `go_repository.bzl` before `gazelle_dependencies()` can run, and `gazelle_dependencies()` is what defines `@package_metadata`, a cycle.

`purl.builder()` also would have needed either forking `go_repository` by mode or passing a precomputed PURL in as data, both a larger change than the qualifier itself justified.

The present handcrated formatting keeps `go_repository.bzl` free of any `@package_metadata` dependency, so the cycle cannot occur, and needs no `package_metadata` version bump.

### DIFF
--- a/internal/go_repository.bzl
+++ b/internal/go_repository.bzl
@@ -436,6 +436,7 @@ repo(
             _generate_package_info(
                 importpath = ctx.attr.importpath,
                 version = ctx.attr.version,
+                sum = ctx.attr.sum,
             ),
         )
 
@@ -444,7 +445,7 @@ repo(
     else:
         return None
 
-def _generate_package_info(*, importpath, version):
+def _generate_package_info(*, importpath, version, sum):
     package_name = importpath
 
     # TODO: Consider adding support for custom remotes.
@@ -459,6 +460,15 @@ def _generate_package_info(*, importpath, version):
             namespace_and_name = importpath,
             version = version,
         )
+
+        # sum may only be set when version is.
+        if sum:
+            # TODO(rdesgroppes): Use @package_metadata//purl:purl.bzl's `purl.builder()` once
+            # WORKSPACE support is dropped. Until then, using it here would require extending the
+            # WORKSPACE stanza with a `package_metadata` declaration before loading
+            # `gazelle_dependencies`, so the interim solution below only percent-encodes the
+            # characters in a go.sum `h1:` digest that must be escaped in a PURL qualifier value.
+            purl += "?checksum=" + sum.replace("+", "%2B").replace("/", "%2F").replace("=", "%3D")
     else:
         purl = "pkg:golang/{namespace_and_name}".format(
             namespace_and_name = importpath,
--- a/tests/bcr/go_mod/tests.bzl
+++ b/tests/bcr/go_mod/tests.bzl
@@ -33,7 +33,7 @@ def _test_package_info_for_gazelle_generated_build_impl(env, target):
     subject.package_name().equals("github.com/fmeum/dep_on_gazelle")
     subject.package_version().equals("1.0.0")
     subject.package_url().equals("https://github.com/fmeum/dep_on_gazelle")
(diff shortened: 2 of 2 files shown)


fmeum (approved):
Looks good to me, please fix the tests and ping me.

fmeum (approved):
@jayconrod Are you okay with the slightly more complex WORKSPACE stanza? I expect other rulesets to be facing the same problem.

fmeum (approved):

PR #2415, opened 2026-08-27
go_deps: add new test cases, make test format machine-editable

jayconrod:
**What type of PR is this?**

> Test

> go_deps

This change adds new unit test cases for //tests/bzlmod:go_deps_test. Some of the new tests are disabled, since they don't work with the current go_deps for reasons listed in the TODO.

The test format has changed significantly, now based on JSON, since it can be parsed by Starlark and other tools. The next PR will add a conversion tool that lets us unpack a test case to a directory then pack it back to a .bzl file with generated output for executed commands.

For #2344

**Other notes for review**

n/a


tests/bzlmod/go_deps_test.bzl
```diff
@@ -1,108 +1,355 @@
...
 
+# TODO(#2344): enable skipped tests for new go_deps implementation
+#
+# Blockers with the old implementation
+# - go_deps_impl calls fail directly, so tests can't check failures
+# - gazelle_override overrides settings on attributes not explicitly set
+# - dep_files not set correctly
+# - root_module_direct_deps not set in isolated go_deps
+# - root_module_direct_deps includes indirect requirements
+# - go_deps may select higher versions than before (bug fix)
```
fmeum, line 44: Great finds, this will make me feel so much better about the extension!

tests/bzlmod/go_deps_test.bzl
```diff
@@ -1,108 +1,355 @@
...
+    state.files[filename] = content
+
 def _mock_module_ctx_path(case, v):
-    repo_name = case.modules[0].name
     if type(v) == "struct" and hasattr(v, "_name"):
         # mock path, created with _mock_path. Return as-is.
         return v
     elif type(v) == "Label":
-        return _mock_path(case, _label_to_path(repo_name, v))
+        return _mock_path(case, _label_to_path("!!!", v))
```
fmeum: Why this value?

fmeum (approved):

tests/bzlmod/go_deps_test.bzl
jayconrod: Ah, I think this was a debug string to check if we ever used the default repo name when converting labels to paths. We don't. I think this can sometimes be incorrect because the label's repo name would ordinarily be a canonical name, but we can't actually tell if it's canonical other than checking the `+` suffix which is not stable across versions. I think this stopped being a problem when I started referencing the main Bazel module's go.mod files directly, since they don't need to be modified.

Issue #2413, opened 2026-08-26
go_repository_config (bzlmod): refactor configuration file format and arguments


jayconrod:
I've found this profoundly confusing while working on #2344. `go_repository_config` generates a `WORKSPACE` file that contains `go_repository` declarations. `gazelle` reads this when it starts up within `go_repository`. It's not literally used as a `WORKSPACE` file. It's effectively a table with the columns:

- `name`: either the name of a `go_repository` or the canonical name of a Bazel module providing a Go module. So could be `org_golang_x_mod` or `@gazelle+` (with a leading `@` and trailing `+`).
- `importpath`: the Go module path
- `module_name`: for Bazel modules, this is the actual name of the module (may differ from the apparent name). Note that `module_name` is not a real attribute on `go_repository`.

Changes we should make:

- Express the table as a JSON file instead of `WORKSPACE` for simpler parsing and understanding. We still need to be able to parse configuration from `WORKSPACE` files in `WORKSPACE` mode, but hopefully we can drop that functionality soon.
- `name` should just be the repo name.
- `is_bazel` should indicate whether the module is provided by Bazel. If true and Gazelle is running inside `go_repository`, generate canonical labels like `@@gazelle+//label` instead of `@gazelle//label`. This is needed since we don't know module apparent names.
- Drop `module_name`.
- Add `prefix_dir`: relative path to the module root directory from the repo root. This lets us map `github.com/bazel-contrib/bazel-gazelle/v2` to `@gazelle//v2/...`. It's generally useful for multi-module workspaces in Bazel modules with dependents.

PR #2418, opened 2026-08-27
go_deps: rewrite using 'go list -m'

### LINKED ISSUE #2344: go_deps underresolves transitive dep versions in a multi-module workspace

### LINKED ISSUE #2394: `go_deps` materializes a main repo module as a redundant `go_repository` when another module imports it

0.52.2

0.61.1

9.2.0

Yes. I built a minimal repro against current `bazel-contrib/bazel-gazelle` master with a real `bazel build`, not just by reading the code, and it reproduces there too.

Linux x86_64, but this is not OS- or architecture-specific: it is a pure label-resolution issue in the `go_deps` Bzlmod extension.

### What did you do?

We have a monorepo (`github.com/DataDog/datadog-agent`) that:
* declares a single Bazel module at the repo root,
* uses `go_deps.from_file(go_work = "//:go.work")`, where `go.work` lists dozens of `use ()` entries, each its own Go module with its own `go.mod`,
* has checked-in `BUILD.bazel` files for every one of those in-tree modules, reachable at `//<path>`, produced by our own root `//:gazelle` invocation with a `gazelle:prefix` directive matching the repo's module path,
* has these `go.work` members `replace`-ing each other locally in their own `go.mod` files on top of `go.work`. For example, one member's `go.mod` has a `replace` directive pointing another member's import path at a relative local path, and that member is in turn `replace`d the same way by several sibling modules' `go.mod` files.

Separately, an external Go module is consumed via the same `go_deps` extension and imports one of those same in-tree packages by its fully-qualified import path.

I built a minimal, real, `bazel build`-verified repro of this (not just a code reading) with the following shape:
```
repo/
├── MODULE.bazel          # bazel_dep(gazelle), go_deps.from_file(go_work = "//:go.work")
├── go.work                # use ( . ./pkg/foo )
├── go.mod                  # module example.com/repo
│                            # require example.com/repo/pkg/foo v0.0.0-fake // indirect
│                            # require example.com/thirdparty v0.0.0-fake
│                            # replace example.com/repo/pkg/foo => ./pkg/foo
└── pkg/foo/
    ├── BUILD.bazel          # go_library(name = "foo", importpath = "example.com/repo/pkg/foo", ...), checked in
    ├── foo.go
    └── go.mod               # module example.com/repo/pkg/foo
```

Plus a `thirdparty` Go module (injected hermetically via `go_deps.archive_override` so the repro needs no network access) whose `go.mod` requires `example.com/repo/pkg/foo` and which imports it. A `go_binary` at the repo root depends on both `//pkg/foo` and, transitively, on `thirdparty`.

The crucial, easy-to-miss ingredient is the `replace example.com/repo/pkg/foo => ./pkg/foo` line in the root's own `go.mod`. Without it, `go_deps` correctly avoids creating a duplicate repo for `example.com/repo/pkg/foo`. I verified this empirically too: removing that line makes the build fail differently, with an unrelated dangling-repository error rather than a duplicate-package one (see "Related, but distinct" below). With it, `bazel build //...` fails with:
```
package conflict error: example.com/repo/pkg/foo: multiple copies of package passed to linker:
    @@gazelle++go_deps+com_example_repo_pkg_foo//
    @@//pkg/foo
```

I traced why in `internal/bzlmod/go_deps.bzl`:
* `go_deps` walks every `module_tag` derived from parsing all of the workspace's go.mod files (via `deps_from_go_mod`) to compute MVS resolutions. For each `module_tag.path`, if `module_tag.path in replace_map` (`replace_map` is the union of every `replace` directive found in any go.work member's go.mod, collected earlier), the resulting `module_resolutions[path]` struct is given `local_path = replacement.local_path`, but no `module_name` attribute.
* The "unify" loop meant to make root-owned `go.work` members resolve to themselves instead of being fetched looks like this:
  ```starlark
  for path, bazel_go_module in bazel_go_modules.items():
      if path in archive_overrides or path in gazelle_overrides or path in module_overrides or path in replace_map:
          continue
      if path in modules_from_go_work:
          module_resolutions[path] = bazel_go_module
          continue
  ```
  The override/replace check fires before the `modules_from_go_work` branch below it. So if the path is also a `replace_map` target (as `pkg/foo` is here, from some other go.work member's `go.mod`), the loop `continue`s without ever overwriting `module_resolutions[path]` with the root-owned struct. The struct from the MVS pass, carrying `local_path` and no `module_name`, survives untouched.
* At repo creation time, `if hasattr(module, "module_name"): continue` is the only thing that skips declaring a `go_repository` for a root-owned path. Because the surviving struct has no `module_name`, this check does not fire, and `go_deps` calls `go_repository(local_path = "pkg/foo", importpath = "example.com/repo/pkg/foo", ...)`, a second, independent `go_library` for source that's already tracked in-tree.

Any `go_binary`/`go_test` depending on both label spaces then fails at link time exactly as shown above.

Since `example.com/repo/pkg/foo` (or, in our real case, the equivalent in-tree package) is a `go.work` member with its own in-tree, already-`BUILD.bazel`'d package, I expected `go_deps` to never create a second `go_repository` for it, regardless of whether some other go.work member's `go.mod` also happens to `replace` it. Concretely, the `path in replace_map` short-circuit in the unify loop should not bypass the `path in modules_from_go_work` unification when the path in question is a root-owned `go.work` member. Root ownership should take priority.

### What did you see instead?

`go_deps` creates a second, independent `go_repository` (via `local_path`) for a `go.work` member whenever (a) some other module imports it and (b) any go.mod in the workspace has a `replace` directive targeting that same import path. In a monorepo where submodules commonly `replace` each other locally (as ours does, and as I'd expect most non-trivial `go.work`-based monorepos to, since it is the standard way to make a multi-module repo buildable both with and without `go.work`), this can affect any workspace member that another module happens to import. The result is two distinct Bazel labels providing the exact same Go import path from the exact same source files, and `rules_go`'s linker correctly rejects the resulting duplicate with "multiple copies of package passed to linker."

Today we work around this in `datadog-agent` by hand-maintaining `gazelle:resolve` and `gazelle:resolve_regexp` directives in `go_deps.gazelle_default_attributes(directives = [...])` that redirect every such import back to the in-tree label. This works, but:
* it has to be updated by hand every time a new external module starts importing one of our in-tree submodules,
* it silently stops working (falls back to the duplicate-label failure) if anyone forgets to update it,
* it does not address the case where two of our own `go.work` members cross-reference each other through the same external label space.

Related, but distinct, findings from building this repro (separate bugs, not covered by the analysis above, but worth mentioning since a real fix will likely touch the same code):

* Without the `replace` line above (no `replace_map` entry for the colliding path), the same setup fails differently. The colliding path is silently omitted from the shared `bazel_gazelle_go_repository_config` (filtered by `if not getattr(module, "go_mod_dir", None)`), so `thirdparty`'s own Gazelle-generated `BUILD.bazel` ends up depending on a Bazel repo name that was never declared, and the build fails at analysis time with `no such package '@@[unknown repo ...]'`.
* A single-`go.mod` root module (no `go.work` at all) whose own top-level import path is required by another `go_deps`-materialized module hits yet another variant: the path is kept in `bazel_gazelle_go_repository_config`, but under the bogus repo name `"@" + module.name`, which is not the module's actual canonical repo name (`@@`, i.e. empty). The generated consumer `BUILD.bazel` then references a nonexistent `@@<module_name>` repo and fails with `Repository '@@<module_name>' is not defined`.

Related, but distinct, existing reports:
* #1658 (general "multiple `go.mod` files in a Bzlmod monorepo" support question, unresolved since 2023),
* #1958 (a different symptom of the same broad "labels generated by a `go_deps`-internal Gazelle invocation don't match what the root module expects" class of problem: `repo_name` mismatch rather than label duplication),
* #2084 (same linker symptom, different trigger: WORKSPACE/Bzlmod mixing for build tools),
* #2336 / #2337 (a narrower, already-fixed bug: dangling `use_repo` entries for locally-`replace`d `go.work` members. That fix addressed `bazel mod tidy` bookkeeping, not repo generation, so it does not help here).

jayconrod:
**What type of PR is this?**

> Feature

> go_deps

This is a complete rewrite of go_deps. While Go MVS is simple to describe, it is unfortunately remarkably difficult to simulate in Starlark, especially considering go.work edge cases that require reading go.mod files from outside any Bazel module. This should make version selection more closely match what Go does.

In short, the new process is:

1. Construct a synthetic go.work workspace, linking each go.mod across Bazel modules.
2. Run 'go list -m -json' and parse the output to select versions.
3. Declare go_repository for each module not provided by a Bazel module.

Fixes #2344
Fixes #2394
For #2413

Should also help with many other subtle go.work issues.

**Other notes for review**

Some compatibility differences:

- `go_deps` can select higher versions of Go modules when `go_deps.from_file` adds multiple `go.mod` files, especially with `go.work` when the same module is required at different versions. It now uses `go list -m`, which pulls in `go.mod` files from requirements, which can add additional version constraints. Use `go work sync` to minimize ambiguity. Use `go get <module>@<version>` to add missing sums to your `go.sum`.
- `gazelle_overrides` no longer overrides global or builtin defaults for attributes that are not set.
- from_file.fail_on_version_conflict is deprecated, now equivalent to a new flag, go_deps.config(checks = "error"). This makes checks more strict, since this applies to multiple checks, across all root module requirements.
- `module.build_naming_convention` and `build_file_proto_mode` are removed. They reported deprecation errors for a long time.
- `module.local_path` is no longer allowed outside the root Bazel module, similar to a `replace` directive in a `go.mod` file.


internal/bzlmod/go_deps.bzl
```diff
@@ -12,925 +12,1393 @@
...
+    for module in module_ctx.modules:
+        if _module_acts_as_root(module_ctx, module):
+            root_module = module
+            if len(module.tags.config) > 1:
+                module_ctx.fail("Multiple go_deps.config tags defined in module {}".format(module.name))
+                return None
+            if len(module.tags.config) == 1:
+                config_tag = module.tags.config[0]
+            if len(module.tags.gazelle_default_attributes) > 1:
+                module_ctx.fail("Multiple go_deps.gazelle_default_attributes tags defined in module {}".format(module.name))
```
fmeum, line 61: Compared to the old implementation, this lacks the references to the tags. The same applies to all other `fail`s that no longer use the (admittedly arcane) `zip` trick to get the pretty print behavior of module extension tags. Maybe this could be extracted into a helper that would allow us to keep the more helpful error messages?

internal/bzlmod/go_deps.bzl
```diff
@@ -12,925 +12,1393 @@
...
+                return
+            go_mod_version = _parse_go_version(go_mod_json["Go"])
+            if go_mod_version > go_version:
+                module_ctx.fail("""\
+In {go_mod_label}, go version is {go_mod_version}, but Bazel is using {go_version}.
+To correct this:
+    1. Upgrade the Bazel Go version in MODULE.bazel:
+
+        go_sdk = use_extension("@rules_go//go:extensions.bzl", "go_sdk")
+        go_sdk.download("{go_mod_version}")
```
fmeum, line 671:
```suggestion
        go_sdk.download(version = "{go_mod_version}")
```

fmeum (commented):
After a first readthrough (and actually even before that) it's clear that this is a very thorough and readable rewrite (+bugfix) of the extension. Great work! I'll give it a more careful read, but I would first like to understand whether we can retain some of the desirable properties of the old implementation.

Most importantly, the old impl didn't require any network access. While that likely can't be avoided for the reasons you point out in the PR description, I wonder whether we could at least arrange for these network accesses to go through Bazel's `rctx.download`, which would guarantee offline evaluation with an up-to-date repository cache or a downloader config rewriting all downloads to a mirror.

For example, we could use module extension `facts` to map Go module + version pairs of relevant deps to their `go.mod`'s download URL + hash. Could we mix `go list -m` and some Starlark logic to do this? Or, more long term, is there a feature request for `golang/go` that would make it more offline friendly? It is a bit surprising to me that `go.mod` purposefully uses a very simple algorithm to resolve versions, but then applies optimizations that make it very difficult to determine the full set of data that it requires as inputs.

internal/bzlmod/go_deps.bzl
```diff
@@ -12,925 +12,1393 @@
...
+    else:
+        checks_level = WARNING
+        check_direct_dependencies_level = OFF
+    from_file_level = OFF
+    for tag in root_module.tags.from_file:
+        if tag.fail_on_version_conflict:
+            from_file_level = ERROR
+    level = max(checks_level, check_direct_dependencies_level, from_file_level)
+    if level == OFF:
+        return lambda *args: None
```
fmeum, line 461:
```suggestion
        return lambda *args, **kwargs: None
```

just so that we don't get unexpected errors when we end up using `sep = `.

internal/bzlmod/go_deps.bzl
```diff
@@ -12,925 +12,1393 @@
...
-        _check_directive(directive)
+def _gazelle_attributes_for_module(importpath, gazelle_overrides, gazelle_default_attributes):
+    """Returns Gazelle settings for a Go module path.
 
-    return struct(**{
-        attr: getattr(gazelle_override_tag, attr)
-        for attr in _GAZELLE_ATTRS.keys()
-        if hasattr(gazelle_override_tag, attr)
-    })
+    Priority, lowest to highest:
```
fmeum, line 494: Should we follow this priority order individually for each attribute or use the highest level among them for everything? For example, generation mode and extra args are somewhat coupled. I'm not sure what's best though, so feel free to stick with this as it's easiest.

internal/bzlmod/go_deps.bzl
```diff
@@ -12,925 +12,1393 @@
...
 
-BUILD files that load other symbols from @rules_proto can be regenerated by adding a
-go_deps.gazelle_override tag with build_file_generation = "clean" for the Go module to the
-relevant MODULE.bazel file.
-"""
+If you need this override for a Bazel module that will be available in a public \
+registry (such as the Bazel Central Registry), please file an issue at \
+https://github.com/bazelbuild/bazel-gazelle/issues/new or submit a PR adding \
+the required directives to the "default_gazelle_overrides.bzl" file at \
+https://github.com/bazelbuild/bazel-gazelle/tree/master/internal/bzlmod/default_gazelle_overrides.bzl.
```
fmeum, line 542: Let's take this chance to update the URL

```suggestion
https://github.com/bazel-contrib/bazel-gazelle/tree/master/internal/bzlmod/default_gazelle_overrides.bzl.
```

internal/bzlmod/go_deps.bzl
jayconrod: I think it should be individually for each attribute. If I remember correctly, this was an intentional change in behavior.

It seems intuitive that the most narrowly scoped setting would take priority over broad defaults, but that can be surprising when a narrowly scoped setting has implicit defaults.

For example, if you set:

```
go_deps.gazelle_override(
    importpath = "example.com/m",
    directives = [...],
)
```

Previously, that would *also* override `build_extra_args` from `go_deps.gazelle_default_attributes`. I don't think the user intends to override it here, so now they must set it explicitly.

internal/bzlmod/go_deps.bzl
jayconrod: Added `_get_only_tag` for this.

The zip trick took me a while to understand. It seems like `str` / `repr` don't work on tags, which is an unfortunate gap in functionality. Added a comment.

internal/bzlmod/go_deps.bzl
```diff
@@ -12,925 +12,1417 @@
...
-                *[t for p in zip(module.tags.gazelle_default_attributes, len(module.tags.gazelle_default_attributes) * ["\n"]) for t in p]
-            )
+    Returns:
+        The tag, if there is only 1. None if there are 0.
+    """
+    tags = getattr(root_module.tags, tag_name, [])
+    if len(tags) > 1:
+        # Pass tags as arguments to fail instead of formatting them as strings.
+        # The str and repr functions print nonsense for these, but the fail
+        # builtin prints something informative.
```
fmeum, line 434: `print` and `fail` handle these specially when they are passed in as `*args`, which is unfortunately necessary to avoid non-hermetic information from leaking into extension evaluation.

You can reference https://bazel.build/rules/lib/builtins/bazel_module_tags for this behavior.

fmeum (approved):
This has some release-blocking issues that Claude found. I have a PR with all the fixes (https://github.com/bazel-contrib/bazel-gazelle/pull/2441). We can merge this now and I'll go through them, filter out slop and send a follow-up PR. I can also send a PR against this branch first - whatever you prefer.

Thanks again for this thorough rewrite and, perhaps even more importantly, the new test framework. This will make our lives so much easier. :-)

internal/bzlmod/go_deps.bzl
```diff
@@ -12,925 +12,1417 @@
...
+    """
 
-                module_resolutions[module_tag.path] = struct(
-                    repo_name = get_repo_name(module_tag.path, module_overrides),
-                    version = version,
-                    raw_version = raw_version,
-                    to_path = to_path,
-                    local_path = local_path,
-                )
+    list_result = env_execute(module_ctx, [go_tool, "list", "-m", "-json", "all"], go_env)
```
rdesgroppes, line 1059: 👍

jayconrod: @rdesgroppes I added test cases based on #2403 verifying #2394 is fixed. Would you mind giving the new implementation and try to see if it fixes the real issue?

This isn't released yet, and there are some follow-ups still in flight, so there's time to change things.

rdesgroppes: On it, please bear with me.

rdesgroppes: @jayconrod sorry for the delay, but I have bad news.

I tried the new implementation against our actual monorepo ([DataDog/datadog-agent](https://github.com/DataDog/datadog-agent)), dozens of `go.work` members) via `local_path_override`, and `go_deps` fails outright before it ever reaches any label resolution:
```
go: cannot load module <member> listed in go.work file: open <member>/go.mod: no such file or directory
```
repeated for every `go.work` member. This looks like the synthetic `go.work` the new implementation constructs keeps each member's path relative to the *original* `go.work`'s location rather than rebasing it to wherever the synthetic copy actually gets written, so every relative `use` entry breaks.

I bisected it with `git bisect run` against a real build in our monorepo, classifying each candidate by whether that exact error reappeared. Last good: `417a00fc` ("language/proto: migrate to v2", #2443). First bad: `f5c890a2`, #2418 itself. Still broken on current `master`. It reproduces regardless of which subset of targets I build, so it isn't specific to any particular target, and I confirmed it's specific to the rewrite, not to `local_path_override` itself, by running the identical override against `v0.54.0`, which gets past workspace setup cleanly.

This blocks verifying #2394 against our actual repo at all. I could only test the two standalone repros from #2403, and both still fail there too:

`tests/bcr/go_work_minimal` (`bazel build //...`) fails identically to the original report:
```
package conflict error: example.com/repro/pkg/foo: multiple copies of package passed to linker:
    @@gazelle++go_deps+com_example_repro_pkg_foo//
    @@//pkg/foo
```

`tests/bcr/go_work/consumer_check` (`bazel build //consumer_check/...`) also still fails, though the specific symptom changed. It no longer creates a literal duplicate repo, but the fetched module's own generated `BUILD.bazel` now emits `deps = ["@@gazelle_bcr_go_work_tests//pkg/data"]`, referencing the root module by its own declared name from outside it, which is never a valid canonical reference (the root's canonical name is empty). Same underlying root-ownership resolution gap, different manifestation.

jayconrod: @rdesgroppes Thanks for checking this. Sorry it didn't work yet.

I'll spend some time tomorrow fixing the issues you mentioned here and will try to verify using #2403 directly.

PR #2420, opened 2026-08-29
language/go: resolve imports of packages in multi-module repos

jayconrod:
**What type of PR is this?**

> Bug fix

> language/go

For example, github.com/bazel-contrib/bazel-gazelle/v2/label now resolves to @gazelle//v2/label.

go_repository_config now adds a fake "prefix_dir" attribute to each go_repository it writes into the fake WORKSPACE configuration file. This is the subdirectory of the Go module within a Bazel module.

The WORKSPACE file is now a list of directives instead of rules, to bypass validation that two rules can't have the same name. With this change, two rules specified as directives CAN have the same name if prefix_dir is different. This might be confusing for update-repos, but update-repos only works in WORKSPACE mode, which is not affected by this change.

The repo.Repo type now has PrefixDir, and the repo.RemoteCache.Root, RootState, and Mod methods all return it. This is incompatible, but the repo package is marked as UNSTABLE. Obviously not ideal, but it's very difficult to bring this functionality into language/go while we still support WORKSPACE.

language/go now uses this for external dependency resolution.

For #2413

**Other notes for review**


internal/bzlmod/go_repository_config.bzl
```diff
@@ -15,20 +15,34 @@
...
-        ))
+            "name=" + repo_name,
+            "importpath=" + importpath,
+        ]
+        module_name = ctx.attr.module_names.get(repo_name)
+        if module_name:
+            words.append("module_name=" + module_name)
+        prefix_dir = ctx.attr.prefix_dirs.get(importpath)
+        if prefix_dir:
+            words.append("prefix_dir=" + prefix_dir)
```
fmeum, line 39: I guess we should just ignore this case, but technically speaking Bazel package names can contain spaces.

fmeum (commented):
Could we add `go_deps` test coverage for this case?

internal/bzlmod/go_repository_config.bzl
jayconrod: Oh, yikes. I hope that'd be rare enough that we can ignore it for now. But #2413 should fix it properly by writing this configuration as JSON rather than directives.

jayconrod (commented):
Added a `config_prefix_dirs.bzl` for coverage in `go_deps_test`.

tyler-french (approved):

PR #2424, opened 2026-09-02
v2/cmd/gazelle/update: update builtin Configurers to v2

jayconrod:
**What type of PR is this?**

> Refactor

> v2/cmd/gazelle/update
> v2/resolve

- v2/config.CommonConfigurer: already v2 with wrapper
- v2/walk.Configurer: already v2 with wrapper
- v2/resolve.Configurer: migrated to v2, added wrapper
- v2/cmd/gazelle/update.updateConfigurer: migrated to v2

Also updated some tests. Flags are deprecated in v2, so dropped RegisterFlags and CheckFlags where possible. Moved initialization to Configure when rel == "", but this required adding calls in a couple places.

For #2272

**Other notes for review**


v2/resolve/resolve_test.go
```diff
@@ -142,14 +142,19 @@ func getConfig(t *testing.T, path string, directives []rule.Directive, parent *c
...
-	configurer.Configure(cfg, path, &rule.File{Directives: directives})
+	err := configurer.Configure(t.Context(), config.ConfigureArgs{
+		Config: cfg,
+		Rel:    path,
+		File:   &rule.File{Directives: directives},
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
 	return cfg
```
tyler-french: nit
```suggestion
	if err := configurer.Configure(t.Context(), config.ConfigureArgs{
		Config: cfg,
		Rel:    path,
		File:   &rule.File{Directives: directives},
	}); err != nil {
		t.Fatal(err)
	}
return cfg
```

v2/cmd/gazelle/update/update.go
tyler-french, line 907: nice!

tyler-french: nit
```suggestion
			for lang := range filterLanguages(c, languages) {
				if err := lang.Fix(ctx, language.FixArgs{
					Config: c,
					Rel:    rel,
					File:   f,
				}); err != nil {
					uc.handleError(err)
				}
}
```

tyler-french (approved):

v2/cmd/gazelle/update/update.go
jayconrod: The new language features are so nice! This saves a slice allocation and copy.

PR #2426, opened 2026-09-03
go_repository: retain shared module trees

blico-v2:
**What type of PR is this?**

Bug fix

`go_repository`

#2389 made `fetch_repo` prune extracted module trees from the default shared module cache. When multiple `go_repository` rules reference the same module and version, Bazel may fetch them concurrently. One process can delete the shared tree while another is copying it, causing errors such as:

```text
fetch_repo: failed copying repo:
lstat .../pkg/mod/example.com/driver@v0.1.5/file.go: no such file or directory
```

I confirmed this with a local reproduction in a large monorepo using three `go_repository` aliases of the same module and version. From a fresh Bazel output base, Gazelle v0.53.0 failed on the first query with the missing-file error above. With this patch, the same fresh-cache query fetched all three repositories successfully and retained the complete 152-file extracted module tree. Enabling `GO_REPOSITORY_EPHEMERAL_MODCACHE=1` also avoided the race by isolating each fetch.

This change disables pruning for the default shared cache and defaults the underlying `fetch_repo -prune` flag to false, restoring the behavior before #2389. Ephemeral mode remains opt-in and its private cache is still removed with `ctx.delete(ephemeral_modcache)`. Host caches remain unmodified. This is a simpler alternative to the locking proposed in #2425.

**Testing**

```text
bazel test //cmd/fetch_repo:fetch_repo_test //internal:bazel_test
```

### DIFF
--- a/internal/go_repository.bzl
+++ b/internal/go_repository.bzl
@@ -321,6 +321,9 @@ def _go_repository_impl(ctx):
         elif ctx.getenv("GO_REPOSITORY_EPHEMERAL_MODCACHE") == "1":
             ephemeral_modcache = ctx.path("_gomodcache_tmp")
             fetch_repo_env["GOMODCACHE"] = str(ephemeral_modcache)
+        else:
+            # This cache is shared by go_repository rules.
+            fetch_repo_args.append("-prune=false")
 
     result = env_execute(
         ctx,
--- a/cmd/fetch_repo/main.go
+++ b/cmd/fetch_repo/main.go
@@ -47,7 +47,7 @@ var (
 	// Module flags
 	version = flag.String("version", "", "module version. Must be semantic version or pseudo-version.")
 	sum     = flag.String("sum", "", "hash of module contents")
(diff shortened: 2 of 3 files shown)


internal/go_repository.bzl
```diff
@@ -321,6 +321,9 @@ def _go_repository_impl(ctx):
         elif ctx.getenv("GO_REPOSITORY_EPHEMERAL_MODCACHE") == "1":
             ephemeral_modcache = ctx.path("_gomodcache_tmp")
             fetch_repo_env["GOMODCACHE"] = str(ephemeral_modcache)
+        else:
+            # This cache is shared by go_repository rules.
+            fetch_repo_args.append("-prune=false")
```
fmeum, line 326: Can we additionally default `prune` to false?

fmeum (commented):
@jayconrod Is this an argument for verifying the dirhash post fetch? I think that we used to do this but optimized it out.

internal/go_repository.bzl
blico-v2: Done. The underlying fetch_repo flag now defaults to false as well.

fmeum (approved):

PR #2429, opened 2026-09-04
gazelle_binary: add generate_gazelle_binary_languages to make v2 list

jayconrod:
**What type of PR is this?**

> Refactor

> gazelle_binary

Instead of generating the code that initializes the list of extensions in Starlark, we now use //cmd/generate_gazelle_binary_languages. This new Go tool parses the export data of each compiled library to determine whether it contains a v1 or v2 extension. It then generates the appropriate function call, wrapping it with an adapter if necessary. The new list has a v2 type.

With this change, v2 extensions can actually be used with either gazelle v1 or v2 (though there aren't any such extensions yet).

For #2272

**Other notes for review**


fmeum (approved):
I wondered how you would do this, nice approach!

internal/gazelle_binary.bzl
```diff
@@ -80,23 +75,18 @@ def gazelle_binary(name, languages, version = 0, **kwargs):
         testonly = kwargs.get("testonly", False),
         visibility = ["//visibility:private"],
     )
+    binary_deps = list(languages) + [
+        "//v2/compat",
+        "//v2/language",
```
dsymonds: I think these two might need a `Label(...)` so that `gazelle_binary` remains usable in other workspaces. I just tested this PR out, and it fails building trying to find a `v2/language` package in my workspace (which, of course, doesn't exist).

dsymonds (changes requested):

internal/gazelle_binary.bzl
jayconrod: Oh, good catch, thanks for testing this.

This is a gap in test coverage. I've added a custom `gazelle_binary` in `tests/bcr/go_mod` so we don't miss this again.

PR #2432, opened 2026-09-08
v2/rule: add load information to KindInfo

jayconrod:
**What type of PR is this?**

> Refactor

> v2/rule, v2/cmd/gazelle/update

v2 drops the language.Language.Loads and language.ApparentLoader.ApparentLoads methods. v2 extensions should now report the same information through new fields KindInfo.Name and KindInfo.LoadedFrom. There is no After field added: it was only used for `WORKSPACE`, since order mattered there.

The v2 Kinds method now returns a slice rather than a map for simplicity: we need to re-index and manipulate this anyway.

For #2272

**Other notes for review**

### DIFF
--- a/cmd/gazelle/update-repos.go
+++ b/cmd/gazelle/update-repos.go
@@ -170,9 +170,7 @@ func updateRepos(
 	loads := []rule.LoadInfo{}
 	for _, lang := range languages {
 		for _, load := range lang.ApparentLoads(c.ModuleToApparentName) {
-			// Remove excess cap so that we can append to Symbols in addKindToLoadList
-			// without mutating the original slice.
-			load.Symbols = load.Symbols[0:len(load.Symbols):len(load.Symbols)]
+			load.Symbols = slices.Clone(load.Symbols)
 			loads = append(loads, load)
 		}
 		for _, kind := range lang.Kinds() {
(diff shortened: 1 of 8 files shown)


cmd/gazelle/update-repos.go
```diff
@@ -169,9 +169,15 @@ func updateRepos(
 	kinds := make(map[string]rule.KindInfo)
 	loads := []rule.LoadInfo{}
 	for _, lang := range languages {
-		loads = append(loads, lang.ApparentLoads(c.ModuleToApparentName)...)
-		for kind, info := range lang.Kinds() {
-			kinds[kind] = info
+		for _, load := range lang.ApparentLoads(c.ModuleToApparentName) {
+			// Remove excess cap so that we can append to Symbols in addKindToLoadList
```
fmeum: This could be `slices.Clip`, but given that this needs two comments about safety, how about we just eagerly `copy` instead?

jayconrod: Changed to `slices.Clone`.

I didn't know about `slices.Clip`, that's neat. I feel like that would make more sense if the write were closer, but you're right, it was confusing.

fmeum (approved):

PR #2441, opened 2026-09-11
go_deps: follow-up fixes for the 'go list -m' rewrite

fmeum:
Follow-up fixes for the `go_deps` rewrite in #2418, based on a review of that PR. Each commit fixes one issue (a few closely related ones are combined where they share the same code path) and comes with unit test coverage in `//tests/bzlmod:go_deps_test` where that was reasonably possible; the commit message says so where it wasn't.

The first commit changes the test harness so that every message printed by `go_deps` must be expected by the test case (previously, printed messages were only checked when a test case listed some), and adds `want.files` to assert on the synthetic `go.work`/`go.mod` files. The conversion tool learned two optional test case fields (`no_go_deps_usage` per module, `go_version_output` per case) and mirrors the extension's new `replace` directives so that recorded `go list` output stays consistent.

Fixes, in commit order:

**Hard failures on previously working setups**
- Root module without a `go_deps` usage failed with "root module not found".
- `go list -m` ran without the host's `GOPROXY`/`GOPRIVATE`/`HTTPS_PROXY`/`SSL_CERT_*`/`PATH`/`HOME`, which `go_repository` forwards. Shared `host_env` helper in `env.bzl`.
- `require` directives for Go modules provided by Bazel modules got no `replace`, so `go list` fetched them from the proxy (fails for private modules and `GOPROXY=off`).
- `go_deps.module(local_path = ...)` got no `replace` either; relative paths are now resolved from the root module and passed to `go_repository` as absolute paths.
- `from_file` under a `dev_dependency = True` usage reported non-dev direct deps (a hard Bazel error for dev-only usages).
- Paths with spaces were written unquoted into the synthetic `go.work`/`go.mod`.
- `go.mod` without a `go` directive crashed; devel toolchain versions produced invalid `go` directives.
- `go.work` without `use` directives crashed (`"Use": null`).
- Bare relative `use foo` paths were copied verbatim; absolute and `..` paths are now rejected with a clear message.
- `go_deps.module` versions without a leading `v` broke `go list`.
- `sum` was mandatory, breaking `module` + `archive_override`.
- A versioned replacement without a zip checksum in `go.sum` (what `go mod tidy` leaves for unneeded modules) failed unconditionally.
- Windows `C:/` paths were turned into `./C:/...` in replace directives (no test: the harness only models POSIX-relative paths).

**Silent wrong behavior**
- Overrides and root `replace` directives targeting Bazel-provided modules were silently ignored; now an error / a check warning.
- Tool labels for Bazel-provided modules used `@<canonical>` instead of `@@<canonical>`.
- Tools in a Bazel-provided sub-module (e.g. `v2/`) resolved to the wrong package.
- Tool targets in `/vN` module roots were named after the invocation name instead of the last path element that Gazelle uses.
- Coverage for module tags merged with `from_file` requirements (the field-dropping merge was reworked by the earlier commits).
- `go.work.sum` was looked up as `go.sum`; `tests/bcr/go_work/workspace.go.sum` is renamed accordingly.
- The synthetic `go_deps_module_tags` module was indexed for Gazelle.
- Absolute `GOROOT`/`GOPATH`/`GOCACHE` were persisted into `@bazel_gazelle_go_repository_config`.
- `gazelle_default_attributes` in non-root modules was silently ignored.

**Spurious diagnostics**
- bazel_dep version checks: relaxed BCR versions (`1.2.3.bcr.1`), empty versions of overridden modules, and newer Bazel modules for indirect requirements were all reported as conflicts.
- Version conflict checks compared against the replacement version instead of the version Go selected (and printed `selected by Go: None` for directory replacements).
- `check_direct_dependencies = "off"` was inert; it now takes the place of `checks` when set, with a deprecation notice.

**Dead code and dropped validation**
- `go_repository_cache` no longer writes the invalid `def.bzl` (no test: dead output).
- Gazelle directives in overrides are validated again.
- `from_file(go_mod = ...)` must name a `go.mod` file again.

`go_deps_test` is now tagged manual in WORKSPACE mode (first commit): go_deps is a module extension, and labels in its test cases and messages use Bzlmod's canonical form. The conversion tool treats POSIX absolute paths as absolute on Windows too (found by the first CI run).

`extensions.md` was updated by hand because stardoc doesn't build on my machine; please run `bazel run //docs:update` if the diff test disagrees.

https://claude.ai/code/session_019GvKuZLSb2AVpavZpHCrP2


tests/bzlmod/go_deps/mvs.bzl
```diff
@@ -53,6 +53,9 @@ TEST = r"""
   },
   "want": {
     "main": {
+      "print": [
+        "Version conflict found for Go module golang.org/x/sync:\n    requested in root module: v0.3.0\n    selected by Go:           v0.11.0"
```
jayconrod, line 57: Let's drop the specific versions from checked messages. I'm worried about these becoming change detector tests if they're too specific.

tests/bzlmod/go_deps_test.bzl
```diff
@@ -114,12 +114,25 @@ def _run_go_deps_instance(env, expect, case, instance_name, isolated, isolate_mo
...
+    # Every printed message must be expected by the test case so that
+    # spurious warnings don't go unnoticed.
+    _assert_messages_contain_substrings(
+        case_expect,
+        module_ctx._state.printed_messages,
+        want.print,
+        "printed_messages",
+    )
+
+    for filename, want_content in want.files.items():
```
jayconrod, line 126: I'd prefer not to check the contents of files in unit tests. `go_deps` can only write files into its scratch directory, which is removed when evaluation finishes. There's no user-observable surface here, only internal implementation detail. Tests that assert on this will be too brittle.

internal/bzlmod/go_deps.bzl
```diff
@@ -461,13 +464,21 @@ def _should_declare_go_repository(module_ctx, go_module):
...
+            not use go_deps.
+    """
     OFF, WARNING, ERROR = 0, 1, 2
     LEVEL = {
         "off": OFF,
         "warning": WARNING,
         "error": ERROR,
     }
+    if not root_module:
+        return module_ctx.print
```
jayconrod, line 481: Let's treat this as `off` and return a no-op. If the user didn't ask for `go_deps`, then none of the warnings are actionable.

internal/bzlmod/go_deps.bzl
```diff
@@ -87,10 +87,15 @@ def go_deps_impl(module_ctx):
     go_tool = go_env["GOROOT"] + "/bin/go" + executable_extension(module_ctx)
     watch(module_ctx, go_tool)
 
+    # 'go list -m' may need to download go.mod files, so like go_repository,
+    # pass proxy, sumdb, and VCS settings through from the host environment.
+    # Explicit settings from the cache repo and go_deps.config take precedence.
+    go_exec_env = host_env(module_ctx.os.environ) | go_env
```
jayconrod, line 93: This commit (`go_deps: inherit proxy, sumdb, and VCS settings from the host for 'go list'`) needs a different approach. Could you pull this out of the stack and submit it after #2438? It will interact with the way I'm managing `GOPROXY` and `GOPRIVATE` there.

Let's have `compute_env` gather any relevant bits of the host environment, to be shared by `go_deps`, `go_repository`, and eventually `@rules_go//:go`. It's important to get the environment in one place so we have consistent settings across tools. And it's important to get the environment from `go env`, not the actual host environment, because the user could store these settings in a config file with `go env -w`.

`compute_env` is only used by `go_repository_cache`, which runs before `go_deps`, so that's the root of the environment variables for this configuration. Unfortunately, that means we can't merge `go_deps.config(go_env = ...)` there, but `resolve_env` does that here in `go_deps`, and the resolved configuration ends up in `go_repository_config`, which is used by `go_repository` and `@rules_go//go`.

We should be careful not to pick up any environment variables that change frequently, since they'll invalidate basically everything. Most of the `HOST_ENV_KEYS` list looks fine, but I'm particularly worried about `PATH`. Let's make sure only `go_repository` looks at `PATH`; nothing else should need it.

internal/bzlmod/go_deps.bzl
```diff
@@ -622,10 +622,15 @@ def _create_workspace_from_tags(module_ctx, go_tool, go_env):
       its 'use' directives, normalizing paths as needed. We only copy 'replace'
       directives if the tag is from the root Bazel module.
     - For each 'module' tag, we add a 'require' directive to a dummy go.mod
-      file, referenced from our go.work file with 'use .'. If the required
-      module is also provided by a Bazel module (via from_file), we add a
-      matching 'replace' directive pointing at that module's workspace
-      directory so 'go list -m' can resolve the version.
+      file, referenced from our go.work file with 'use .'.
+    - For each version of a Go module required by any go.mod file or module
```
jayconrod, line 626: This commit (`go_deps: replace required versions of Go modules provided by Bazel modules`) doesn't make sense to me.

If a Go module is provided by a Bazel module with `from_file`, then its directory will be referenced in the synthetic `go.work` by a `use` directive. There should be no need for a `require` or `replace`. Go won't select a version for that module since it's one of the main modules.

internal/bzlmod/go_deps.bzl
```diff
@@ -672,6 +683,7 @@ def _create_workspace_from_tags(module_ctx, go_tool, go_env):
     module_tag_requires = {}  # Go module path => go_deps.module tag with highest version
     required_versions = {}  # Go module path => dict of required versions from all go.mod files
     root_replaced_paths = {}  # Go module paths replaced by the root module's go.mod or go.work files
+    local_path_dirs = {}  # Go module path => absolute directory from go_deps.module.local_path
```
jayconrod, line 686: Is it important to distinguish `local_path_dirs` and `root_replaced_paths`? I guess so, because we need to generate additional `replace` directives for `module` tags, which we don't need to do for `replace` directives that come from `from_file`.

Let's make the purpose of this variable clearer, naming it `module_tag_local_path_dirs`.

internal/bzlmod/go_deps.bzl
```diff
@@ -943,6 +949,15 @@ def _format_go_mod_json(go_mod_json):
     ])
     return "\n".join(lines)
 
+def _format_replace(r):
+    """Formats a replace directive parsed by 'go mod edit -json' or 'go work edit -json'"""
+    return "replace {}{} => {}{}".format(
+        r["Old"]["Path"],
```
jayconrod, line 955: Do we need to apply `_modfile_token` to each of these?

internal/bzlmod/go_deps.bzl
```diff
@@ -678,9 +678,14 @@ def _create_workspace_from_tags(module_ctx, go_tool, go_env):
         module_ctx.fail("determining go version: {}".format(result.stderr))
         return None
 
-    # example output: go version go1.27rc3 darwin/arm64
-    go_version_str = result.stdout.split(" ")[2][len("go"):]
+    go_version_str = _parse_go_version_output(module_ctx, result.stdout)
```
jayconrod, line 681: Let's pull the execution of `go version` and the check below for `-` below into `_parse_go_version_output` (maybe rename to `_go_version`). Better to get more code out of this huge function.

internal/bzlmod/go_deps.bzl
```diff
@@ -739,7 +744,11 @@ def _create_workspace_from_tags(module_ctx, go_tool, go_env):
             go_mod_json = _parse_go_mod_json(module_ctx, go_tool, go_env, go_mod_path)
             if module_ctx.failed():
                 return
-            go_mod_version = _parse_go_version(go_mod_json["Go"])
+
+            # "As of the Go 1.17 release, if the go directive is missing,
+            # go 1.16 is assumed." 'go mod edit -json' omits the key then.
```
jayconrod, line 749: In the parsing function, let's just fail if there's no Go version instead of trying to support such old edge cases. The `go` directive should be mandatory in any `go.mod` files we read, certainly in any synthetic files we emit.

This only affects `go.mod` files referenced by `from_file`, so the user likely has some control over them.

internal/bzlmod/go_deps.bzl
```diff
@@ -1582,7 +1593,10 @@ _module_tag = tag_class(
             doc = """The module path.""",
             mandatory = True,
         ),
-        "version": attr.string(mandatory = True),
+        "version": attr.string(
+            doc = """The module version, like "v1.2.3". The leading "v" may be omitted.""",
```
jayconrod, line 1597: On `go_deps: accept go_deps.module versions without a leading 'v'`, I regret that we accepted versions without the leading `v` prefix: Go versions should always have it. You're right that we shouldn't break compatibility by rejecting versions without it, but let's not document this behavior.

jayconrod (commented):
Partial review: I got as far as `15b37e7 go_deps: don't fail when a versioned replacement has no go.sum entry`.

PR #2442, opened 2026-09-11
language/go: migrate to v2

jayconrod:
**What type of PR is this?**

> Refactor

> language/go

Migrate to the v2 extension API.

For #2272

**Other notes for review**

Almost entirely agent-written. The tests were pretty overfitted to the v1 API, so there's a pretty big diff there.

I initially intended to do `//language/proto` first, but Go's tests depend on proto, so that would have required a v2-to-v1 compatibility layer, which was ugly when we already have a v1-to-v2 compatibility layer.


language/go/kinds.go
```diff
@@ -16,203 +16,178 @@ limitations under the License.
...
-			"embed":       true,
-			"embedsrcs":   true,
-			"gc_goopts":   true,
-			"gc_linkopts": true,
-			"pgoprofile":  true,
-			"srcs":        true,
-		},
-		ResolveAttrs: map[string]bool{"deps": true},
+var goKinds = []rule.KindInfo{{
+	Name:       "cgo_library",
```
fmeum, line 30: Should we drop this for v2?

language/go/kinds.go
```diff
@@ -16,203 +16,178 @@ limitations under the License.
...
-			"tag":          true,
-			"type":         true,
-			"urls":         true,
-			"vcs":          true,
-			"version":      true,
-		},
+	ResolveAttrs: map[string]bool{"deps": true},
+}, {
+	Name:       "go_grpc_library",
+	LoadedFrom: rulesGoProtoDefBzl,
```
fmeum, line 111: Should we drop this too?

fmeum (approved):
Your output is impressive

language/go/kinds.go
jayconrod: Yes. I don't think we've emitted `cgo_library` for like 8 years, I doubt there's an example left in the wild.

I'll send this as a follow-up PR. It ended up being a decent sized changed on its own because it touches docs and integration tests.

language/go/kinds.go
jayconrod: Yes. I'll remove this in the follow-up PR with `cgo_library`.

jayconrod (commented):
Thanks! I was able to carve out a few weeks of project time this quarter for v2, so I'm trying to make the most of it. I'm trying to wrap up most of it by the end of this week, though I'm sure there will still be a long tail of things to fix.

PR #2449, opened 2026-09-14
language/defaults: move from language/bazel/visibility, promote to be built-in

jayconrod:
**What type of PR is this?**

> Refactor

> language/defaults née language/bazel/visibility

language/bazel/visibility provides some core functionality: it generates package statements with 'default_visibility' and 'features' attributes. This functionality should be built into Gazelle, but it still fits well within the extension architecture, so let's make it a built-in extension so that users don't need to specify it explicitly.

- Moved from language/bazel/visibility to language/defaults since default_visibility is not its only responsibility. And 'bazel' doesn't really add information.
- Left behind a small package that forwards the NewV2() call. Aspect imports it directly, so we can't delete it.
- In v2, gazelle_binary silently drops language/bazel/visibility and adds language/defaults if it's not present.

For #2272

**Other notes for review**

No substantial changes to the actual functionality.


internal/gazelle_binary.bzl
```diff
@@ -110,3 +112,39 @@ def _get_gazelle_major_version():
         return 2
     else:
         fail("Unsupported Gazelle major version: {}. Only versions 0, 1, and 2 are supported.".format(major))
+
+def _languages_for_version(languages, version):
```
tyler-french: optional: would it be possible to add a small test for these functions?

tyler-french (approved):

internal/gazelle_binary.bzl
jayconrod: Done. These feel a little too internal for testing to me, but it's probably better than not: `gazelle_binary` is hard to test otherwise.

PR #2456, opened 2026-09-18
v2/merger: MergeFile: move args into struct, support errors

jayconrod:
**What type of PR is this?**

> Refactor

> v2/merger

Support compatible changes in the future. No change in behavior now. This would have let us implement #2232 without such an ugly hack. And it seems like we're always changing things about alias / map kinds.

For #2272

**Other notes for review**


tyler-french (approved):
Nice! This is much cleaner

